PR #3, opened 2016-10-10
Atomic class definition redundant

Ethanlm:
In include/services/daal_atomic_int.h, the Atomic definition is redundant. And also the return type of inc() and dec() should be dataType instead of int


iburyl:
Thank you for your contribution.
We will review your contribution. When accepted, your pull request will be merged into our internal and GitHub repositories.
Intel DAAL is licensed under Apache License, Version 2.0. By contributing to the project, you agree to the license and copyright terms therein and release your contribution under these terms.

iburyl:
Ethanlm, change base for your commit to 01org:daal_2017_update1 resulted in lost history between commits. I will recommit it separately to daal_2017_update1. Thank you.

PR #34, opened 2017-08-04
adding daal4scripting TP

fschlimb:
Initial TP release


dr-pain:
Need to test it before integration. In process...

fschlimb:
When can this be merged?

PR #62, opened 2019-09-13
aos_numeric_table.h check offset and index values

ntfshard:
Checks that offset and index in setFeature and setOffset is in valid range


averbukh:
make test

averbukh:
make tests

ntfshard:
> make tests

jobs/e9d62343-0326-f161-ab23-8cdcd4b723a1

PR #78, opened 2019-09-23
Config file for clang-format

averbukh:
Config file for clang-format was added to use it for automatic formatting


SmirnovEgorRu:
Thank you for the PR!

As it was discussed offline - please, include in PR an example of work for the tool for 1 random DAAL algorithm to test correctness of set parameters.

PR #93, opened 2019-09-26
Fixed impurity decrease computation in GBT.

vepifanov:
Fixed typos leading to incorrect impurity decrease computation and incorrect early stop which reduces final accuracy.


vepifanov:
make test

PR #106, opened 2019-10-01
Fix nullptr dereference if not ok

ntfshard:
If not ok, then we should read status first, after than we should remove object. Not vice-verse.


prutskov:
make test

ntfshard:
make test

ntfshard:
restart

ntfshard:
restart

PR #116, opened 2019-10-08
Enable one cpu build for CI

napetrov:
This change switch CI build process from all CPUs to single CPU


SmirnovEgorRu:
@napetrov What is level of confidence that all machines in CI supports at least avx2?
If all of them support - it is approved

napetrov:
@SmirnovEgorRu - actually current implementation of one CPU target build get both defined and minimal supported versions.  So even in case we will not have AVX2 system dispatcher will fail-back to lowest one. 
Also at this stage we do not run tests/examples so change itself is safe as build will succeed even on system without AVX2 support.

SmirnovEgorRu:
@napetrov in this case, I think we need to build only minimum supported version w/o avx2 to speedup CI

PR #122, opened 2019-10-09
Updated compilers option

maria-Petrova:
Updated compilers option for preventing buffer overflows to a new one. 
The change should not affect performance.


maria-Petrova:
make test

PR #133, opened 2019-10-10
log loss fix unstable (merge to rls)

ShvetsKS:


ShvetsKS:
precommit has passed:
e9eb5ba9-6cc5-f1e3-8556-8cdcd4b723a1

PR #134, opened 2019-10-10
variable importance names were aligned with DF

ShvetsKS:


ShvetsKS:
warning in precommit happened due to pydaal build (but it was fixed in test branch)

PR #136, opened 2019-10-10
Fix the bug in svm_train

agorshk:


algorithms/kernel/svm/svm_train_boser_impl.i
```diff
@@ -914,7 +915,6 @@ Status SVMCache<simpleCache, algorithmFPType, cpu>::updateShrinkingRowIndices(
         result |= daal::services::daal_memcpy_s(_tmp.get(), lineSizeInBytes, _cache.get() + i * _lineSize, lineSizeInBytes);
         result |= daal::services::daal_memcpy_s(_cache.get() + i * _lineSize, lineSizeInBytes, _cache.get() + j * _lineSize, lineSizeInBytes);
         result |= daal::services::daal_memcpy_s(_cache.get() + j * _lineSize, lineSizeInBytes, _tmp.get(), lineSizeInBytes);
-        DAAL_CHECK(!result, ErrorMemoryCopyFailedInternal);
```
ShvetsKS: put DAAL_ASSERT at least here, if it affects performance.

agorshk: I suggest to merge it and add this assert later

PR #137, opened 2019-10-11
Fix performance degradation for apriori, knn_train, pca_transform

agorshk:


agorshk:
make test

agorshk:
Performance was checked, it's ok. ID: e9ebfd48-03c3-f127-8d16-8cdcd4b723a1

PR #162, opened 2019-10-22
errorID fixed collision, backward compatible fix

ntfshard:
This fix should remove collision with ErrorMethodNotSupported(-1) error.
Picked value are belong to announced range [-2k..-2999] and didn't move all previous values.
Old user's code still have -1 as a error value which leads to not right error message in parseResourceFile function in cpp file which doesn't change code behaviour in case of dynamic linkage.

### DIFF
--- a/include/services/error_indexes.h
+++ b/include/services/error_indexes.h
@@ -145,7 +145,7 @@ enum ErrorID
     ErrorBufferSizeIntegerOverflow = -80,                               /*!< Integer oveflow is occured during buffer size calculation */
 
     // Environment errors: -2000..-2999
-    ErrorCpuIsInvalid = -1,                                             /*!< Invalid CPU value used */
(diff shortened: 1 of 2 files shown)


include/services/error_indexes.h
```diff
@@ -145,7 +145,7 @@ enum ErrorID
     ErrorBufferSizeIntegerOverflow = -80,                               /*!< Integer oveflow is occured during buffer size calculation */
 
     // Environment errors: -2000..-2999
-    ErrorCpuIsInvalid = -1,                                             /*!< Invalid CPU value used */
+    ErrorCpuIsInvalid = -2999,                                          /*!< Invalid CPU value used */
```
averbukh: Looks like it breaks backward compatibility

kiselik: @ntfshard do you plan fix the backward compatibility problem or close the PR?

napetrov:
@ntfshard  What is current status for this request? do we need it?

PR #165, opened 2019-10-25
optimized spark sample KMean

PivovarA:
optimized spark sample KMeans for case, when data doesn't fit in memory


PovelikinRostislav:
Seems that EOL settings  were changed in the file and the diff is not readable itself. (Github shows it as you changed every line of the file)
Have you configured your local git settings to commit EOL in an appropriate manner?

vepifanov:
> Is it OK that EOL were changed in the file and merged into master?

No, that should be fixed next time.

PovelikinRostislav:
@vepifanov, thanks

PR #172, opened 2019-10-30
configured KMeans spark sample code in appropriate manner

PivovarA:
configured KMeans spark sample code in appropriate manner


PivovarA:
make test

PR #178, opened 2019-11-01
Fix of makefile for one algorithm build

orazve:
PCm: http://intel-ci.intel.com/e9fc81f8-aa46-f13c-b0c9-8cdcd4b723a1


orazve:
make test

napetrov:
As we have 1cpu build broken now - do we  think that this PR can fix this problem?

orazve:
@napetrov no, this fix is only for one algorithm build, i.e. when CORE.ALGORITHMS.CUSTOM makefile option is specified

PR #184, opened 2019-11-18
helper scripts fixed

vmeshche:
# Description
This PR is to fix issues in helper scripts.

- scripts/tbb.sh calls without parameters. 
- scripts/mklfpk.sh parameters modified to be able to work without parameters. Arch parameter need to exclude gpu fpk in case of ia32.


dr-pain:
1. mklfpk shouldn`t request any parameters
2. testing isn`t passed

.ci/pipeline/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     displayName: 'apt-get'
   - script: |
       scripts/mklfpk.sh ${$(platform.type):3}
```
napetrov: The problem is that this construction doesn't work - ${$(platform.type):3} results in empty string.
Please fix it

scripts/mklfpk.sh
```diff
@@ -19,17 +19,15 @@
 MKLFPK_URL_ROOT="https://github.com/intel/daal/releases/download/Dependencies/"
 MKLFPK_VERSION="20180112_10"
 MKLGPUFPK_VERSION="20191109"
-MKLFPK_ARCH=32e
+WITH_GPU=false
```
napetrov: We position oneDAL as primary thing in this repo. So i suggest to change this to default true

vmeshche: np.
done

dr-pain:
I see that tbb.sh do tbb download even if tbb already downloaded. needs to be fixed.

dr-pain:
The same wrong behavior observed for tbb.bat.

dr-pain:
Can`t build DAAL after tbb.bat is used:
makefile:196: *** Can`t find TBB neither in ./externals/tbb nor on $TBBROOT.  Stop.

possible reason - tbb downloaded to .\externals\tbb\. Right location is .\externals\tbb\win

PR #187, opened 2019-11-19
Added ittnotify tasks for training of Linear Regression CPU/GPU

prutskov:


algorithms/kernel/linear_model/linear_model_train_normeq_finalize_impl.i
```diff
@@ -66,6 +68,7 @@ Status FinalizeKernel<algorithmFPType, cpu>::compute(const NumericTable &xtxTabl
 
         if (&xtxTable != &xtxFinalTable)
         {
+            DAAL_ITTNOTIFY_SCOPED_TASK(computeFinalize.copyToOnline);
```
michael-smirnov: This task has the same name as the one on line 82. It`s confusing for me, do you really need to consider these parts of code as a single task? Also, the name **copyToOnline** says nothing to me, could you please name it more clear?

algorithms/kernel/linear_model/linear_model_train_normeq_update_impl.i
```diff
@@ -87,12 +88,16 @@ Status ThreadingTask<algorithmFPType, cpu>::update(DAAL_INT startRow, DAAL_INT n
...
+    {
+        DAAL_ITTNOTIFY_SCOPED_TASK(computeUpdate.syrkX);
+        Blas<algorithmFPType, cpu>::xxsyrk(&up, &notrans, &nFeatures, &nRows, &alpha,
+                                        const_cast<algorithmFPType *>(x), &nFeatures, &alpha,
+                                        _xtx, &_nBetasIntercept);
+    }
 
     if (nFeatures < _nBetasIntercept)
     {
+        DAAL_ITTNOTIFY_SCOPED_TASK(computeUpdate.gemm1X);
```
michael-smirnov: Is this really a gemm stage?

prutskov: Yes, it is equivalent of gemm with vector of ones from GPU version of implementation

algorithms/kernel/linear_model/linear_model_train_normeq_finalize_impl.i
prutskov: fixed

prutskov:
make test

PR #202, opened 2019-11-27
Support for Unified Shared Memory (USM)

rlnx:
🎉  **Introducing USM support** 🎉 
- Introduced new buffer implementation based on USM (see `buffer_impl_sycl.h`, `UsmBuffer`).
- Added new constructors to `Buffer` for creation from USM and the `Buffer<T>::toUSM()` method to extract USM pointer from the buffer.
- Overall improvement of error handling in `Buffer`, e.g., `Status` as optional parameter.
- Refactoring of `Buffer` interface and implementation classes. `services/buffer.h` is split into three files:
   ```sh
   services/buffer.h  # User-level interface 
   services/internal/buffer_impl.h # Base impl. classes and cpu-specific buffer impl.
   services/internal/buffer_impl_sycl.h # SYCL-specific impl.
   ```
- Added new constructor to `SyclNumericTable` for creation from USM.
- Refactoring of `SyclExecutionContext`, move `SyclKernelScheduler` to `oneapi/internal/kernel_scheduler_sycl.h`.
- Added SYCL example `data_management/numeric_table_usm.cpp`.

CI: http://intel-ci.intel.com/ea150516-c2b4-f187-a00c-8cdcd4b723a1
More info on USM: https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/USM/USM.adoc


include/services/internal/collection.h
```diff
@@ -28,8 +29,112 @@ namespace services
...
+
+    virtual ~PrimitiveCollection()
+    {
+        destroy();
+    }
+
+    void destroy()
+    {
+        services::daal_free((void *)_buffer);
+        _buffer = NULL;
```
amgrigoriev: nullptr is used in other parts of DAAL

rlnx: This is not SYCL part of the library, C++98 is required

amgrigoriev: Ok

include/data_management/data/numeric_table_sycl_homogen.h
```diff
@@ -225,29 +251,31 @@ class DAAL_EXPORT SyclHomogenNumericTable : public SyclNumericTable
...
-        }
+            if (getNumberOfRows())
+            {
+                const size_t size = getNumberOfColumns() * getNumberOfRows();
 
-        if (status)
-        {
-            _memStatus = internallyAllocated;
+                const auto universalBuffer = oneapi::internal::getDefaultContext()
+                    .allocate(oneapi::internal::TypeIds::id<DataType>(), size, &status);
```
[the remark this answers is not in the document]
rlnx: There is a check at line 264. Situation when the number of rows is zero is not error, hence the method `allocateDataMemoryImpl` just deallocates the memory and does not allocates the new. This behaviour is aligned with the other numeric tables.

Data source relies on this behaviour when calls `nt.resize(0)` intentionally, which causes invocation of `allocateDataMemoryImpl()` with zero numbers of rows.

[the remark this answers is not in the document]
rlnx: To be honest, I'm not sure whether this is right behaviour or not, but anyway it's сounterintuitive. Unfortunately, it cannot be fixed locally in this class and requires fixes in data sources, `NumericTable` and its descendants.

examples/cpp_sycl/makefile_lnx
```diff
@@ -135,7 +135,7 @@ sointel64:
 _make_ex: $(RES)
 
 vpath
-vpath %.cpp $(addprefix ./source/,datasource covariance linear_regression logistic_regression optimization_solvers pca kmeans moments)
+vpath %.cpp $(addprefix ./source/,datasource data_management covariance linear_regression logistic_regression optimization_solvers pca kmeans moments)
```
michael-smirnov: Probably we should locate an example for `syclSOANumericTable` (in datasource folder) and example on USM in the same folder

PR #209, opened 2019-12-02
change storage levels and modified SVD

PivovarA:
Сhanged storage levels in kmeans and svd.
Modified SVD for performance improvement in case when data does not fit into memory.
Updated comparison function in sorting for ALS, to speed up conversion.

### DIFF
--- a/samples/scala/spark/sources/SVD.scala
+++ b/samples/scala/spark/sources/SVD.scala
@@ -23,6 +23,7 @@ import org.apache.spark.SparkContext
 import org.apache.spark.api.java._
 import org.apache.spark.api.java.function._
 import org.apache.spark.rdd.RDD
+import org.apache.spark.storage.StorageLevel
 
 import scala.Tuple2
 import scala.collection.mutable.ArrayBuffer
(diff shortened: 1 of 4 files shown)


PivovarA:
make test

samples/scala/spark/sources/SVD.scala
```diff
@@ -102,100 +102,78 @@ object SVD {
...
+                tables += table
+
+                context.dispose()
+                arrays.clear
                 tables.iterator
             }
-        )
-
-        tablesRDD.zipWithIndex.cache
+        ).persist(StorageLevel.MEMORY_AND_DISK)
```
Alexsandruss: Why exactly this storage level is used? Is it optimal for any dimensions?

samples/scala/spark/sources/SampleKMeans.scala
```diff
@@ -30,7 +31,7 @@ object SampleKMeans extends App {
     val sc = new SparkContext(conf)
 
     val data = sc.textFile("/Spark/KMeans/data/KMeans.txt")
-    val dataRDD = data.map(s => Vectors.dense(s.split(' ').map(_.toDouble))).cache()
+    val dataRDD = data.map(s => Vectors.dense(s.split(' ').map(_.toDouble))).persist(StorageLevel.MEMORY_AND_DISK)
```
Alexsandruss: Same questions for storage level

samples/scala/spark/sources/SampleSVD.scala
```diff
@@ -35,12 +36,15 @@ object SampleSVD extends App {
     val sc = new SparkContext(conf)
 
     val data = sc.textFile("/Spark/SVD/data/SVD.txt")
-    val dataRDD = data.map(s => Vectors.dense(s.split(' ').map(_.toDouble))).cache()
+    val dataRDD = data.map(s => Vectors.dense(s.split(' ').map(_.toDouble))).persist(StorageLevel.MEMORY_AND_DISK)
```
Alexsandruss: Same questions for storage level

samples/scala/spark/sources/SampleSVD.scala
```diff
@@ -35,12 +36,15 @@ object SampleSVD extends App {
...
 
     val nRows = data.count()
     val nCols = dataRDD.first.size
 
-    val result = SVD.computeSVD(new RowMatrix(dataRDD, nRows, nCols))
+    val rowMatrix = new RowMatrix(dataRDD, nRows, nCols)
+    rowMatrix.rows.persist(StorageLevel.MEMORY_AND_DISK)
```
Alexsandruss: Same questions for storage level

samples/scala/spark/sources/SVD.scala
PivovarA: This storage level is used, because in case, when data doesn't fit into memory, it will be written to disk. The memory_only level, that used before, recomputes some partitions if they don't fit into memory, which may be inefficient.

PivovarA: On small dimensions, this storage level = Memory_only
At large dimensions, those partitions that don't fit into memory will be written to disk. Those partisans that did not fit in memory on disk will be recomputed.

PivovarA:
make test

PR #225, opened 2019-12-09
Added blockSize which accelerates GEMM stage in Linear Regression GPU

prutskov:


PetrovKP:
For double type it is optimal size block?

prutskov:
@PetrovKP yes, size of block is optimal for double type also.

prutskov:
make test

PR #228, opened 2019-12-10
Check arguments of getVectorUpCast function

ntfshard:
### Description
We should check the first argument before invoke the `getVectorUpCast` function. We can't apply UpCast to _Other_ type.


ntfshard:
make test

PR #231, opened 2019-12-11
Added requirements for doc generation

outoftardis:
# Description
- Added a readme file to /docs folder with instructions on how to build documentation locally + requirements
- Automated: include all existing SYCL examples to the doc


docs/Makefile
```diff
@@ -27,9 +27,13 @@ BUILDDIR      = build
...
+add:
+	@echo "Creating rST files for SYCL* Examples..."
+	py -3 rst_examples.py
+
+.PHONY: help Makefile 
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile
+%: add Makefile
```
dr-pain: in case of parallel build make may reorder targets. so, Makefile ,ay be executed before add.
it is also suggested to rename targets to get more convenient names.

outoftardis: I renamed the target.

Documentation is not built with parallel builds, I don't think there is a problem.

napetrov:
make test

napetrov:
@outoftardis  - please fix content in 
docs/rst_examples.py

docs/Makefile
dr-pain: I can run this make with -j option. so, it may be run in parallel.

docs/Makefile
```diff
@@ -27,9 +27,13 @@ BUILDDIR      = build
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile
+add:
+	@echo "Creating rST files for SYCL* Examples..."
+	py -3 rst_examples.py
```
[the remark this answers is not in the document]
outoftardis: It creates rst files for existing SYCL examples, not just a list of files. I know how to do it in Python, I have no idea how to do it in a make file. 
Also, Sphinx works with Python anyway, why adding another Python script to the build process is a bad idea?

docs/Makefile
outoftardis: ok, is there a way to fix this or prevent from happening?

outoftardis:
@napetrov  I don't think I fixed everything

napetrov:
make test

napetrov:
restart

docs/Makefile
```diff
@@ -27,9 +27,13 @@ BUILDDIR      = build
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile
+create_rst_examples:
+	@echo "Creating rST files for SYCL* Examples..."
+	py -3 rst_examples.py
```
averbukh: You introduced dependency to 'py' here. Please do it carefully with appropriate diagnostics for users, if she have no correct environment. Also, introduce appropriate variable to make it possible to specify full path via command line argument for users.

docs/Makefile
```diff
@@ -27,9 +27,13 @@ BUILDDIR      = build
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile
+create_rst_examples:
```
averbukh: Add .PHONY for 'create_rst_examples'

docs/rst_examples.py
```diff
@@ -0,0 +1,57 @@
...
+    if os.path.exists(rst_path):
+        shutil.rmtree(rst_path)
+    os.makedirs(rst_path)
+
+    files = list_examples(example_path)
+    write_examples(files, rst_path)
+
+if __name__ == "__main__":
+    script_dir = os.path.dirname(os.path.abspath(__file__))
+    run(script_dir)
```
averbukh: Looks like it can be easily done in the make file without python. See Andrey's comment.

docs/Makefile
napetrov: Users are not expected to use this document generation scripts

docs/Makefile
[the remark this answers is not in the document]
napetrov: Solved - we will go with python script for now

docs/rst_examples.py
napetrov: We've sync and agree to keep python script

napetrov:
make test

napetrov:
restart

docs/Makefile
outoftardis: added

napetrov:
make test

docs/rst_examples.py
averbukh: What are the reasons?

napetrov: This is something that already works and this python call introduce minimal dependency.  The plan is to enable automatic generation and then align final automation with build system that we will endup using

PR #260, opened 2020-01-13
Fix gradient computation

ShvetsKS:
# Description
Fix was applied to clear/rewrite buffer allocated for gradient before compute.


ShvetsKS:
make test

PR #258, opened 2020-01-10
MKL functions are enabled by default

prutskov:
Macro **ONEAPI_DAAL_USE_MKL_GPU_FUNC** was replaced by **ONEAPI_DAAL_NO_MKL_GPU_FUNC** with inversing of benaviour. MKL functions are used by default now.

It is fix isuue #245

PR #270, opened 2020-01-14
MKL functions are enabled by default (#258)

prutskov:
MKL functions are enabled by default and gbt vsproj fixed


prutskov:
PCm: http://intel-ci.intel.com/ea370ff4-f81e-f181-8627-8cdcd4b723a1?

PR #272, opened 2020-01-15
Code cleanup

ntfshard:


ntfshard:
make test

ntfshard:
make test

algorithms/kernel/assocrules/assoc_rules_apriori_tree.i
```diff
@@ -157,7 +157,7 @@ struct hash_tree
         if (!is_child(node_idx, hash_res))
         {
             /* Update children mask of the root node if needed */
-            root[node_idx].mask |= (1 << hash_res);
+            root[node_idx].mask |= (__int64)(1) << hash_res;
```
Vika-F: Too much braces. It's better to have just __int64(1). Anyway, I'm Ok with the change.

algorithms/kernel/optimization_solver/lbfgs/lbfgs_dense_default_impl.i
```diff
@@ -97,7 +97,7 @@ services::Status LBFGSKernel<algorithmFPType, defaultDense, cpu>::compute(
 
     if (useWolfeConditions)
     {
-        gradientFunction->sumOfFunctionsParameter->resultsToCompute |= objective_function::value;
+        gradientFunction->sumOfFunctionsParameter->resultsToCompute |= (DAAL_UINT64)(objective_function::value);
```
Vika-F: Too much braces again.

algorithms/kernel/assocrules/assoc_rules_apriori_tree.i
ntfshard: Semantically it's different construtions. We must put type inside () due it's not 1 word.

ntfshard: (Type)1  - c-style cast
Type(1)  construction of Type with passed value to ctor's parameter

include/data_management/data_source/file_data_source.h
```diff
@@ -196,7 +196,7 @@ class FileDataSource : public CsvDataSource<FeatureManager, SummaryStatisticsTyp
             return services::throwIfPossible(services::ErrorOnFileOpen);
         }
 #else
-        _file = fopen((char *)(fileName.c_str()), "r");
+        _file = std::fopen((char *)(fileName.c_str()), "r");
```
Vika-F: If this worked previously, then we have "using namespace std" or "#include <stdio.h>" somewhere in the includes. But I can't find any of those. I am Ok with the change, but can you please explain how this worked previously?

samples/java/hadoop/sources/LinearRegressionNormEq.java
```diff
@@ -51,8 +51,14 @@
...
 
-    public static void main(String[] args) throws Exception {
-        int res = ToolRunner.run(new Configuration(), new LinearRegressionNormEq(), args);
+    public static void main(String[] args) {
+        int res = -1;
+        try {
+            res = ToolRunner.run(new Configuration(), new LinearRegressionNormEq(), args);
+        } catch (Exception e) {
+            System.out.println("Error");
+            e.printStackTrace();
```
Vika-F: Let's print more meaningful diagnostic. I think that stack trace is not enough.
            System.out.println("Exception caught: " + e);
            System.out.println("Exception message: " + e.getMessage());
            System.out.println("Exception cause: " + e.getCause());
            e.printStackTrace();

Same comment to the rest of the samples.

include/data_management/data_source/file_data_source.h
ntfshard: First of all, according to [iso cpp core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rs-using-directive) we mustn't write `using namespace std` inside headers.
The second one, fopen without namespace is available for compatibility with C. We have included cstdio in the begging of the file. So we should use [std::fopen](https://en.cppreference.com/w/cpp/io/c/fopen)

Vika-F: I do know that "using namespace std" is prohibited. But I didn't know that fopen exists in global namespace. Thanks for explanation!

algorithms/kernel/assocrules/assoc_rules_apriori_tree.i
ntfshard: replaced by static_cast

KalyanovD:
restart

include/data_management/data_source/file_data_source.h
averbukh: OK to have 'std::' here. The possible reason why it worked before is inclusion of stdio.h in another file that included here indirectly.

ntfshard:
restart

ntfshard:
make test

PR #309, opened 2020-01-27
Fix gap in WindowsDLL libs structure

dr-pain:


dr-pain:
PCm - http://intel-ci.intel.com/ea3e6f9c-58f9-f169-bb00-8cdcd4b723a1

PR #311, opened 2020-01-28
Optimization for one row inference in DF

ShvetsKS:
# Description
Optimization for one row inference in DF classification


ShvetsKS:
make test

PR #312, opened 2020-01-28
Optimization for one row inference in DF

ShvetsKS:
# Description
Optimization for one row inference in DF classification


ShvetsKS:
CPm: http://intel-ci.intel.com/ea41a770-84da-f18c-9e74-8cdcd4b723a1

PR #318, opened 2020-01-29
replace int to align types in compare operation

ntfshard:
and unused variable removed in java


ntfshard:
make test

algorithms/kernel/cosdistance/cosdistance_lp_impl.i
```diff
@@ -69,7 +69,7 @@ services::Status cosDistanceLowerPacked(const NumericTable * xTable, NumericTabl
 
         /* compute inverse of sqrt of gemm result and save for use in computation off-diagonal blocks */
         PRAGMA_VECTOR_ALWAYS
-        for (int i = 0; i < blockSize1; i++)
+        for (DAAL_INT i = 0; i < blockSize1; i++)
```
ShvetsKS: seems 'size_t' would be more preferable here.

ntfshard: Yep, but intention of this patch is to make it consistent. Changing it could lead to huge code refactoring and I not sure that it should be implemented in such context.

PR #326, opened 2020-01-30
Disable op= for Batch* classes with copy ctor

ntfshard:
For Batch classes with non-trivial copy ctor we should use specific copy ctor or disable it;


ntfshard:
make test

PR #327, opened 2020-01-30
replace int by same type (#318)

ntfshard:
Using the exact same type inside for loops;
Removed unused variable;
Small fix for printf


ntfshard:
make test

ntfshard:
restart

PR #329, opened 2020-01-30
Disable op= for InputDataArchive class

ntfshard:


ntfshard:
make test

ntfshard:
Ok. I was not right about denying op= of this classes. All derived classes is fine with default op=.
The main question is about `interface1::Argument` class: it has interesting copy ctor and default op=

ntfshard:
restart

ntfshard:
make test

PR #331, opened 2020-01-31
Fix of sporadic adaboost multiclass samme java example [to releases branch]

orazve:
Fixed:
- hidden nested threader_for that causes incorrect results
- removed unnecessary includes


orazve:
make test

PR #335, opened 2020-02-03
Fix CEV build after optimizations in DF

ShvetsKS:
# Description
Fix CEV build after optimizations in DF


ShvetsKS:
make test

ShvetsKS:
ea4679cd-e056-f1a6-b1b9-8cdcd4b723a1?

PR #350, opened 2020-02-06
Replace relative includes with full paths (#346)

rlnx:
see #346
- Replaced all chaotic includes in all .cpp/.h/.i files with the includes relative to <root>
and<root>/include directories
- Updated Makefile by removing `-I` flags to include paths


napetrov:
make test

napetrov:
I see noticeable build speedup with this change. :fire:

rlnx:
make test

napetrov:
restart

rlnx:
This is valid one http://intel-ci.intel.com/ea4988fe-88cf-f168-9c22-8cdcd4b723a1

rlnx:
> I see noticeable build speedup with this change. 🔥

@napetrov, I also noticed we compile twice the same files with the same options -- one for static, another for dynamic lib. On Linux and Mac it is not necessary, make sense only on Windows. Probably, we can speedup build 2x on Linux and Mac for free.

napetrov:
@rlnx  yes, i've also check that we do strange things like building knl optimizations on mac using AVX2 target.   But for sure changing logic for staic->dynamic build will help.

PR #352, opened 2020-02-06
add setNFeatures to df and gbt regression models

PivovarA:
add setNFeatures method to regression model and add features check to decision forest and gbt prediction input check


PivovarA:
make test

PivovarA:
make test

PR #354, opened 2020-02-06
disable operator= for Distributed classes

ntfshard:
This classes has some pointer manipulation and clone() method;


ntfshard:
make test

PR #357, opened 2020-02-06
Disable operator= in Online and other classes

ntfshard:
This classes has ptr field, or has clone method, or non-trivial copy-ctor


ntfshard:
make test

PR #359, opened 2020-02-07
Fix SDL gaps in knn

dmitry741:
Fix SDL gaps in KNN algorithm. Added some instructions according to SDL T204 requirements.


dmitry741:
make test

PR #364, opened 2020-02-09
Fix for DF training

masdevas:
- Merge redundant memory allocations into existing allocations (and repair memory leaks)
- Fill undefined values "featureValueOrResponse" and "leftIndexOrClass" in leaves for classification and regression

### DIFF
--- a/algorithms/kernel/dtrees/dtrees_model_impl_common.h
+++ b/algorithms/kernel/dtrees/dtrees_model_impl_common.h
@@ -68,7 +68,6 @@ void nodeToTable(const NodeBase& node, size_t iRow, size_t& iCur, DecisionTreeNo
             }
         }
         row.featureIndex = -1;
-        row.featureValueOrResponse = 0;
         writeLeaf<typename NodeType::Leaf>(l, row);
     }
 }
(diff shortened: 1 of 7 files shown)


masdevas:
CI: http://intel-ci.intel.com/ea4b8249-be95-f1a9-b8dd-8cdcd4b723a1

ShvetsKS:
Please provide some evidence with fixed fail in weekly testing.

algorithms/kernel/dtrees/dtrees_model_impl_common.h
```diff
@@ -68,6 +68,7 @@ void nodeToTable(const NodeBase& node, size_t iRow, size_t& iCur, DecisionTreeNo
             }
         }
         row.featureIndex = -1;
+        row.featureValueOrResponse = 0;
```
ShvetsKS: in regression case we have:
void writeLeaf(const TLeaf& l, DecisionTreeNode& row)
{
    row.featureValueOrResponse = l.response;
}
This field `featureValueOrResponse` is not  initialized only for classification flow.
Seems we have to add the same for regression flow: `row.leftIndexOrClass = 0`

ShvetsKS:
Did you plane to fix failed test cases in weekly scope?
Is it fix will be covered in test system?

PR #371, opened 2020-02-11
[2020 mnt] Removed avx512_mic instructions from win64e

teabagk7:
MKL doesn't support avx512_mic on win64e since ~ 2019u1


teabagk7:
[CI here!] (http://intel-ci.intel.com/ea4c6806-a110-f1c9-ba17-8cdcd4b723a1)

PR #376, opened 2020-02-13
Fix for 32bit ssystem: without aligned memory support

ShvetsKS:
# Description
Fix for 32bit system: without aligned memory support


ShvetsKS:
make test

ShvetsKS:
CEV: ea4e7a2f-4503-f171-9b93-8cdcd4b723a1

PR #381, opened 2020-02-17
Fix serialization and deserializattion in DF classification

ShvetsKS:
# Description
Fix serialization and deserializattion in DF classification


ShvetsKS:
make test

ShvetsKS:
make test

ShvetsKS:
make test

PR #387, opened 2020-02-18
Update macOS to 10.15 in public CI build

napetrov:
# Description
Update macOS version in public CI to 10.15

### DIFF
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -41,7 +41,7 @@ jobs:
 - job: macOS
   timeoutInMinutes: 0
   pool:
-    vmImage: 'macOS-10.14'
+    vmImage: 'macOS-10.15'
   strategy:
     matrix:
       mac32e:


algorithms/kernel/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i
```diff
@@ -91,8 +90,7 @@ public:
                     exp[i] = -f[i];
                     /* make all values less than threshold as threshold value
                     to fix slow work on vExp on large negative inputs */
-                    if (exp[i] < expThreshold)
-                        exp[i] = expThreshold;
+                    if (exp[i] < expThreshold) exp[i] = expThreshold;
```
outoftardis: current coding guidelines say that the statement that follows conditional statement should be on a separate line.
why clang formats your code in a way that contradicts the guidelines?

@averbukh: do we need to change this in coding guidelines or clang configs?

PR #391, opened 2020-02-18
move code from pipline to script

napetrov:
# Description
moving code from pipeline to scripts

- moving code from pipeline to scripts
- enable example run


napetrov:
make test

PR #411, opened 2020-02-25
Improve performance for columnar inputs for lasso and elastic net

agorshk:
Improve performance for columnar input data for lasso and elastic net

### DIFF
--- a/algorithms/kernel/objective_function/mse/mse_dense_default_batch_impl.i
+++ b/algorithms/kernel/objective_function/mse/mse_dense_default_batch_impl.i
@@ -104,10 +104,15 @@ inline services::Status MSEKernel<algorithmFPType, method, cpu>::compute(Numeric
                     }
                     else
                     {
-                        XPtr.set(dataNT, 0, nDataRows);
-                        DAAL_CHECK_BLOCK_STATUS(XPtr);
-                        X = const_cast<algorithmFPType*>(XPtr.get());
-                        transposedData = false;
+                        X = new algorithmFPType[nTheta*nDataRows];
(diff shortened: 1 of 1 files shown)


agorshk:
make test

algorithms/kernel/objective_function/mse/mse_dense_default_batch_impl.i
```diff
@@ -104,10 +104,15 @@ inline services::Status MSEKernel<algorithmFPType, method, cpu>::compute(Numeric
                     }
                     else
                     {
-                        XPtr.set(dataNT, 0, nDataRows);
-                        DAAL_CHECK_BLOCK_STATUS(XPtr);
-                        X = const_cast<algorithmFPType*>(XPtr.get());
-                        transposedData = false;
+                        X = new algorithmFPType[nTheta*nDataRows];
```
SmirnovEgorRu: Here a few concerns from my side: 
1) "new", but there are no "delete"
2) "new" is bad practice, better to use RAII classes - TArray. For example "new" doesnt guarantee mem alignment
3) We use n_samples x n_features additional memory in the worst case. It's not necessary.

Good practice across DAAL algorithms:
Use getBlockOfColumnValues / getBlockOfRows in threader for which do some algorithmic computations (additional memory is only for current block, faster because you keep data in cache) for MSE - it should be GEMM call
``` cpp
daal::threader_for(... [&] (...) {
  data_management::BlockDescriptor<algorithmFPType> bd;
  dataNT->getBlockOfRows(0, nDataRows, readOnly, bd);
  gemm(bd.getBlockPtr(), ...);
});
```

PR #412, opened 2020-02-25
[2020-u1-rls] Finiteness checker fix

Alexsandruss:
# Description

- Change of OR operator from bitwise to logical in all data values finiteness check
- Addition of lost memory allocation error check in summing

### DIFF
--- a/service/kernel/data_management/finiteness_checker.cpp
+++ b/service/kernel/data_management/finiteness_checker.cpp
@@ -39,6 +39,17 @@ const uint64_t doubleExpMask  = 0x7ff0000000000000uLL;
 const uint64_t doubleFracMask = 0x000fffffffffffffuLL;
 const uint64_t doubleZeroBits = 0x0000000000000000uLL;
 
+template <typename DataType>
+DataType getInf()
+{
+    DataType inf;
+    if (sizeof(DataType) == 4)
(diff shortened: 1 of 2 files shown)


Alexsandruss:
http://intel-ci.intel.com/ea57dd7a-4236-f1ba-a40a-8cdcd4b723a1

service/kernel/data_management/finiteness_checker.cpp
```diff
@@ -129,6 +140,7 @@ DataType computeSumAVX512Impl(size_t nDataPtrs, size_t nElementsPerPtr, const Da
 
     daal::services::internal::TArray<DataType, avx512> partialSumsArr(nTotalBlocks);
     DataType * pSums = partialSumsArr.get();
+    if (!pSums) return getInf<DataType>();
```
ShvetsKS: It`s not clear. Why we return inf in case of memory allocation error.

Alexsandruss: If infinity is returned as sum, finiteness checker goes to next step (finiteness check of all elements). Comment added.

ShvetsKS: seems status should be propagated here (by parameter for example )

PR #438, opened 2020-03-06
Fix for soa optimizations

KalyanovD:
Fix for SOA optimizations(https://github.com/intel/daal/pull/422) to be able to build examples with pedantic mode(https://github.com/intel/daal/pull/435).


KalyanovD:
make test

PR #443, opened 2020-03-10
Dev/okiselev/master licence

kiselik:
# Description
Files with licence and TTP were updated according to the last internal guidelines. 

- separate licence and oneDAL TPP 
- update list of components in oneDAL TPP, fix wording
- update info about TBB in IPP, MKL TPPs


THIRD-PARTY-PROGRAMS-MKL
```diff
@@ -259,51 +259,215 @@ This computer code material was prepared, in part, as an account of work sponsor
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.    
 
+-------------------------------------------------------------
 
+The following third party programs have their own third party programs. These additional third party program files are as follows: 
+    1. Intel� Threading Building Blocks (Intel� TBB): https://github.com/intel/tbb/blob/tbb_2020/third-party-programs.txt
```
outoftardis: Not sure, but I think Intel(R) is the correct way to introduce the registered mark. Applicable in other files/parts of the file as well.

THIRD-PARTY-PROGRAMS-IPP
```diff
@@ -0,0 +1,456 @@
+Intel® Integrated Performance Primitives (Intel® IPP) Third Party Programs File
```
napetrov: what is point of using capital letters in filenames? 
Do we have a guidelines here?

THIRD-PARTY-PROGRAMS-MKL
outoftardis: I've checked in our Style Guide. Here's a quote:
> Do not use superscript for the registered trademark symbol.

THIRD-PARTY-PROGRAMS-IPP
outoftardis: related question: why these files do not have a .txt extension? Is this because of the new guidelines?

napetrov:
Do we need this change? @kiselik

napetrov:
New request will be created to handle license update

PR #451, opened 2020-03-12
PCA Optimization for Column-Major Tables

owerbat:
Some optimizations for PCA have been done in Correlation

### DIFF
--- a/algorithms/kernel/covariance/covariance_impl.i
+++ b/algorithms/kernel/covariance/covariance_impl.i
@@ -143,7 +143,7 @@ inline size_t getBlockSize<avx512>(size_t nrows)
 
 /********************* updateDenseCrossProductAndSums ********************************************/
 template <typename algorithmFPType, Method method, CpuType cpu>
-services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatures, size_t nVectors, algorithmFPType * dataBlock,
+services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatures, size_t nVectors, NumericTable * dataTable,
                                                 algorithmFPType * crossProduct, algorithmFPType * sums, algorithmFPType * nObservations)
 {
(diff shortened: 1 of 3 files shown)


owerbat:
make test

algorithms/kernel/covariance/covariance_impl.i
```diff
@@ -276,8 +276,11 @@ services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatu
         default: break;
         }
 
+        DEFINE_TABLE_BLOCK(ReadRows, dataBlock, dataTable);
```
SmirnovEgorRu: So, here we are still using sequential conversion, right?

algorithms/kernel/covariance/covariance_impl.i
```diff
@@ -184,29 +185,28 @@ services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatu
             algorithmFPType alpha = 1.0;
             algorithmFPType beta  = 1.0;
 
-            size_t startRow = iBlock * numRowsInBlock;
-            size_t endRow   = startRow + numRowsInBlock;
-            if (endRow > nVectors)
-            {
-                endRow = nVectors;
-            }
+            size_t _nRows    = (iBlock < (numBlocks - 1)) ? numRowsInBlock : numRowsInLastBlock;
```
SmirnovEgorRu: naming of variables with _* prefix is used for private fields in classes only

algorithms/kernel/covariance/covariance_impl.i
SmirnovEgorRu: Does it affect PCA?

algorithms/kernel/covariance/covariance_impl.i
owerbat: fixed

PR #452, opened 2020-03-12
fix helper scripts

KalyanovD:
In case we try to install mklfpk at the same place for another system we receive message that mklfpk was already installed. To avoid that situation we may put each mklfpk in its own folder with name of system they are installed.

### DIFF
--- a/scripts/mklfpk.bat
+++ b/scripts/mklfpk.bat
@@ -28,27 +28,32 @@ set MKLGPUPACKAGE=mklgpufpk_win_%MKLGPUVERSION%
 set MKLURL=%MKLURLROOT%%MKLPACKAGE%.zip
 set MKLGPUURL=%MKLURLROOT%%MKLGPUPACKAGE%.zip
 if /i "%1"=="" (
-	set CPUDST=%~dp0..\externals\mklfpk
-	set GPUDST=%~dp0..\externals\mklgpufpk
+	set CPUCOND=%~dp0..\externals\mklfpk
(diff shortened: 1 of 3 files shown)


KalyanovD:
make test

KalyanovD:
make test

KalyanovD:
make test

napetrov:
In general - have no objections as far CI will be passing.  Think that this review should be done by @dr-pain

scripts/mklfpk.bat
```diff
@@ -48,7 +48,17 @@ set FILENAME=%~3
...
+if not exist "%DST%\win\" (
+	set is_exist=true
+	if not exist "%DST%\lib\" (
+		set is_exist=true
+	) else (
+		set is_exist=false
+	)
+)
+
+if "%is_exist%" == "true" (
```
napetrov: Will it make sense to flip true with false here?  
Variable name have opposite meaning

KalyanovD:
make test

scripts/mklfpk.bat
KalyanovD: After conversation with @dr-pain logic was changed. This variable was excluded.

napetrov:
for some reason public CI wasn't started. You can try to squash your commits and push as a single one

KalyanovD:
http://intel-ci.intel.com/ea677e82-0d95-f1d6-b5f2-8cdcd4b723a1

PR #454, opened 2020-03-13
PCA Transform Optimization for Column-Major Tables

owerbat:
Some optimizations for SOA numeric tables in PCA Transform have been done


owerbat:
make test

PR #459, opened 2020-03-16
fix inconsistency in brownboost and svm

PivovarA:


PivovarA:
make test

ShvetsKS:
make test

ShvetsKS:
Could you clarify why `Batch` class in `kdtree_knn_classification_training_batch.h` was not extended by similar constructor?

PivovarA:
make test

PivovarA:
make test

PivovarA:
> Could you clarify why `Batch` class in `kdtree_knn_classification_training_batch.h` was not extended by similar constructor?

In general, I can add such a constructor, but kdtree-knn counts the number of classes within itself. Therefore, adding this constructor will not change anything.
Adding nClasses parameter specifically to kdtree-knnis is a bit more difficult process and will require more detailed consideration.

PR #461, opened 2020-03-17
[releases] fix helper scripts

KalyanovD:
In case we try to install mklfpk at the same place for another system we receive message that mklfpk was already installed. To avoid that situation we may put each mklfpk in its own folder with name of system they are installed.

Cherry-pick of https://github.com/intel/daal/pull/452


KalyanovD:
make test

PR #465, opened 2020-03-18
[master] PCA Transform Bug Fix

owerbat:
A bug which was a reason of PCA Transform degradation with SOA tables has been fixed


owerbat:
make test

PR #473, opened 2020-03-20
SVM Predict Optimization for Column-Major Tables

owerbat:
Some optimizations for SVM Predict have been done in Kernel Function

### DIFF
--- a/algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
+++ b/algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
@@ -120,8 +120,8 @@ services::Status KernelImplLinear<defaultDense, algorithmFPType, cpu>::computeIn
 {
     SafeStatus safeStat;
 
-    const char trans   = 'T';
-    const char notrans = 'N';
+    char trans   = 'T';
+    char notrans = 'N';
 
     const size_t nFeatures = a1->getNumberOfColumns();
     const size_t nVectors1 = a1->getNumberOfRows();
@@ -134,21 +134,29 @@ services::Status KernelImplLinear<defaultDense, algorithmFPType, cpu>::computeIn
 
     if (a1 != a2)
     {
-        ReadRows<algorithmFPType, cpu> mtA1(*const_cast<NumericTable *>(a1), 0, nVectors1);
-        DAAL_CHECK_BLOCK_STATUS(mtA1);
-        const algorithmFPType * dataA1 = const_cast<algorithmFPType *>(mtA1.get());
-
         ReadRows<algorithmFPType, cpu> mtA2(*const_cast<NumericTable *>(a2), 0, nVectors2);
         DAAL_CHECK_BLOCK_STATUS(mtA2);
         const algorithmFPType * dataA2 = const_cast<algorithmFPType *>(mtA2.get());
 
-        WriteOnlyRows<algorithmFPType, cpu> mtR(r, 0, nVectors1);
-        DAAL_CHECK_BLOCK_STATUS(mtR);
-        algorithmFPType * dataR = mtR.get();
+        const size_t blockSize = 256;
+        const size_t blockNum  = nVectors1 / blockSize + !!(nVectors1 % blockSize);
 
-        Blas<algorithmFPType, cpu>::xgemm(&trans, &notrans, (DAAL_INT *)&nVectors2, (DAAL_INT *)&nVectors1, (DAAL_INT *)&nFeatures, &alpha,
-                                          (algorithmFPType *)dataA2, (DAAL_INT *)&nFeatures, (algorithmFPType *)dataA1, (DAAL_INT *)&nFeatures, &beta,
-                                          dataR, (DAAL_INT *)&nVectors2);
+        daal::threader_for(blockNum, blockNum, [&](const size_t iBlock) {
+            const size_t startRow       = iBlock * blockSize;
+            const size_t numRowsInBlock = (iBlock != blockNum - 1) ? blockSize : nVectors1 - iBlock * blockSize;
+
+            ReadRows<algorithmFPType, cpu> mtA1(*const_cast<NumericTable *>(a1), startRow, numRowsInBlock);
+            DAAL_CHECK_BLOCK_STATUS_THR(mtA1);
+            const algorithmFPType * blockPtr = const_cast<algorithmFPType *>(mtA1.get());
+
+            WriteOnlyRows<algorithmFPType, cpu> mtR(r, startRow, numRowsInBlock);
+            DAAL_CHECK_BLOCK_STATUS_THR(mtR);
+            algorithmFPType * dataR = mtR.get();
+
+            Blas<algorithmFPType, cpu>::xxgemm(&trans, &notrans, (DAAL_INT *)&nVectors2, (DAAL_INT *)&numRowsInBlock, (DAAL_INT *)&nFeatures, &alpha,
+                                               (algorithmFPType *)dataA2, (DAAL_INT *)&nFeatures, (algorithmFPType *)blockPtr, (DAAL_INT *)&nFeatures,
+                                               &beta, dataR, (DAAL_INT *)&nVectors2);
+        });
     }
     else
     {
--- a/algorithms/kernel/kernel_function/kernel_function_rbf_dense_default_impl.i
+++ b/algorithms/kernel/kernel_function/kernel_function_rbf_dense_default_impl.i
@@ -138,15 +138,6 @@ services::Status KernelImplRBF<defaultDense, algorithmFPType, cpu>::computeInter
         char trans = 'T', notrans = 'N';
         algorithmFPType zero = 0.0, negTwo = -2.0;
 
-        services::Status retStat = Blas<algorithmFPType, cpu>::xgemm_blocked(&trans, &notrans, (DAAL_INT *)&nVectors2, (DAAL_INT *)&nVectors1,
-                                                                             (DAAL_INT *)&nFeatures, &negTwo, a2, (DAAL_INT *)&nFeatures, a1,
-                                                                             (DAAL_INT *)&nFeatures, &zero, r, (DAAL_INT *)&nVectors2);
-        if (!retStat) return retStat;
-
-        WriteRows<algorithmFPType, cpu> mtR(r, 0, nVectors1);
-        DAAL_CHECK_BLOCK_STATUS(mtR);
-        algorithmFPType * dataR = mtR.get();
-
         DAAL_OVERFLOW_CHECK_BY_ADDING(size_t, nVectors1, nVectors2);
         DAAL_OVERFLOW_CHECK_BY_MULTIPLICATION(size_t, nVectors1 + nVectors2, sizeof(algorithmFPType));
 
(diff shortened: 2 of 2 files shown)


owerbat:
make test

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
```diff
@@ -134,21 +134,29 @@ services::Status KernelImplLinear<defaultDense, algorithmFPType, cpu>::computeIn
...
         DAAL_CHECK_BLOCK_STATUS(mtA2);
         const algorithmFPType * dataA2 = const_cast<algorithmFPType *>(mtA2.get());
 
-        WriteOnlyRows<algorithmFPType, cpu> mtR(r, 0, nVectors1);
-        DAAL_CHECK_BLOCK_STATUS(mtR);
-        algorithmFPType * dataR = mtR.get();
+        const size_t blockSize = 256;
+        const size_t blockNum  = nVectors1 / blockSize + !!(nVectors1 % blockSize);
+
+        daal::threader_for(blockNum, blockNum, [&](const size_t iBlock) {
```
SmirnovEgorRu: Do you see perf difference with xgemm_blocked method?

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
```diff
@@ -134,21 +134,29 @@ services::Status KernelImplLinear<defaultDense, algorithmFPType, cpu>::computeIn
...
-        Blas<algorithmFPType, cpu>::xgemm(&trans, &notrans, (DAAL_INT *)&nVectors2, (DAAL_INT *)&nVectors1, (DAAL_INT *)&nFeatures, &alpha,
-                                          (algorithmFPType *)dataA2, (DAAL_INT *)&nFeatures, (algorithmFPType *)dataA1, (DAAL_INT *)&nFeatures, &beta,
-                                          dataR, (DAAL_INT *)&nVectors2);
+            Blas<algorithmFPType, cpu>::xxgemm(&trans, &notrans, (DAAL_INT *)&nVectors2, (DAAL_INT *)&numRowsInBlock, (DAAL_INT *)&nFeatures, &alpha,
+                                               (algorithmFPType *)dataA2, (DAAL_INT *)&nFeatures, (algorithmFPType *)blockPtr, (DAAL_INT *)&nFeatures,
+                                               &beta, dataR, (DAAL_INT *)&nVectors2);
+        });
     }
     else
     {
```
SmirnovEgorRu: Do we see slow conversion in this case?

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
```diff
@@ -120,8 +120,8 @@ services::Status KernelImplLinear<defaultDense, algorithmFPType, cpu>::computeIn
 {
     SafeStatus safeStat;
 
-    const char trans   = 'T';
-    const char notrans = 'N';
+    char trans   = 'T';
```
SmirnovEgorRu: Better to revert const :)

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
```diff
@@ -120,8 +120,8 @@ services::Status KernelImplLinear<defaultDense, algorithmFPType, cpu>::computeIn
 {
```
SmirnovEgorRu: Overall comment - you optimized linear kernel only. But DAAL has RBF kernel as well.
Do we see the regression in SVM with RBF kernel?

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
owerbat: No, I don't see it because parallel reading is using in blocked_gemm. But in case of our homogen/soa metric threader_for+gemm gets better result (~10%) than blocked_gemm

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
owerbat: threader_for+gemm performs a little bit better (~5%) than xgemm_blocked. So I think I shouldn't use xgemm_blocked at all

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
owerbat: Yes, I see a place in RBF kernel where the regression may be. So I'm going to optimize it too

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
SmirnovEgorRu: xgemm_blocked - it's the same as threader_for + gemm :)

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
SmirnovEgorRu: Okay, let's do it

owerbat:
make test

algorithms/kernel/kernel_function/kernel_function_rbf_dense_default_impl.i
```diff
@@ -174,52 +159,92 @@ services::Status KernelImplRBF<defaultDense, algorithmFPType, cpu>::computeInter
...
+        });
     }
     else
     {
-        char uplo, trans;
+        char uplo, trans, notrans;
         algorithmFPType zero = 0.0, one = 1.0, two = 2.0;
-        uplo  = 'U';
-        trans = 'T';
-        if (isInParallel)
```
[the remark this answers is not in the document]
SmirnovEgorRu: Please, clarify this moment with @Vika-F

algorithms/kernel/kernel_function/kernel_function_linear_dense_default_impl.i
Vika-F: Agree with Egor's comment about xgemm_blocked. Also, you can pass block sizes for matrices a1 and a2 into xgemm_blocked as parameters. And 5% performance difference can be a deviation in measurements.

algorithms/kernel/kernel_function/kernel_function_rbf_dense_default_impl.i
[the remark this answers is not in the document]
Vika-F: Removing isInParallel should be Ok.
This condition was added due to a bug in multiclass classifier. That bug caused failures when an underlying 2-class classifier run in parallel. The bug is fixed already.

algorithms/kernel/kernel_function/kernel_function_rbf_dense_default_impl.i
```diff
@@ -124,46 +124,28 @@ template <typename algorithmFPType, CpuType cpu>
...
-        {
-            Blas<algorithmFPType, cpu>::xgemm(&trans, &notrans, (DAAL_INT *)&nVectors2, (DAAL_INT *)&nVectors1, (DAAL_INT *)&nFeatures, &negTwo,
-                                              (algorithmFPType *)dataA2, (DAAL_INT *)&nFeatures, (algorithmFPType *)dataA1, (DAAL_INT *)&nFeatures,
-                                              &zero, dataR, (DAAL_INT *)&nVectors2);
-        }
+
+        services::Status retStat = Blas<algorithmFPType, cpu>::xgemm_blocked(&trans, &notrans, (DAAL_INT *)&nVectors2, (DAAL_INT *)&nVectors1,
+                                                                             (DAAL_INT *)&nFeatures, &negTwo, a2, (DAAL_INT *)&nFeatures, a1,
+                                                                             (DAAL_INT *)&nFeatures, &zero, r, (DAAL_INT *)&nVectors2);
+        if (!retStat) return retStat;
```
Vika-F: In case of RBF kernel, use of xgemm_blocked is not the best option.
In your variant of code you have additional passes through a1 and a2 to compute sqrDataA1 and sqrDataA2.
In my opinion, it is better to embed those computations into threader_for which does matrix multiplication a1 x a2 in cache-friendly manner.

algorithms/kernel/kernel_function/kernel_function_rbf_dense_default_impl.i
```diff
@@ -174,72 +156,142 @@ services::Status KernelImplRBF<defaultDense, algorithmFPType, cpu>::computeInter
...
-        else
-        {
-            Blas<algorithmFPType, cpu>::xsyrk(&uplo, &trans, (DAAL_INT *)&nVectors1, (DAAL_INT *)&nFeatures, &one, (algorithmFPType *)dataA1,
-                                              (DAAL_INT *)&nFeatures, &zero, dataR, (DAAL_INT *)&nVectors1);
-        }
-        for (size_t i = 0; i < nVectors1; i++)
-        {
-            for (size_t k = 0; k < i; k++)
+
+        services::Status retStat =
```
Vika-F: Here you replaced xsyrk with xgemm for the case a1 == a2.
xgemm does 2X more computations than xsyrk because it computes full resulting matrix. In case a1 == a2, the result is a symmetric matrix. That means only half (i.e. upper or lower triangle) of the matrix can be computed, as it is done with xsyrk.

algorithms/kernel/kernel_function/kernel_function_rbf_dense_default_impl.i
```diff
@@ -174,72 +156,142 @@ services::Status KernelImplRBF<defaultDense, algorithmFPType, cpu>::computeInter
...
+            for (size_t i = startRow; i < finishRow; i++)
             {
-                dataR[i * nVectors1 + k] = dataR[k * nVectors1 + i];
+                dataR[i * nVectors1 + i] = zero;
+                daal::internal::Math<algorithmFPType, cpu>::vExp(i + 1, dataR + i * nVectors1, dataR + i * nVectors1);
             }
-        }
+        });
+
+        daal::threader_for(blockNum, blockNum, [&](const size_t iBlock) {
```
Vika-F: This and the previous threader_for can be merged together to have more cache-friendly implementation.

owerbat: I think it's not possible because all the computations must be finished before the last threader_for begins. Otherwise an element of the matrix may be copied before it's been computed.
Or if you meant that I should copy elements and run `vExp(...)` after, then we get 2 times more computations. So I think it's not the best idea

algorithms/kernel/kernel_function/kernel_function_rbf_dense_default_impl.i
owerbat: I modified blocked gemm. So now gemm and sqrDataA1 calculated in the same threader_for. But I couldn't do the same for sqrDataA2 because of parallelism

algorithms/kernel/kernel_function/kernel_function_rbf_dense_default_impl.i
owerbat: I modified blocked gemm. So now only lower triangle of the matrix is computed

PetrovKP:
General comment. If I will call of Kernel Function in parallel block case. Will a bunch of paralleled regions be worked out correctly? In the case of a predict SVM, I would parallel the external function `SVMPredictImpl`, at the same time, the problem with memory allocation would disappear and it would be several times faster.

PR #455, opened 2020-03-13
Add ability to train linear and ridge regression with n_observations …

Vika-F:
The changes aim to pass arguments properly to LAPACK XORMRQ function in case n_observations < n_betas.
The results of XORMRQ are appended: R - with unit diagonal sub-matrix; QtY - with zero rows; to make the resulting linear system solvable.

PR #475, opened 2020-03-23
[releases] Add ability to train linear and ridge regression with n_observations < n_beta in QR method

Vika-F:
Same as, but to releases: #455

The changes aim to pass arguments properly to LAPACK XORMRQ function in case n_observations < n_betas.
The results of XORMRQ are appended: R - with unit diagonal sub-matrix; QtY - with zero rows; to make the resulting linear system solvable.

CI (successful): http://intel-ci.intel.com/ea6cecf3-a379-f1a9-8625-8cdcd4b723a1

### DIFF
--- a/algorithms/kernel/linear_model/linear_model_train_qr_common_impl.i
+++ b/algorithms/kernel/linear_model/linear_model_train_qr_common_impl.i
@@ -78,29 +78,60 @@ Status CommonKernel<algorithmFPType, cpu>::computeQRForBlock(DAAL_INT p, DAAL_IN
     DAAL_CHECK(info == 0, services::ErrorLinearRegressionInternal);
 
     /* Copy result into matrix R */
-    const DAAL_INT rOffset       = (n - p) * p;
-    const algorithmFPType * xPtr = x + rOffset;
-    for (size_t i = 0; i < p; i++)
+    const DAAL_INT xOffset             = (n > p ? (n - p) * p : 0);
+    const DAAL_INT rOffset             = (n > p ? 0 : (p - n) * p);
+    const DAAL_INT nRowsInR            = daal::services::internal::min<cpu, DAAL_INT>(p, n);
+    const DAAL_INT jOffset             = (n > p ? 0 : p - n);
+    const algorithmFPType * const xPtr = x + xOffset;
+    algorithmFPType * const rPtr       = r + rOffset;
+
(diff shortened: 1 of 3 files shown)


algorithms/kernel/linear_regression/linear_regression_training_input.cpp
```diff
@@ -88,7 +88,14 @@ services::Status Input::check(const daal::algorithms::Parameter * par, int metho
     NumericTablePtr dataTable = get(data);
     size_t nRowsInData        = dataTable->getNumberOfRows();
     size_t nColumnsInData     = dataTable->getNumberOfColumns();
-    DAAL_CHECK(nRowsInData >= nColumnsInData + (int)(method == qrDense && parameter->interceptFlag == true), ErrorIncorrectNumberOfRows);
+    if (method == normEqDense)
+    {
+        DAAL_CHECK(nRowsInData >= nColumnsInData + (int)(parameter->interceptFlag == true), ErrorIncorrectNumberOfRows);
```
ShvetsKS: nRowsInData == 0 and nColumnsInData == 0 is valid case for this check, if intercept was disabled

algorithms/kernel/linear_model/linear_model_train_qr_common_impl.i
```diff
@@ -78,29 +78,60 @@ Status CommonKernel<algorithmFPType, cpu>::computeQRForBlock(DAAL_INT p, DAAL_IN
...
+    const DAAL_INT jOffset             = (n > p ? 0 : p - n);
+    const algorithmFPType * const xPtr = x + xOffset;
+    algorithmFPType * const rPtr       = r + rOffset;
+
+    for (size_t i = 0; i < nRowsInR; ++i)
     {
         PRAGMA_IVDEP
         PRAGMA_VECTOR_ALWAYS
-        for (size_t j = 0; j <= i; j++)
+        for (size_t j = 0; j <= i + jOffset; ++j)
```
ShvetsKS: why not use daal_memcpy_s instead?

algorithms/kernel/linear_regression/linear_regression_training_input.cpp
Vika-F: The case of nRowsInData == 0 or nColumnsInData == 0 is checked in regression::training::Input::check, which is called on line #84 in this check.

PR #484, opened 2020-03-26
Fix for threading init.

KalyanovD:
This PR fixes case when we use internal threading before using getInstance()->setNumberOfThreads(num_threads) function to change number of using threads.

### DIFF
--- a/algorithms/threading/threading.cpp
+++ b/algorithms/threading/threading.cpp
@@ -26,12 +26,10 @@
 
 #if defined(__DO_TBB_LAYER__)
     #include <stdlib.h> // malloc and free
-    #include <tbb/tbb.h>
     #ifndef TBB_PREVIEW_GLOBAL_CONTROL
         #define TBB_PREVIEW_GLOBAL_CONTROL 1
     #endif
-    #include <tbb/global_control.h>
-    #include <tbb/task_arena.h>
+    #include <tbb/tbb.h>
     #include <tbb/spin_mutex.h>
     #include "tbb/scalable_allocator.h"
 #else
@@ -61,7 +59,7 @@ DAAL_EXPORT void _daal_tbb_task_scheduler_free(void *& init)
 #if defined(__DO_TBB_LAYER__)
     if (init)
     {
-        delete (tbb::global_control *)init;
+        delete static_cast<tbb::global_control *>(init);
         init = nullptr;
     }
 #endif
@@ -75,7 +73,7 @@ DAAL_EXPORT size_t _setNumberOfThreads(const size_t numThreads, void ** init)
     if (numThreads != 0)
     {
         _daal_tbb_task_scheduler_free(*init);
-        *init = (void *)(new tbb::global_control(tbb::global_control::max_allowed_parallelism, numThreads));
+        *init = static_cast<void *>(new tbb::global_control(tbb::global_control::max_allowed_parallelism, numThreads));
         daal::threader_env()->setNumberOfThreads(numThreads);
         return numThreads;
     }
--- a/algorithms/threading/threading.cpp
+++ b/algorithms/threading/threading.cpp
@@ -27,6 +27,11 @@
 #if defined(__DO_TBB_LAYER__)
     #include <stdlib.h> // malloc and free
     #include <tbb/tbb.h>
+    #ifndef TBB_PREVIEW_GLOBAL_CONTROL
+        #define TBB_PREVIEW_GLOBAL_CONTROL 1
+    #endif
+    #include <tbb/global_control.h>
+    #include <tbb/task_arena.h>
     #include <tbb/spin_mutex.h>
     #include "tbb/scalable_allocator.h"
 #else
@@ -56,7 +61,7 @@ DAAL_EXPORT void _daal_tbb_task_scheduler_free(void *& init)
 #if defined(__DO_TBB_LAYER__)
     if (init)
     {
-        delete (tbb::task_scheduler_init *)init;
+        delete (tbb::global_control *)init;
         init = nullptr;
     }
 #endif
@@ -70,7 +75,7 @@ DAAL_EXPORT size_t _setNumberOfThreads(const size_t numThreads, void ** init)
     if (numThreads != 0)
     {
         _daal_tbb_task_scheduler_free(*init);
-        *init = (void *)(new tbb::task_scheduler_init(numThreads));
+        *init = (void *)(new tbb::global_control(tbb::global_control::max_allowed_parallelism, numThreads));
         daal::threader_env()->setNumberOfThreads(numThreads);
         return numThreads;
     }
@@ -130,7 +135,7 @@ DAAL_EXPORT void _daal_threader_for_optional(int n, int threads_request, const v
 DAAL_EXPORT int _daal_threader_get_max_threads()
 {
 #if defined(__DO_TBB_LAYER__)
-    return tbb::task_scheduler_init::default_num_threads();
+    return tbb::this_task_arena::max_concurrency();
 #elif defined(__DO_SEQ_LAYER__)
     return 1;
 #endif
--- a/service/kernel/env_detect.cpp
+++ b/service/kernel/env_detect.cpp
@@ -148,7 +148,7 @@ DAAL_EXPORT void daal::services::Environment::initNumberOfThreads()
         {
             daal::services::Environment::setNumberOfThreads(ncores);
         }
-    } 
+    }
     else
     {
         daal::services::Environment::setNumberOfThreads(_daal_threader_get_max_threads());
--- a/service/kernel/env_detect.cpp
+++ b/service/kernel/env_detect.cpp
@@ -137,17 +137,14 @@ DAAL_EXPORT void daal::services::Environment::initNumberOfThreads()
 {
     if (isInit) return;
 
-    /* if HT enabled - set _numThreads to physical cores num */
-    if (daal::internal::Service<>::serv_get_ht())
-    {
-        /* Number of cores = number of cpu packages * number of cores per cpu package */
-        int ncores = daal::internal::Service<>::serv_get_ncpus() * daal::internal::Service<>::serv_get_ncorespercpu();
+    /* if HT enabled - set number of threads to physical cores num */
+    /* Number of cores = number of cpu packages * number of cores per cpu package */
+    int ncores = daal::internal::Service<>::serv_get_ncpus() * daal::internal::Service<>::serv_get_ncorespercpu();
 
-        /*  Re-set number of threads if ncores is valid and different to _numThreads */
-        if ((ncores > 0) && (ncores < _daal_threader_get_max_threads()))
-        {
-            daal::services::Environment::setNumberOfThreads(ncores);
-        }
+    /*  Re-set number of threads if ncores is valid */
+    if ((ncores > 0) && (ncores <= _daal_threader_get_max_threads()))
+    {
+        daal::services::Environment::setNumberOfThreads(ncores);
     }
     isInit = true;
 }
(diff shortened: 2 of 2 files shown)


dr-pain:
This change is breaking current logic - to set Number of threads only if HT enabled.
if HT disabled TBB should manage max thread  numbers.

SmirnovEgorRu:
@dr-pain,
TBB uses max threads by default, so the logic is the same as before, except that we create task scheduler on our side explicitly.

dr-pain:
We create task scheduler with thread num which calculated based on cores num.
in case if we use TBB for it, It may reduce number of threads for daal because other app which use tbb also is run to avoid thread oversubscription.

KalyanovD:
We can't manage number of threads from our side if task_scheduler_init was created on TBB side, that's why our functionality for changing number of threads isn't working for now in case described above. This PR targets to fix it. If I understand correctly, we can use tbb::task_scheduler_init::default_num_threads() to cover case described by you.

dr-pain:
@iburyl promised to do review tomorrow.

iburyl:
**First thing.**
Current implementation of setNumOfThreads thing is doing it via creating and deleting of own task_scheduler_init class. It was never TBB best practice way to do that. The right way to do that was to init TBB scheduler once and then control the number of threads used via arenas, but it was considered too intrusive.

**Second thing.**
https://software.intel.com/en-us/node/506296

> task_scheduler_init Class
> CAUTION
> This feature is deprecated and will be removed in the future.

And I can tell you, that it is already removed in oneTBB mainline.
It is time to look at alternative approach.

averbukh:
> We can't manage number of threads from our side if task_scheduler_init was created on TBB side, that's why our functionality for changing number of threads isn't working for now in case described above. This PR targets to fix it. If I understand correctly, we can use tbb::task_scheduler_init::default_num_threads() to cover case described by you.

No.

averbukh:
Our product is the library. We have no "TBB + DAL" case. It's always "Application + DAL + TBB". Application. So you should not assume that Application didn't reduced number of threads for TBB.

service/kernel/env_detect.cpp
```diff
@@ -137,18 +137,22 @@ DAAL_EXPORT void daal::services::Environment::initNumberOfThreads()
...
     if (daal::internal::Service<>::serv_get_ht())
     {
         /* Number of cores = number of cpu packages * number of cores per cpu package */
-        int ncores = daal::internal::Service<>::serv_get_ncpus() * daal::internal::Service<>::serv_get_ncorespercpu();
+        const size_t ncores = daal::internal::Service<>::serv_get_ncpus() * daal::internal::Service<>::serv_get_ncorespercpu();
 
-        /*  Re-set number of threads if ncores is valid and different to _numThreads */
-        if ((ncores > 0) && (ncores < _daal_threader_get_max_threads()))
+        /*  Re-set number of threads if ncores is valid */
+        if ((ncores > 0) && (ncores <= _daal_threader_get_max_threads()))
```
averbukh: Your modification looks wrong. What are reasons to set it, if ncores == _daal_threader_get_max_threads() ?

service/kernel/env_detect.cpp
```diff
@@ -137,18 +137,22 @@ DAAL_EXPORT void daal::services::Environment::initNumberOfThreads()
...
         {
             daal::services::Environment::setNumberOfThreads(ncores);
         }
     }
+    else
+    {
+        daal::services::Environment::setNumberOfThreads(_daal_threader_get_max_threads());
+    }
```
averbukh: What are reasons for that?

averbukh:
Moving from deprecated task_scheduler_init to global_control looks like a right way to fix the original issue, because it allows to get current value of the parameter (instead of wrong logic to override application settings by maximum value).

iburyl:
TBB logic with global_control is the following - among all the alive global_control objects with some value, the minimum will be taken.

Thus one would see the following behavior with existing thread setting logic in oneDAL:
```cpp
// assuming 4 cores and 8 ht

tbb::parallel_for(blocked_range<size_t>(0,n), [](auto r){/*do user work*/}); // 8 threads

// first oneDAL usage here
training::Batch<> algorithm;
algorithm.compute();

// from now on, user is not able to set TBB threads to larger values, than in oneDAL
// and oneDAL defaulted to 4
tbb::parallel_for(blocked_range<size_t>(0,n), [](auto r){/*do user work*/}); // 4 threads

{
  tbb::global_control(max_allowed_parallelism, 8);
  tbb::parallel_for(blocked_range<size_t>(0,n), [](auto r){/*do user work*/}); // still 4 threads

  tbb::global_control(max_allowed_parallelism, 2);
  tbb::parallel_for(blocked_range<size_t>(0,n), [](auto r){/*do user work*/}); // that is fine - 2 threads
}
```
It is not obvious and, if this limitation is chosen to stay, should be somewhere documented.

With task_scheduler_init I think it was even less obvious - the first who managed to initialize TBB wins, and the second guy has no control (as it was happening in the issue being fixed, I guess).

averbukh:
@KalyanovD, do you have any progress for this PR?

KalyanovD:
@averbukh, hello! Sorry for delay in answer. I was switched to other activity for these days and I'm going to return to this work today. I plan to use global_control as you recommended before and check if it really help to solve original issue.

averbukh:
> @averbukh, hello! Sorry for delay in answer. I was switched to other activity for these days and I'm going to return to this work today. I plan to use global_control as you recommended before and check if it really help to solve original issue.

@KalyanovD, do you have any progress since you returned to this work?

KalyanovD:
Hello @averbukh,
I've changed task_scheduler_init to global_control and checked if the original issue was resolved by this change. But with deprecating task_scheduler_init function tbb::task_scheduler_init::default_num_threads() was removed too. Please, review if my using of tbb::this_task_arena::max_concurrency() instead right or wrong?

service/kernel/env_detect.cpp
KalyanovD: Original logic was restored

service/kernel/env_detect.cpp
KalyanovD: Original logic was restored

service/kernel/env_detect.cpp
```diff
@@ -144,15 +144,11 @@ DAAL_EXPORT void daal::services::Environment::initNumberOfThreads()
...
+        if ((ncores > 0) && (ncores < _daal_threader_get_max_threads()))
         {
             daal::services::Environment::setNumberOfThreads(ncores);
         }
     }
-    else
-    {
-        daal::services::Environment::setNumberOfThreads(_daal_threader_get_max_threads());
-    }
     isInit = true;
```
SmirnovEgorRu: Does it solve the original problem?

KalyanovD: Actually, solution is the using global_control instead of task_scheduler_init for setting number of threads. This diff between commits is just shows that previous changes were reverted =)

algorithms/threading/threading.cpp
```diff
@@ -56,7 +61,7 @@ DAAL_EXPORT void _daal_tbb_task_scheduler_free(void *& init)
 #if defined(__DO_TBB_LAYER__)
     if (init)
     {
-        delete (tbb::task_scheduler_init *)init;
+        delete (tbb::global_control *)init;
```
averbukh, line 64: Use C++ casting instead.

algorithms/threading/threading.cpp
```diff
@@ -70,7 +75,7 @@ DAAL_EXPORT size_t _setNumberOfThreads(const size_t numThreads, void ** init)
     if (numThreads != 0)
     {
         _daal_tbb_task_scheduler_free(*init);
-        *init = (void *)(new tbb::task_scheduler_init(numThreads));
+        *init = (void *)(new tbb::global_control(tbb::global_control::max_allowed_parallelism, numThreads));
```
averbukh, line 78: Use C++ casting instead.

algorithms/threading/threading.cpp
```diff
@@ -130,7 +135,7 @@ DAAL_EXPORT void _daal_threader_for_optional(int n, int threads_request, const v
 DAAL_EXPORT int _daal_threader_get_max_threads()
 {
 #if defined(__DO_TBB_LAYER__)
-    return tbb::task_scheduler_init::default_num_threads();
+    return tbb::this_task_arena::max_concurrency();
```
averbukh, line 138: The behavior is modified. It looks wrong.

service/kernel/env_detect.cpp
```diff
@@ -137,13 +137,13 @@ DAAL_EXPORT void daal::services::Environment::initNumberOfThreads()
 {
     if (isInit) return;
 
-    /* if HT enabled - set _numThreads to physical cores num */
+    /* if HT enabled - set number of threads to physical cores num */
     if (daal::internal::Service<>::serv_get_ht())
     {
         /* Number of cores = number of cpu packages * number of cores per cpu package */
-        int ncores = daal::internal::Service<>::serv_get_ncpus() * daal::internal::Service<>::serv_get_ncorespercpu();
+        const size_t ncores = daal::internal::Service<>::serv_get_ncpus() * daal::internal::Service<>::serv_get_ncorespercpu();
```
averbukh, line 144: SDL violation.

averbukh:
@KalyanovD, do you have any progress for this PR? After discussion with TBB team, I have some additional comments for you code, and not sure, is it makes sense to discuss here, if even previous comments are not addressed.

algorithms/threading/threading.cpp
KalyanovD: Thanks! Casting was changed.

algorithms/threading/threading.cpp
KalyanovD: Thanks! Casting was changed.

service/kernel/env_detect.cpp
KalyanovD: This change was reverted and added casting to size_t for ncores in setNumberOfThreads() call

algorithms/threading/threading.cpp
```diff
@@ -56,7 +59,7 @@ DAAL_EXPORT void _daal_tbb_task_scheduler_free(void *& init)
 #if defined(__DO_TBB_LAYER__)
     if (init)
     {
-        delete (tbb::task_scheduler_init *)init;
+        delete static_cast<tbb::global_control *>(init);
```
averbukh, line 62: init is wrong name for the parameter after your changes

algorithms/threading/threading.cpp
```diff
@@ -70,7 +73,7 @@ DAAL_EXPORT size_t _setNumberOfThreads(const size_t numThreads, void ** init)
     if (numThreads != 0)
     {
         _daal_tbb_task_scheduler_free(*init);
-        *init = (void *)(new tbb::task_scheduler_init(numThreads));
+        *init = static_cast<void *>(new tbb::global_control(tbb::global_control::max_allowed_parallelism, numThreads));
```
averbukh, line 76: init is wrong name for the parameter after your changes. It also applicable for corresponding field name

service/kernel/env_detect.cpp
```diff
@@ -137,16 +137,16 @@ DAAL_EXPORT void daal::services::Environment::initNumberOfThreads()
...
     {
         /* Number of cores = number of cpu packages * number of cores per cpu package */
         int ncores = daal::internal::Service<>::serv_get_ncpus() * daal::internal::Service<>::serv_get_ncorespercpu();
 
-        /*  Re-set number of threads if ncores is valid and different to _numThreads */
+        /*  Re-set number of threads if ncores is valid */
         if ((ncores > 0) && (ncores < _daal_threader_get_max_threads()))
         {
-            daal::services::Environment::setNumberOfThreads(ncores);
+            daal::services::Environment::setNumberOfThreads(static_cast<size_t>(ncores));
```
averbukh, line 149: Double-check if your casting violates SDL

KalyanovD:
As I see this PR may be closed because all necessary changes were included in https://github.com/oneapi-src/oneDAL/pull/559

PR #488, opened 2020-03-27
[releases] fix inconsistency in brownboost and svm (#459)

PivovarA:
* fix inconsistency in brownboost and svm

* change code according to clang-format

* add description to new svm and brownboost constructors


PivovarA:
make test

PR #481, opened 2020-03-25
Added GPU backend for DBSCAN

amgrigoriev:

PR #496, opened 2020-03-31
Added GPU backend for DBSCAN (#481)

amgrigoriev:
-
-

### DIFF
--- a/examples/cpp_sycl/makefile_lnx
+++ b/examples/cpp_sycl/makefile_lnx
@@ -116,7 +116,7 @@ sointel64:
 _make_ex: $(RES)
 
 vpath
-vpath %.cpp $(addprefix ./source/,datasource covariance linear_regression logistic_regression optimization_solvers pca pca_transform kmeans moments k_nearest_neighbors gradient_boosted_trees)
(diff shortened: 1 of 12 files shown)


examples/cpp_sycl/makefile_lnx
```diff
@@ -116,7 +116,7 @@ sointel64:
 _make_ex: $(RES)
 
 vpath
-vpath %.cpp $(addprefix ./source/,datasource covariance linear_regression logistic_regression optimization_solvers pca pca_transform kmeans moments k_nearest_neighbors gradient_boosted_trees)
+vpath %.cpp $(addprefix ./source/,datasource covariance linear_regression logistic_regression optimization_solvers pca pca_transform kmeans moments k_nearest_neighbors gradient_boosted_trees dbscan)
```
michael-smirnov, line 119: Don't we need to have this list sorted alphabetically?

amgrigoriev: Almost every method is added at the end for now. All are in historical order, I guess

michael-smirnov: No, at the beginning all methods were sorted: problems started from `k_nearest_neighbors`. It is not critical but should be fixed later for both master and releases

PR #499, opened 2020-04-01
Remove C++11 features from oneapi headers exposed in classical DAAL

rlnx:
#### Problem statement
oneAPI header files are included to some classical DAAL headers, namely `env_detect.h`, because  `Environment` class depends on `ExecutionContext` that defined in `oneapi/execution_context.hpp`, where C++11 features are used.

#### Solution
C++11 features are removed from `oneapi/execution_context.hpp` and other oneAPI included header files.

### DIFF
--- a/examples/cpp/makefile_lnx
+++ b/examples/cpp/makefile_lnx
@@ -97,7 +97,12 @@ endif
 
 DAAL_LIB := $(DAAL_PATH)/libdaal_core.$(RES_EXT) $(DAAL_LIB_T)
 
-COPTS := -std=c++11 -pedantic -Wall -Wextra -Werror=return-type -Werror=uninitialized -Werror=unknown-pragmas -Werror -I./source/utils
+COPTS := -pedantic -Wall -Wextra -Werror \
+         -Werror=return-type \
+         -Werror=uninitialized \
+         -Werror=unknown-pragmas \
+         -I./source/utils
+
 LOPTS := -Wl,--start-group $(DAAL_LIB) $(EXT_LIB) -Wl,--end-group
 
 RES_DIR=_results/$(compiler)_$(_IA)_$(threading)_$(RES_EXT)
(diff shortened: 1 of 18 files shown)


napetrov:
Overall this looks good for me. Suggest to add other folks to review DPC++ part

@SmirnovEgorRu  - there still C++11 features left on C++ interfaces headers, so we can't enforce C++ 3 standard. We need to take a look on this separately.

dr-pain:
Do you plan to exclude stdc++11 from linux examples makefiles and check your fix with gcc5?

rlnx:
@dr-pain, I've checked fix with gcc 9.1.0 and `-std=c++03`, there are no errors related to oneAPI headers, but there are still many errors in classical DAAL headers, e.g., variadic macro and some minor things like comma at the end of  the last enum value.

So, I cannot remove `-std=c++11` option from the examples.

rlnx:
@dr-pain, actually the errors with variadic macro are coming from warning and combination of  `-Werror` and `-pedantic`. Technically, we can override it by adding `-Wno-variadic-macros`, but formally, classical DAAL headers are not compliant with ISO C++03 and we must rely on some compiler extensions.

rlnx:
We could give recommendation for the users who want to use DAAL with `-std=c++03` and `-pedantic` to add `-Wno-variadic-macros`. @SmirnovEgorRu, @dr-pain are you ok with it?

rlnx:
One more observation, ICC 19.xxx does not support option `-std=c++03` :smile:

dr-pain:
I`m OK with -Wno-variadic-macros addition to examples makefiles.
I also suggest to remove std option from makefiles.

rlnx:
> I also suggest to remove std option from makefiles.

@dr-pain, just to double check, do you mean remove `-std=c++11` or remove `-std=` at all?

dr-pain:
-std at all. I have added -std=c++11 to "workaround" issue from this pr.

rlnx:
@dr-pain, I suggest to add `-std=c++03` for GCC only, otherwise adding `-Wno-variadic-macro` changes nothing. Starting from GCC 6.1 the default C++ standard is 14 (we use GCC 7.4.0 as the default for testing environment in infra scripts and GCC 8.3 in CI to build examples). See more details [here](https://gist.github.com/ax3l/53db9fa8a4f4c21ecc5c4100c0d93c94).

Having `-std=c++03 -pedantic -Wno-variadic-macro` for GCC allows us to check whether DAAL can be compiled along with C++ ISO 2003 (+ variadic macro).

PR #510, opened 2020-04-06
Update version to beta07

homksei:


dr-pain:
Why nobody from DAAL developers was not invited for review?
Do you plan to change version in infra repo?

homksei:
@dr-pain, If necessary, we will change version in the infra repository.

PR #394, opened 2020-02-19
fix DAAL build by newest msvs2019

dr-pain:

PR #513, opened 2020-04-07
Fix msvs2019 build

dr-pain:
the same as https://github.com/oneapi-src/oneDAL/pull/394


dr-pain:
PCm - http://intel-ci.intel.com/ea78cb44-bbe5-f12d-8d09-8cdcd4b723a1

PR #518, opened 2020-04-08
kNN predict optimizations

SmirnovEgorRu:
The PR optimizes knn predict algorithm if training data set in model is represented as a homogen SOA numeric table (default behavior for C++ and only one available for IDP Sklearn).


KalyanovD:
http://intel-ci.intel.com/ea799f27-2134-f1ff-b845-8cdcd4b723a1

PR #520, opened 2020-04-08
Fixes for DBSCAN documentation

vepifanov:


outoftardis:
@ShvetsKS could you review? I don't know whether or not this is enough to fix the problem with DBSCAN.

ShvetsKS:
please create similar PR to another branches

PR #521, opened 2020-04-08
[beta 6] kNN predict optimizations

KalyanovD:
This is a cherry-pick from https://github.com/oneapi-src/oneDAL/pull/518

The PR optimizes knn predict algorithm if training data set in model is represented as a homogen SOA numeric table (default behavior for C++ and only one available for IDP Sklearn).


KalyanovD:
http://intel-ci.intel.com/ea79cc24-68d2-f1ad-8457-8cdcd4b723a1

PR #534, opened 2020-04-13
Change compiler options

maria-Petrova:


dr-pain:
What are the reasons of these changes?

napetrov:
This change is required to workaround changes in new compiler that exposed problem with zero relocation in core library

PR #538, opened 2020-04-15
tbb atomics removal

dr-pain:


include/services/daal_atomic_int.h
```diff
@@ -51,39 +58,66 @@ class DAAL_EXPORT Atomic
      * Returns an increment of atomic object
      * \return An increment of atomic object
      */
-    dataType inc();
+    dataType inc()
+    {
+        tbb::atomic<dataType> * atomicPtr = (tbb::atomic<dataType> *)(this->_ptr);
```
SmirnovEgorRu, line 63: 1) Better to use static_cast instead of C-cast
2) check for null  pointer is required

The same for other places

include/services/daal_atomic_int.h
```diff
@@ -51,39 +58,66 @@ class DAAL_EXPORT Atomic
...
+        tbb::atomic<dataType> * atomicPtr = new tbb::atomic<dataType>();
+        *atomicPtr                        = value;
+        this->_ptr                        = atomicPtr;
+    }
 
     /** Destructor */
-    ~Atomic();
+    ~Atomic()
+    {
+        delete (tbb::atomic<dataType> *)(this->_ptr);
```
SmirnovEgorRu, line 119: Set ptr to nullptr according to SDL

include/services/daal_atomic_int.h
```diff
@@ -51,39 +58,66 @@ class DAAL_EXPORT Atomic
...
+    }
 
     /**
      * Constructs an atomic object from a value
      * \param[in] value The value to be assigned to the atomic object
      */
-    Atomic(dataType value);
+    Atomic(dataType value) : _ptr(nullptr)
+    {
+        tbb::atomic<dataType> * atomicPtr = new tbb::atomic<dataType>();
```
SmirnovEgorRu, line 111: It's bad practice if you have default copy constructor and a filed with pointer. It can lead to "double free" error.
C++03 Solution: make copy constructor private

SmirnovEgorRu:
Probably stupid question, But you still use tbb::atomic.
What is the main idea?

dr-pain:
> Probably stupid question, But you still use tbb::atomic.
> What is the main idea?
It is the first change in set of changes. Its main idea of this exact change - to move our threading layer to daal atomics usage. Because of this, I need to move atomics implementation from kernel to headers.
Next change will replace tbb usage to system functions calls. So, I will leave your comments without answers, but will take them into accaunt.

include/services/daal_atomic_int.h
iburyl: existing constructors ensure no null pointer is possible

include/services/daal_atomic_int.h
```diff
@@ -37,68 +40,293 @@ namespace interface1
...
+ */
+template <typename dataType>
+class DAAL_EXPORT Atomic
+{
+public:
+    /**
+     * Returns an increment of atomic object
+     * \return An increment of atomic object
+     */
+    inline dataType inc() { return (dataType)(__atomic_add_fetch(reinterpret_cast<volatile dataType *>(&my_storage), 1, __ATOMIC_SEQ_CST)); }
```
alexey-katranov, line 273: The casts seem surplus.

include/services/daal_atomic_int.h
```diff
@@ -37,68 +40,293 @@ namespace interface1
...
  *
  * \tparam dataType Data type of the atomic object
  */
-template <typename dataType = int>
+template <typename dataType>
 class DAAL_EXPORT Atomic
+{};
+
+template <>
+class DAAL_EXPORT Atomic<int>
```
alexey-katranov, line 130: Win32 implementation can be reused.

algorithms/threading/service_thread_pinner.cpp
```diff
@@ -37,12 +37,13 @@
...
         #include <tbb/task_arena.h>
         #include <tbb/task_scheduler_observer.h>
-        #include <tbb/atomic.h>
         #include <tbb/task_scheduler_init.h>
         #include <tbb/parallel_reduce.h>
         #include <tbb/blocked_range.h>
         #include <tbb/tick_count.h>
         #include <tbb/scalable_allocator.h>
+        #include "services/daal_atomic_int.h"
+using namespace daal::services;
```
averbukh, line 46: If you introduces 'using namespace', it makes sense to remove this namespace across entire file from lines where it explicitly used.

include/services/daal_atomic_int.h
```diff
@@ -37,68 +40,293 @@ namespace interface1
...
+class DAAL_EXPORT Atomic
+{
+public:
+    /**
+     * Returns an increment of atomic object
+     * \return An increment of atomic object
+     */
+    inline dataType inc()
+    {
+        DAAL_ASSERT(sizeof(my_storage) == sizeof(long))
```
averbukh, line 61: If the class to be moved into cpp, it can be replaced by static_assert (due to c++11 usage).

include/services/daal_atomic_int.h
```diff
@@ -37,68 +40,293 @@ namespace interface1
...
 
     /**
      * Returns a decrement of atomic object
      * \return An decrement of atomic object
      */
-    dataType dec();
+    inline size_t dec()
+    {
+        DAAL_ASSERT(sizeof(my_storage) == sizeof(size_t))
+        return (size_t)(_InterlockedExchangeAdd64((__int64 *)(&my_storage), -1) - 1);
```
averbukh, line 214: The result can be wrong if highest bit is 1 in the original number. Also, it violates SDL.

dr-pain: I don`t see problems here:
1. we don`t loose number sign here - no sdl violations
2. 0x80000001. size_t(32bit) is 2147483649. int(32bit) is -2147483647. int(-2147483647 - 1) = 0x80000001 = size_t(2147483649 - 1)

include/services/daal_atomic_int.h
dr-pain: I don`t plan to move it into cpp

include/services/daal_atomic_int.h
dr-pain: done.

dr-pain:
PCms:
http://intel-ci.intel.com/ea8870b4-8e98-f13e-8923-8cdcd4b723a1
http://intel-ci.intel.com/ea8870ad-7f7f-f163-b6cc-8cdcd4b723a1
http://intel-ci.intel.com/ea8870a3-6bb1-f1e2-8a22-8cdcd4b723a1

PR #549, opened 2020-04-20
switched to the approach with loading ze_loader lib on demand

makart19:
switched to the approach with loading ze_loader lib on demand


napetrov:
/intelci run
set infra_branch=dev/amk_level0_v

napetrov:
/intelci restart
set infra_branch=dev/amk_level0_v

napetrov:
/intelci: run
set infra_branch=dev/amk_level0_v

PR #556, opened 2020-04-24
Fix memory allocation in SVM Predict  & Performance optimization

PetrovKP:
1. Fixed bugs for SVM Memory allocation failed for big data in prediction step. We allocated `n_samples` x `n_support_vectors`. For `n_samples = 1M` and `n_support_vectors = 500K` we tried allocated  **~3725 GB**. In this PR we work with data in blocks (`block_size = 256`). For `n_samples = 1M` and `n_support_vectors = 500K` we now allocate **~238MB**. 
2. Parallelization blocks for Dense and CSR cases. 
3. Added safe dispatcher for parallel blocks.


PetrovKP:
@Vika-F , @owerbat , @PivovarA  Please, look changes 1,2:

1. Fixed bugs for SVM Memory allocation failed for big data in prediction step. We allocated n_samples x n_support_vectors. For n_samples = 1M and n_support_vectors = 500K we tried allocated ~3725 GB. In this PR we work with data in blocks (block_size = 256). For n_samples = 1M and n_support_vectors = 500K we now allocate ~238MB.
2. Parallelization blocks for Dense and CSR cases.

@masdevas Please, look change 3:
3. Added safe dispatcher for parallel blocks.

algorithms/kernel/svm/svm_predict_impl.i
```diff
@@ -43,6 +43,134 @@ using namespace daal::internal;
...
+
+protected:
+    PredictTaskDense(const size_t nRowsPerBlock, const NumericTablePtr & xTable, const NumericTablePtr & svTable,
+                     kernel_function::KernelIfacePtr & kernel)
+        : Super(nRowsPerBlock, xTable, svTable, kernel)
+    {}
+
+    NumericTablePtr getBlockNTData(const size_t startRow, const size_t nRows, services::Status & status) override
+    {
+        _xBlock.set(*Super::_xTable, startRow, nRows);
```
owerbat, line 120: You should check the block: ```DAAL_CHECK_BLOCK_STATUS_THR(_xBlock);```
And the same for CSR case

algorithms/kernel/svm/svm_predict_impl.i
```diff
@@ -43,6 +43,134 @@ using namespace daal::internal;
...
+        return _kernel->computeNoThrow();
+    }
+
+    const algorithmFPType * getBuff() const { return _buff.get(); }
+
+protected:
+    PredictTask(const size_t nRowsPerBlock, const NumericTablePtr & xTable, const NumericTablePtr & svTable, kernel_function::KernelIfacePtr & kernel)
+        : _xTable(xTable), _nSV(svTable->getNumberOfRows()), _nFeatures(svTable->getNumberOfColumns())
+    {
+        services::Status status;
```
owerbat, line 76: Probably ```status``` is unnecessary here

PetrovKP: Thanks, delete.

algorithms/kernel/svm/svm_predict_impl.i
PetrovKP: I added check for status. But I don't use `DAAL_CHECK_BLOCK_STATUS_THR`, because `getBlockNTData` return not status

masdevas:
I think that it's nice approach for threading sequential parts of algorithms that also supports code version is equal to safe version! LGTM

PR #558, opened 2020-04-26
Hot fixes for dbscan

amgrigoriev:
- Fixed behavior for very small datasets (nRows < 64)
- Optimized copyCoreObservations for current oneapi state

### DIFF
--- a/algorithms/kernel/dbscan/oneapi/dbscan_dense_default_batch_ucapi_impl.i
+++ b/algorithms/kernel/dbscan/oneapi/dbscan_dense_default_batch_ucapi_impl.i
@@ -26,6 +26,8 @@
 #include "algorithms/kernel/dbscan/oneapi/cl_kernels/dbscan_cl_kernels.cl"
 #include "externals/service_ittnotify.h"
 
+#include <iostream>
+
 using namespace daal::services;
 using namespace daal::oneapi::internal;
 using namespace daal::data_management;
@@ -39,7 +41,7 @@ namespace dbscan
 namespace internal
 {
 template <typename algorithmFPType>
-services::Status DBSCANBatchKernelUCAPI<algorithmFPType>::initializeBuffers(uint32_t nRows)
+services::Status DBSCANBatchKernelUCAPI<algorithmFPType>::initializeBuffers(uint32_t nRows, uint32_t chunkNumber)
 {
     Status s;
     auto & context = Environment::getInstance()->getDefaultExecutionContext();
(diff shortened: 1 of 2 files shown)


algorithms/kernel/dbscan/oneapi/dbscan_kernel_ucapi.h
```diff
@@ -82,14 +83,16 @@ class DBSCANBatchKernelUCAPI : public Kernel
...
 
     services::Status buildProgram(oneapi::internal::ClKernelFactoryIface & kernel_factory);
 
-    static const size_t _minSubgroupSize  = 16;
-    static const size_t _maxWorkgroupSize = 256;
-    static const size_t _chunkNumber      = 64;
-    static const size_t _queueBlockSize   = 64;
+    static const size_t _minSubgroupSize              = 16;
+    static const size_t _maxWorkgroupSize             = 256;
+    static const size_t _minRecommendedNumberOfChunks = 64;
```
michael-smirnov, line 92:
```suggestion
    static const size_t _minNumberOfChunks = 64;
```
probably will be better this way?

algorithms/kernel/dbscan/oneapi/dbscan_dense_default_batch_ucapi_impl.i
```diff
@@ -202,7 +203,14 @@ Status DBSCANBatchKernelUCAPI<algorithmFPType>::compute(const NumericTable * x,
...
 
     DAAL_OVERFLOW_CHECK_BY_MULTIPLICATION(uint32_t, _queueBlockSize, nRows);
-    DAAL_CHECK_STATUS_VAR(initializeBuffers(nRows));
+    uint32_t chunkSize   = _recommendedChunkSize;
+    uint32_t chunkNumber = nRows / chunkSize + uint32_t(bool(nRows % chunkSize));
+    while (chunkNumber < _minRecommendedNumberOfChunks && chunkSize > _minChunkSize)
+    {
+        chunkSize /= 2;
+        chunkNumber = nRows / chunkSize + uint32_t(bool(nRows % chunkSize));
+    }
```
michael-smirnov, line 212: You can extract method `calculateChunkNumber()` from this code, make `chunkNumber` variable a field of `DBSCANBatchKernelUCAPI` object. In this case, I would recommend:
- remove an additional parameter `chunkNumber` from several functions like `initializeBuffers`.
- initialize `_chunkNumber` in the constructor with minimum/recommended value. It will be recalculated in the `compute()` method.

algorithms/kernel/dbscan/oneapi/dbscan_kernel_ucapi.h
amgrigoriev: It might be confusing because real value can be less than _minNumberOfChunks in that case

PR #563, opened 2020-04-28
Fix deadlock in daal4py 

rlnx:
The following Python code hangs due to deadlock in DPC++ runtime:
```py
import pandas as pd
import numpy as np
import daal4py as d4p
from daal4py.oneapi import sycl_context

x = np.ascontiguousarray(pd.read_csv('examples/data/batch/kmeans_dense.csv',
                         header=None, dtype=np.float64).values)

with sycl_context('gpu'):
    r = d4p.kmeans_init(10).compute(x)
    c = r.centroids

c_host = r.centroids

with sycl_context('gpu'):
    r = d4p.kmeans(10, 10, assignFlag=True).compute(x, c_host)
    c = r.centroids
```

It's hard to explain why, but it's related to invalid order of host accessor and buffer destruction. The following code reproduces the same problem on C++ side:
```cpp
#include <vector>
#include <iostream>
#include "CL/sycl.hpp"

int main(int argc, char const *argv[]) {
    using namespace cl::sycl;

    auto q = queue(gpu_selector{});

    auto x_ptr = new buffer<float, 1>(10);
    auto& x = *x_ptr;

    q.submit([&](handler& h) {
        auto x_acc = x.get_access<access::mode::write>(h);
        h.parallel_for(range<1>{x.get_count()}, [=](id<1> index) {
            x_acc[index] = 1.0;
        });
    }).wait_and_throw();

    auto x_host_acc = accessor<float, 1, access::mode::read,
                               access::target::host_buffer>(x);

    delete x_ptr;

    return 0;
}
```
 
**The fix proposed in this PR solves the problem by tracking reference to buffer in `HostAccessorDeleter`.**


rlnx:
@Alexander-Makaryev, please take a look

PetrovKP:
```c++
#include <vector>
#include <iostream>
#include "CL/sycl.hpp"

int main(int argc, char const *argv[]) {
    using namespace cl::sycl;

    auto q = queue(gpu_selector{});

    auto x_ptr = new buffer<float, 1>(10);
    auto& x = *x_ptr;

    q.submit([&](handler& h) {
        auto x_acc = x.get_access<access::mode::write>(h);
        h.parallel_for(range<1>{x.get_count()}, [=](id<1> index) {
            x_acc[index] = 1.0;
        });
    }).wait_and_throw();

    auto x_host_acc = accessor<float, 1, access::mode::read,
                               access::target::host_buffer>(x);

    delete x_ptr;

    return 0;
}
```

A small question. Is this a problem only with the current version of the compiler? Have you tried to run and build the example with previous versions of compiler?

rlnx:
> Precommit is Failed

I'm branched from the outdated master, so I'll rebase it.

rlnx:
> A small question. Is this a problem only with the current version of the compiler? Have you tried to run and build the example with previous versions of compiler?

@PetrovKP, Alexander Makaryev claims that the same problem takes place with the older compiler. 

Seems like, the situation shown in the reproducer above is undefined behavior in DPC++. The problem is that we delete the buffer while a host accessor holds a reference/pointer (not shared) to it. Spec says nothing about should host accessor hold shared pointer, raw pointer or reference and of course it says nothing about how to handle the situation when the buffer is already dead, but the accessor is still alive.

PR #568, opened 2020-04-29
[master] added a virtual destructor to fix memory leak in gbt

PivovarA:
I detected a memory leak in GBT.
This was due to the fact that the destructor of the ColIndexTaskBins object is not called.
I added a virtual destructor  for the parent ColIndexTask object to fix this.


ShvetsKS:
please change description to valid one

PR #572, opened 2020-05-06
Update of modulefiles

maria-Petrova:


maria-Petrova:
/intelci: run
set infra_branch=dev/mariapet-modulefiles-update

PR #575, opened 2020-05-09
Fix logreg predict instance for GPU

PetrovKP:
In windows testing we have inlegal symbols in SVM Predict from Logistic Regression Predict. It changes fixed instance Logistic Regression Predict.


PetrovKP:
http://intel-ci.intel.com/ea922145-31f8-f122-ae0f-8cdcd4b723a1

PR #594, opened 2020-05-18
WIP: Improving the mechanism for selecting device parameters

vlad-nazarov:


include/oneapi/internal/execution_context.h
```diff
@@ -313,6 +313,7 @@ class ClKernelFactoryIface
 struct InfoDevice
 {
     size_t maxWorkGroupSize;
+    uint32_t maxNumSubGroups;
```
PetrovKP, line 316: Why do use different type?

PR #599, opened 2020-05-21
KMeans block size optimizations have been implemented.

masdevas:
This request includes block size optimizations for KMeans.

### DIFF
--- a/algorithms/kernel/kmeans/kmeans_lloyd_helper.h
+++ b/algorithms/kernel/kmeans/kmeans_lloyd_helper.h
@@ -34,6 +34,7 @@
 #include "externals/service_blas.h"
 #include "externals/service_spblas.h"
 #include "service/kernel/service_data_utils.h"
+#include "service/kernel/service_environment.h"
 
 namespace daal
 {
@@ -85,7 +86,7 @@ struct TlsTask
         }
     }
 
-    static TlsTask<algorithmFPType, cpu> * create(int dim, int clNum, int maxBlockSize)
+    static TlsTask<algorithmFPType, cpu> * create(const size_t dim, const size_t clNum, const size_t maxBlockSize)
     {
         TlsTask<algorithmFPType, cpu> * result = new TlsTask<algorithmFPType, cpu>(dim, clNum, maxBlockSize);
         if (!result)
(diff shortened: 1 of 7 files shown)


algorithms/kernel/kmeans/kmeans_lloyd_helper.h
```diff
@@ -109,6 +110,47 @@ struct TlsTask
...
+template <Method method, typename algorithmFPType, CpuType cpu>
+struct BSHelper
+{
+    static size_t kmeansGetBlockSize(const size_t nRows, const size_t dim, const size_t clNum);
+};
+
+template <typename algorithmFPType, CpuType cpu>
+struct BSHelper<lloydDense, algorithmFPType, cpu>
+{
+    static size_t kmeansGetBlockSize(const size_t nRows, const size_t dim, const size_t clNum)
```
SmirnovEgorRu, line 122: Now it's okay. But we need to think about extending this for any DAAL algorithm.
I suppose it will be applicable for large list of algorithms if we will set some function which compute memory consumption based on nRows.

service/kernel/service_topo.cpp
```diff
@@ -2023,6 +2032,11 @@ unsigned getL1CacheSize()
     return 32 * 1024;
 }
 
+unsigned getL2CacheSize()
```
SmirnovEgorRu, line 2035: This path of code is related to Mac, right?
If so - I propose to set this to 256kB, because it's more relevant for desktop cpus (main use case for Mac).

algorithms/kernel/kmeans/kmeans_lloyd_helper.h
```diff
@@ -109,6 +110,47 @@ struct TlsTask
...
+    {
+        const double cacheFullness     = 0.8;
+        const size_t maxRowsPerBlock   = 512;
+        const size_t minRowsPerBlockL1 = 256;
+        const size_t minRowsPerBlockL2 = 8;
+        const size_t rowsFitL1         = (getL1CacheSize() / sizeof(algorithmFPType) - (clNum * dim)) / (clNum + dim) * cacheFullness;
+        const size_t rowsFitL2         = (getL2CacheSize() / sizeof(algorithmFPType) - (clNum * dim)) / (clNum + dim) * cacheFullness;
```
SmirnovEgorRu, line 129: I suppose it's applicable for Homogen mostly without data copying. Is it applicable for SOA, when we need to copy data?

PR #602, opened 2020-05-25
fix the wrong jni method names in homogen_numeric_table_byte_buffer_impl.cpp file

JkSelf:
# Description
Fix the `java.lang.UnsatisfiedLinkError` issue when calling the set method in HomogenNumericTableByteBufferImpl file. It is caused by the wrong jni function names described in homogen_numeric_table_byte_buffer_impl.cpp file. And this PR will correct the wrong jni function names in the native codes.

-only correct the wrong function names in native codes
-
-

### DIFF
--- a/lang_service/java/com/intel/daal/data_management/data/homogen_numeric_table_byte_buffer_impl.cpp
+++ b/lang_service/java/com/intel/daal/data_management/data/homogen_numeric_table_byte_buffer_impl.cpp
@@ -594,7 +594,7 @@ JNIEXPORT void JNICALL Java_com_intel_daal_data_1management_data_HomogenNumericT
  * Method:    set
  * Signature: (JJD)V
  */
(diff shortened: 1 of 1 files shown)


PivovarA:
Hi @JkSelf 
Thanks so much for your PR. Fix really looks appropriate. 
But can you format your code according to clang-format?
You can do this as follows:
*clang-format -i your_file.cpp*

xwu99:
@JkSelf @PivovarA For those functions, I think need to add additional "long cObject" param for "jlong numTableAddr" in the JNI call. and fix all related calls in Java. could you double check?

    private native void cSetDouble(long row, long column, double value);
    private native void cSetFloat(long row, long column, float value);
    private native void cSetLong(long row, long column, long value);
    private native void cSetInt(long row, long column, int value);

    private native double cGetDouble(long row, long column);
    private native float cGetFloat(long row, long column);
    private native long cGetLong(long row, long column);
    private native int cGetInt(long row, long column);

napetrov:
@Mergify rebase

PR #603, opened 2020-05-25
Fix for Random Forest deserialization in daal4py

KalyanovD:
There is difference between master and rls/daal-2020-mnt branches which leads to failing serialization tests in daal4py.

### DIFF
--- a/algorithms/kernel/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
+++ b/algorithms/kernel/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
@@ -813,14 +813,14 @@ Status PredictClassificationTask<float, avx512>::predictOneRowByAllTrees(size_t 
                 for (size_t i = 0; i < 16; ++i)
                 {
                     prob_ptr[iTree + i] = const_cast<double *>(_model->getProbas(iTree + i));
-                    prob_pd = _mm512_add_pd(prob_pd, _mm512_mask_load_pd(zero_pd, tail_mask, prob_ptr[iTree + i] + displaces[i] * _nClasses));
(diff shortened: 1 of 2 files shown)


algorithms/kernel/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
```diff
@@ -839,16 +839,16 @@ Status PredictClassificationTask<float, avx512>::predictOneRowByAllTrees(size_t
             {
                 for (size_t i = 0; i < 16; ++i)
                 {
-                    prob_pd = _mm512_add_pd(prob_pd, _mm512_load_pd(prob_ptr[iTree + i] + displaces[i] * _nClasses + iBlock * 8));
+                    prob_pd = _mm512_add_pd(prob_pd, _mm512_loadu_pd(prob_ptr[iTree + i] + displaces[i] * _nClasses + iBlock * 8));
                 }
-                _mm512_store_pd(prob_d + iBlock * 8, _mm512_add_pd(prob_pd, _mm512_mask_load_pd(zero_pd, tail_mask, prob_d + iBlock * 8)));
+                _mm512_store_pd(prob_d + iBlock * 8, _mm512_add_pd(prob_pd, _mm512_load_pd(prob_d + iBlock * 8)));
```
SmirnovEgorRu, line 844: Better to use ```_mm512_loadu_pd``` even here, if we can't guarantee alignment of the memory.
The same for ```_mm512_store_pd```, why don't use ```_mm512_storeu_pd```?

KalyanovD: `services::internal::TArray<double, cpu> _probas_d;`
And following
`double * prob_d = _probas_d.get();`
And we can guarantee that it is aligned as it is created on our side.
 
But `prob_ptr` may be not aligned as it comes from user.

PR #612, opened 2020-05-28
PCA block size optimizations

masdevas:
PCA block size optimizations: fit and transform. Covariance optimizations were created for fit stage optimization.

### DIFF
--- a/algorithms/kernel/covariance/covariance_helper.h
+++ b/algorithms/kernel/covariance/covariance_helper.h
@@ -1,0 +1,64 @@
+/* file: covariance_helper.h */
+/*******************************************************************************
+* Copyright 2014-2020 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef __COVARIANCE_HELPER_H__
+#define __COVARIANCE_HELPER_H__
+
+#include "service/kernel/service_environment.h"
+
+namespace daal
+{
+namespace algorithms
+{
+namespace covariance
+{
+namespace internal
+{
+template <typename algorithmFPType, CpuType cpu>
+struct BSHelper
+{
+    static size_t getBlockSize(const size_t n, const size_t k)
+    {
+        const double cacheFullness     = 0.8;
+        const size_t maxRowsPerBlock   = 512;
+        const size_t minRowsPerBlockL1 = 256;
+        const size_t minRowsPerBlockL2 = 8;
+        size_t rowsFitL1               = (services::internal::getL1CacheSize() / sizeof(algorithmFPType) * cacheFullness - n * n) / n;
+        size_t rowsFitL2               = (services::internal::getL2CacheSize() / sizeof(algorithmFPType) * cacheFullness - n * n) / n;
+        size_t blockSize               = 96;
+
+        if (rowsFitL1 >= minRowsPerBlockL1 && rowsFitL1 <= maxRowsPerBlock)
+        {
+            blockSize = rowsFitL1;
+        }
+        else if (rowsFitL2 >= minRowsPerBlockL2 && rowsFitL2 <= maxRowsPerBlock)
+        {
+            blockSize = rowsFitL2;
+        }
+        else if (rowsFitL2 >= maxRowsPerBlock)
+        {
+            blockSize = maxRowsPerBlock;
+        }
+        return blockSize;
+    }
+};
+} // namespace internal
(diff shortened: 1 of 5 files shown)


algorithms/kernel/covariance/covariance_helper.h
```diff
@@ -0,0 +1,64 @@
...
+namespace algorithms
+{
+namespace covariance
+{
+namespace internal
+{
+template <typename algorithmFPType, CpuType cpu>
+struct BSHelper
+{
+    static size_t getBlockSize(const size_t n, const size_t k)
```
SmirnovEgorRu, line 34: I don't see a difference btw helper for covariance/pca transform/k-means.
Can we extract this to a common class?

algorithms/kernel/covariance/covariance_helper.h
```diff
@@ -0,0 +1,64 @@
...
+    {
+        const double cacheFullness     = 0.8;
+        const size_t maxRowsPerBlock   = 512;
+        const size_t minRowsPerBlockL1 = 256;
+        const size_t minRowsPerBlockL2 = 8;
+        size_t rowsFitL1               = (services::internal::getL1CacheSize() / sizeof(algorithmFPType) * cacheFullness - n * n) / n;
+        size_t rowsFitL2               = (services::internal::getL2CacheSize() / sizeof(algorithmFPType) * cacheFullness - n * n) / n;
+        size_t blockSize               = 96;
```
PetrovKP, line 42: Why 96?

algorithms/kernel/covariance/covariance_impl.i
```diff
@@ -145,7 +133,8 @@ services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatu
         algorithmFPType nVectorsInv = 1.0 / (double)(nVectors);
 
         /* Split rows by blocks */
-        size_t numRowsInBlock = getBlockSize<cpu>(nVectors);
+        using BSHelper        = BSHelper<algorithmFPType, cpu>;
+        size_t numRowsInBlock = BSHelper::getBlockSize(nFeatures, nVectors);
```
PetrovKP, line 137: Intel dispatcher?

algorithms/kernel/covariance/covariance_helper.h
masdevas: Block size calculations between covariance and kmeans are different: in kmeans we split left matrix by rows, in covariance we split right matrix by columns. That has impact to algorithm block size calculation. Also we have different conditional extra memory usage in iteration of threader_for for each of GEMM based algorithms. For example that is case of pca transform when flag "isNormalized" is used. In future we have to look at for delivering the best optimizations. So, I think that we can't create general strategy for block size calculation across all algorithms. We don't have a lot of GEMM base algorithms so I think that library size won't be greatly increased.

algorithms/kernel/covariance/covariance_helper.h
masdevas: That is "default" approach for doing that optimization that is used in kmeans. We will calculate optimal value for this parameter in future.

algorithms/kernel/covariance/covariance_impl.i
masdevas: Sure, I will fix it, thanks!

algorithms/kernel/covariance/covariance_helper.h
PetrovKP: I don't understand how k-means are related to covariance.

teabagk7: This value is obtained experimentally as most favorable for cases that exceed L2$ capacity.

PetrovKP: The reader does not know about this. but he must understand where such magic numbers come from.

algorithms/kernel/covariance/covariance_helper.h
teabagk7: The difference between algorithms will be in approach of calculation of `rowsFitL1` and `rowsFitL2`.
So, we can extract part that exclude variables mentioned above.
This can be noted as future improvements, I think at the moment we can do without it.

algorithms/kernel/covariance/covariance_helper.h
PetrovKP: > This value is obtained experimentally as most favorable for cases that exceed L2$ capacity.

Moreover, unless the gemm has a different form, it will not affect your experiments?

teabagk7: We will check it ofc.

PetrovKP: I want you to add comments to these numbers later. To make it clear to anyone.

teabagk7: Thanks, we will append some clarifying comments.

algorithms/kernel/pca/transform/pca_transform_helper.h
```diff
@@ -0,0 +1,67 @@
+/* file: pca_transform_helper.h */
+/*******************************************************************************
+* Copyright 2014-2020 Intel Corporation
```
PetrovKP, line 3: minnor: new file must have copyright this year:

```c++
* Copyright 2020 Intel Corporation
```

masdevas: Thanks!

PR #616, opened 2020-06-01
[mnt] Fix memory allocation in SVM Predict & Performance optimization

PetrovKP:
Cherri-pick from https://github.com/oneapi-src/oneDAL/pull/556

### DIFF
--- a/algorithms/kernel/svm/svm_predict_impl.i
+++ b/algorithms/kernel/svm/svm_predict_impl.i
@@ -44,6 +44,134 @@ using namespace daal::services;
 using namespace daal::services::internal;
 
 template <typename algorithmFPType, CpuType cpu>
+class PredictTask
+{
+public:
+    DAAL_NEW_DELETE();
+    virtual ~PredictTask() {}
+
+    bool isValid() const { return _buff.get(); }
+
+    services::Status kernelCompute(const size_t startRow, const size_t nRows)
+    {
+        services::Status status;
+        NumericTablePtr shResNT = HomogenNumericTableCPU<algorithmFPType, cpu>::create(_buff.get(), _nSV, nRows, &status);
+        DAAL_CHECK_STATUS_VAR(status);
+
+        auto xBlockNT = getBlockNTData(startRow, nRows, status);
+        DAAL_CHECK_STATUS_VAR(status);
+        _shRes->set(kernel_function::values, shResNT);
+        _kernel->getInput()->set(kernel_function::X, xBlockNT);
+        _kernel->getParameter()->computationMode = kernel_function::matrixMatrix;
+
+        return _kernel->computeNoThrow();
+    }
+
+    const algorithmFPType * getBuff() const { return _buff.get(); }
+
+protected:
+    PredictTask(const size_t nRowsPerBlock, const NumericTablePtr & xTable, const NumericTablePtr & svTable, kernel_function::KernelIfacePtr & kernel)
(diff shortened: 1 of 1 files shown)


algorithms/kernel/svm/svm_predict_impl.i
```diff
@@ -67,57 +195,76 @@ struct SVMPredictImpl<defaultDense, algorithmFPType, cpu> : public Kernel
...
+            DAAL_INT n_ = nRowsPerBlockReal;
+            algorithmFPType alpha(1.0);
+            DAAL_INT lda = m_;
+            DAAL_INT incx(1);
+            algorithmFPType beta(1.0);
+            DAAL_INT incy(1);
+
+            if (nBlocks == 1)
+            {
+                Blas<algorithmFPType, cpu>::xgemv(&trans, &m_, &n_, &alpha, buf, &lda, svCoeff, &incx, &beta, distance + startRow, &incy);
```
ShvetsKS, line 258: why we need threaded gemv version here inside threader_for?

algorithms/kernel/svm/svm_predict_impl.i
```diff
@@ -43,6 +43,134 @@ using namespace daal::internal;
...
+        const size_t * rows      = _xBlock.rows();
+        if (!values || !cols || !rows) status |= services::Status(services::ErrorMemoryAllocationFailed);
+        _rowOffsets[0] = 1;
+        for (size_t i = 0; i < nRows; i++)
+        {
+            const size_t nNonZeroValuesInRow = rows[i + 1] - rows[i];
+            _rowOffsets[i + 1]               = _rowOffsets[i] + nNonZeroValuesInRow;
+        }
+
+        return CSRNumericTable::create(values, cols, _rowOffsets.get(), Super::_nFeatures, nRows, CSRNumericTableIface::CSRIndexing::oneBased,
```
ShvetsKS, line 165: potential problem with illegal instruction here. Is it possible to use CPU specific create like for `HomogenNumericTableCPU`?

PetrovKP: I doubt that the CSR tables use specific instructions. Because in the code we constantly use them inside the library - like here. There have never been problems. If we want to make a specific CPU, then we need to do it within the entire library. Not within the PR.

algorithms/kernel/svm/svm_predict_impl.i
PetrovKP: Yes, for `nBlock = 1`. This means that there will be one thread.

algorithms/kernel/svm/svm_predict_impl.i
```diff
@@ -43,6 +43,134 @@ using namespace daal::internal;
...
+    PredictTaskCSR(const size_t nRowsPerBlock, const NumericTablePtr & xTable, const NumericTablePtr & svTable,
+                   kernel_function::KernelIfacePtr & kernel)
+        : Super(nRowsPerBlock, xTable, svTable, kernel), _rowOffsets(nRowsPerBlock + 1)
+    {}
+
+    NumericTablePtr getBlockNTData(const size_t startRow, const size_t nRows, services::Status & status) override
+    {
+        _xBlock.set(dynamic_cast<CSRNumericTableIface *>(Super::_xTable.get()), startRow, nRows);
+        algorithmFPType * values = const_cast<algorithmFPType *>(_xBlock.values());
+        size_t * cols            = const_cast<size_t *>(_xBlock.cols());
```
ShvetsKS, line 155: if possible add const

PR #628, opened 2020-06-04
Fix SVM predict not deterministic result & small refactor

PetrovKP:
Replace TLS struct on LS struct

### DIFF
--- a/algorithms/kernel/kmeans/kmeans_lloyd_postprocessing.h
+++ b/algorithms/kernel/kmeans/kmeans_lloyd_postprocessing.h
@@ -192,8 +192,6 @@ struct PostProcessing<lloydCSR, algorithmFPType, cpu>
     static Status computeAssignments(const size_t p, const size_t nClusters, const algorithmFPType * const inClusters, const NumericTable * ntData,
                                      algorithmFPType * catCoef, NumericTable * ntAssign, const size_t blockSizeDefault)
     {
-        CSRNumericTableIface * ntDataCsr = dynamic_cast<CSRNumericTableIface *>(const_cast<NumericTable *>(ntData));
-
         const size_t n       = ntData->getNumberOfRows();
         const size_t nBlocks = n / blockSizeDefault + !!(n % blockSizeDefault);
 
@@ -217,6 +215,7 @@ struct PostProcessing<lloydCSR, algorithmFPType, cpu>
         }
 
         SafeStatus safeStat;
+        CSRNumericTableIface * ntDataCsr = dynamic_cast<CSRNumericTableIface *>(const_cast<NumericTable *>(ntData));
 
         daal::threader_for(nBlocks, nBlocks, [&](int iBlock) {
             algorithmFPType * x_clusters = tlsTask.local();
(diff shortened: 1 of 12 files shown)


algorithms/kernel/svm/svm_predict_impl.i
```diff
@@ -194,23 +194,19 @@ struct SVMPredictImpl<defaultDense, algorithmFPType, cpu> : public Kernel
...
 
         ReadColumns<algorithmFPType, cpu> mtSVCoeff(*svCoeffTable, 0, 0, nSV);
         DAAL_CHECK_BLOCK_STATUS(mtSVCoeff);
-        const algorithmFPType * svCoeff = mtSVCoeff.get();
+        const algorithmFPType * const svCoeff = mtSVCoeff.get();
 
         size_t nRowsPerBlock = 1;
-        DAAL_SAFE_CPU_CALL((nRowsPerBlock = 256), (nRowsPerBlock = nVectors));
-
+        DAAL_SAFE_CPU_CALL((nRowsPerBlock = 2048), (nRowsPerBlock = nVectors));
```
owerbat, line 204: Does changing ```nRowsPerBlock``` affect performance? Do you have any measurements?

PetrovKP: @owerbat 
Yes. The increase in the block is associated with a change in the TLS structure (TLS -> LS). To ensure stability. But at the same time, an overhead per block has grown. To compensate for it, I had to reduce the number of blocks.

algorithms/kernel/kmeans/kmeans_lloyd_postprocessing.h
```diff
@@ -217,6 +215,7 @@ struct PostProcessing<lloydCSR, algorithmFPType, cpu>
         }
 
         SafeStatus safeStat;
+        CSRNumericTableIface * ntDataCsr = dynamic_cast<CSRNumericTableIface *>(const_cast<NumericTable *>(ntData));
```
ShvetsKS, line 218: should be checked on `nullptr`

algorithms/kernel/kmeans/kmeans_lloyd_postprocessing.h
```diff
@@ -272,15 +271,14 @@ struct PostProcessing<lloydCSR, algorithmFPType, cpu>
...
-        CSRNumericTableIface * ntDataCsr = dynamic_cast<CSRNumericTableIface *>(const_cast<NumericTable *>(ntData));
-
         const size_t n       = ntData->getNumberOfRows();
         const size_t nBlocks = n / blockSizeDefault + !!(n % blockSizeDefault);
 
         TArrayScalable<algorithmFPType, cpu> goalLocal(nBlocks);
         algorithmFPType * goalLocalData = goalLocal.get();
         DAAL_CHECK_MALLOC(goalLocalData);
 
+        CSRNumericTableIface * ntDataCsr = dynamic_cast<CSRNumericTableIface *>(const_cast<NumericTable *>(ntData));
```
ShvetsKS, line 281: similar as above

algorithms/kernel/kmeans/kmeans_lloyd_postprocessing.h
PetrovKP: Applied

algorithms/kernel/kmeans/kmeans_lloyd_postprocessing.h
PetrovKP: Applied

algorithms/kernel/svm/svm_predict_impl.i
ShvetsKS: seems it would be better to have member field of method which returns optimal block size depending on cache sizes.
 let it be for ex. :
```
size_t getBlockSize() {
  return 2048;
}
```
and when we will return to optimization we could change just this functionality.
please create task or issue for experimenting with block size for future optimizations!

PR #622, opened 2020-06-04
[mnt][Part1] Implemented Thunder method for SVM algorithm on CPU

PetrovKP:
Merge with https://github.com/oneapi-src/oneDAL/pull/600

PR #634, opened 2020-06-08
Fix memory allocation in SVM Predict & Performance optimization & small refactor

PetrovKP:
From https://github.com/oneapi-src/oneDAL/pull/628 and https://github.com/oneapi-src/oneDAL/pull/622


ShvetsKS:
it would be better to have similar PR head line as in   #628

PR #654, opened 2020-06-17
Add test logs of conformance testing to artifact in public CI

Pahandrovich:
Logs of pytest was included in artifact, and scikit-learn version was changed to last.

### DIFF
--- a/.ci/scripts/conformance_test.sh
+++ b/.ci/scripts/conformance_test.sh
@@ -38,5 +38,5 @@ source ${BUILD_DIR}/daal/latest/env/vars.sh intel64
 
 # testing
 cd .ci/scripts/conformance-scripts/
-python run_tests.py 0.22.1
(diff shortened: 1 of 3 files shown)


.ci/scripts/conformance_test.sh
```diff
@@ -27,15 +27,16 @@ source ${CONDA_DIR}/etc/profile.d/conda.sh
 export PATH=${CONDA_DIR}/bin:$PATH
 conda create -y -n conf python=3.7
 source activate conf
-conda install -y -c intel mpich tbb-devel numpy pytest scikit-learn pandas
+conda install -y -c intel mpich tbb-devel numpy pytest pandas
```
napetrov, line 30: Per my understanding tbb is not requered for daal4py now.

Pahandrovich: Thanks, fixed

PR #673, opened 2020-06-21
new cmp interop API for OpenCL and workaround for perf degradation in handler::fill

makart19:
Added support for new cmp API for interoperability with OpenCL and workaround to fix performance degradation in case of using handler::fill method.
GPU nightly testing passed for linux and windows:
http://intel-ci.intel.com/eab34118-2e60-f1e7-a8c4-8cdcd4b723a1


cpp/daal/include/sycl/internal/types_utils_cxx11.h
```diff
@@ -163,12 +166,8 @@ class BufferFiller
...
         void operator()(Typelist<T>)
         {
-            auto dst              = dstUnivers.get<T>().toSycl();
-            cl::sycl::event event = queue.submit([&](cl::sycl::handler & cgh) {
-                auto acc = dst.template get_access<cl::sycl::access::mode::write>(cgh);
-                cgh.fill(acc, static_cast<T>(value));
-            });
-            event.wait();
+            auto dst = dstUnivers.get<T>();
+            fillBuffer(dst, dst.size(), static_cast<T>(value));
```
michael-smirnov, line 170: I'd rather comment the lines you've deleted because it'd be easier to turn them back when the bug disappears. I suggest creating an internal task to make it.

cpp/daal/src/sycl/fill_buffer_helper.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include "sycl/internal/utils.h"
+#include "sycl/internal/types.h"
+
+namespace daal
+{
+namespace oneapi
+{
+namespace internal
+{
+namespace interface1
```
PetrovKP, line 30: Why interaface instead library? Need to delete

cpp/daal/src/sycl/fill_buffer_helper.cpp
```diff
@@ -0,0 +1,82 @@
...
+        KernelRange global_range(nElems);
+
+        ctx.run(global_range, kernel, args, &status);
+        DAAL_CHECK_STATUS_VAR(status);
+    }
+
+    return status;
+}
+
+template DAAL_EXPORT services::Status fillBuffer<float>(services::Buffer<float> & buf, size_t nElems, float val);
```
PetrovKP, line 68: Can do based on a universe buffer? to avoid such multiple instantiations?

cpp/daal/include/sycl/internal/types_utils_cxx11.h
makart19: added it as a commented block

cpp/daal/src/sycl/fill_buffer_helper.cpp
makart19: fixed

cpp/daal/src/sycl/fill_buffer_helper.cpp
makart19: In case of using universal buffer additional casting of setting val need to be done in runtime. The goal is to move it on compilation stage and it is reached in current variant.

PR #674, opened 2020-06-22
Implemented training and prediction of SVM for new interfaces

PetrovKP:
- Added Interface for training of SVM 
- Added common interfaces of Infer for all algorithms
- Added Interface for infer for SVM
- Implemented thunder method for training of SVM
- Implemented infer of SVM
- Implemented convector `convert_from_daal_homogen_table`
- Created example for SVM

❓  Naming parameters `tau` and `c` can changed
⚡ Using `kernel function` as parameter will be in next PR
⚡ Support for GPU and DPC++ will be next PR


PetrovKP:
@rlnx , @michael-smirnov Please, look a prototype example: naming parameters, using algorithm and consistency with other examples and with style.

examples/oneapi/cpp/source/svc/svm_thunder_dense_batch.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include <iomanip>
+#include <iostream>
+
+#include "oneapi/dal/data/table.hpp"
+#include "oneapi/dal/data/accessor.hpp"
+#include "oneapi/dal/algo/svm.hpp"
+
+using namespace oneapi;
+
+std::ostream &operator <<(std::ostream& stream, const dal::table& table) {
```
rlnx, line 26: Move it to `utils` folder, see how it is done in classical examples

examples/oneapi/cpp/source/svc/svm_thunder_dense_batch.cpp
```diff
@@ -0,0 +1,113 @@
...
+        .set_tau(0.01);
+
+   const auto result = onedal::train(svm_desc, x_table, y_table);
+   return result.get_model();
+}
+
+onedal::table run_inference(const onedal::table& x_table,
+                                const onedal::table& y_table) {
+
+    const auto svm_desc = dal::svm::descriptor<float, task::classification>{}
```
rlnx, line 60: I'd suggest to reuse descriptor: the same descriptor for both training and inference

examples/oneapi/cpp/source/svc/svm_thunder_dense_batch.cpp
```diff
@@ -0,0 +1,113 @@
...
+                      << std::setiosflags(std::ios::fixed)
+                      << std::setprecision(3)
+                      << x[i * table.get_column_count() + j];
+        }
+        std::cout << std::endl;
+    }
+    return stream;
+}
+
+onedal::svm::model run_training(const onedal::table& x_table,
```
rlnx, line 42: I think, there is no reason to have separate method for training, new API is much more compact

examples/oneapi/cpp/source/svc/svm_thunder_dense_batch.cpp
PetrovKP: But in predict can use other method. For example: multi class have two method: `wu`, `vote` and training have only method: `oneAgainstOne`. Some situation for SVM, training have two method: `thunder` and `boser` and predict have only `DefaultDense`. By the way, I’m thinking of renaming methods for training SVM: `boser` -> `smo`, `thunder` -> `block_smo`. What do you think?

examples/oneapi/cpp/source/svc/svm_thunder_dense_batch.cpp
PetrovKP: I am trying to do as in the specification. But I liked the idea)

examples/oneapi/cpp/source/svc/svm_thunder_dense_batch.cpp
PetrovKP: Sure, I'm only afraid of conflicts. For good I’ll try to do it in a separate PR.

cpp/oneapi/dal/backend/interop/table_conversion.hpp
```diff
@@ -46,6 +46,7 @@ template <typename T>
 inline auto convert_to_daal_homogen_table(array<T>& data,
                                           std::int64_t row_count,
                                           std::int64_t column_count) {
+    if (!data) return daal::data_management::NumericTablePtr();
```
michael-smirnov, line 49: What does this check mean? `array` has no `operator bool()` implementation.

cpp/oneapi/dal/algo/svm/backend/cpu/infer_kernel_predict_fast.cpp
```diff
@@ -0,0 +1,124 @@
...
+    auto& set_coefficients(daal::data_management::NumericTablePtr coefficients) {
+        _SVCoeff = coefficients;
+        return *this;
+    }
+
+    auto& set_bias(double bias) {
+        _bias = bias;
+        return *this;
+    }
+};
```
michael-smirnov, line 53: We are trying not to propagate old code to the new oneDAL library in every implementation file. For oneDAL-DAAL integration, we have an `interop` module: https://github.com/oneapi-src/oneDAL/tree/master/cpp/oneapi/dal/backend/interop.
I'd suggest creating converters/adapters on DAAL in this module. It would be easier to find such conversions, rather than they will be distributed across all the library. The only exception, in my opinion, should be `call_daal_kernel`, but it can be also moved to the interop module.

cpp/oneapi/dal/algo/svm/detail/train_ops.cpp
```diff
@@ -0,0 +1,42 @@
...
+                            const descriptor_base& desc,
+                            const train_input& input) const {
+        using kernel_dispatcher_t = dal::backend::kernel_dispatcher<
+            backend::train_kernel_cpu<Float, Task, Method>>;
+        return kernel_dispatcher_t()(ctx, desc, input);
+    }
+};
+
+#define INSTANTIATE(F, T, M) \
+  template struct train_ops_dispatcher<default_execution_context, F, T, M>;
```
michael-smirnov, line 35: I think it's time to make a common header with this define.

cpp/oneapi/dal/backend/interop/table_conversion.hpp
```diff
@@ -54,4 +55,18 @@ inline auto convert_to_daal_homogen_table(array<T>& data,
         daal_data, column_count, row_count);
 }
 
+template <typename T>
+inline auto convert_from_daal_homogen_table(const daal::data_management::NumericTablePtr& nt) {
```
michael-smirnov, line 59: Please do not use `auto` everywhere. If the type is simple, use the correct type.

examples/oneapi/cpp/makefile_lnx
```diff
@@ -147,7 +147,7 @@ sointel64:
 _make_ex: $(RES)
 
 vpath
-vpath %.cpp $(addprefix ./source/,pca)
+vpath %.cpp $(addprefix ./source/,pca svc)
```
michael-smirnov, line 150: svc, not svm?

cpp/oneapi/dal/algo/svm/backend/cpu/infer_kernel_predict_fast.cpp
PetrovKP: Thanks for the comment. In this case, there is no conversion. There is only an extension of the existing functionality of the old daal. Is it worth moving to `interop`?

cpp/oneapi/dal/algo/svm/detail/train_ops.cpp
PetrovKP: But it in namespace svm and I think for other algorithm it can change. For example PCA:
```c++
#define INSTANTIATE(F, M) \
  template struct train_ops_dispatcher<default_execution_context, F, M>;
```

cpp/oneapi/dal/backend/interop/table_conversion.hpp
PetrovKP: Ok, thanks!

cpp/oneapi/dal/algo/svm/backend/cpu/infer_kernel_predict_fast.cpp
michael-smirnov: Yes. All staff that interacts with DAAL shall be located inside `interop` module.

cpp/oneapi/dal/algo/svm/detail/train_ops.cpp
michael-smirnov: The only difference is presence of a task parameter. I think your define will be used in the same way for all algorithms we are implementing now. It's better to avoid copy-paste.

cpp/oneapi/dal/algo/svm/backend/cpu/infer_kernel_predict_fast.cpp
PetrovKP: Applied

cpp/oneapi/dal/algo/svm/detail/train_ops.cpp
PetrovKP: ok, let's do the next pull request.

PetrovKP:
@rlnx , @michael-smirnov , @makart19 , @amgrigoriev, @mchernov-intel Please, make review for this PR.

cpp/oneapi/dal/algo/svm/backend/cpu/train_kernel_thunder.cpp
```diff
@@ -0,0 +1,116 @@
...
+using daal_svm_thunder_kernel_t = daal_svm::training::internal::
+    SVMTrainImpl<daal_svm::training::thunder, Float, daal_svm::Parameter, Cpu>;
+
+template <typename Float>
+static train_result call_daal_kernel(const context_cpu& ctx,
+                                     const descriptor_base& desc,
+                                     const table& data,
+                                     const table& labels,
+                                     const table& weights) {
+    const std::int64_t row_count    = data.get_row_count();
```
michael-smirnov, line 42: nip: you have `using std::int64_t` on 25 line
```suggestion
    const int64_t row_count    = data.get_row_count();
```

cpp/oneapi/dal/algo/svm/backend/cpu/train_kernel_thunder.cpp
```diff
@@ -0,0 +1,116 @@
...
+    auto table_classification_coefficients = interop::convert_from_daal_homogen_table<Float>(
+        daal_model->getClassificationCoefficients());
+    const double bias                        = daal_model->getBias();
+    const std::int64_t support_vectors_count = table_support_vectors.get_row_count();
+
+    auto onedal_model = model()
+                            .set_support_vectors(table_support_vectors)
+                            .set_coefficients(table_classification_coefficients)
+                            .set_bias(bias)
+                            .set_support_vectors_count(support_vectors_count);
```
michael-smirnov, line 89: I think it is better to extract it into a separate method like `interop::convert_from_daal_svm_model(daal_model)`

cpp/oneapi/dal/algo/svm/backend/cpu/train_kernel_thunder.cpp
```diff
@@ -0,0 +1,116 @@
...
+namespace daal_svm             = daal::algorithms::svm;
+namespace daal_kernel_function = daal::algorithms::kernel_function;
+namespace interop              = dal::backend::interop;
+
+template <typename Float, daal::CpuType Cpu>
+using daal_svm_thunder_kernel_t = daal_svm::training::internal::
+    SVMTrainImpl<daal_svm::training::thunder, Float, daal_svm::Parameter, Cpu>;
+
+template <typename Float>
+static train_result call_daal_kernel(const context_cpu& ctx,
```
michael-smirnov, line 37: I think it is better to move this inside `interop` (I think that we should do it in PCA too)

cpp/oneapi/dal/algo/svm/backend/cpu/infer_kernel.cpp
```diff
@@ -0,0 +1,109 @@
...
+namespace daal_svm             = daal::algorithms::svm;
+namespace daal_kernel_function = daal::algorithms::kernel_function;
+namespace interop              = dal::backend::interop;
+
+template <typename Float, daal::CpuType Cpu>
+using daal_svm_predict_kernel_t =
+    daal_svm::prediction::internal::SVMPredictImpl<daal_svm::prediction::defaultDense, Float, Cpu>;
+
+template <typename Float>
+static infer_result call_daal_kernel(const context_cpu& ctx,
```
michael-smirnov, line 38: I think it is better to move it inside `interop`

cpp/oneapi/dal/algo/svm/backend/gpu/infer_kernel.cpp
```diff
@@ -0,0 +1,33 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "oneapi/dal/algo/svm/backend/gpu/infer_kernel.hpp"
+
+namespace oneapi::dal::svm::backend {
+
+template <typename Float>
+struct infer_kernel_gpu<Float, task::classification, method::by_default> {
```
michael-smirnov, line 22: It's not critical, but in specializations/explicit instantiations I'd prefer to use exact `method` tags rather than aliases

cpp/oneapi/dal/algo/svm/common.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "oneapi/dal/algo/svm/common.hpp"
+
+namespace oneapi::dal::svm {
+
+class detail::descriptor_impl : public base {
+public:
+    double c                         = 1.0;
+    double accuracy_threshold        = 0.001;
+    std::int64_t max_iteration_count = 100000;
+    double cache_size                = 200.0;
```
michael-smirnov, line 26: don't we want to focus that it is in megabytes? For example, `cache_size_mb`. Asking because many entities in DPC++ and our library have `get_size()` method returning the size in bytes. I see possible confusion here.

cpp/oneapi/dal/algo/svm/common.hpp
```diff
@@ -0,0 +1,152 @@
...
+    using method_t = method::by_default;
+
+    descriptor_base();
+
+    double get_c() const;
+    double get_accuracy_threshold() const;
+    std::int64_t get_max_iteration_count() const;
+    double get_cache_size() const;
+    double get_tau() const;
+    bool get_shrinking() const;
```
michael-smirnov, line 56: Propably better:

```suggestion
    bool do_shrinking() const;
```

cpp/oneapi/dal/algo/svm/common.cpp
PetrovKP: I think this should be reflected in the specification? I will think about your proposal, while in this PR I propose to leave it as it is.

cpp/oneapi/dal/algo/svm/common.hpp
PetrovKP: Will it be very inconsistent with other methods of library parameters? This name of the parameter was in the old interface. This will also need to be discussed.

cpp/oneapi/dal/algo/svm/backend/gpu/infer_kernel.cpp
PetrovKP: For the infer, we do not want to create a separate method. So, I decided this trick by default value. Maybe there is a better way?

cpp/oneapi/dal/algo/svm/backend/cpu/train_kernel_thunder.cpp
PetrovKP: @rlnx  What do you think?

cpp/oneapi/dal/algo/svm/backend/cpu/train_kernel_thunder.cpp
PetrovKP: Applied

cpp/oneapi/dal/algo/svm/backend/cpu/train_kernel_thunder.cpp
PetrovKP: Done

cpp/oneapi/dal/algo/svm/common.hpp
rlnx: In PCA we have the same situation, `isDeterministic` in old API and `get/set_deterministic` in the new. I'd keep `get/set_shrinking`

cpp/oneapi/dal/algo/svm/backend/cpu/train_kernel_thunder.cpp
rlnx: Let's keep as is.

cpp/oneapi/dal/algo/svm/common.hpp
```diff
@@ -0,0 +1,152 @@
...
+    double get_bias() const;
+
+    auto &set_bias(const double value) {
+        set_bias_impl(value);
+        return *this;
+    }
+
+    std::int64_t get_support_vectors_count() const;
+
+    auto &set_support_vectors_count(const std::int64_t value) {
```
rlnx, line 138:
```suggestion
    auto &set_support_vector_count(const std::int64_t value) {
```

cpp/oneapi/dal/algo/svm/backend/utils/model.hpp
```diff
@@ -0,0 +1,61 @@
+/*******************************************************************************
```
rlnx, line 1: Let's not create even more directories in backend, I'd suggest to rename it to `interop_model.hpp` and put to `svm/backend`

cpp/oneapi/dal/algo/svm/detail/infer_ops.cpp
```diff
@@ -0,0 +1,40 @@
...
+            dal::backend::kernel_dispatcher<backend::infer_kernel_cpu<Float, Task, Method>>;
+        return kernel_dispatcher_t()(ctx, desc, input);
+    }
+};
+
+#define INSTANTIATE(F, T, M) \
+    template struct ONEAPI_DAL_EXPORT infer_ops_dispatcher<default_execution_context, F, T, M>;
+
+INSTANTIATE(float, task::classification, method::by_default)
+INSTANTIATE(double, task::classification, method::by_default)
```
rlnx, line 38: I'm afraid we'll get the problem with backward compatibility here if we change default method. It's ok for Beta, but for Gold we'll need to figure out more solid approach.

PetrovKP: We can add a new method for the prediction and instantiate it with any parameter of the train method. Will it be better?

cpp/oneapi/dal/algo/svm/backend/utils/model.hpp
PetrovKP: Applied

cpp/oneapi/dal/algo/svm/common.hpp
PetrovKP: Applied

examples/oneapi/cpp/source/svm/svm_two_class_thunder_dense_batch.cpp
```diff
@@ -0,0 +1,89 @@
...
+                            .set_accuracy_threshold(0.01)
+                            .set_max_iteration_count(1000)
+                            .set_cache_size(200.0)
+                            .set_tau(1e-6);
+
+  const auto result_train = dal::train(svm_desc, x_train_table, y_train_table);
+
+  std::cout << "Bias:" << std::endl << result_train.get_bias() << std::endl;
+  std::cout << "Support indices:" << std::endl
+            << result_train.get_support_indices() << std::endl;
```
michael-smirnov, line 61: You are using the same pattern for all outputs: 
```cpp
std::cout << "Some description" << std::endl << some_value << std::endl;
``` 
Probably we should move it to utils as special output helper

PR #678, opened 2020-06-24
Clang-format for new interfaces

PetrovKP:
Don't changes folder:
- data
- examples

### DIFF
--- a/cpp/oneapi/dal/.clang-format
+++ b/cpp/oneapi/dal/.clang-format
@@ -1,0 +1,145 @@
+---
+Language: Cpp
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: false
+AlignConsecutiveMacros: true
+AlignEscapedNewlines: Left
+AlignOperands: true
+AlignTrailingComments: false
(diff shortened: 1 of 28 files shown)


cpp/oneapi/dal/.clang-format
```diff
@@ -0,0 +1,145 @@
...
+SpaceBeforeParens: ControlStatements
+SpaceBeforeRangeBasedForLoopColon: true
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles: false
+SpacesInContainerLiterals: false
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard: Cpp11
```
YarShev, line 139: Since we use C++17 for all files in new interfaces, please, fix onto `Cpp17`.

YarShev: Currently, clang-format 9 version is used (it doesn't support `Cpp17`), I think, we can keep standard as `Cpp11` for now based on [doc](https://releases.llvm.org/9.0.0/tools/clang/docs/ClangFormatStyleOptions.html). But, it worth to think on its update in the future.

PR #679, opened 2020-06-24
DPC++ support for dal::array

michael-smirnov:
This PR contains:
- :hatching_chick: public `#define ONEAPI_DAL_DATA_PARALLEL` to enable DPC++ computations
- :wrench: `array` API changes: `get_size()` now returns size in bytes
- `get_count()` implemented: returns number of elements in array
- static construction helpers `full()`, `zeros()` implemented
- :toolbox: all array member function support DPC++ computation
- Common memory management functions between C++ and DPC++: [memory.hpp](https://github.com/oneapi-src/oneDAL/pull/679/files#diff-7b2063dc42fdbd656e87ac9276190b41) 
- Generalizators `host_policy` and `host_only_alloc` for C++ to call memory management functions consistently with DPC++


PetrovKP:
Support make for DPC++ we planned in this PR or in new?

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_cov.cpp
```diff
@@ -41,10 +41,10 @@ static train_result call_daal_kernel(const context_cpu& ctx,
     const int64_t component_count = desc.get_component_count();
 
     auto arr_data = row_accessor<const Float>{ data }.pull();
-    array<Float> arr_eigvec{ column_count * component_count };
-    array<Float> arr_eigval{ 1 * component_count };
-    array<Float> arr_means{ 1 * component_count };
-    array<Float> arr_vars{ 1 * component_count };
+    auto arr_eigvec = array<Float>::zeros(column_count * component_count);
```
PetrovKP, line 44: Why need to spend time for initializing of array? when will we rewrite the contents anyway? This can take a lot of time in some algorithms. Especially if the filling is in sequential mode.

cpp/oneapi/dal/data/array.hpp
```diff
@@ -117,70 +160,132 @@ class array {
...
+private:
+    template <typename Policy, typename AllocKind>
+    static array<T> zeros_impl(Policy&& policy, std::int64_t count, AllocKind&& kind) {
+        auto* data = detail::malloc<T>(policy, count, kind);
+        detail::memset(policy, data, 0, sizeof(T)*count);
+        return array<T> { data, count, detail::default_delete<T, Policy>{ policy } };
+    }
+
+    template <typename K, typename Policy, typename AllocKind>
+    static array<T> full_impl(Policy&& policy,
```
PetrovKP, line 232: Why name is `full`, not `fill`? It confuses me a little :)

cpp/oneapi/dal/detail/memory_impl_dpc.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+inline void memcpy(sycl::queue& queue, void* dest, const void* src, std::int64_t size) {
+    auto event = queue.memcpy(dest, src, size);
+    event.wait();
+}
+
+template <typename T>
+inline void fill(sycl::queue& queue, T* dest, std::int64_t count, const T& value) {
+    auto event = queue.submit([&](sycl::handler& cgh) {
+        cgh.parallel_for<class oneapi_dal_memory_fill>(sycl::range<1>(count),
```
PetrovKP, line 51: Maybe using from standard functions?
```c++
   auto event = queue.fill(value);
```

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_cov.cpp
michael-smirnov: fixed

cpp/oneapi/dal/data/array.hpp
michael-smirnov: It's from @rlnx request to make these functions like in numpy: https://numpy.org/doc/1.18/reference/generated/numpy.full.html
But numpy has ndarray.fill() function too. I can rename it if @rlnx don't mind.

cpp/oneapi/dal/detail/memory_impl_dpc.hpp
michael-smirnov: Unfortunately, this I cannot find it in new DPC++ sources: https://github.com/intel/llvm/blob/sycl/sycl/include/CL/sycl/queue.hpp
That`s why I used custom kernel

cpp/oneapi/dal/data/array.hpp
michael-smirnov: But I think that `full()` is more appropriate here since in numpy it constructs new ndarray, but `fill()` only fills existing one

cpp/oneapi/dal/detail/memory_impl_dpc.hpp
PetrovKP: Ok, sorry. I mean `fill` for handler: https://github.com/intel/llvm/blob/sycl/sycl/include/CL/sycl/handler.hpp#L1626
```c++
inline void fill(sycl::queue& queue, T* dest, std::int64_t count, const T& value) {
    cgh.fill(dest, sycl::range<1>(count));
}
```

cpp/oneapi/dal/data/array.hpp
PetrovKP: Ok, agree with you

cpp/oneapi/dal/detail/memory_impl_dpc.hpp
michael-smirnov: There is no `fill()` method in the `handler` for USM pointers. Your link points to the method accepting accessor. Onse again: I checked the existence of an appropriate `fill()` method before implementing my own custom kernel but didn't find anything. There is a proposal to create `fill()` method for USM pointers in `queue` class, but it is not implemented yet.

cpp/oneapi/dal/data/array.hpp
```diff
@@ -117,70 +160,132 @@ class array {
...
+    const T& operator [](std::int64_t index) const {
+        return get_data()[index];
+    }
+
+    T& operator [](std::int64_t index) {
+        return get_mutable_data()[index];
+    }
+
+private:
+    template <typename Policy, typename AllocKind>
```
rlnx, line 224: I think `AllocKind` can be replaced by `Allocator`

cpp/oneapi/dal/data/array.hpp
```diff
@@ -37,30 +40,71 @@ class array {
...
 public:
-    using default_delete = std::default_delete<T[]>;
+    template <typename K>
+    static array<T> full(std::int64_t count, K&& element) {
+        return full_impl(detail::host_policy{},
+            count, std::forward<K>(element), detail::host_only_alloc{});
+    }
+
+    static array<T> zeros(std::int64_t count) {
+        return zeros_impl(detail::host_policy{}, count, detail::host_only_alloc{});
```
rlnx, line 50: Why not to call `full(count, 0)`? I know, it may be not optimal now, but I'd like to minimize amount of code to support. In the future, `full` can be optimized for that case (read the comment about `memset`).

rlnx:
> Support make for DPC++ we planned in this PR or in new?
@PetrovKP, I'm working on PR right now

cpp/oneapi/dal/detail/memory_impl_host.hpp
```diff
@@ -0,0 +1,52 @@
...
+inline T* malloc(const host_policy&, std::int64_t count, host_only_alloc) {
+    return new T[count];
+}
+
+template <typename T>
+inline void free(const host_policy&, T* pointer) {
+    delete[] pointer;
+}
+
+inline void memset(const host_policy&, void* dest, std::int32_t value, std::int64_t size) {
```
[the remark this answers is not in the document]
michael-smirnov: A few comments from my side:
- "Implementation is not quite safe" - I agree with you, but we haven't done services to check SDL it new interfaces yet. I think this is a common problem and do not need to be solved in this PR. In my vision, we should check SDL compliance for all new sources in next PRs, when a minimal set of functionality is implemented.
- "it's always true for 32bit systems" - AFAIK, we do not support 32bit systems anymore.
- "Put TODO here to optimize it in the future" - yes, it's a good solution, I'll follow it.
- "Ideally, it should be like this" - I started just like you with templated functions with `Policy` template. But this is not necessary, so I decided to keep it as simple as possible. In my opinion, it should look like this:
```cpp
template <typename T>
inline void fill(const host_seq_policy&, T* dest, std::int64_t count, const T& value) { ... }

template <typename T>
inline void fill(const host_par_policy&, T* dest, std::int64_t count, const T& value) { ... }

// somewhere in another file

template <typename T>
inline void fill(const dpcpp_policy&, T* dest, std::int64_t count, const T& value) { ... }
```
universal reference for `value` does not make sense here in my opinion, because we don't want to move it inside `fill()`, but copy to many elements of the `dest` array.

cpp/oneapi/dal/data/array.hpp
michael-smirnov: I think `AllocKind` gives more flexibility rather than an `Allocator`:
- Allocator is always an object of a class with a defined set of methods.
- AllocKind is just a flag indicating how we want to allocate memory. Depends on the value of this flag, we can create an appropriate allocator or use a set of malloc/free methods, for example. Without creating any allocator.
I think we should support allocators, but not in this PR. It is not a critical feature. Internally, if we want to have common codebase for Allocators and AllocKinds, we should create allocators for each AllocKind value and use them:
```cpp
template <typename T>
inline T* malloc(const host_seq_policy& policy, std::int64_t count, host_only_alloc) {
   return malloc(policy, count, backend::host_only_allocator{});
}

template <typename T, typename Allocator>
inline T* malloc(const host_policy&, std::int64_t count, Allocator&& alloc) { ... }
```

cpp/oneapi/dal/data/array.hpp
michael-smirnov: Ok, I will remove `zeros_impl()`. 
But your comments confuse me: for `memset()` you are saying that I don't have an optimized branch and should put a TODO comment. Here I have such a branch, but you want to delete it.

cpp/oneapi/dal/data/array.hpp
michael-smirnov: We even can make internal functions like malloc/memset/fill/free accepting only allocators, but for public API we need to have both versions with AllocKind and Allocator.

cpp/oneapi/dal/detail/memory_impl_dpc.hpp
PetrovKP: Ok, thanks

cpp/oneapi/dal/detail/memory_impl_host.hpp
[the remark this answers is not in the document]
rlnx: >  We haven't done services to check SDL. I think this is a common problem and do not need to be solved in this PR.

SDL is not one time activity or automated tool, but the guidelines that should maintained throughout whole process of development. Do no create problems if you can obviously avoid them. 

> "it's always true for 32bit systems" - AFAIK, we do not support 32bit systems anymore.

It does not matter, standard defines `int` as something between `short` and `long`:
```c++
sizeof(char) <= sizeof(short) <= sizeof(int) <= sizeof(long) <= sizeof(long long)
```

I'd better stick to standard rather to assumptions about hardware.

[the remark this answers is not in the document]
michael-smirnov: Argree

PR #686, opened 2020-06-28
Enable onedal_dpc build on Linux

rlnx:
Initial support for `onedal_dpc` build on Linux


rlnx:
/intelci: run
set infra_branch=dev/risrafil-dpcpp

rlnx:
/intelci: restart
set infra_branch=dev/risrafil-dpcpp

rlnx:
/intelci: run
set infra_branch=dev/risrafil-dpcpp

rlnx:
/intelci: run
set infra_branch=dev/risrafil-dpcpp

rlnx:
/intelci: run
set infra_branch=dev/risrafil-dpcpp

rlnx:
/intelci: run
set infra_branch=dev/risrafil-dpcpp

rlnx:
/intelci: run
set infra_branch=dev/risrafil-dpcpp

rlnx:
/intelci: run
set infra_branch=dev/risrafil-dpcpp

rlnx:
@PetrovKP, @michael-smirnov, @amgrigoriev, @makart19, please take a look.

examples/oneapi/dpc/source/pca/pca_cor_dense_batch.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include <iostream>
+#include <CL/sycl.hpp>
+
+#define ONEAPI_DAL_DATA_PARALLEL
+#include "oneapi/dal/algo/pca.hpp"
+#include "oneapi/dal/data/accessor.hpp"
+
+using namespace oneapi;
+
+std::ostream &operator <<(std::ostream& stream, const dal::table& table) {
```
PetrovKP, line 27: Use utills

examples/oneapi/dpc/source/pca/pca_cor_dense_batch.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    const float data_host[] = {
+        1.f,  2.f,  3.f,
+        1.f,  -1.f, 0.f,
+        4.f,  5.f,  6.f,
+        1.f,  2.f,  5.f,
+        -4.f, 3.f,  0.f
+    };
+    auto data = sycl::malloc_shared<float>(row_count * column_count, queue);
+    queue.memcpy(data, data_host, sizeof(float) * row_count * column_count).wait();
```
PetrovKP, line 71: For host or cpu device we have real copy?

examples/oneapi/dpc/daal_lnx.lst
```diff
@@ -0,0 +1,21 @@
+#===============================================================================
+# Copyright 2014-2020 Intel Corporation
```
PetrovKP, line 2:
```suggestion
# Copyright 2020 Intel Corporation
```

examples/oneapi/dpc/source/pca/pca_cor_dense_batch.cpp
```diff
@@ -0,0 +1,96 @@
...
+    std::cout << "Eigenvalues:" << std::endl
+              << result.get_eigenvalues() << std::endl;
+
+    sycl::free(data, queue);
+}
+
+int main(int argc, char const *argv[]) {
+    for (auto device : list_devices()) {
+        auto queue = sycl::queue{device};
+        run(queue);
```
PetrovKP, line 93: this does not start now for gpu now?

dev/make/cmplr.dpcpp.mk
```diff
@@ -0,0 +1,41 @@
+# file: cmplt.dpcpp.mk
+#===============================================================================
+# Copyright 2012-2020 Intel Corporation
```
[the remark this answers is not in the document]
rlnx: Why should I change it? This file exists or ages in the repo

examples/oneapi/dpc/source/pca/pca_cor_dense_batch.cpp
rlnx: We always have real copy, because `data` is a pointer a newly allocated shared memory.

examples/oneapi/dpc/source/pca/pca_cor_dense_batch.cpp
rlnx: It starts, but produces empty tables in the result.

examples/oneapi/dpc/source/pca/pca_cor_dense_batch.cpp
PetrovKP: with sycl buffer were the same problems?

examples/oneapi/dpc/source/pca/pca_cor_dense_batch.cpp
PetrovKP: Ok, thanks

examples/oneapi/dpc/source/pca/pca_cor_dense_batch.cpp
rlnx: I don't know, because it depends on buffer implementation. Per our discussion with compiler team, there are cases when host-side memory is copied inside the buffer. Standard itself guaranties no-copy if special property `use_host_ptr` is passed at the buffer creation, but says nothing about performance. 

I think it's artifact of example, if we had data source, the dataset would be written into shared memory w/o intermediate host copy. In this particular example, we could even use device memory (`malloc_device`). And probably we should since we use explicit copy.

rlnx:
/intelci: restart
set infra_branch=dev/risrafil-dpcpp

rlnx:
/intelci: restart
set infra_branch=dev/risrafil-dpcpp

cpp/oneapi/dal/detail/common_helpers.hpp
```diff
@@ -18,6 +18,7 @@
 
 #include <stdexcept>
 #include "oneapi/dal/common.hpp"
+#include "oneapi/dal/policy.hpp"
```
michael-smirnov, line 21: Why it's needed here?

cpp/oneapi/dal/policy.hpp
```diff
@@ -39,25 +40,47 @@ enum class cpu_extension : uint64_t {
     avx512 = 1U << 5
 };
 
-class ONEAPI_DAL_EXPORT default_execution_context : public base {
+class ONEAPI_DAL_EXPORT default_policy : public base {
```
michael-smirnov, line 43: Why do we need to have `default_policy` and `data_parallel_policy` in public API? I thought we discussed it and agreed that if user wants running with DPC++ he sends a queue to the `train()`, and sends nothing when wants to run on CPU. It's inconsistent with data management API

michael-smirnov, line 43: I think that for DPC++ world `default_policy` is a policy with a queue created from `default_selector`. Suggest to make its name like in data management, `host_seq_policy`, or `host_par_policy`

cpp/oneapi/dal/policy.hpp
```diff
@@ -39,25 +40,47 @@ enum class cpu_extension : uint64_t {
...
+
+#ifdef ONEAPI_DAL_DATA_PARALLEL
+class ONEAPI_DAL_EXPORT data_parallel_policy : public base {
+public:
+    data_parallel_policy(const sycl::queue& queue);
+
+    sycl::queue& get_queue() const noexcept;
+
+private:
+    dal::detail::pimpl<detail::data_parallel_policy_impl> impl_;
```
michael-smirnov, line 68: Why we need to hide this policy under the PIMPL? We need to align here between your and mine implementation: https://github.com/michael-smirnov/oneDAL/blob/dev/dpcpp-data-management/cpp/oneapi/dal/detail/common_dpc.hpp#L23

rlnx: To be consistent with the approach we use in other classes, what if we add one more member variable to the class? (binary compatibility)

cpp/oneapi/dal/policy.hpp
rlnx: We still need to discuss `seq` and `par` policies, because they slightly contradicts with the approach with two threading layers `daal_thread.so`/`daal_sequential.so`. User selects threading layer at link time (when they link their application), initial DAAL design assumes user don't need to change their code to use sequential or parallel version.

rlnx: I'm afraid we'll need to have it in public API...

The reason is that policy for CPU is used to control at least threading (number of threads) and available instruction sets.

rlnx: You are right, maybe `host_policy` is more suitable name.

PR #687, opened 2020-06-29
Sync .clang-format with the latest xLib version

rlnx:
This PR gets `.clang-format` synchronized with the latest xLib version, which was updated to address oneCCL feedback.

**Key changes**
- Do not align consecutive assignments
- Do not indent preprocessor directives
- Do not sort includes
- Do not align escaped new lines (in multiple macros)


cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_cov.cpp
```diff
@@ -27,7 +27,7 @@ using dal::backend::context_cpu;
```
KulikovNikita, line 27: Is it a good Idea to commit both clang format specification and results of it's applying to code?

cpp/oneapi/dal/backend/interop/common.hpp
```diff
@@ -46,21 +46,14 @@ struct to_daal_cpu_type<cpu_dispatch_avx512> : daal_cpu_value<daal::avx512> {};
...
-            return cpu_extension::avx2;
-        case daal::avx512:
-            return cpu_extension::avx512;
+        case daal::sse2: return cpu_extension::sse2;
+        case daal::ssse3: return cpu_extension::ssse3;
+        case daal::sse42: return cpu_extension::sse42;
+        case daal::avx: return cpu_extension::avx;
+        case daal::avx2: return cpu_extension::avx2;
+        case daal::avx512: return cpu_extension::avx512;
         case daal::avx512_mic:
```
KulikovNikita, line 55: Is it correct not to return anything? If yes why do this lines are presented here?

napetrov:
@rlnx  -  should we still integrate this?

PR #688, opened 2020-06-29
tbb update preparations

teabagk7:
Changes in makefile.


dr-pain:
Are you sure that it is needed to keep compatibility with both tbbs .2 and .12? Will mkl_fpk will work with both tbbs?

teabagk7:
> Are you sure that it is needed to keep compatibility with both tbbs .2 and .12? Will mkl_fpk will work with both tbbs?
> Why behavior made by this change isn`t compatible with MacOS?

Yeah, we need it to be able to build with old and new tbb packages.
mkl fpk won't work with different tbb versions but we are already solving this problem.
MacOS doesn't require any changes.
I'm glad you still with us! 👍

dr-pain:
> I'm glad you still with us! 👍
>  teabagk7 merged commit 10e7eda into oneapi-src:master 8 hours ago
as I see, you have no interest to discuss this question any more

PR #689, opened 2020-06-29
GBT observationsPerTreeFraction bug fix

PivovarA:
# Description
We found that using *observationsPerTreeFraction* <1.0 leads to some very wrong predictions.
After some research, it became clear that the reason for this is incorrectly built trees.
At the first iteration, the **initial estimation of y** is initialized by single value: *sum (column) / nRows*.
The final step is updating the values ​​of this estimation for OOB objects. At the first iteration, this value is obtained *sum (column) / nRows + inc*. This leads to a very large value of OOB objects, which leads to the incorrect construction of subsequent trees.
I believe that the best way to fix this is to initialize the **current estimation of y** for OOB objects with zeros in the first iteration.


cpp/daal/src/algorithms/dtrees/gbt/gbt_train_dense_default_impl.i
```diff
@@ -307,6 +307,7 @@ services::Status TrainBatchTaskBase<algorithmFPType, BinIndexType, cpu>::run(gbt
 template <typename algorithmFPType, typename BinIndexType, CpuType cpu>
 void TrainBatchTaskBase<algorithmFPType, BinIndexType, cpu>::updateOOB(size_t iTree, TreeType & t)
 {
+    double res            = _initialF;
```
ShvetsKS, line 310: const

cpp/daal/src/algorithms/dtrees/gbt/gbt_train_dense_default_impl.i
```diff
@@ -317,7 +318,7 @@ void TrainBatchTaskBase<algorithmFPType, BinIndexType, cpu>::updateOOB(size_t iT
         auto pNode = dtrees::prediction::internal::findNode<algorithmFPType, TreeType, cpu>(t, x.get());
         DAAL_ASSERT(pNode);
         algorithmFPType inc = TreeType::NodeType::castLeaf(pNode)->response;
-        pf[iRow * _nTrees + iTree] += inc;
+        pf[iRow * _nTrees + iTree] += inc - res;
```
ShvetsKS, line 321: I think clarification comment should be added to justify value decreasing.
e.g. " 'pf' buffer was already initialized by '_initialF' before first iteration"

PR #691, opened 2020-06-30
New TBB API was used

averbukh:
# Description
Temporary TBB interface usage was replaced by brand-new public API


cpp/daal/src/threading/threading.cpp
```diff
@@ -31,11 +31,15 @@
     #include <stdlib.h> // malloc and free
     #include <tbb/tbb.h>
```
alexey-katranov, line 32: If tbb.h is included, there is no big sense to include other TBB headers.

averbukh: I'll investigate it in the future. Probably, it's better to include particular headers of TBB instead, and get rid of tbb/tbb.h

alexey-katranov: Great!

cpp/daal/src/threading/threading.cpp
```diff
@@ -247,8 +251,8 @@ DAAL_EXPORT void _daal_del_mutex(void * mutexPtr)
 DAAL_EXPORT bool _daal_is_in_parallel()
 {
 #if defined(__DO_TBB_LAYER__)
-    #if defined(TBB_INTERFACE_VERSION) && TBB_INTERFACE_VERSION >= 12001
-    return tbb::detail::d1::task::current_execute_data() != nullptr;
+    #if defined(TBB_INTERFACE_VERSION) && TBB_INTERFACE_VERSION >= 12002
+    return tbb::task::current_context() != nullptr;
```
Vika-F, line 255: Shouldn't it be something like:
    #if defined(TBB_INTERFACE_VERSION) && TBB_INTERFACE_VERSION == 12001
    return tbb::detail::d1::task::current_execute_data() != nullptr;
    #elif defined(TBB_INTERFACE_VERSION) && TBB_INTERFACE_VERSION >= 12002
    return tbb::task::current_context() != nullptr;
    #else
    return tbb::task::self().state() == tbb::task::executing;
    #endif

averbukh: No, because there is no TBB with TBB_INTERFACE_VERSION == 12001. It's canceled.

PR #697, opened 2020-07-03
CPU optimizations for DBSCAN

SmirnovEgorRu:


SmirnovEgorRu:
Speedup relies on dimensions highly, but here are some numbers just for reference:

100k rows x nFeatures | master, ms | this PR, ms | Speedup
-- | -- | -- | --
10 | 4100 | 846 | 4.84
50 | 6160 | 1485 | 4.15


Out of scope of this PR:
1) noMemSafe mode optimizations
2) Distributed mode optimizations

it's quite easy to do, but need to allocate additional time for this.

PR #704, opened 2020-07-05
Fix illegal instructions on Windows

rlnx:
The problem is that on Windows Intel compiler replaces the `new BlockDescritptor<T>[n]` expression with the custom compiler-generated function that allocates and initializes all `BlockDesriptor` elements in the array. The function is compiled multiple times with different compiler options and behaves as C++ `inline` that leads to ODR violation. Some versions of that function being compiled for multiple CPUs contain AVX2 instructions. When user's application (in fact, example or test system) is linked statically against DAL, the linker selects 'wrong' version of the function (the one which contains AVX2) that leads to 'illegal' instruction on old architectures.

The fix consists in moving `new BlockDescritptor<T>[n]` to separate object file that will be compiled once to satisfy ODR. Additionally, it introduces `BlockDescriptorArray` that simplifies resource management and fixes possible memory leaks in the existing decision tree code.

### DIFF
--- a/cpp/daal/src/algorithms/decision_tree/decision_tree_train_impl.i
+++ b/cpp/daal/src/algorithms/decision_tree/decision_tree_train_impl.i
@@ -650,11 +650,14 @@ public:
         TreeNodeIndex nodeIndex = pushBack(statusPushBack);
         DAAL_CHECK_STATUS_VAR(statusPushBack)
 
-        BlockDescriptor<IndependentVariableType> * xBD = new BlockDescriptor<IndependentVariableType>[xColumnCount];
-        const IndependentVariableType ** dx            = new const IndependentVariableType *[xColumnCount];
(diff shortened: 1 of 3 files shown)


PetrovKP:
As far as I remember, PCA tests failed. Why is it fixed in the DF?

PetrovKP:
Do we really have a fails for DF?

rlnx:
> Do we really have a fails for DF?

Not DF, but decision trees, which are used within (brow|ada|logit)boost. A lot of them: http://intel-ci.intel.com/eabef4d8-0ade-f150-a92f-8cdcd4b723a1, see `new_ts_bkw`

PetrovKP:
Sorry, thanks! 😄

rlnx:
new_ts_bkw is passed on Windows.

http://intel-ci.intel.com/eabf566e-e141-f13b-a8ca-8cdcd4b723a1

PR #707, opened 2020-07-06
DBSCAN optimizations for CPU

SmirnovEgorRu:
Speedup relies on dimensions highly, but here are some numbers just for reference:

100k rows x nFeatures | master, ms | this PR, ms | Speedup
-- | -- | -- | --
10 | 4100 | 846 | 4.84
50 | 6160 | 1485 | 4.15

Out of scope of this PR:
- noMemSafe mode optimizations
- Distributed mode optimizations

it's quite easy to do, but need to allocate additional time for this.

### DIFF
--- a/cpp/daal/src/algorithms/dbscan/dbscan_dense_default_batch_impl.i
+++ b/cpp/daal/src/algorithms/dbscan/dbscan_dense_default_batch_impl.i
@@ -40,6 +40,7 @@ namespace dbscan
 namespace internal
 {
 #define __DBSCAN_PREFETCHED_NEIGHBORHOODS_COUNT 64
+#define __DBSCAN_MAXIMUM_NESTED_STACK_LEVEL     200
 
 template <typename algorithmFPType, Method method, CpuType cpu>
 Status DBSCANBatchKernel<algorithmFPType, method, cpu>::processNeighborhood(size_t clusterId, int * const assignments,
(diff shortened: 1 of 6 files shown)


vepifanov:
Looks great!
Right now, I don't have any objections on code.

Btw, it would be interesting to see performance comparison on various scenarios, that can appear during hyperparameters tuning:
- Small epsilon, when all neighborhoods are very small and excessive nested parallelism can give big overhead;
- Large espilon, when all neighborhoods are very big and data racing effects can appear (simultineous checks if point is already clustered can lead to situation, when point will be added to multiple queues in different threads).

SmirnovEgorRu:
@vepifanov 

Eps | master, ms | this PR, ms | Speedup
-- | -- | -- | --
->0 | 2440.84 | 200.148 | 12.2
36 | 4018.27 | 690.756 | 5.82
1e10 | 27167.3 | 20293.1 | 1.34

100k rows x 10 features

When all samples are one cluster I suppose we spend most of time in memory allocation/memcopy/deallocation. Potentially it can be optimized by developing special allocator with some logic.

SmirnovEgorRu:
There are some issues in CI, but there is nothing new introduced with my PR.

SmirnovEgorRu:
Ping @vepifanov, @ShvetsKS, @PetrovKP

PR #708, opened 2020-07-06
fix multiple deffinition in error handling

mchernov-intel:
moved from: #706


mchernov-intel:
@PetrovKP please, approve

PetrovKP:
Here fix this problem https://github.com/oneapi-src/oneDAL/pull/705/files#diff-c995c6a5b788b22535a0a0927fff956dR1. Maybe we need to do also?

cpp/oneapi/dal/backend/interop/error_converter.hpp
```diff
@@ -20,253 +20,9 @@
...
-        case ErrorID::ErrorBzip2MemoryAllocationFailed:
-            throw bad_alloc();
-        default:
-            throw internal_error("Unknown status code");
-    }
-}
+void status_to_exception(const daal::services::Status& s);
 
 template <class StatusConverter>
-void status_to_exception(const daal::services::Status& s, StatusConverter alg_converter) {
```
rlnx, line 264: This is template function. It must be defined in header files with `inline`:
```cpp
template <class StatusConverter>
inline void status_to_exception(const daal::services::Status& s, const StatusConverter& alg_converter) {
   ...
}
```

cpp/oneapi/dal/backend/interop/error_converter.cpp
```diff
@@ -0,0 +1,270 @@
...
+        case ErrorID::ErrorZlibMemoryAllocationFailed:
+        case ErrorID::ErrorBzip2MemoryAllocationFailed:
+            throw bad_alloc();
+        default:
+            throw internal_error("Unknown status code");
+    }
+}
+
+template <class StatusConverter>
+void status_to_exception(const daal::services::Status& s, StatusConverter alg_converter) {
```
rlnx, line 262: Template function in `*cpp` without explicit instantiation will not work.

PR #723, opened 2020-07-09
made solution generator cross python

vmeshche:
# Description
I have fixed a problem: solution generator support python 3.6+ only. made tool cross python.

- python 3.6+ specific calls replaced with universal.


deploy/local/generate_win_solution.py
```diff
@@ -140,9 +140,14 @@ def generate_proj(config, relative_example_path):
 
 def find_all_examples(examples_dir):
     examples = []
-    with cd(examples_dir):
-        for relative_path in glob.iglob('**/*.cpp', recursive=True):
-            examples.append(relative_path)
+
+    print('Walking dir {}'.format(examples_dir))
+    for root, dirnames, filenames in os.walk(examples_dir):
+        print('-> checking file: {}{}'.format(root,filenames))
```
rlnx, line 146: Is this print really necessary or it was added for debugging? We already have prints in  `generate_sln` and `write_proj`.

vmeshche: I will remove it. Let me complete testing

PR #726, opened 2020-07-10
Removed NULLs in GPU K-Means

amgrigoriev:
-
-
-


napetrov:
GNU build looks ok for this change - there is no warnings

PR #735, opened 2020-07-13
MPI samples launcher fix for VS2017 [beta 08 rls]

Alexsandruss:
# Description
Fix for infinite compilation of MPI samples with oneDAL built by VS2017.

Old line `call mpiicc %LFLAGS% %RESULT_DIR%\%%T.obj %LFLAGS_DAAL% -Fe%RESULT_DIR%\%%T.exe 2>&1 >> %MPI_LOGFILE%` leads to infinite compilation and error 10014 (code -42) on cancelling.

Example of failed job:
http://intel-ci.intel.com/eac15e34-a163-f19a-92ed-8cdcd4b723a1

Successful PCm with changes:
http://intel-ci.intel.com/eac4f9c0-00cd-f13e-8adb-8cdcd4b723a1

Duplicate of https://github.com/oneapi-src/oneDAL/pull/734

### DIFF
--- a/samples/daal/cpp/mpi/launcher.bat
+++ b/samples/daal/cpp/mpi/launcher.bat
@@ -73,10 +73,8 @@ setlocal enabledelayedexpansion enableextensions
 for %%T in (%MPI_SAMPLE_LIST%) do (
 
     if not "%RMODE%"=="run" (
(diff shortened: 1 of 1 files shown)


SmirnovEgorRu:
For hotfix it's okay.
But we need to align behavior across operation systems. Now you don't create intermediate obj files on Win, we need to do this for other OS too.

vepifanov:
Actually I think it does create .obj files, otherwise it would fail on next step where we combine .obj file with another set of libs (LFLAGS_DAAL_DLL).

PR #738, opened 2020-07-13
Removing old tbb dependencies 

napetrov:
* fix of spark and hadoop launcher scripts

* remove tbb deprecated header

Co-authored-by: konstantin.nektyagaev @teabagk7


cpp/daal/src/threading/service_thread_pinner.cpp
```diff
@@ -37,7 +37,6 @@
         #include "tbb/tbb.h"
         #include <tbb/task_arena.h>
```
teabagk7, line 38: Potentially it'll be good to remove all of them excluding "tbb/tbb.h"
But let's do it next time 🤪

PR #764, opened 2020-07-23
Modifying the input check for normEqDense for linear regression 

PivovarA:
condition for normEqDense has changed,  so no need to add interceptFlag.


pdxjohnny:
Any update on this?

ShvetsKS (approved):

PR #768, opened 2020-07-24
[DOC] Fixed paths to examples

outoftardis:
# Description

- Updated paths to examples
- Fixed a typo in the name of the lib file


outoftardis:
I removed the updates to README from this PR, I'll add them to #767

PR #788, opened 2020-08-04
Update interfaces in headers for SVM algorithm

PetrovKP:


napetrov:
@PetrovKP  do we need this request?

PR #790, opened 2020-08-05
Fixing CSS style code problems

napetrov:
Fixing CSS code problems


napetrov:
@aegorova651 - thanks.  I don't think quotation is required for fonts - only for font names with spaces.  Overall this is not the only place - i see lot of inconsistency across document on formatting anyway.  + other linters generate other warnings for this css. So wide overall question would  be - do we have a good one that should be used across components

outoftardis (approved):

Issue #243, opened 2019-12-19
env/vars.sh script does not work as expected with zsh


rlnx:
Incorrect `LIBRARY_PATH`, `CPATH`, etc. after calling env script in **Z shell**
```sh
source ./__release_mac_clang/daal/latest/env/vars.sh`
```

After that I see _wrong paths_ to the libs and includes: 
```sh
CPATH=/Users/ruslanis/Locrepo/daal/include
LIBRARY_PATH=/Users/ruslanis/Locrepo/daal/lib
DYLD_LIBRARY_PATH=/Users/ruslanis/Locrepo/daal/lib
```

Expected paths are:
```sh
CPATH=/Users/ruslanis/Locrepo/daal/__release_mac_clang/daal/latest/include
LIBRARY_PATH=/Users/ruslanis/Locrepo/daal/__release_mac_clang/daal/latest/lib
DYLD_LIBRARY_PATH=/Users/ruslanis/Locrepo/daal/__release_mac_clang/daal/latest/lib
```

**Environment:**
- OS: macOS Catalina 10.15.2
- DAL version: master, 246591d
- zsh version: zsh 5.6.2 (x86_64-apple-darwin18.0.0)

PR #813, opened 2020-08-12
Cleaning up env scripts to support zsh

napetrov:
# Description
- Cleaning remains of arch support from env scripts 
- Enable zsh support 
- Unify shell based env scripts
- Fixes https://github.com/oneapi-src/oneDAL/issues/243


napetrov:
/intelci: restart
set infra_branch=dev/napetrov_zsh

PR #815, opened 2020-08-12
Added basic csv reader implementation

vepifanov:
Added basic csv reader implementation using existing implementation from old interfaces.
Currently limited to reading into float32 memory.


cpp/oneapi/dal/data/backend/csv_table_reader_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+        data_source.getFeatureManager().setDelimiter(delimiter_);
+        data_source.loadDataBlock();
+
+        auto nt = data_source.getNumericTable();
+
+        daal::data_management::BlockDescriptor<DAAL_DATA_TYPE> block;
+        const std::int64_t row_count    = nt->getNumberOfRows();
+        const std::int64_t column_count = nt->getNumberOfColumns();
+
+        nt->getBlockOfRows(0, row_count, daal::data_management::readOnly, block);
```
amgrigoriev, line 77: Check return code

cpp/oneapi/dal/data/backend/csv_table_reader_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+        DAAL_DATA_TYPE* data = block.getBlockPtr();
+
+        auto arr = array<DAAL_DATA_TYPE>::empty(queue, row_count * column_count);
+        detail::memcpy(queue, arr.get_mutable_data(), data, sizeof(DAAL_DATA_TYPE) * row_count * column_count);
```
amgrigoriev, line 81: Does it return status?

cpp/oneapi/dal/data/backend/csv_table_reader_impl.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+        nt->releaseBlockOfRows(block);
```
amgrigoriev, line 83: Check return code

cpp/oneapi/dal/data/backend/csv_table_reader_impl.hpp
vepifanov: Fixed.

cpp/oneapi/dal/data/backend/csv_table_reader_impl.hpp
vepifanov: Fixed.

cpp/oneapi/dal/data/backend/csv_table_reader_impl.hpp
vepifanov: It calls std::memcpy which should throw exception.

cpp/oneapi/dal/io/csv/common.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include "oneapi/dal/table/common.hpp"
+
+namespace oneapi::dal::csv {
+
+namespace detail {
+struct data_source_tag {};
+class data_source_impl;
+} // namespace detail
+
+class ONEAPI_DAL_EXPORT data_source_base : public base {
```
michael-smirnov, line 29: Why do you need `data_source_base`? Asking because this technique applied with templated descriptors, but `data_source` has no template params

cpp/oneapi/dal/io/csv/common.hpp
```diff
@@ -0,0 +1,67 @@
...
+public:
+    using tag_t = detail::data_source_tag;
+
+    data_source_base(const char* file_name);
+
+    auto get_delimiter() const -> char;
```
michael-smirnov, line 35:
```suggestion
    char get_delimiter() const;
```
Consistent style

michael-smirnov: Probably, we need to use `std::uint8_t` instead of `char`. @rlnx ?

cpp/oneapi/dal/io/csv/detail/read_ops.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include "oneapi/dal/table/common.hpp"
+
+namespace oneapi::dal::csv::detail {
+
+template <typename table, typename Context, typename... Options>
+struct ONEAPI_DAL_EXPORT read_ops_dispatcher {
+    table operator()(const Context&, const data_source_base&, const read_args<table>&) const;
+};
+
+template <typename table, typename Descriptor>
```
michael-smirnov, line 30:
```suggestion
template <typename Result, typename Descriptor>
```

cpp/oneapi/dal/io/csv/common.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+protected:
+    void set_delimiter_impl(char value);
+    void set_parse_header_impl(bool value);
+    void set_file_name_impl(const char*);
+
+    dal::detail::pimpl<detail::data_source_impl> impl_;
+};
+
+class data_source : public data_source_base {
```
michael-smirnov, line 47: ONEAPI_DAL_EXPORT ?

examples/oneapi/cpp/source/pca/pca_cor_dense_batch.cpp
```diff
@@ -15,27 +15,22 @@
...
-    const float data[] = {
-        1.f,  2.f,  3.f,
-        1.f,  -1.f, 0.f,
-        4.f,  5.f,  6.f,
-        1.f,  2.f,  5.f,
-        -4.f, 3.f,  0.f
-    };
+int main(int argc, char const *argv[]) {
+    const auto data_source = dal::csv::data_source(data_file_name)
+        .set_delimiter(',');
```
michael-smirnov, line 28: Can we avoid creating data source explicitly by the use of default params and leave only the call of the `read()` function?
Or this is done specifically for some reason?

examples/oneapi/dpc/source/pca/pca_cor_dense_batch.cpp
```diff
@@ -20,29 +20,22 @@
```
michael-smirnov, line 20: @vepifanov do you plan to update all examples with use of `data_source`?

vepifanov: This is discussable - we can use data source in all examples or just introduce separate example specifically for data source functionality. I modified pca example for testing for now.

examples/oneapi/cpp/source/pca/pca_cor_dense_batch.cpp
vepifanov: I believe there are now mechanisms during evaluating of read parameters to dispatch string into constructor of data_source, so it must be created explicitly. Moreover, data_source itself is a starting point for dispatcher to evaluate whole 'read' operation.

cpp/oneapi/dal/io/csv/common.hpp
vepifanov: Removed _base one.

cpp/oneapi/dal/io/csv/common.hpp
vepifanov: Fixed style.
I think `char` is more reasonable, because it refers to symbols rather than some integer value.

cpp/oneapi/dal/io/csv/detail/read_ops.hpp
vepifanov: Fixed.

cpp/oneapi/dal/io/csv/common.hpp
vepifanov: Fixed.

examples/oneapi/cpp/source/pca/pca_cor_dense_batch.cpp
vepifanov: Turned into one-liner.

cpp/oneapi/dal/io/csv/BUILD
```diff
@@ -0,0 +1,16 @@
...
+dal_public_includes(
+    name = "public_includes",
+)
+
+dal_module(
+    name = "csv",
+    dpc = True,
+    auto = True,
+    deps = [
+        "@onedal//cpp/oneapi/dal:common",
```
rlnx, line 13: `core` already contains `common`, so can be removed

cpp/oneapi/dal/BUILD
```diff
@@ -56,10 +57,18 @@ dal_algos(
     ]
 )
 
+dal_ios(
```
rlnx, line 60: Can we generalize `dal_algos` to not duplicate code in `dal_ios`, for example `dal_modules_group`?

vepifanov: Done.

cpp/oneapi/dal/io/csv/BUILD
vepifanov: Fixed.

cpp/oneapi/dal/io/csv/common.cpp
```diff
@@ -0,0 +1,65 @@
...
+class detail::data_source_impl : public base {
+public:
+    char delimiter    = ',';
+    bool parse_header = false;
+    char *file_name   = nullptr;
+};
+
+using detail::data_source_impl;
+
+data_source::data_source(const char *file_name) : impl_(new data_source_impl{}) {
```
rlnx, line 31: Why not `std::string`, but `const char*`?

rlnx: If you care about dependencies on `std`, then we already have dependency on string in exceptions (in constructors of `std::invalid_argument`, `std::runtime_error`, etc.)

PR #829, opened 2020-08-16
Build and run tests using Bazel

rlnx:
Add all GoogleTest-based unit test to bazel.

Run tests for your piece of functionality (for example accessors) using this command: 
```sh
bazel test //cpp/oneapi/dal/table:accessor_tests
```

Test targets are listed in corresponding BUILD files, for example, see `cpp/oneapi/dal/table/BUILD`


cpp/oneapi/dal/table/BUILD
```diff
@@ -14,3 +18,29 @@ dal_module(
...
+    deps = [ ":table" ],
+)
+
+dal_test_suite(
+    name = "builder_tests",
+    srcs = [
+        "detail/table_builder_test.cpp",
+    ],
+    deps = [ ":table" ],
+)
```
michael-smirnov, line 46: For building, we have only one rule for `table` folder named the same. For tests, we have separate test suites for every table sub-component. I think we should have a bijection between components and test rules - it is much simpler. Also, it would be good if new tests will switch on automatically - now I have to specify them manually, which adds a risk if I forget to do it. For builders, we can have a separate test rule, though, while they are internal.

cpp/daal/BUILD
```diff
@@ -65,8 +65,11 @@ daal_module(
 
 daal_module(
     name = "services",
-    srcs = glob(
-        ["src/services/**/*.cpp", "src/externals/**/*.cpp"],
+    srcs = glob([
+            "src/services/**/*.cpp",
+            "src/externals/**/*.cpp",
+            "src/algorithms/*.cpp",
```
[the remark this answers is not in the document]
rlnx: Not all algorithms, but only service functionality in algorithms. The problem is `src/algorithms/(*.h|*.cpp)` defines such classes as `Argument`, `HashTable`, etc., which are mostly services rather than algorithms (don't ask my why they are in algorithms). In fact functionality in `src/services` depends on it.

cpp/oneapi/dal/table/BUILD
rlnx: Well, I'd like to have meaningful categories for the tests. So `table_tests` are collection of tests for tables, `accessor_tests` is a collection of tests for all accessors.

> bijection between components and test rules - it is much simpler

I'm afraid not "simpler" for the person who develops new functionality, because otherwise he/she will not be able run all tests for accessors, all tests for tables and all tests for builders, but have to run all the tests at once or run every test manually (run manually test for `row_accessor`, `column_accessor`, etc).

> Also, it would be good if new tests will switch on automatically

Manual specification is required for categorization into test suites.

> which adds a risk if I forget to do it

I don't see risk, because you'll need to add them to BUILD file in order to run. I'm not considering the case then someone adds the test without running it 😄

cpp/daal/BUILD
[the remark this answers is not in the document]
michael-smirnov: Ok, it's fine

cpp/oneapi/dal/table/BUILD
michael-smirnov: I think here we might have two scenarios:
1. Someone wants to run a particular test to see how it is going - in this case we can provide rules for every .cpp test file or make the user run them manually.
2. Run all the tests for the whole component to make sure it works fine - In this case, I suggest to think about component with the relation of directories - we have `io` component, `algo` compoment with sub-components like `knn`, `pca`, `svm` and `table` component. Do we have a reason to divide in into sub-components like accessors? I'm not sure. Moreover, I think it's too confusing to have `table` sub-component into the `table` folder.

cpp/oneapi/dal/algo/svm/BUILD
```diff
@@ -16,3 +20,14 @@ dal_module(
         "@onedal//cpp/daal/src/algorithms/svm:kernel",
     ]
 )
+
+dal_test_suite(
+    name = "cpu_tests",
+    srcs = [
+        # Diabled
+        # "backend/cpu/train_kernel_test.cpp",
```
michael-smirnov, line 28: Need to uncomment? 
Especially since we can disable tests straight into their sources: https://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test

cpp/oneapi/dal/table/BUILD
rlnx: 1. It's done, you can run each test using name of the file (without extension).

2. It's available, you can run all tests from component using `//cpp/oneapi/dal/table:all`. 

3. We need more granular specification for the tests. It may be to much for the developer to run all the tests per component. Especially it's important for CPU/GPU backends, user must be able to build/run CPU and GPU tests separately.

> I think it's too confusing to have table sub-component into the table folder.

Why do you think it's confusing? In fact, you have different components in table directory: `table`, `accessor`, `builder`. I cannot remember we discussed rules about one component per directory.

#### One more observation
Manual specification is required if you'll need to specify addition options such as dependencies or [tags](https://docs.bazel.build/versions/master/user-manual.html#flag--test_tag_filters).

```bazel
dal_test_suite(
   name = "component_tests",
   srcs = [
      "component_first_test.cpp",
      "component_second_test.cpp",
   ],
   deps = [
      "subcomponetns_which_those_particular_tests_depends_on",
   ],
   test_deps = [
      "extra_dependencies_for_that_test_suite",
   ],
   tags = [ "gemm", "functional" ],
)
```

You are proposing to have one test suite for all the tests in directory

```bazel
dal_test_suite(
   name = "component_tests",
   srcs = grep([*_test.cpp]),
   deps = [
      "coarse_grained_component",
   ],
   test_deps = [
      "???",
   ],
   tags = [ "???" ],
)
```

What should I do, if I want to specify tags/dependencies for particular tests?

cpp/oneapi/dal/algo/svm/BUILD
rlnx: Ok

cpp/oneapi/dal/table/BUILD
michael-smirnov: I think it is ok for now, your improvements look great. In the future, we can adapt how to organize test suites to the current situation. It is ok for now to go with your proposal.
For my point of view, here are two concerns:
1. Having a table component inside the table directory seems very strange for me (in terms of naming - just because of repeating the same name. It could be named common, for example, to specify that other components in that directory are "auxiliary" for the main one). For table builders - it is very arguable that they are independent component since they are very bound to each specific table type.

michael-smirnov (approved):
Approved since only minor questions still exist. Please pay attention to them.

napetrov (approved):

makart19 (approved):

PR #831, opened 2020-08-17
Fix conformance testing

Pahandrovich:


.ci/scripts/conformance-scripts/run_tests.py
```diff
@@ -7,7 +7,7 @@
 
 algs_filename = "algorithms.txt"
 report_filename = "report.html"
-sklearn_version = "0.23.1" if len(sys.argv) == 1 else sys.argv[1]
+sklearn_version = "0.23.2" if len(sys.argv) == 1 else sys.argv[1]
```
PetrovKP, line 10: Can we download the latest release version automatically?

Pahandrovich: We can download latest 0.23.X version, but if this version will change on 0.24, then we must manually fix it

PetrovKP (approved):
I see now conformance testing works. And added the pulling of tests from the last package. Good job!

PR #834, opened 2020-08-17
enable initial mergify config for labels

napetrov:
# Description
In this change we enable mergify configuration to use 'backport-2020' and 'backport-2021' to be used for automatic backport PRs creation
https://doc.mergify.io/examples.html#using-labels-to-backport-pull-requests

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,17 @@
+pull_request_rules:
+  - name: backport patches to 2021 branch
+    conditions:
+      - base=master
+      - label=backport-2021
+    actions:
+      backport:
+        branches:
+          - rls/2021-gold-mnt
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -0,0 +1,17 @@
...
+        branches:
+          - rls/2021-gold-mnt
+  - name: backport patches to 2020 branch
+    conditions:
+      - base=master
+      - label=backport-2020
+    actions:
+      backport:
+        branches:
+          - rls/daal-2020-mnt
```
PetrovKP, line 17: I think this is already pointless. the branches diverged colossally and do not merge automatically.

napetrov: Well -a leas this might help with some cases

PR #835, opened 2020-08-17
Reduce of calls memory allocate for cache in SVM

PetrovKP:


PetrovKP:
nightly build: http://intel-ci.intel.com/eae0e0c3-9fb1-f101-8ffc-8cdcd4b723a1

PR #851, opened 2020-08-20
Dispatching by offset in Buffer type casting

KulikovNikita:
-

### DIFF
--- a/cpp/daal/include/sycl/internal/buffer_utils.h
+++ b/cpp/daal/include/sycl/internal/buffer_utils.h
@@ -129,11 +129,11 @@ public:
     {
         _st = services::Status();
 
-        auto buffer    = _src.template get<DataType>();
+        auto buffer = _src.template get<DataType>();
         if (_offset != 0)
         {
             auto subbuffer = buffer.getSubBuffer(_offset, _size);
-            _dest = subbuffer;
+            _dest          = subbuffer;
         }
         else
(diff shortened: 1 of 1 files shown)


cpp/daal/include/sycl/internal/buffer_utils.h
```diff
@@ -129,10 +129,16 @@ class BufferConverterTo
     {
         _st = services::Status();
 
-        auto buffer    = _src.template get<DataType>();
-        auto subbuffer = buffer.getSubBuffer(_offset, _size);
-
-        _dest = subbuffer;
+        auto buffer = _src.template get<DataType>();
```
michael-smirnov, line 132: Can you please also create a specialization of `operator()` in `BufferConverterFrom` when it is no need to perform conversion?

cpp/daal/include/sycl/internal/buffer_utils.h
```diff
@@ -129,10 +129,16 @@ class BufferConverterTo
...
+        if ((_offset == 0) && (_size == buffer.size()))
```
rlnx, line 133: Why this dispatching is done here, but not [there](https://github.com/oneapi-src/oneDAL/blob/c19d88eca7c627e4cb79833062c695bdddfbf986/cpp/daal/include/services/buffer.h#L190)?

PR #875, opened 2020-08-29
Build dynamic libraries using Bazel

rlnx:


cpp/oneapi/dal/algo/svm/backend/cpu/train_kernel_test.cpp
```diff
@@ -356,7 +356,7 @@ TEST(svm_thunder_dense_test, can_classify_any_two_labels) {
 
         auto support_indices_table = result_train.get_support_indices();
         const auto support_indices = row_accessor<const float>(support_indices_table).pull();
-        for (size_t i = 0; i < support_indices.get_count(); i++) {
+        for (std::int64_t i = 0; i < support_indices.get_count(); i++) {
             ASSERT_EQ(support_indices[i], i);
         }
```
KulikovNikita, line 362: Does this changes affect process building dynamic libraries anyhow?

KulikovNikita (approved):

cpp/oneapi/dal/algo/svm/backend/cpu/train_kernel_test.cpp
rlnx: No, this is the fix of compiler's warning

napetrov:
There is copyrights issue that have to be fixed. also might be good to make a rebase - there are some strange failures in ci.

BUILD
```diff
@@ -23,6 +23,10 @@ release(
         "@onedal//cpp/daal:core_static",
         "@onedal//cpp/daal:thread_static",
         "@onedal//cpp/daal:sequential_static",
+        "@onedal//cpp/daal:core_dynamic",
+        "@onedal//cpp/daal:thread_dynamic",
+        "@onedal//cpp/daal:sequential_dynamic",
         "@onedal//cpp/oneapi/dal:static",
+        "@onedal//cpp/oneapi/dal:dynamic",
```
michael-smirnov, line 30: Does it mean we don't want to have several static/dynamic libs of oneDAL (thread/sequential)?

cpp/daal/BUILD
```diff
@@ -1,13 +1,40 @@
...
     "daal_generate_version",
     "daal_patch_kernel_defines",
 )
 
+daal_module(
+    name = "microvmlipp",
+    override_deps_lib_tag = True,
+    deps = [
+        "@micromkl//:vml_ipp",
+        # FIXME: Currently vml_ipp lib depends on TBB, but it shouldn't
```
michael-smirnov, line 16: FIXME vs TODO - I think we should use one word in the library

examples/oneapi/cpp/BUILD
```diff
@@ -49,10 +49,21 @@ dal_example_suite(
     test_deps = _TEST_DEPS,
 )
 
+dal_example_suite(
```
michael-smirnov, line 52: Don't we want to say it's `dal_preview_example_suite`?

BUILD
rlnx: We already have `onedal_thread.so`/`onedal_sequential.so` as part of the old DAAL build. 

See lines 
```bazel
"@onedal//cpp/daal:thread_static",
"@onedal//cpp/daal:sequential_static",
"@onedal//cpp/daal:thread_dynamic",
"@onedal//cpp/daal:sequential_dynamic",
```
The libraries that correspond to these targets are named `onedal_thread/sequential.so/a`.

examples/oneapi/cpp/BUILD
rlnx: It makes sense only if the rules are different, currently they are the same for preview/stable functionality. I don't see much sense in having alias for `dal_example_suite`.

cpp/daal/BUILD
rlnx: Ok

examples/oneapi/cpp/BUILD
rlnx: We have `jaccard` examples in separate rule, just because the example explicitly use TBB functions, so we need dependency on TBB for this particular example.

BUILD
michael-smirnov: I see that we have this for DAAL. Does it mean that all the kernels we develop in new interfaces will be placed in `onedal_thread/sequential.so/a` alongside with old kernels? How this will be organized?

rlnx: There is no kernels inside `onedal_thread` or `onedal_sequential`, this is abstraction layer over threading routines.

PR #892, opened 2020-09-03
Update version to 2021.2

homksei:


makefile.ver
```diff
@@ -15,10 +15,10 @@
 #===============================================================================
 
 MAJOR   =       2021
-MINOR   =       1
-UPDATE  =       9
+MINOR   =       0
```
napetrov, line 18: should be 2

makefile.ver
```diff
@@ -15,10 +15,10 @@
...
+UPDATE  =       2
```
napetrov, line 19: should be 0

maria-Petrova (approved):

PR #894, opened 2020-09-07
Conformance between spec and API: error hadnling

mchernov-intel:
- unimplemented_error -> unimplemented
- erase unavailable_error
- add unsupported_device
- add host_bad_alloc & device_bad_alloc

### DIFF
--- a/cpp/oneapi/dal/algo/knn/backend/gpu/train_kernel_brute_force_dpc.cpp
+++ b/cpp/oneapi/dal/algo/knn/backend/gpu/train_kernel_brute_force_dpc.cpp
@@ -67,7 +67,7 @@ static train_result call_daal_kernel(const context_gpu& ctx,
 
     daal::algorithms::classifier::ModelPtr model_ptr(new daal_knn::Model(column_count));
     if (!model_ptr) {
-        throw bad_alloc();
+        throw host_bad_alloc();
     }
 
     auto knn_model = static_cast<daal_knn::Model*>(model_ptr.get());
--- a/cpp/oneapi/dal/exceptions.hpp
+++ b/cpp/oneapi/dal/exceptions.hpp
@@ -82,9 +82,17 @@ public:
     const char* what() const noexcept override;
 };
 
-class ONEAPI_DAL_EXPORT bad_alloc : public exception, public std::bad_alloc {
+class ONEAPI_DAL_EXPORT bad_alloc : public exception, public std::bad_alloc {};
+
+class ONEAPI_DAL_EXPORT host_bad_alloc : public bad_alloc {
 public:
-    bad_alloc() noexcept = default;
+    host_bad_alloc() noexcept = default;
+    const char* what() const noexcept override;
+};
+
+class ONEAPI_DAL_EXPORT device_bad_alloc : public bad_alloc {
+public:
+    device_bad_alloc() noexcept = default;
     const char* what() const noexcept override;
 };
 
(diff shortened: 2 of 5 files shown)


cpp/oneapi/dal/exceptions.hpp
```diff
@@ -82,9 +82,17 @@ class ONEAPI_DAL_EXPORT internal_error : public runtime_error, public std::runti
...
-class ONEAPI_DAL_EXPORT bad_alloc : public exception, public std::bad_alloc {
+class ONEAPI_DAL_EXPORT bad_alloc : public exception, public std::bad_alloc {};
+
+class ONEAPI_DAL_EXPORT host_bad_alloc : public bad_alloc {
+public:
+    host_bad_alloc() noexcept = default;
+    const char* what() const noexcept override;
+};
+
+class ONEAPI_DAL_EXPORT device_bad_alloc : public bad_alloc {
```
michael-smirnov, line 93: Don't we want to add information about which device cause the exception?

cpp/oneapi/dal/algo/knn/backend/gpu/train_kernel_brute_force_dpc.cpp
```diff
@@ -67,7 +67,7 @@ static train_result call_daal_kernel(const context_gpu& ctx,
 
     daal::algorithms::classifier::ModelPtr model_ptr(new daal_knn::Model(column_count));
     if (!model_ptr) {
```
michael-smirnov, line 69: Why you think that `model_ptr` will be `nullptr` when memory allocation is failed? I think `new daal_knn::Model(column_count)` would throw `std::bad_alloc` if `operator new` is not overloaded. Is it overloaded by DAAL in the way it simply returns `nullptr` in case of bad allocation?

mchernov-intel: fixed

cpp/oneapi/dal/exceptions.hpp
mchernov-intel: My belief - no. `std::bad_alloc` provides only default message to save the memory. So, if we see strong need, new sub-classes like `dal::gpu_bad_alloc` will be implemented.

mchernov-intel: please, resolve if agree

cpp/oneapi/dal/algo/knn/backend/gpu/train_kernel_brute_force_dpc.cpp
michael-smirnov: @mchernov-intel so what's the conclusion? I see that you changed this behavior, but why? I just asked you if you are sure you are doing right. Have you checked that new operator used in `Model` really throws `bad_alloc`? I remember that for DAAL objects `new` and `delete` are overloaded.

cpp/oneapi/dal/exceptions.hpp
michael-smirnov: Seems very strange for me to have an exception class for each device type like gpu, fpga. etc. What if I have two GPUs? Which device cases exception? How to know that? 
Have you looked at how `device_bad_alloc` is organized in other oneAPI libraries?

cpp/oneapi/dal/algo/knn/backend/gpu/train_kernel_brute_force_dpc.cpp
mchernov-intel: resolved via status

cpp/oneapi/dal/exceptions.hpp
mchernov-intel: decided to postpone.

mchernov-intel:
@michael-smirnov please, review

michael-smirnov (commented):
@mchernov-intel what's about `unimplemented_error`? I remember there was a discussion about rename it to `unimplemented` to be aligned with MKL.

mchernov-intel:
@michael-smirnov precommit failed, but my vision this changes doesn't connect to precommit failures.

michael-smirnov (approved):
My recommendation is to wait until the master is fixed, rebase, re-run CI, and then merge.

PR #895, opened 2020-09-07
Update version to 2020.4

homksei:


SmirnovEgorRu:
It looks like that we can keep Update3 as is.

PR #906, opened 2020-09-10
USM support for accessors

michael-smirnov:
In this PR I introduced support of USM memory allocations inside pull/push methods of `row_accessor` and `column_accessor` accepting tables (empty and homogen) and homogen builder.


cpp/oneapi/dal/table/backend/homogen_table_impl.cpp
```diff
@@ -23,250 +23,288 @@ namespace oneapi::dal::backend {
...
-        detail::get_data_type_size(feature_type) * (idx + rows.start_idx * column_count);
+        detail::get_data_type_size(table_dtype) * (column_index + rows.start_idx * column_count);
 
-    if (block_dtype == feature_type && column_count == 1) {
+    if (block_dtype == table_dtype && column_count == 1) {
         if (reinterpret_cast<const void*>(data_.get_data() + row_offset) !=
             reinterpret_cast<const void*>(block.get_data())) {
-            data_.need_mutable_data();
+            make_mutable_data(policy, data_);
             auto dst_ptr = data_.get_mutable_data() + row_offset;
```
amgrigoriev, line 252: It doesn't look consistent with check at line 249: data_.get_data() + row_offset against data_.get_mutable_data() + row_offset

michael-smirnov: If array has mutable data, they both return the same pointer. Otherwise `get_mutable_data()` throws an exception.
Why it's inconsistent? Condition at the 249 line checks is block use the memory directly from the table. If not, we shall copy the data from the block to the table.

rlnx (approved):

PR #918, opened 2020-09-11
Java Interfaces for Brute-Force kNN

owerbat:


PetrovKP (commented):
Need aligned examples with cpp. For example struct directories:

k_nearest_neighbors
- bf_knn_dense_batch
- kdtree_knn_dense_batch

ShvetsKS (approved):

PR #928, opened 2020-09-14
Bug fix for equal indices for kd-tree kNN

averbukh:
# Description
Bug fix with equal indices


Alexsandruss (approved):
I checked on few cases, bug disappeared.

averbukh:
Issue with Java is not related to this PR

PR #930, opened 2020-09-14
Initial version of tbb env in build

napetrov:
# Description
In this change we enable initial version of windows ci

- Fix download scripts on windows
- Enable NuGet package download for tbb
- Cleanup makes from 32 bit targets
- Update install instructions


.ci/env/tbb.bat
```diff
@@ -0,0 +1,67 @@
+@echo off
+rem ============================================================================
+rem Copyright 2018-2020 Intel Corporation
```
averbukh, line 3: 2020

INSTALL.md
```diff
@@ -100,21 +101,21 @@ For details, see [System Requirements for oneDAL](https://software.intel.com/con
 
 8. Build oneDAL via command-line interface. Choose the appropriate commands based on the platform and the compiler you use:
```
outoftardis, line 102: I'd add here that not only the platform and the compiler matter but also the interfaces (DAAL vs. oneAPI)

INSTALL.md
```diff
@@ -19,6 +19,7 @@
 
 Required Software:
 * C/C++ Compiler
+* DPC++ Compiler
```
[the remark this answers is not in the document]
napetrov: Any. Although there is only intel implementation only at the moment. But assuming there will be another one implemented - this should work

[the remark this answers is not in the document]
outoftardis: sure

INSTALL.md
napetrov: fixed

outoftardis (approved):

homksei (approved):

PR #954, opened 2020-09-17
[2020-mnt] kNN kdtree class voting fix

Alexsandruss:
# Description
Fix for k-Nearest Neighbors kdtree class voting


Alexsandruss:
Close due to end of support of 2020-mnt branch. See #962

PR #956, opened 2020-09-17
Reimplementing setters for Jaccard result (align API with ML part)

orrrrtem:
Aligning of the result constructing with oneDAL algos


orrrrtem:
Rejected due possible performance degradation.

PR #964, opened 2020-09-18
[DOC] Added known issues for beta09

outoftardis:


averbukh (approved):

napetrov:
@outoftardis  - could you please rename files so they will not be triggered for fail keyword?

napetrov (approved):

PR #970, opened 2020-09-18
Enable PRODUCT_STATUS update for headers version

napetrov:
# Description
Product status is missed in header definition now

### DIFF
--- a/cpp/daal/include/services/library_version_info.h
+++ b/cpp/daal/include/services/library_version_info.h
@@ -29,6 +29,7 @@
 #define __INTEL_DAAL__        2199
 #define __INTEL_DAAL_MINOR__  9
 #define __INTEL_DAAL_UPDATE__ 9
+#define __INTEL_DAAL_STATUS__ 'A'
 
 #define INTEL_DAAL_VERSION (__INTEL_DAAL__ * 10000 + __INTEL_DAAL_MINOR__ * 100 + __INTEL_DAAL_UPDATE__)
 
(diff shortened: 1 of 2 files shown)


cpp/daal/include/services/library_version_info.h
```diff
@@ -29,6 +29,7 @@
 #define __INTEL_DAAL__        2199
 #define __INTEL_DAAL_MINOR__  9
 #define __INTEL_DAAL_UPDATE__ 9
+#define __INTEL_DAAL_STATUS__ 'A'
```
michael-smirnov, line 32: 1. Can you please explain what these versions mean? 
2. Status 'A' - what the letter belongs to? Alpha? 🤔 
3. Shall we have something similar for new interfaces?

napetrov: 1. Yes - A- alpha, B- beta, P- production. It defines status of library. the problem is that it's only avaialbe in getlibraryversion() call, but not in headers.  daal4py do parse headers to understand versioning, but currently it can't distinguish beta and product releases

3. As i see we don't have library versioning there. So we should do this eventually

averbukh: We need to move into semantic versioning for oneAPI that means major.minor.patch schema.

averbukh (approved):

cpp/daal/include/services/library_version_info.h
napetrov: Agree - think we can update this as separate request.

PR #975, opened 2020-09-20
rename default dir  daal->dal

napetrov:
# Description
rename default dir daal -> dal
create link from daal -> dal

### DIFF
--- a/makefile
+++ b/makefile
@@ -1031,11 +1031,11 @@ endif
 #       `cd`to the lib directory first and then create symlink.
 $(RELEASEDIR)\dal:
 	ifeq ($(if $(or $(OS_is_lnx),$(OS_is_mac)),yes,),yes)
-	cd $(RELEASEDIR) && ln -sf daal dal
+	cd $(RELEASEDIR) && ln -sf dal daal
 	endef
 	
 	ifeq ($(OS_is_win),yes)
-	cd  $(RELEASEDIR) && mklink /D daal dal
+	cd  $(RELEASEDIR) && mklink /J daal dal
 	endef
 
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -1022,6 +1022,19 @@ ifeq ($(OS_is_win),yes)
...
+# Note: The `ln` command does not support `-r` on MacOS, so we
+#       `cd`to the lib directory first and then create symlink.
+ifeq ($(if $(or $(OS_is_lnx),$(OS_is_mac)),yes,),yes)
+_release_c: 
+  cd $(RELEASEDIR) && ln -sf daal dal
+endef
+
+ifeq ($(OS_is_win),yes)
+_release_c: 
+  cd $(RELEASEDIR) && mklink /D daal dal
```
rlnx, line 1035: Does it work without admin access?

napetrov: yes. as we create link in some user directory and it's sufficient to have use right there

makefile
```diff
@@ -1022,6 +1022,21 @@ ifeq ($(OS_is_win),yes)
...
 endif
 
+# Adds symlink to the old library directory
+# Note: The `ln` command does not support `-r` on MacOS, so we
+#       `cd`to the lib directory first and then create symlink.
+ifeq ($(if $(or $(OS_is_lnx),$(OS_is_mac)),yes,),yes)
+_release_c:
+	cd $(RELEASEDIR) && ln -sf dal daal
+
+endef
```
KulikovNikita, line 1032:
```suggestion
endif
```
Misprint?

napetrov: Yes - thanks!

PR #985, opened 2020-09-21
Force backports in case of conflicts

napetrov:
# Description
In case of merge conflicts backports are get lost - they will be created anyway now


PetrovKP (approved):

PetrovKP:
I think daal-mnt-2020 can be deleted already?

napetrov:
let's do release first

PR #986, opened 2020-09-21
Aligning graph examples for using const + auto

orrrrtem:
auto vs excplicit. const is added where it is required.

### DIFF
--- a/examples/oneapi/cpp/source/example_util/output_helpers.hpp
+++ b/examples/oneapi/cpp/source/example_util/output_helpers.hpp
@@ -55,8 +55,8 @@ std::ostream &operator<<(std::ostream &stream,
   return stream;
 }
 
-void print_vertex_similarity_result(const oneapi::dal::table &table, const int64_t& nnz_count) {
-  auto arr = oneapi::dal::row_accessor<const float>(table).pull();
+void print_vertex_similarity_result(const oneapi::dal::table &table, int64_t nnz_count) {
+  const auto arr = oneapi::dal::row_accessor<const float>(table).pull();
   const auto x = arr.get_data();
 
   if (table.get_row_count() <= 10) {
--- a/examples/oneapi/cpp/source/graph/graph_service_functions.cpp
+++ b/examples/oneapi/cpp/source/graph/graph_service_functions.cpp
@@ -28,19 +28,19 @@ using namespace oneapi::dal::preview;
 int main(int argc, char **argv) {
     const std::string filename = get_data_path("graph.csv");
 
-    graph_csv_data_source ds(filename);
-    load_graph::descriptor<> d;
-    auto my_graph = load_graph::load(d, ds);
+    const graph_csv_data_source ds(filename);
+    const load_graph::descriptor<> d;
+    const auto my_graph = load_graph::load(d, ds);
     std::cout << "Number of vertices: " << get_vertex_count(my_graph) << std::endl;
     std::cout << "Number of edges: " << get_edge_count(my_graph) << std::endl;
 
-    auto node_id = 0;
+    std::int32_t node_id = 0;
     std::cout << "Degree of " << node_id << ": " << get_vertex_degree(my_graph, node_id)
               << std::endl;
 
-    for (unsigned int j = 0; j < get_vertex_count(my_graph); ++j) {
+    for (std::uint64_t j = 0; j < get_vertex_count(my_graph); ++j) {
         std::cout << "Neighbors of " << j << ": ";
-        auto neigh = get_vertex_neighbors(my_graph, j);
+        const auto neigh = get_vertex_neighbors(my_graph, j);
         for (auto i = neigh.first; i != neigh.second; ++i) {
             std::cout << *i << " ";
         }
--- a/examples/oneapi/cpp/source/jaccard/jaccard_batch_app.cpp
+++ b/examples/oneapi/cpp/source/jaccard/jaccard_batch_app.cpp
@@ -39,22 +39,22 @@ using namespace std;
 /// @param [in]   block_column_count The size of block by columns
 template <class Graph>
 void vertex_similarity_block_processing(const Graph &g,
-                                        const int32_t block_row_count,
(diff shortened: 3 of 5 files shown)


examples/oneapi/cpp/source/graph/graph_service_functions.cpp
```diff
@@ -28,19 +28,19 @@ using namespace oneapi::dal::preview;
...
-    load_graph::descriptor<> d;
-    auto my_graph = load_graph::load(d, ds);
+    const graph_csv_data_source ds(filename);
+    const load_graph::descriptor<> d;
+    const auto my_graph = load_graph::load(d, ds);
     std::cout << "Number of vertices: " << get_vertex_count(my_graph) << std::endl;
     std::cout << "Number of edges: " << get_edge_count(my_graph) << std::endl;
 
-    auto node_id = 0;
+    std::int32_t node_id = 0;
```
bysheva, line 37: Use the appropriate type which will work for any graph type (not int32 only)

bysheva, line 37: And additionally change "node" terminology to "vertex"
Please also check if it is used in some other code

examples/oneapi/cpp/source/graph/graph_service_functions.cpp
```diff
@@ -28,19 +28,19 @@ using namespace oneapi::dal::preview;
...
     std::cout << "Degree of " << node_id << ": " << get_vertex_degree(my_graph, node_id)
               << std::endl;
 
-    for (unsigned int j = 0; j < get_vertex_count(my_graph); ++j) {
+    for (std::uint64_t j = 0; j < get_vertex_count(my_graph); ++j) {
```
bysheva, line 41: Use the appropriate type which is really returned by get_vertex_count for any graph type

examples/oneapi/cpp/source/jaccard/jaccard_batch_app.cpp
```diff
@@ -39,22 +39,22 @@ using namespace std;
...
+                                        int32_t block_row_count,
+                                        int32_t block_column_count);
 
 int main(int argc, char **argv) {
   // load the graph
-  std::string filename = get_data_path("graph.csv");
-  graph_csv_data_source ds(filename);
-  load_graph::descriptor<> d;
-  auto graph = load_graph::load(d, ds);
+  const std::string filename = get_data_path("graph.csv");
```
michael-smirnov, line 47: line 32: `using namespace std` - need to be removed?

examples/oneapi/cpp/source/graph/graph_service_functions.cpp
```diff
@@ -28,19 +28,20 @@ using namespace oneapi::dal::preview;
...
+    const graph_csv_data_source ds(filename);
+    using my_graph_type = undirected_adjacency_array_graph<>;
+    const load_graph::descriptor<edge_list<int32_t>, my_graph_type> desc;
+    const auto my_graph = load_graph::load(desc, ds);
     std::cout << "Number of vertices: " << get_vertex_count(my_graph) << std::endl;
     std::cout << "Number of edges: " << get_edge_count(my_graph) << std::endl;
 
-    auto node_id = 0;
-    std::cout << "Degree of " << node_id << ": " << get_vertex_degree(my_graph, node_id)
+    vertex_type<my_graph_type> vertex_id = 0;
```
michael-smirnov, line 38: you make `using namespace oneapi::dal` and `using namespace oneapi::dal::preview`. Do not we want to mark in exaples which functionality is preview and which is not? Suggest not to use `oneapi::dal::preview` namespace and explicitly mark every preview feature `preview::vertex_type<my_graph_type>`

examples/oneapi/cpp/source/jaccard/jaccard_batch_app.cpp
orrrrtem: "using namespace std" was used as there are a lot of number variables, like, std::int32_t and using namespace std reduces amount of code. This problem can be resolved in another way, if we put a using especially on std::int32_t with purpose not to extend each init of int32_t with std:: prefix. But such approach was rejected by the team as it complicates the example. 

Suggest to remove std:: from each functionality as using do it.

orrrrtem: > line 32: `using namespace std` - need to be removed?

"using namespace std" was used as there are a lot of number variables, like, std::int32_t and using namespace std reduces amount of code. This problem can be resolved in another way, if we put a using especially on std::int32_t with purpose not to extend each init of int32_t with std:: prefix. But such approach was rejected by the team as it complicates the example.

Suggest not to write using, and extend all variables with std::

examples/oneapi/cpp/source/example_util/output_helpers.hpp
```diff
@@ -55,8 +55,8 @@ std::ostream &operator<<(std::ostream &stream,
   return stream;
 }
 
-void print_vertex_similarity_result(const oneapi::dal::table &table, const int64_t& nnz_count) {
-  auto arr = oneapi::dal::row_accessor<const float>(table).pull();
+void print_vertex_similarity_result(const oneapi::dal::table &table, int64_t nnz_count) {
```
michael-smirnov, line 58: std::int64_t?

michael-smirnov (approved):

examples/oneapi/cpp/source/jaccard/jaccard_batch_app.cpp
michael-smirnov: Why `using std::int64_t` complicates the code?

michael-smirnov:
approved with the condition of minor questions addressed

examples/oneapi/cpp/source/example_util/output_helpers.hpp
orrrrtem: > std::int64_t?

Fixed.

PR #1023, opened 2020-09-29
added maxBins and minBinSize params support for DF (classic and oneapi interfaces)

makart19:


cpp/daal/include/algorithms/decision_forest/decision_forest_training_parameter.h
```diff
@@ -176,6 +176,11 @@ class DAAL_EXPORT Parameter
     double minImpurityDecreaseInSplitNode; /*!< A node will be split if this split induces a decrease of the impurity
                                                   greater than or equal to the value, non-negative. Default is 0.0 */
     size_t maxLeafNodes;                   /*!< Maximum number of leaf node. Default is 0 (unlimited) */
+    size_t maxBins;                        /*!< Used with 'hist' split finding method only.
```
PetrovKP, line 179: 1. need to check dynamic BC. 
2. you need to get approval for expanding the interfaces (for both interfaces: old and new).

averbukh (approved):
While we have strong reasons to have this level of flexibility for Gold timeline (to achieve good perf), I'm ok to skip formal approval of API extension by the Board of reviewers, because as @makart19 commented offline, we have approval to include this extension into Beta 10. However, binarization is not something DF-related, and there are a lot of technics to perform similar preprocessing (including smarter   histograms and another technics like clustering), so it makes sense to split it as we do for k-means - we need additional discussion for that.

cpp/daal/include/algorithms/decision_forest/decision_forest_training_parameter.h
makart19: Checked that dynamic BC is not affected with these changes.

makart19:
@Mergifyio backport rls/2021-gold-mnt

PR #1050, opened 2020-10-04
[Draft] Stepless distributed approach + common code of batch and distributed k-means

SmirnovEgorRu:
Some initial sketch


cpp/daal/src/algorithms/kmeans/kmeans_container.h
```diff
@@ -83,9 +85,11 @@ services::Status BatchContainer<algorithmFPType, method, cpu>::compute()
     Parameter * par                        = static_cast<Parameter *>(_par);
     daal::services::Environment::env & env = *_env;
 
+    oneapi::dal::network::empty_network net;
+
     if (deviceInfo.isCpu || method != lloydDense)
     {
-        __DAAL_CALL_KERNEL(env, internal::KMeansBatchKernel, __DAAL_KERNEL_ARGUMENTS(method, algorithmFPType), compute, a, r, par);
+        __DAAL_CALL_KERNEL(env, internal::KMeansBatchKernel, __DAAL_KERNEL_ARGUMENTS(method, algorithmFPType), compute, a, r, par, net);
```
amgrigoriev, line 92: Maybe it make sense to take network from env like it's done for execution context? In that cases user will set proper network object to environment. By default, empty network will be set. In that case, there is no need to pass network as a parameter to kernel - kernel will be able to take it from environment

examples/oneapi/cpp/source/kmeans/kmeans_lloyd_dense_batch.cpp
```diff
@@ -17,12 +17,33 @@
...
+#include "oneapi/dal/network/mpi.hpp"
 
 using namespace oneapi;
 
+#define MPI
+
+int main(int argc, char* argv[]) {
+    int myRank = 0;
+    #ifdef MPI
+        MPI_Init(&argc, &argv);
```
amgrigoriev, line 29: No way for user to select a particular communicator/network? Direct call to MPI to obtain ranks? Something like:
MPI_network network;
int myRank = network.get_communicator()->rank();
env->setDefaultNetwork(network);
algo.compute();

PR #1063, opened 2020-10-08
Add Jaccard Similarity avx512-specific code branch

bysheva:
# Description

-
-
-


cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_hsw.cpp
```diff
@@ -41,8 +271,78 @@ vertex_similarity_result call_jaccard_default_kernel<undirected_adjacency_array_
...
+                }
+            }
+        }
+        vertex_similarity_result res(
+            homogen_table::wrap(first_vertices, 2, number_elements_in_block),
+            homogen_table::wrap(jaccard, 1, number_elements_in_block),
+            nnz);
+        return res;
+    }
+} // namespace detail
```
KulikovNikita, line 346: Looks like duplication

bysheva: @KulikovNikita  Where do you see the duplication?

cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_skx.cpp
```diff
@@ -31,17 +30,688 @@ namespace oneapi::dal::preview {
...
+        __m512i v_u = _mm512_loadu_si512((void *)(neigh_u + i_u)); // load 16 neighbors of u
+        __m512i v_v = _mm512_loadu_si512((void *)(neigh_v + i_v)); // load 16 neighbors of v
+        if (maxu >= maxv)
+            i_v += 16;
+        if (maxu <= maxv)
+            i_u += 16;
+
+        __mmask16 match = _mm512_cmpeq_epi32_mask(v_u, v_v);
+        if (_mm512_mask2int(match) != 0xffff) { // shortcut case where all neighbors match
+            __m512i circ1 = _mm512_set_epi32(0, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1);
```
SmirnovEgorRu, line 89: I suppose it will be faster to replace this:
```cpp
__m512i circ1 = _mm512_set_epi32(0, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1);
..
__m512i circ15 = _mm512_set_epi32(14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0, 15);

__m512i v_v1 = _mm512_permutexvar_epi32(circ1, v_v);
...
__m512i v_v15 = _mm512_permutexvar_epi32(circ15, v_v);
```
By:
```cpp
alignas(256) __m512i buff[2];
buff[0] =  v_v;
buff[1] =  v_v;

__m512i v_v1 = _mm512_loadu_si512((void *)(((int*)buff) + 0));
...
__m512i v_v1 =_mm512_loadu_si512((void *)(((int32_t*)buff) + 14));
...
```
I hope that reverse of element's order isn't important for you.

Throughput of ```_mm512_permutexvar_epi32``` is equal to 1 clocks/instruction (CPI), but ```_mm512_loadu_si512``` is equal to 0.5 CPI. Also, ```_mm512_set_epi32``` has undefined CPI because it's some sequence of instructions defined by a compiler.

cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_hsw.cpp
```diff
@@ -41,8 +271,78 @@ vertex_similarity_result call_jaccard_default_kernel<undirected_adjacency_array_
...
 } // namespace detail
```
KulikovNikita: I'm pretty sure that in this exact version there were two `} // namespace detail` lines.

KulikovNikita: But it was fixed later

bysheva: Ok, thanks, it was fixed in the next commits

cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_hsw.cpp
```diff
@@ -31,17 +30,308 @@ namespace oneapi::dal::preview {
...
+DAAL_FORCEINLINE std::int32_t _popcnt32_redef(const std::int32_t &x) {
+    std::int32_t count = 0;
+    std::int32_t a = x;
+    while (a != 0) {
+        a = a & (a - 1);
+        count++;
+    }
+    return count;
+}
+#endif
```
KulikovNikita, line 47:
```suggestion
#if defined(__INTEL_COMPILER)
DAAL_FORCEINLINE std::int32_t _popcnt32_redef(const std::int32_t &x) {
    return _popcnt32(x);
}
#elif defined(_MSC_VER)
DAAL_FORCEINLINE std::int32_t _popcnt32_redef(const std::int32_t &x) {
    return __popcnt(x);
}
#elif (defined(__GNUC__) || defined(__clang__)) && ((defined(__has_builtin) && __has_builtin(__builtin_popcnt)) || defined(__builtin_popcnt))
DAAL_FORCEINLINE std::int32_t _popcnt32_redef(const std::int32_t &x) {
    return __builtin_popcount(x);
}
#else
DAAL_FORCEINLINE std::int32_t _popcnt32_redef(const std::int32_t &x) {
    std::int32_t count = 0;
    std::int32_t a = x;
    while (a != 0) {
        a = a & (a - 1);
        count++;
    }
    return count;
}
#endif
```
Docs:
1. https://docs.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64?view=vs-2019
2. https://clang.llvm.org/docs/LanguageExtensions.html
3. https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
4. https://gcc.gnu.org/onlinedocs/gcc-10.1.0/cpp/_005f_005fhas_005fbuiltin.html

Actually situation with `__has_builtin` is a little bit complicated since it's deprecated in Clang since 10.0 version (they recommend to use `defined(__builtin_NAME)` instaed) and added into GCC spec since 10 version (magic of number 10).

cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_skx.cpp
bysheva: >>> I hope that reverse of element's order isn't important for you. 
No, it's important. We have to compare all the permutations of the set. 
Probably we can't apply the suggested change.

SmirnovEgorRu: Your idea is just iterate over all possible combinations and do comparisons. The order of the comparisons doesn't matter here.

bysheva: @SmirnovEgorRu ok, yes, you're right, got it! will check today

orrrrtem: > Your idea is just iterate over all possible combinations and do comparisons. The order of the comparisons doesn't matter here.

It should works. With alignas we can use _mm512_load_si512.

SmirnovEgorRu: @orrrrtem, for the first element - yes, for other - no, because we shift buff by some INT's.

bysheva: @SmirnovEgorRu checked this option - got ~1% improvement, so looks like on-par. 
But the code looks simpler, probably we can use this version.

orrrrtem: @SmirnovEgorRu, this algorithm is bounded by memory. The last part of our perf improvements was about memory. Any suggestions for prefetch parts?

SmirnovEgorRu: If the hotspot is memory - acceleration of the computation will not bring anything. Yes, it's clear. Probably for Icelake (with higher BW) computation portion will take more % of time.

About work with memory acceleration - need to profile, if the hotspot is memory BW, prefetch shouldn't help.

orazve (approved):

PR #1065, opened 2020-10-08
Jaccard result is in a column major format

orazve:
+ example refactoring

### DIFF
--- a/cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_avx2.hpp
+++ b/cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_avx2.hpp
@@ -330,9 +330,10 @@ vertex_similarity_result call_jaccard_default_kernel_avx2(
             }
         }
     }
-    vertex_similarity_result res(homogen_table::wrap(first_vertices, number_elements_in_block, 2, data_layout::column_major),
-                                 homogen_table::wrap(jaccard, number_elements_in_block, 1, data_layout::column_major),
-                                 nnz);
+    vertex_similarity_result res(
+        homogen_table::wrap(first_vertices, number_elements_in_block, 2, data_layout::column_major),
+        homogen_table::wrap(jaccard, number_elements_in_block, 1, data_layout::column_major),
+        nnz);
     return res;
 }
 } // namespace detail
--- a/cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_avx2.hpp
+++ b/cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_avx2.hpp
@@ -330,8 +330,8 @@ vertex_similarity_result call_jaccard_default_kernel_avx2(
             }
         }
     }
-    vertex_similarity_result res(homogen_table::wrap(first_vertices, 2, number_elements_in_block),
-                                 homogen_table::wrap(jaccard, 1, number_elements_in_block),
+    vertex_similarity_result res(homogen_table::wrap(first_vertices, number_elements_in_block, 2, data_layout::column_major),
+                                 homogen_table::wrap(jaccard, number_elements_in_block, 1, data_layout::column_major),
                                  nnz);
     return res;
 }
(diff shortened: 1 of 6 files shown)


cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_avx2.hpp
```diff
@@ -330,9 +330,10 @@ vertex_similarity_result call_jaccard_default_kernel_avx2(
             }
         }
     }
-    vertex_similarity_result res(homogen_table::wrap(first_vertices, 2, number_elements_in_block),
-                                 homogen_table::wrap(jaccard, 1, number_elements_in_block),
-                                 nnz);
+    vertex_similarity_result res(
+        homogen_table::wrap(first_vertices, number_elements_in_block, 2, data_layout::column_major),
+        homogen_table::wrap(jaccard, number_elements_in_block, 1, data_layout::column_major),
+        nnz);
```
michael-smirnov, line 336: It's confusing for me - why it works. `data_layout::column_major` does not transpose the data or do another complex logic - it just says that the data stored column by column. Why you swapped the number of rows and columns?

cpp/oneapi/dal/algo/jaccard/vertex_similarity_types.hpp
```diff
@@ -67,22 +67,21 @@ class ONEAPI_DAL_EXPORT vertex_similarity_result {
...
+    ///                                   similarity coefficients
+    /// @param [in]   coeffs              The table of size [nonzero_coeff_count x 1] with
+    ///                                   non-zero Jaccard similarity coefficients
+    /// @param [in]   nonzero_coeff_count The number of non-zero Jaccard coefficients
     vertex_similarity_result(const table& vertex_pairs,
                              const table& coeffs,
                              std::int64_t nonzero_coeff_count);
 
-    /// Returns the table of size 1*nonzero_coeff_count with non-zero Jaccard
+    /// Returns the table of size [nonzero_coeff_count x 1] with non-zero Jaccard
```
michael-smirnov, line 80: The shape of the table does not depend on the layout - it has `n` rows and `m` columns in both row major and column major formats. The only difference is that in rowmajor format adjacent elements correspond to the same row, and in colmajor - to the same column.
That's the right shape of that table?

bysheva (approved):

examples/oneapi/cpp/source/example_util/output_helpers_graph.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+void print_vertex_similarity_result(
+    const oneapi::dal::preview::jaccard::vertex_similarity_result &result) {
+    using namespace oneapi::dal;
+    const std::int64_t nonzero_coeff_count = result.get_nonzero_coeff_count();
+    const std::int64_t lines_count = 20;
+    const std::int64_t print_count =
+        (nonzero_coeff_count < lines_count) ? nonzero_coeff_count : lines_count;
+
+    std::vector<std::int64_t> print_range(print_count);
```
michael-smirnov, line 35: `#include <vector>` at the head of the file? For now, I suppose this dependency goes from the graph component

examples/oneapi/cpp/source/example_util/output_helpers_graph.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include <iomanip>
+#include <iostream>
+#include <numeric>
+
+#include "oneapi/dal/table/common.hpp"
+#include "oneapi/dal/algo/jaccard/vertex_similarity_types.hpp"
+#include "oneapi/dal/table/homogen.hpp"
+
+void print_vertex_similarity_result(
```
michael-smirnov, line 27: We assume that helpers can be reused between several components. 
Can we make it less specific? Now it's targeted only for similarity and only for jaccard algorithm. What if we want to print out the same similarity results for other graph algorithms?

cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_avx2.hpp
orazve: It is the requirement to store the first vertices and the second vertices in a continuous arrays. But they have to be represented as [n x 2] table. Column-major format is a good option for this case.

cpp/oneapi/dal/algo/jaccard/vertex_similarity_types.hpp
orazve: Yes, it is right

examples/oneapi/cpp/source/example_util/output_helpers_graph.hpp
orazve: Added

PR #1068, opened 2020-10-12
parallel_sort instead of std::sort

orazve:
# Description

-
-
-


cpp/oneapi/dal/io/detail/meas.hpp
```diff
@@ -0,0 +1,31 @@
...
+#define CR_INIT()                                        \
+    auto t0 = std::chrono::high_resolution_clock::now(); \
+    auto t1 = std::chrono::high_resolution_clock::now();
+#define CR_ST() t0 = std::chrono::high_resolution_clock::now();
+#define CR_END(name)                                                                \
+    t1 = std::chrono::high_resolution_clock::now();                                 \
+    {                                                                               \
+        std::chrono::duration<double> elapsed = t1 - t0;                            \
+        std::cout << "Elapsed time " << #name << ": " << elapsed.count() << " s\n"; \
+    }
```
KulikovNikita, line 30:
```suggestion
#include <type_traits>

template<typename Function, typename... Arguments>
inline double measure(Function func, Arguments&... args) {
    const auto start_time = std::chrono::high_resolution_clock::now();
    func(std::forward<Arguments>(args)...);
    const auto end_time = std::chrono::high_resolution_clock::now();
    return std::chrono::duration<double>(end_time - start_time).count();
} 
template<typename Function, typename... Arguments>
inline std::pair<double, typename std::result_of<Function(Arguments...)>::type> measure_with_result(Function func, Arguments&... args) {
    const auto start_time = std::chrono::high_resolution_clock::now();
    auto function_result = func(std::forward<Arguments>(args)...);
    const auto end_time = std::chrono::high_resolution_clock::now();
    const auto time_delta = std::chrono::duration<double>(end_time - start_time).count();
    return std::pair(time_delta, function_result); 
} 
```

PR #1114, opened 2020-10-16
parallel_sort for graph loading functionality

orazve:
parallel_sort functionality is used during load graph. It improves graph converting performance.


cpp/daal/src/algorithms/service_qsort.h
```diff
@@ -0,0 +1,134 @@
+/* file: service_qsort.h */
+/*******************************************************************************
+* Copyright 2014-2020 Intel Corporation
```
averbukh, line 3: 2020

averbukh (commented):
Minor comment only

cpp/daal/src/algorithms/service_qsort.h
```diff
@@ -1,6 +1,6 @@
 /* file: service_qsort.h */
 /*******************************************************************************
-* Copyright 2014-2020 Intel Corporation
+* Copyright 2020-2020 Intel Corporation
```
averbukh, line 3: Just "Copyright 2020 Intel Corporation". See the rest of the library and follow for consistency.

cpp/daal/src/externals/core_threading_win_dll.cpp
```diff
@@ -277,8 +277,8 @@ typedef void (*_daal_tbb_task_scheduler_free_t)(void *& globalControl);
 typedef size_t (*_setNumberOfThreads_t)(const size_t, void **);
 typedef void * (*_daal_threader_env_t)();
 
-typedef void (*_daal_parallel_sort_int_t)(int *, int *);
-typedef void (*_daal_parallel_sort_size_t_t)(size_t *, size_t *);
+typedef void (*_daal_parallel_sort_int32_t)(std::int32_t *, std::int32_t *);
+typedef void (*_daal_parallel_sort_uint64_t)(std::uint64_t *, std::uint64_t *);
```
averbukh, line 281: In fact we prefer to use int and size_t in DAAL.

cpp/daal/src/threading/threading.h
```diff
@@ -25,6 +25,7 @@
 #define __THREADING_H__
 
 #include "services/daal_defines.h"
+#include <cstdint>
```
averbukh, line 28: It isn't look like we need this dependency here. In fact you only need to have 'unsigned long *' and 'unsigned long long *' casting in the safe way.

averbukh (changes requested):
In your particular case, the introduction of additional dependency to standard library is not required.

cpp/daal/src/externals/core_threading_win_dll.cpp
orazve: fixed

cpp/daal/src/threading/threading.h
orazve: removed

averbukh (approved):

PR #1153, opened 2020-10-21
[DRAFT] Prototype of centralized error generator

rlnx:
Current:
- Introduce `error_messages` class with the static functions returning `const char *`  for all possible error messages, name of the function reflects content of the message, but not tightly coupled with it to have flexibility in changing error message content.
-  Introduce `exception_generator` that simplifies throwing exceptions with particular error messages.

Future:
- Support [messages formatting](https://en.cppreference.com/w/cpp/utility/format)
- Generate `error_messages` and `exception_generator` from YAML:
```yaml
input_data_is_empty: >
   Input data is empty

input_labels_are_empty: >
   Input labels are empty
```

### DIFF
--- a/cpp/oneapi/dal/detail/error_reporting.cpp
+++ b/cpp/oneapi/dal/detail/error_reporting.cpp
@@ -1,0 +1,34 @@
+/*******************************************************************************
+* Copyright 2020 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "oneapi/dal/detail/error_reporting.hpp"
+
+#define DEFINE_MESSAGE(id, text)                \
+    static const char* _##id = text;            \
+    const char* error_messages::id() noexcept { \
+        return _##id;                           \
+    }
+
+namespace oneapi::dal::detail {
+
+DEFINE_MESSAGE(input_data_is_empty, "Input data is empty")
+DEFINE_MESSAGE(input_labels_are_empty, "Input labels are empty")
+DEFINE_MESSAGE(input_data_row_count_is_not_equal_to_input_labels_row_count,
+               "Input data row count is not equal to input labels row count")
+DEFINE_MESSAGE(bootstrap_is_incompatible_with_variable_importance_mode,
+               "Parameter `bootstrap` is incompatible with requested variable importance mode")
+
+} // namespace oneapi::dal::detail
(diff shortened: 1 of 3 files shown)


cpp/oneapi/dal/detail/error_reporting.hpp
```diff
@@ -0,0 +1,64 @@
...
+class exception_generator {
+public:
+    exception_generator() = delete;
+    exception_generator(const exception_generator&) = delete;
+    exception_generator& operator=(const exception_generator&) = delete;
+
+    DEFINE_EXCEPTION(domain_error, input_data_is_empty)
+    DEFINE_EXCEPTION(domain_error, input_labels_are_empty)
+    DEFINE_EXCEPTION(invalid_argument, input_data_row_count_is_not_equal_to_input_labels_row_count)
+    DEFINE_EXCEPTION(invalid_argument, bootstrap_is_incompatible_with_variable_importance_mode)
```
michael-smirnov, line 59: Cohesion between `exception_generator` and `error_messages` is very big. I'd suggest to merge them somehow or unbound `exception_generator` from knowledge of error messages or concrete types of exceptions. Is it always true that every error message is bound with a single exception type? 
If think we don't need to have exception_generator. 

I think its more readable to have
```cpp
if (!params.get_bootstrap() &&	      
    (params.get_variable_importance_mode() == variable_importance_mode::mda_raw ||	            
     params.get_variable_importance_mode() == variable_importance_mode::mda_scaled)) {	             
     throw invalid_argument(error_message::bootstrap_is_incompatible_with_variable_importance_mode());	
}
```
instead of
```cpp
 eg::report_bootstrap_is_incompatible_with_variable_importance_mode_if_false( params.get_bootstrap() ||
            (params.get_variable_importance_mode() == variable_importance_mode::mda_raw ||	            
            (params.get_variable_importance_mode() != variable_importance_mode::mda_raw &&
	     params.get_variable_importance_mode() != variable_importance_mode::mda_scaled));
```

rlnx: > Is it always true that every error message is bound with a single exception type?

I assume it's always true. Do you see an example where it's not true?

cpp/oneapi/dal/detail/error_reporting.cpp
```diff
@@ -0,0 +1,34 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "oneapi/dal/detail/error_reporting.hpp"
+
+#define DEFINE_MESSAGE(id, text)                \
+    static const char* _##id = text;            \
```
averbukh, line 20: Is it global variable after macro substitution (for lines 27-32 of this file)? If so, could we avoid it (SDL requires to avoid if possible)? For instance, free function that returns const char * can be used instead?

cpp/oneapi/dal/detail/error_reporting.hpp
averbukh: At least, Michael's example with 'if' has better code readability.

cpp/oneapi/dal/detail/error_reporting.cpp
rlnx: Actually it's not global variable, but global constant, but agree it's better to return directly from function.

averbukh: Not sure that do mean, but "constants" are "literals" in c++, and definitely it isn't literal. If you want to say that the variable type is const-qualified, it does not make sense for corresponding SDL task

PR #1158, opened 2020-10-21
Add dal.hpp

rlnx:
- Add global header that includes all library functionality
- Add test that verifies all algorithms are include in `dal.hpp`. There is no such checks for tables and IO, expect they are not updated so frequently
- Move `.clang-format` one level up, update paths in CI scripts


napetrov (approved):
Looks good

cpp/oneapi/dal.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+/* Algos */
+#include "oneapi/dal/algo/decision_forest.hpp"
+#include "oneapi/dal/algo/kmeans.hpp"
+#include "oneapi/dal/algo/kmeans_init.hpp"
+#include "oneapi/dal/algo/knn.hpp"
+#include "oneapi/dal/algo/linear_kernel.hpp"
+#include "oneapi/dal/algo/pca.hpp"
+#include "oneapi/dal/algo/rbf_kernel.hpp"
+#include "oneapi/dal/algo/svm.hpp"
```
michael-smirnov, line 45: Do you plan to include graphs?

rlnx: Not sure if we have to include preview feature here

rlnx: Added

PR #1173, opened 2020-10-22
Remove debug examples for DPCPP

napetrov:
Remove debug case from examples on Windows for DPC++


rlnx:
> /intelci: run

Shouldn't it be disabled in private CI?

napetrov:
> 
> 
> > /intelci: run
> 
> Shouldn't it be disabled in private CI?

No- it looks only release examples are executed there historically.

rlnx (approved):

PR #1197, opened 2020-10-25
Clean code & add tasks for SVM/Kernel Functions for new interfaces

PetrovKP:
1. Clean code for SVM, kernel function for old & new interfaces
2. Add template parameter `task` for input, result, model for consistency with other algorithms
3. Swap template parameter `Method` and `Task` for SVM algorithm
4. Add template parameter `task` for RBF & Linear kernel for consistency with other algorithms
5. Fix bug for size buffer for SVM prediction


cpp/daal/src/algorithms/kernel_function/oneapi/kernel_function_rbf_dense_default_oneapi_impl.i
```diff
@@ -96,6 +96,11 @@ services::Status KernelImplRBFOneAPI<defaultDense, algorithmFPType>::computeRBF(
 
     KernelArguments args(6, status);
     DAAL_CHECK_STATUS_VAR(status);
+
+    DAAL_ASSERT_UNIVERSAL_BUFFER(sqrMatRight, algorithmFPType, n);
+    DAAL_ASSERT_UNIVERSAL_BUFFER(sqrMatRight, algorithmFPType, m);
+    DAAL_ASSERT(rbf.size() == n * m);
```
michael-smirnov, line 102: overflow check?

cpp/daal/src/algorithms/svm/oneapi/svm_helper_oneapi.h
```diff
@@ -168,6 +171,9 @@ struct HelperSVM
         services::internal::sycl::KernelArguments args(4, status);
         DAAL_CHECK_STATUS_VAR(status);
 
+        DAAL_ASSERT(indX.size() == nWS);
+        DAAL_ASSERT(newX.size() == nWS * p);
```
michael-smirnov, line 175: is overflow check needed here?

cpp/daal/src/algorithms/svm/oneapi/svm_train_thunder_oneapi_impl.i
```diff
@@ -215,10 +222,9 @@ services::Status SVMTrainOneAPI<algorithmFPType, thunder>::compute(const Numeric
     algorithmFPType diffPrev = algorithmFPType(0);
 
     size_t sameLocalDiff = 0;
-    SVMCacheOneAPIPtr<algorithmFPType> cachePtr;
 
     // TODO: support caching for thunder method
-    cachePtr = SVMCacheOneAPI<noCache, algorithmFPType>::create(cacheSize, nWS, nVectors, xTable, kernel, status);
+    SVMCacheOneAPIPtr<algorithmFPType> cachePtr = SVMCacheOneAPI<noCache, algorithmFPType>::create(cacheSize, nWS, nVectors, xTable, kernel, status);
```
michael-smirnov, line 227: checks status?

cpp/daal/src/algorithms/kernel_function/oneapi/kernel_function_rbf_dense_default_oneapi_impl.i
PetrovKP: 1. It should checked [here](https://github.com/oneapi-src/oneDAL/blob/master/cpp/daal/src/algorithms/kernel_function/kernel_function_fpt.cpp#L55)
2. This is in assert check - I would not want to add extra checks for the release build.

cpp/daal/src/algorithms/svm/oneapi/svm_helper_oneapi.h
PetrovKP: It should checked when buffer is allocate. For my case it [here](https://github.com/oneapi-src/oneDAL/blob/master/cpp/daal/src/algorithms/svm/oneapi/svm_train_cache_oneapi.h#L194).

cpp/daal/src/algorithms/svm/oneapi/svm_train_thunder_oneapi_impl.i
PetrovKP: Done

cpp/daal/src/algorithms/kernel_function/oneapi/kernel_function_rbf_dense_default_oneapi_impl.i
michael-smirnov: How you can guarantee that `n` and `m` are always equal to `nVectors2` and `nVectors1`?

cpp/daal/src/algorithms/svm/oneapi/svm_helper_oneapi.h
michael-smirnov: It is not seen that it is safe.By the reference the numeric table is allocated but not a buffer. How you guarantee the safety of that function if some changes in the code that calls it? E.g. it is called with different block. Place a comment here at least why you think overflow cannot happen here

cpp/daal/src/algorithms/kernel_function/oneapi/kernel_function_rbf_dense_default_oneapi_impl.i
PetrovKP: Because sizes of `matLeftBuf` is equal `nVectors1` and `sqrMatRight` is equal `nVectors2` . Is always

rlnx:
I have only one concern regarding task name for kernel functions. @PetrovKP, do you have any thoughts how the name could be generalized? The task name should reflect the type of problem solved by the algorithm.

rlnx (approved):

PR #1205, opened 2020-10-26
Graph traits v2

orrrrtem:
Common graph traits in common.
Graph traits spec for graph_type is below the graph_type. 
Using pimpl is hidden in a private section of public graph. pimpl now uses impl type referenced from graph traits.


cpp/oneapi/dal/graph/undirected_adjacency_array_graph.hpp
```diff
@@ -114,30 +64,89 @@ class ONEDAL_EXPORT undirected_adjacency_array_graph {
...
+template <typename VertexValue,
+          typename EdgeValue,
+          typename GraphValue,
+          typename IndexType,
+          typename Allocator>
+struct graph_traits<
+    undirected_adjacency_array_graph<VertexValue, EdgeValue, GraphValue, IndexType, Allocator>> {
+    using graph_type =
+        undirected_adjacency_array_graph<VertexValue, EdgeValue, GraphValue, IndexType, Allocator>;
+    using impl_type = detail::undirected_adjacency_array_graph_impl<VertexValue,
```
rlnx, line 100: Is it part of public interface available to the user? If so, do you really want to expose API for getting underlying graph impl type?

cpp/oneapi/dal/graph/detail/undirected_adjacency_array_graph_impl.hpp
```diff
@@ -19,16 +19,19 @@
 #include "oneapi/dal/common.hpp"
 #include "oneapi/dal/detail/common.hpp"
 #include "oneapi/dal/graph/detail/graph_container.hpp"
+#include "oneapi/dal/graph/graph_common.hpp"
 
 namespace oneapi::dal::preview::detail {
 
 template <typename Graph>
-ONEDAL_EXPORT typename Graph::pimpl& get_impl(Graph& graph) {
+ONEDAL_EXPORT oneapi::dal::detail::pimpl<typename graph_traits<Graph>::impl_type>& get_impl(
```
rlnx, line 27: Why `dal::detail::get_impl` declared in `dal/detail/common.hpp` cannot be used instead?

cpp/oneapi/dal/graph/undirected_adjacency_array_graph.hpp
orrrrtem: > Is it part of public interface available to the user? If so, do you really want to expose API for getting underlying graph impl type?
We expose defined by impl types here not to expose them in graph. You can see that a lot of things are removed from public class. Concretely this using is service with the purpose not to duplicate long name undirected_adjacency_array_graph<VertexValue, EdgeValue, GraphValue, IndexType, Allocator>.

cpp/oneapi/dal/graph/detail/undirected_adjacency_array_graph_impl.hpp
orrrrtem: /intelci: run

orrrrtem: Updated.

cpp/oneapi/dal/graph/undirected_adjacency_array_graph.hpp
rlnx: Do you expect users to use `graph_traits` directly in their code?

orrrrtem: Expected. The goal of graph traits is to provide the opportunity for adaption of the external graph types. That is the user will create another one specialization of graph traits concretely for its graph.

rlnx: Well, do you want user to know type of the `impl` class hidden behind `undirected_adjacency_array_graph` or any other type?

cpp/oneapi/dal/algo/jaccard/backend/cpu/vertex_similarity_default_kernel_avx2.hpp
```diff
@@ -265,10 +265,12 @@ vertex_similarity_result call_jaccard_default_kernel_avx2(
     const descriptor_base &desc,
     vertex_similarity_input<undirected_adjacency_array_graph<>> &input) {
     const auto &my_graph = input.get_graph();
-    const auto &g = oneapi::dal::preview::detail::get_impl(my_graph);
-    auto g_edge_offsets = g->_edge_offsets.data();
-    auto g_vertex_neighbors = g->_vertex_neighbors.data();
-    auto g_degrees = g->_degrees.data();
+    using graph_type = undirected_adjacency_array_graph<>;
+    const auto &g =
+        oneapi::dal::detail::get_impl<const typename graph_traits<graph_type>::impl_type>(my_graph);
```
rlnx, line 270: Please synchronise your branch with master, there are the new changes that simplifies `get_impl`

```suggestion
        dal::detail::get_impl(my_graph);
```

cpp/oneapi/dal/graph/graph_common.hpp
```diff
@@ -26,55 +26,101 @@
 
 namespace oneapi::dal::preview {
 
+template <typename Graph>
+struct ONEDAL_EXPORT graph_traits {
```
rlnx, line 30: I think, `ONEDAL_EXPORT` should be removed

cpp/oneapi/dal/graph/undirected_adjacency_array_graph.cpp
```diff
@@ -134,31 +138,34 @@ template class ONEDAL_EXPORT undirected_adjacency_array_graph<empty_value,
                                                               std::int32_t,
                                                               std::allocator<char>>;
 
+template struct ONEDAL_EXPORT graph_traits<undirected_adjacency_array_graph<empty_value,
```
rlnx, line 141: Explicit instantiation is useless for `graph_traits`, there is no any method defined in `.cpp`

cpp/oneapi/dal/graph/undirected_adjacency_array_graph.hpp
orrrrtem: It is not only for the get_impl. using vertex_set = typename graph_traits<Graph>::vertex_set; in graph loading. Moreover vertex_size_type is based on its set type and it is anyway exposed from the impl. To avoid duplication, it is exposed directly in graph traits.

bysheva (approved):

PR #1214, opened 2020-10-27
Introduce DALROOT and Major Minor versions in env scripts

napetrov:
# Description
Adding DALROOT variable
Adding Major and Minor version env variables to identify DLL versions
Updating path for installed lib search

### DIFF
--- a/deploy/local/dal
+++ b/deploy/local/dal
@@ -84,6 +84,9 @@ set daal_target_arch         "intel64"
 module-whatis "Intel(R) Data Analytics Library for $daal_target_arch."
 
 # Setup environment variables
+setenv          DAL_MAJOR_BINARY   1
+setenv          DAL_MINOR_BINARY   0
+setenv          DALROOT            $daalroot
 setenv          DAALROOT           $daalroot
 
 prepend-path    LD_LIBRARY_PATH    $daalroot/lib/$daal_target_arch
--- a/deploy/local/dal
+++ b/deploy/local/dal
@@ -81,7 +81,7 @@ if { [ module-info mode ] == "unload" || [ module-info mode ] == "remove" } {
 set daalroot "$componentroot/$modulefilever"
 set daal_target_arch         "intel64"
 
-module-whatis "Intel(R) Data Analytics Library for $daal_target_arch."
+module-whatis "Intel(R) oneAPI Data Analytics Library for $daal_target_arch."
 
 # Setup environment variables
 setenv          DAL_MAJOR_BINARY   1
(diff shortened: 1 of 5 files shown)


rlnx:
Shouldn't it be 'ONEDALROOT'

napetrov:
@rlnx  - on system we are instilled in 'dal' dir to align with other components. 
From variable name - short variant is also aligned with other libs

So the main thought was to align with other components

deploy/local/dal
```diff
@@ -84,6 +84,9 @@ set daal_target_arch         "intel64"
 module-whatis "Intel(R) Data Analytics Library for $daal_target_arch."
```
rlnx, line 84: Shouldn't it be official product name?

```suggestion
module-whatis "Intel(R) oneAPI Data Analytics Library for $daal_target_arch."
```

deploy/local/vars_win.bat
```diff
@@ -52,6 +52,9 @@ if exist "%DAAL_UP_OLD%\redist" (
 )
 set "CLASSPATH=%DAAL%\lib\onedal.jar;%CLASSPATH%"
 endlocal& ^
+set DAL_MAJOR_BINARY=1& ^
+set DAL_MINOR_BINARY=0& ^
+set DALROOT=%DAALROOT%& ^
```
rlnx, line 57: I see section `:GoodArgs` at line 43, don't we need to add `DAL_MAJOR`/`DAL_MINOR` there?

deploy/local/dal
```diff
@@ -84,6 +84,9 @@ set daal_target_arch         "intel64"
...
 
 # Setup environment variables
+setenv          DAL_MAJOR_BINARY   1
+setenv          DAL_MINOR_BINARY   0
```
averbukh, line 88: Much better is to specify values for variables in separate text file in following form:
var1=value1
var2=value2
and use this file to generate code to set environment variables for every OS.

For example, on Linux to do so, you can use sed/awk/... to perform required transformations and use `source <(cat version.txt | <transformations>)` to do actual set of env.variables.

napetrov: yes, and ideally this should be taken from single file with version for entire product and not separate version.txt file. 

But this is work for @Alexsandruss that will be completed after release .  For now i would like to expose variables for users in env

deploy/local/vars_win.bat
napetrov: Thins within setlocal context is get used for different data manipulations and calculations but actual values are set after endlocal section.  As far we have no processing for now - it's sufficient to specify them at the end. 

but as this have to be moved to version parsing - this logic will be moved to setlocal context and GoodArgs point

deploy/local/dal
averbukh: OK

averbukh (approved):

PR #1215, opened 2020-10-27
Centralized generator of error messages

rlnx:
Current:
- Introduced `error_messages` class with the static functions returning `const char *`  for all possible error messages, name of the function reflects content of the message, but not tightly coupled with it to have flexibility in changing error message content.
-  Replace all raw string in throw expressions across the oneAPI part

Future:
- Support [messages formatting](https://en.cppreference.com/w/cpp/utility/format)
- Generate `error_messages` and `exception_generator` from YAML:
```yaml
input_data_is_empty: >
   Input data is empty

input_labels_are_empty: >
   Input labels are empty
```


rlnx:
> do you mean that this should be checked in the model and in the kernel functions?

@PetrovKP, I think the fact that `model.get_support_vectors().row_count() == model.get_support_vector_count()` and `model.get_support_vector_count() >= 0` is invariant of the model, it should be always maintained there. The best way to satisfy these conditions without any checks is:
```c++
std::int64_t model::get_support_vector_count() const {
    return model::get_support_vectors().get_row_count();
}
```

And remove setter for support vector count at all.

mchernov-intel (approved):

michael-smirnov (approved):

cpp/oneapi/dal/detail/error_messages.hpp
```diff
@@ -74,15 +97,28 @@ class error_messages {
...
+    /* Decision Forest */
+    MSG(bootstrap_is_incompatible_with_error_metric);
+    MSG(bootstrap_is_incompatible_with_variable_importance_mode);
+    MSG(decision_forest_train_dense_method_is_not_implemented_for_gpu);
+    MSG(decision_forest_train_hist_method_is_not_implemented_for_cpu);
+
+    /* Jaccard */
+    MSG(column_begin_gt_column_end);
+    MSG(empty_edge_list);
+    MSG(negative_interval);
```
orazve, line 117: not an alphabetical order

orazve (approved):

orrrrtem (approved):

cpp/oneapi/dal/detail/error_messages.hpp
rlnx: Fixed

cpp/oneapi/dal/algo/decision_forest/detail/train_ops.hpp
```diff
@@ -38,28 +38,28 @@ struct train_ops {
...
+            throw domain_error(msg::input_data_is_empty());
         }
         if (!(input.get_labels().has_data())) {
-            throw domain_error("Input labels should not be empty");
+            throw domain_error(msg::input_labels_are_empty());
         }
         if (input.get_data().get_row_count() != input.get_labels().get_row_count()) {
-            throw invalid_argument("Input data row_count should be equal to labels row_count");
+            throw invalid_argument(msg::input_data_rc_neq_input_labels_rc());
         }
```
KulikovNikita, line 51: shouldn't we check if labels have the only column?

cpp/oneapi/dal/algo/kmeans/train_types.cpp
```diff
@@ -107,15 +107,15 @@ void train_result<Task>::set_labels_impl(const table& value) {
 template <typename Task>
 void train_result<Task>::set_iteration_count_impl(std::int64_t value) {
     if (value < 0) {
```
KulikovNikita, line 109: What if user set negative value with assumption that the algorithm should be stopped only by exceeding tolerance?

cpp/oneapi/dal/detail/common.cpp
```diff
@@ -33,7 +33,8 @@ void integer_overflow_ops<Data>::check_mul_overflow(const Data& first, const Dat
         volatile Data tmp = first * second;
         tmp /= first;
         if (tmp != second) {
-            throw range_error("overflow found in multiplication of two values");
+            throw range_error(
+                dal::detail::error_messages::overflow_found_in_multiplication_of_two_values());
```
KulikovNikita, line 37: Shouldn't it be named `overfow_found_in_mul_of_two_values` for consistency with function above?

KulikovNikita (approved):

cpp/oneapi/dal/algo/decision_forest/detail/train_ops.hpp
rlnx: We should, @makart19 need to ass this check to decision forest

cpp/oneapi/dal/algo/kmeans/train_types.cpp
rlnx: This is a result of training, this field is assumed to contain actual number of iteration performed by kmeans, not requested by the user.

I guess in case of `max_iteration_count = 0` in descriptor, we do not check stop condition with respect to iteration count. Don't we @amgrigoriev?

amgrigoriev (approved):

cpp/oneapi/dal/detail/common.cpp
rlnx: Will rename in next PR

PR #1230, opened 2020-10-29
Add badargs tests & safety checks for kf & svm in onedal

PetrovKP:


KulikovNikita (approved):

michael-smirnov (approved):
some of the tests valid for both host and dpc++ interfaces, but some use only host interface. Do you plan to include tests for dpc++ API?

PetrovKP:
@michael-smirnov Done for dpc & host interfaces

PR #1233, opened 2020-10-29
Aligned oneAPI examples and apply clang-format

vlad-nazarov:


examples/oneapi/cpp/source/kmeans/kmeans_lloyd_dense_batch.cpp
```diff
@@ -18,39 +18,43 @@
...
 
-    const auto result_train = dal::train(kmeans_desc, x_train, initial_centroids);
+    const auto result_train = onedal::train(kmeans_desc, x_train, initial_centroids);
 
     std::cout << "Iteration count: " << result_train.get_iteration_count() << std::endl;
     std::cout << "Objective function value: " << result_train.get_objective_function_value()
               << std::endl;
-    std::cout << "Lables:" << std::endl << result_train.get_labels() << std::endl;
-    std::cout << "Centroids:" << std::endl << result_train.get_model().get_centroids() << std::endl;
+    std::cout << "Lables:\n" << result_train.get_labels() << std::endl;
```
michael-smirnov, line 50:
```suggestion
    std::cout << "Labels:\n" << result_train.get_labels() << std::endl;
```

examples/oneapi/cpp/source/decision_forest/df_reg_dense_batch.cpp
```diff
@@ -18,45 +18,47 @@
...
+namespace onedal = oneapi::dal;
 namespace df = oneapi::dal::decision_forest;
 
-
 int main(int argc, char const *argv[]) {
-  const std::string train_data_file_name  = get_data_path("df_regression_train_data.csv");
-  const std::string train_label_file_name = get_data_path("df_regression_train_label.csv");
-  const std::string test_data_file_name   = get_data_path("df_regression_test_data.csv");
-  const std::string test_label_file_name  = get_data_path("df_regression_test_label.csv");
+    const std::string train_data_file_name = get_data_path("df_regression_train_data.csv");
```
michael-smirnov, line 25:
```suggestion
    const auto train_data_file_name = get_data_path("df_regression_train_data.csv");
```
just a nip, you don't have to apply changes like this

examples/oneapi/cpp/source/decision_forest/df_reg_dense_batch.cpp
```diff
@@ -18,45 +18,47 @@
...
+    const std::string train_label_file_name = get_data_path("df_regression_train_label.csv");
+    const std::string test_data_file_name = get_data_path("df_regression_test_data.csv");
+    const std::string test_label_file_name = get_data_path("df_regression_test_label.csv");
 
-  const auto x_train = dal::read<dal::table>(dal::csv::data_source{train_data_file_name});
-  const auto y_train = dal::read<dal::table>(dal::csv::data_source{train_label_file_name});
+    const auto x_train =
+        onedal::read<onedal::table>(onedal::csv::data_source{ train_data_file_name });
```
michael-smirnov, line 31: Can we make it shorter, or compiler error occurs?
```suggestion
    const auto x_train =
        onedal::read<onedal::table>({ train_data_file_name });
```

examples/oneapi/cpp/source/decision_forest/df_reg_dense_batch.cpp
```diff
@@ -18,45 +18,47 @@
 #include "oneapi/dal/algo/decision_forest.hpp"
 #include "oneapi/dal/io/csv.hpp"
 
-using namespace oneapi;
+namespace onedal = oneapi::dal;
 namespace df = oneapi::dal::decision_forest;
```
michael-smirnov, line 22:
```suggestion
namespace df = onedal::decision_forest;
```

examples/oneapi/cpp/source/decision_forest/df_cls_dense_batch.cpp
```diff
@@ -18,50 +18,51 @@
...
-          .set_min_observations_in_leaf_node(8)
-          .set_min_observations_in_split_node(16)
-          .set_min_weight_fraction_in_leaf_node(0.0)
-          .set_min_impurity_decrease_in_split_node(0.0)
-          .set_variable_importance_mode(df::variable_importance_mode::mdi)
-          .set_error_metric_mode(df::error_metric_mode::out_of_bag_error)
-          .set_infer_mode(df::infer_mode::class_labels | df::infer_mode::class_probabilities)
-          .set_voting_mode(df::voting_mode::weighted);
+    const auto df_desc =
+        df::descriptor<float, df::task::classification, df::method::dense>{}
```
michael-smirnov, line 41: What are the default parameter values for the second and third template parameters of df? We can make it simpler if here the default values are used

examples/oneapi/cpp/source/decision_forest/df_reg_dense_batch.cpp
rlnx: @michael-smirnov, not sure it's good idea, `read` is invariant for multiple data sources. This may demonstrate wrong usage scenario in case where multiple data sources can be constructed from a string. Currently it works, because we have only one data source. When a new data source is added, for example `hdf5` that also accepts path to file, the suggested example leads to the compilation error due to ambiguity between `csv::data_source` and `hdf5::data_source` constructors.

examples/oneapi/cpp/source/decision_forest/df_reg_dense_batch.cpp
```diff
@@ -18,45 +18,47 @@
 #include "oneapi/dal/algo/decision_forest.hpp"
 #include "oneapi/dal/io/csv.hpp"
 
-using namespace oneapi;
+namespace onedal = oneapi::dal;
```
rlnx, line 21: Is it xLib suggesting to have `onedal` alias for `oneapi::dal`?

examples/oneapi/cpp/source/decision_forest/df_reg_dense_batch.cpp
michael-smirnov: Agree

examples/oneapi/cpp/source/decision_forest/df_reg_dense_batch.cpp
michael-smirnov: No - they suggest more general things:
1. Do not put `using namespace` in the examples
2. Create and aliases instead

vlad-nazarov: Change onedal alias to dal

michael-smirnov (approved):

PR #1242, opened 2020-10-30
added overflow checks and corresponding casts

makart19:
added mul overflow checks and casts were necessary for PCA, KMeans, KMeansInit


cpp/oneapi/dal/algo/kmeans/BUILD
```diff
@@ -41,6 +52,7 @@ dal_test_suite(
 dal_test_suite(
     name = "tests",
     host_tests = [
+        ":common_tests",
         ":cpu_tests",
     ],
     dpc_tests = [
```
michael-smirnov, line 58: Why you do not include common tests to the `dpc_tests` if they are really common?

michael-smirnov (commented):
Don't forget to add backport label

cpp/oneapi/dal/algo/kmeans/BUILD
amgrigoriev: It seems to me, common tests should be able to run on host only (w/o GPU)

makart19: yes, from my perspective in common tests we are testing common parts/branches for CPU/GPU (like descriptor) on host, i.e. we can avoid launch on GPU.

cpp/oneapi/dal/algo/kmeans/common_kmeans_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "oneapi/dal/algo/kmeans/infer.hpp"
+#include "oneapi/dal/algo/kmeans/train.hpp"
+#include "oneapi/dal/table/row_accessor.hpp"
+
+using namespace oneapi;
+namespace kmn = oneapi::dal::kmeans;
+
+using kmn_desc = kmn::descriptor<float, kmn::method::lloyd_dense, kmn::task::clustering>;
+
+TEST(kmn_bad_arg_tests, set_cluster_count) {
```
amgrigoriev, line 27: Maybe kmeans_bad_arg_tests?

cpp/oneapi/dal/algo/kmeans/BUILD
michael-smirnov: We have two interfaces of oneDAL: host and dpc++. When you say "common tests," I think these tests shall work for both interfaces - okay, if you will test in only on CPUs for dpc++ interface. But seems like you are testing the host interface only.

cpp/oneapi/dal/algo/kmeans/common_kmeans_test.cpp
makart19: done

mchernov-intel (approved):

cpp/oneapi/dal/algo/kmeans/BUILD
makart19: made tests renaming

PR #1252, opened 2020-11-02
Add v1 to all algorithms + consistency

rlnx:
- Add v1 namespace to all algorithms
- Fix variadic templates in `train_ops`/`infer_op`/`compute_ops` to prevent future backward compatibility problem
- Add `static_assert` to validate template parameters to all algorithms
- Add `task_t` type alias to all algorithms including inputs and results
- Consistent approach for returning tables from inputs/results via const reference
- Slightly refactored kernel function handling in SVM 
- Correct `private`/`protected` visibility
- Add type traits to validate tags of the descriptors
- Remove redundant `ONEDAL_EXPORT`

### DIFF
--- a/cpp/oneapi/dal/algo/knn/common.hpp
+++ b/cpp/oneapi/dal/algo/knn/common.hpp
@@ -22,27 +22,61 @@
 namespace oneapi::dal::knn {
 
 namespace task {
+namespace v1 {
 struct classification {};
 using by_default = classification;
+} // namespace v1
+
+using v1::classification;
+using v1::by_default;
+
 } // namespace task
 
-namespace detail {
-struct tag {};
-
-template <typename Task = task::by_default>
-class descriptor_impl;
-
-class model_impl;
-} // namespace detail
-
 namespace method {
+namespace v1 {
 struct kd_tree {};
 struct brute_force {};
 using by_default = brute_force;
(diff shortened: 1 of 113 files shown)


michael-smirnov (commented):
Overall looks good - but there are unresolved externals on Windows

cpp/oneapi/dal/algo/knn/common.hpp
```diff
@@ -22,27 +22,61 @@
...
+template <typename Task>
 class model_impl;
+} // namespace v1
+
+using v1::tag;
+using v1::descriptor_impl;
+using v1::model_impl;
+
+template <typename Float>
+constexpr bool is_valid_float_v = dal::detail::is_one_of_v<Float, float, double>;
```
amgrigoriev, line 63: Is it specific for kNN?

amgrigoriev (approved):

cpp/oneapi/dal/algo/knn/common.hpp
rlnx: There is uncertainty in floating point types need to be supported in the future, this could be different for different algorithms, so I've made algorithm-specific `is_valid_float`

michael-smirnov (approved):

PR #1261, opened 2020-11-03
Using safe arithmetic in graph

orrrrtem:
*safe arithmetic to avoid integer overflow

*handling primitive data types

* extra fixes for coding style.


cpp/oneapi/dal/detail/error_messages.cpp
```diff
@@ -116,6 +117,7 @@ MSG(empty_edge_list, "Empty edge list")
 MSG(interval_gt_vertex_count, "Interval is greater than vertex count")
 MSG(negative_interval, "Negative interval")
 MSG(row_begin_gt_row_end, "Row begin is greater than row end")
+MSG(range_idx_gt_max_int32, "Range indexes are greater than max of int32")
```
bysheva, line 120: indexes -> indices

bysheva (approved):

cpp/oneapi/dal/detail/error_messages.cpp
orrrrtem: Both "indexes" and "indices" are acceptable plural forms of the word "index" or to refer to more than one index. Index is one of those rare words that have two different plurals in English. "Indices" is originally a Latin plural, while "Indexes" has taken the English way of making plurals, using –s or –es

PR #1264, opened 2020-11-04
Add v1 to data management + consistency

rlnx:
- Add v1 namespace to `dal::detail` and data management
- Move `descriptor_base` to `detail`
- Rename `tag` -> `descriptor_tag`


cpp/oneapi/dal/table/common.cpp
```diff
@@ -31,16 +31,19 @@ class detail::table_metadata_impl {
     virtual const data_type& get_data_type(int64_t index) const = 0;
 };
 
-namespace backend {
+namespace v1 {
+
 class empty_metadata_impl : public detail::table_metadata_impl {
```
michael-smirnov, line 36: Isn't it safer to inherit from `detail::va::table_metadata_impl`?

michael-smirnov (approved):
Reviewed the table part only - just one minor question asked

PR #1271, opened 2020-11-10
Fix build issue for GCC compiler

Pahandrovich:


Pahandrovich:
http://intel-ci.intel.com/eb2402f4-48d3-f14f-b1aa-8cdcd4b723a1

Pahandrovich:
gnu 1010
https://intel-ci.intel.com/eb2423d8-0566-f1fc-b061-8cdcd4b723a1

cpp/daal/include/services/daal_defines.h
```diff
@@ -30,7 +30,9 @@
 
 #if !(defined(_MSC_VER) || defined(__INTEL_COMPILER))
     #define __int32 int
-    #define __int64 long long int
+    #if !defined(__int64)
+        #define __int64 long long int
+    #endif
```
rlnx, line 35: Why don't we handle this here for GNU 10, but redefine in makefile? Having one place where these types are defined is much better.

Pahandrovich: I don't quite understand what you mean.
Is it better for all compilers to define __int64 in the makefile?

Pahandrovich:
nightly
http://intel-ci.intel.com/eb253259-5ef6-f106-aacc-8cdcd4b723a1
http://intel-ci.intel.com/eb253256-429a-f153-885d-8cdcd4b723a1

cpp/daal/src/services/library_version_info.cpp
```diff
@@ -25,6 +25,7 @@
 #include "src/services/daal_version.h"
 #include "src/services/service_defines.h"
 #include "services/env_detect.h"
+#include "services/daal_defines.h"
 #include "mkl_daal.h"
```
rlnx, line 29: Why `mkl_daal.h` is required here?

cpp/daal/src/externals/core_threading_win_dll.cpp
```diff
@@ -26,6 +26,7 @@
 #include "src/threading/threading.h"
 #include "src/threading/service_thread_pinner.h"
 #include "services/env_detect.h"
+#include "services/daal_defines.h"
```
rlnx, line 29: Not required here, `env_detect.h` already includes `daal_defines.h`

cpp/daal/src/services/library_version_info.cpp
Pahandrovich: `fpk_serv_cpuisknm` variable defined in `mkl_daal.h`
https://github.com/oneapi-src/oneDAL/blob/master/cpp/daal/src/services/library_version_info.cpp#L48

cpp/daal/src/externals/core_threading_win_dll.cpp
Pahandrovich: Fixed,
thanks

rlnx: Why `mkl_daal.h` is deleted here? Is it needed or not? If it's not needed, why `service/daal_defines.h` is added?

My previous comment was about order of includes, it was
```c++
#include "services/env_detect.h"
#include "mkl_daal.h"
```

You add `service/daal_defines.h`, so it becomes
```
#include "services/env_detect.h"
#include "service/daal_defines.h"
#include "mkl_daal.h"
```

My comment was about `service/daal_defines.h`, it's not needed, because `services/env_detect.h` already includes it.

rlnx (approved):

PR #1274, opened 2020-11-11
change compiler for sycl examples

napetrov:
switching nmake example to dpcpp


michael-smirnov (approved):

michael-smirnov:
@napetrov I think its worth to change compiler from `clang++` to `dpcpp` for linux sycl examples as well

PR #1306, opened 2020-12-02
Logistic regression online example

michael-smirnov:
# Logistic regression online example

- created SYCL examples of logistic regression online algorithm using SGD solver and loss functions
- fixed initialization of starting point in logistic regression batch algorithm
- relaxed conditions of some asserts in logreg - they are valid now in the case of both batch and online computations

### DIFF
--- a/examples/daal/cpp_sycl/source/logistic_regression/log_reg_dense_online.cpp
+++ b/examples/daal/cpp_sycl/source/logistic_regression/log_reg_dense_online.cpp
@@ -29,7 +29,7 @@
  * \example log_reg_dense_online.cpp
  */
 
-#include "logistic_regression_online.hpp"
+#include "logistic_regression_online.h"
 #include "service.h"
 #include "service_sycl.h"
 
@@ -73,7 +73,6 @@ int main(int argc, char * argv[])
 logistic_regression::ModelPtr trainModel()
 {
     constexpr bool isIntercept           = false;
-    constexpr float l1Penalty            = 0.0f;
     constexpr float l2Penalty            = 0.0f;
     constexpr float learningRate         = 1.0f;
     constexpr size_t nEpochs             = 20;
@@ -85,7 +84,7 @@ logistic_regression::ModelPtr trainModel()
     NumericTablePtr mergedDataBlock(new MergedNumericTable(xBlock, yBlock));
 
     LogisticRegressionOnline alg;
-    alg.setParams(nClasses, nFeatures, isIntercept, l1Penalty, l2Penalty, dataBlockRowCount, nIterationsPerBlock);
+    alg.setParams(nClasses, nFeatures, isIntercept, l2Penalty, dataBlockRowCount, nIterationsPerBlock);
 
     for (size_t epoch = 0; epoch < nEpochs; epoch++)
     {
--- a/examples/daal/cpp_sycl/source/logistic_regression/log_reg_dense_online.cpp
+++ b/examples/daal/cpp_sycl/source/logistic_regression/log_reg_dense_online.cpp
@@ -1,0 +1,144 @@
+/* file: log_reg_dense_online.cpp */
+/*******************************************************************************
+* Copyright 2020 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+/*
+!  Content:
+!    C++ example of logistic regression in the online processing mode
+!    with DPC++ interfaces.
+!
+!    The program trains the logistic regression model on a training
+!    datasetFileName and computes classification for the test data.
+!******************************************************************************/
+
+/**
+ * <a name="DAAL-EXAMPLE-CPP-LOG_REG_DENSE_ONLINE"></a>
+ * \example log_reg_dense_online.cpp
+ */
+
+#include "logistic_regression_online.hpp"
+#include "service.h"
+#include "service_sycl.h"
+
+using namespace daal;
+using namespace daal::algorithms;
+using namespace daal::algorithms::optimization_solver;
+using namespace daal::data_management;
+
+/* Input data set parameters */
+const std::string trainDatasetFileName = "../data/batch/logreg_train.csv";
+const std::string testDatasetFileName  = "../data/batch/logreg_test.csv";
+constexpr size_t nFeatures             = 6;
+constexpr size_t nClasses              = 5;
+
+logistic_regression::ModelPtr trainModel();
+void testModel(const logistic_regression::ModelPtr& model);
+
+int main(int argc, char * argv[])
+{
+    checkArguments(argc, argv, 2, &trainDatasetFileName, &testDatasetFileName);
+
+    for (const auto & deviceSelector : getListOfDevices())
+    {
+        const auto & nameDevice = deviceSelector.first;
+        const auto & device     = deviceSelector.second;
+        cl::sycl::queue queue(device);
+        std::cout << "Running on " << nameDevice << "\n\n";
+
+        services::SyclExecutionContext ctx(queue);
+        services::Environment::getInstance()->setDefaultExecutionContext(ctx);
+
+        auto model = trainModel();
+        printNumericTable(model->getBeta(), "Logistic Regression coefficients:");
+
+        testModel(model);
+    }
+
+    return 0;
+}
+
+logistic_regression::ModelPtr trainModel()
+{
+    constexpr bool isIntercept           = false;
+    constexpr float l1Penalty            = 0.0f;
+    constexpr float l2Penalty            = 0.0f;
+    constexpr float learningRate         = 1.0f;
+    constexpr size_t nEpochs             = 20;
+    constexpr size_t nIterationsPerBlock = 5;
+    constexpr size_t dataBlockRowCount   = 500;
+
+    auto xBlock = SyclHomogenNumericTable<>::create(nFeatures, 0, NumericTable::notAllocate);
+    auto yBlock = SyclHomogenNumericTable<>::create(1, 0, NumericTable::notAllocate);
+    NumericTablePtr mergedDataBlock(new MergedNumericTable(xBlock, yBlock));
+
+    LogisticRegressionOnline alg;
+    alg.setParams(nClasses, nFeatures, isIntercept, l1Penalty, l2Penalty, dataBlockRowCount, nIterationsPerBlock);
+
+    for (size_t epoch = 0; epoch < nEpochs; epoch++)
+    {
+        FileDataSource<CSVFeatureManager> dataSource(trainDatasetFileName,
+                                                    DataSource::notAllocateNumericTable,
+                                                    DataSource::doDictionaryFromContext);
+
+        while (dataSource.loadDataBlock(dataBlockRowCount, mergedDataBlock.get()) == dataBlockRowCount)
+        {
+            alg.setIterationParams(learningRate);
+            alg.setInput(xBlock, yBlock);
+            alg.compute();
+        }
+    }
+
+    alg.finalizeCompute();
+    return alg.getModel();
+}
+
+void testModel(const logistic_regression::ModelPtr& model)
+{
+    namespace class_prediction = classifier::prediction;
+    using LogRegPrediction = logistic_regression::prediction::Batch<>;
+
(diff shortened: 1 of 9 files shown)


examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
```diff
@@ -0,0 +1,219 @@
...
+    SGDMinibatchSolverPtr _solver;
+
+    LogLossPtr _logLoss;
+    CrossEntropyLossPtr _crossEntropyLoss;
+
+    daal_dm::NumericTablePtr _workPoint;
+    daal_dm::NumericTablePtr _learningRateTable;
+    daal_dm::NumericTablePtr _finalBetas;
+
+    size_t _batchRowCount;
```
PetrovKP, line 213: Can some fields be obtained from tables `_workPoint`, `_learningRateTable`, `_finalBetas` methods getNumberOfRows, getNumberOfColumns?

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
```diff
@@ -0,0 +1,219 @@
...
+    using LogLossPtr = daal::services::SharedPtr<LogLoss>;
+
+    using CrossEntropyLoss = daal_solver::cross_entropy_loss::Batch<Fptype>;
+    using CrossEntropyLossPtr = daal::services::SharedPtr<CrossEntropyLoss>;
+
+    using SGDMinibatchSolver = daal_sgd::Batch<Fptype, daal_sgd::miniBatch>;
+    using SGDMinibatchSolverPtr = daal::services::SharedPtr<SGDMinibatchSolver>;
+
+public:
+    LogisticRegressionOnline()
```
PetrovKP, line 44: I would suggest to align with batch

```suggestion
    LogisticRegressionOnline(size_t nClasses)
```

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
```diff
@@ -0,0 +1,219 @@
...
+        : _batchRowCount(0), _nFeatures(0), _nClasses(0), _nBetas(0), _nBetaCols(0), _isIntercept(false) {}
+
+    // Sets the algorithm parameters
+    void setParams(size_t nClasses,
+                   size_t nFeatures,
```
PetrovKP, line 49: This information can be obtained from the table when setting the input.

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
```diff
@@ -0,0 +1,219 @@
...
+            }
+
+            _workPoint->releaseBlockOfRows(bd);
+        }
+    }
+
+private:
+    SGDMinibatchSolverPtr _solver;
+
+    LogLossPtr _logLoss;
```
PetrovKP, line 206: LogLossPtr  and CrossEntropyLossPtr  have a base class objective_function::Batch why not store one instance of this pointer?

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
```diff
@@ -0,0 +1,219 @@
+/* file: logistic_regression_online.hpp */
```
PetrovKP, line 1: in all examples we use the extension *.h for headers.

```suggestion
/* file: logistic_regression_online.h */
```

PetrovKP:
You forgot to add new examples to the file: daal_lnx.lst, without this your examples will not run in CI.

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+public:
+    LogisticRegressionOnline()
+        : _batchRowCount(0), _nFeatures(0), _nClasses(0), _nBetas(0), _nBetaCols(0), _isIntercept(false) {}
+
+    // Sets the algorithm parameters
+    void setParams(size_t nClasses,
+                   size_t nFeatures,
+                   bool isIntercept,
+                   Fptype l1_penalty,
```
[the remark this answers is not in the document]
michael-smirnov: Do not see this in the documentation. The only thing I've found is that L1 is not supported on GPU: https://docs.oneapi.com/versions/latest/onedal/dev_guide/algorithms/logistic_regression/logistic-regression.html
Neither loss functions nor solver docs say that L1 regularization is not supported for SGD: https://docs.oneapi.com/versions/latest/onedal/dev_guide/algorithms/optimization-solvers/solvers/stochastic-gradient-descent-algorithm.html

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
michael-smirnov: No. `_workPoint` stores betas as 1D array, `_learningRateTable` has the shape of 1x1, `_finalBetas` are only allocated in finalize().

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
michael-smirnov: Do not see the reason for this. I would like to initialize all the parameters at once instead of separating them.

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
michael-smirnov: But I need it here to initialize table for betas.

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
michael-smirnov: Because as you see [here](https://github.com/oneapi-src/oneDAL/blob/e8e3085b6c69c595e1f29f13d852fe06e100bce7/examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp#L113) I must set input to the loss function. Base class do not allow to set it - `data` and `dependentVariables` are different enum values for logLoss and CrossEntropy. 
Moreover, in setParameter() I  set loss function parameters after extracting it from the solver. It is also cannot be done via base class. Rather than having the same `if`s with casts from base class to the original one I would like to make the code simpler.

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+    using CrossEntropyLoss = daal_solver::cross_entropy_loss::Batch<Fptype>;
+    using CrossEntropyLossPtr = daal::services::SharedPtr<CrossEntropyLoss>;
+
+    using SGDMinibatchSolver = daal_sgd::Batch<Fptype, daal_sgd::miniBatch>;
+    using SGDMinibatchSolverPtr = daal::services::SharedPtr<SGDMinibatchSolver>;
+
+public:
+    LogisticRegressionOnline()
+        : _batchRowCount(0), _nFeatures(0), _nClasses(0), _nBetas(0), _nBetaCols(0), _isIntercept(false) {}
```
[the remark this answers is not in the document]
michael-smirnov: Removed, thanks

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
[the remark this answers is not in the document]
michael-smirnov: I removed L1 regularization from parameters anyway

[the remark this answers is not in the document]
PetrovKP: I'm very surprised. We kind of added this information, @ShvetsKS didn't we?

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
PetrovKP: ok

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
PetrovKP: I see you divided on parts: constructors, set_params, set_inputs, compute, finalize by analogy with container of algorithms. So, in this case, it would look consistency. Otherwise, why not pass all these parameters directly to the constructor?

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
PetrovKP: It's a pity, it would be nice.

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.h
```diff
@@ -0,0 +1,265 @@
...
+            _finalBetas->getBlockOfRows(0, _nBetas, daal_dm::writeOnly, bFinalBetas);
+
+            auto src = bBetas.getBlockPtr();
+            auto dst = bFinalBetas.getBlockPtr();
+
+            for (size_t i = 0; i < _nBetas; i++)
+            {
+                for (size_t j = 0; j < _nFeatures; j++)
+                {
+                    dst[i*_nFeatures + j] = src[i*_nBetaCols + j + 1];
```
PetrovKP, line 176: [minor] For good, this code should be on sycl.

examples/daal/cpp_sycl/source/logistic_regression/log_reg_dense_online.cpp
```diff
@@ -0,0 +1,143 @@
...
+    /* Create an algorithm object to predict values of logistic regression */
+    LogRegPrediction algorithm(nClasses);
+    algorithm.parameter().resultsToEvaluate |= classifier::computeClassProbabilities | classifier::computeClassLogProbabilities;
+
+    /* Pass a testing data set and the trained model to the algorithm */
+    algorithm.input.set(class_prediction::data, testX);
+    algorithm.input.set(class_prediction::model, model);
+
+    /* Predict values of logistic regression */
+    algorithm.compute();
```
PetrovKP, line 133: [minor] prediction would also be good to show online, especially it's very easy to follow.

examples/daal/cpp_sycl/source/logistic_regression/log_reg_dense_online.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+        testModel(model);
+    }
+
+    return 0;
+}
+
+logistic_regression::ModelPtr trainModel()
+{
+    constexpr bool isIntercept           = false;
```
PetrovKP, line 75: [style] Cool of course, but in other examples, no one does that. Looks probably not consistency.

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.h
```diff
@@ -0,0 +1,265 @@
...
+            for (size_t i = 0; i < bd.getNumberOfRows(); i += nBetaCols)
+            {
+                ptr[i] = Fptype(1e-3);
+            }
+
+            _workPoint->releaseBlockOfRows(bd);
+        }
+    }
+
+    void checkParameterDomains(size_t nClasses,
```
PetrovKP, line 225: [minor] I think this is superfluous - we are not writing a full-product class/algorithm, but just an simple example, otherwise the checks will not be enough

PetrovKP (approved):

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.hpp
michael-smirnov: It's done with intent to have default ctor of the class - this avoids user the need to create multiple objects to solve the tasks with different parameters, and it's easy to initialize user-owned classes which hold `LogisticRegressionOnline` object - no need to store it by pointer if we have a default ctor.

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.h
michael-smirnov: It's done this way for simplicity - if you insist on the change, I can apply it

examples/daal/cpp_sycl/source/logistic_regression/log_reg_dense_online.cpp
michael-smirnov: It's done this way for simplicity - if you insist on the change, I can apply it

examples/daal/cpp_sycl/source/logistic_regression/log_reg_dense_online.cpp
michael-smirnov: Where do you see the inconsistency?

examples/daal/cpp_sycl/source/logistic_regression/logistic_regression_online.h
michael-smirnov: Why do you think these checks are not enough?

KulikovNikita (approved):

PR #1310, opened 2020-12-03
Cleanup a few makefile issues for examples/oneapi/dpc

william-r-dieter:
There was a syntax error, which prevents the Linux Makefile working, and also several files appear to have been renamed recently.  These changes adapt to those issues.


rlnx:
@william-r-dieter, `onedal_lnx.lst` and `makefile_lnx.` are copied to the build directory as `onedal.lst` and `makefile`, so there is no issue here. Examples are not expected to be launched from this dir. The idea to have templates that should be substituted with some values when build happens.

PR #1318, opened 2020-12-07
Add additional safe checks for CSR table

PetrovKP:


cpp/daal/include/data_management/data/csr_numeric_table.h
```diff
@@ -1214,7 +1214,10 @@ class DAAL_EXPORT CSRNumericTable : public NumericTable, public CSRNumericTableI
             }
 
             size_t * row_offsets = block.getBlockRowIndicesSharedPtr().get();
-
+            if (row_offsets == NULL)
+            {
+                return services::Status(services::ErrorMemoryAllocationFailed);
```
michael-smirnov, line 1219: It's not clear why:
1. This particular error message is used. If there is a problem with memory allocation, why we cannot check it directly where this memory is allocated?
2. Why this check is done only for row indices? What's about column indices, values on lines 1064, 1073, 1134, 1199, 1249?

PetrovKP: Because only `row_offsets` pointer use next lines. The static analyzer swears that  `row_offsets` can be `nullptr`.

michael-smirnov (approved):

PR #1351, opened 2020-12-23
PCA Covariance Method

owerbat:


cpp/daal/include/algorithms/covariance/covariance_batch.h
```diff
@@ -341,6 +341,16 @@ class DAAL_EXPORT Batch : public BatchImpl
...
+    /**
+     * Constructs an algorithm for correlation or variance-covariance matrix computation
+     * \param[in] outMatrixType The type of matrix (covariance or correlation) to calculate
+     */
+    Batch(OutputMatrixType outMatrixType)
+    {
+        initialize();
+        parameter.outputMatrixType = outMatrixType;
+    }
+
```
SmirnovEgorRu, line 353: Do we really need to change APIs?

cpp/daal/src/algorithms/pca/pca_dense_correlation_batch_impl.i
```diff
@@ -111,6 +114,7 @@ services::Status PCACorrelationKernel<batch, algorithmFPType, cpu>::compute(
             DAAL_CHECK_STATUS(status, this->copyVarianceFromCovarianceTable(covarianceTable, variances));
         }
 
+        if (needCorrelation)
         {
             DAAL_ITTNOTIFY_SCOPED_TASK(compute.full.correlationFromCovariance);
             DAAL_CHECK_STATUS(status, this->correlationFromCovarianceTable(covarianceTable));
```
SmirnovEgorRu, line 120: Can we compute this by setting:
```cpp
covarianceAlg->parameter.outputMatrixType = covariance::correlationMatrix;
```

cpp/daal/src/algorithms/pca/transform/pca_transform_dense_default_batch_impl.i
```diff
@@ -82,9 +82,18 @@ services::Status ComputeInvSigmas(NumericTable * pVariances, TArray<algorithmFPT
...
-            pInvSigmas[varianceId] = pRawVariances[varianceId] ?
-                                         algorithmFPType(1.0) / daal::internal::Math<algorithmFPType, cpu>::sSqrt(pRawVariances[varianceId]) :
-                                         algorithmFPType(0.0);
+            if (pRawVariances[varianceId] > 0)
+            {
+                pInvSigmas[varianceId] = algorithmFPType(1.0) / daal::internal::Math<algorithmFPType, cpu>::sSqrt(pRawVariances[varianceId]);
+            }
+            else if (pRawVariances[varianceId] < 0)
+            {
+                pInvSigmas[varianceId] = algorithmFPType(1.0) / daal::internal::Math<algorithmFPType, cpu>::sSqrt(-pRawVariances[varianceId]);
```
SmirnovEgorRu, line 91: Is it mathematically correct to reverse variance?

cpp/daal/src/algorithms/pca/pca_dense_correlation_batch_impl.i
owerbat: No, because we need to compute variances which can't be computed by correlation algorithm. So, firstly we compute covariance, secondly we compute variances from covariance and then correlation if needed.
Actually, we can compute means and variances from the original data and then compute covariance/correlation depending on `covarianceAlg->parameter.outputMatrixType`. But I don't think it's the best option from the performance perspective

cpp/daal/src/algorithms/pca/transform/pca_transform_dense_default_batch_impl.i
owerbat: Values from `pRawVariances` are negative in some cases which is not mathematically correct at all. Anyway, MKL solver returns these values.
May be we should raise an exception in these cases? Is it a better option?

mchernov-intel:
@owerbat if you are working on the algorithm please add algorithm-owners from GPU team too.

cpp/daal/src/algorithms/pca/pca_dense_correlation_batch_impl.i
```diff
@@ -91,6 +91,9 @@ services::Status PCACorrelationKernel<batch, algorithmFPType, cpu>::compute(
     {
         DAAL_ITTNOTIFY_SCOPED_TASK(compute.full);
 
+        bool needCorrelation                      = covarianceAlg->parameter.outputMatrixType == covariance::correlationMatrix;
```
mchernov-intel, line 94: const

cpp/daal/src/algorithms/pca/pca_dense_correlation_batch_impl.i
```diff
@@ -91,6 +91,9 @@ services::Status PCACorrelationKernel<batch, algorithmFPType, cpu>::compute(
...
+        covarianceAlg->parameter.outputMatrixType = covariance::covarianceMatrix;
```
mchernov-intel, line 95: Is it really OK to change output matrix type inside the kernel?

cpp/daal/src/algorithms/pca/transform/pca_transform_dense_default_batch_impl.i
```diff
@@ -82,9 +82,18 @@ services::Status ComputeInvSigmas(NumericTable * pVariances, TArray<algorithmFPT
         PRAGMA_VECTOR_ALWAYS
         for (size_t varianceId = 0; varianceId < numFeatures; ++varianceId)
         {
-            pInvSigmas[varianceId] = pRawVariances[varianceId] ?
-                                         algorithmFPType(1.0) / daal::internal::Math<algorithmFPType, cpu>::sSqrt(pRawVariances[varianceId]) :
-                                         algorithmFPType(0.0);
+            if (pRawVariances[varianceId] > 0)
```
mchernov-intel, line 85: Please, provide math explanation in the description.

cpp/daal/src/algorithms/pca/transform/pca_transform_dense_default_batch_impl.i
SmirnovEgorRu: Probably, better to cast negative values to zeros?

cpp/daal/src/algorithms/pca/pca_dense_correlation_batch_impl.i
SmirnovEgorRu: OK, but please restore prev value of outputMatrixType after computing of covariance.

owerbat:
I found a way to patch covariance PCA without any changes in C++. So, I'm going to close this PR

PR #1353, opened 2020-12-24
Update env script to support mksh

napetrov:
# Description
Updating env scripts to support mksh

### DIFF
--- a/deploy/local/vars_lnx.sh
+++ b/deploy/local/vars_lnx.sh
@@ -161,7 +161,7 @@ fi
 __daal_tmp_dir="<INSTALLDIR>"
 __daal_tmp_dir=$__daal_tmp_dir/dal
 if [ ! -d $__daal_tmp_dir ]; then
-    __daal_tmp_dir=$(dirname -- "$(rreadlink "$vars_sourced_nm")")
+    __daal_tmp_dir=$(dirname -- "$(dirname -- "$(rreadlink "$vars_sourced_nm")")../")
 fi
 
 export DAL_MAJOR_BINARY=__DAL_MAJOR_BINARY__
--- a/deploy/local/vars_lnx.sh
+++ b/deploy/local/vars_lnx.sh
@@ -88,8 +88,6 @@ rreadlink() (
 # code block below that contain elements that are specific to a shell. The
 # shell-specific elements are needed to identify the sourcing shell.
 
-vars_script_name="vars.sh"
-
 vars_sourced=0 ;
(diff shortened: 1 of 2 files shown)


PetrovKP:
How can we check the functionality of the generated scripts for mksh?

napetrov:
@PetrovKP  - looks it works now and per my check works for mksh. could you please review?

deploy/local/vars_lnx.sh
```diff
@@ -15,77 +15,153 @@
...
-  while [ -L "$script" ] ; do
-    # combining next two lines fails in zsh shell
-    script_dir=$(command dirname -- "$script")
-    script_dir=$(cd "$script_dir" && command pwd -P)
-    script="$(readlink "$script")"
-    case $script in
-      (/*) ;;
-       (*) script="$script_dir/$script" ;;
-    esac
+rreadlink() (
```
PetrovKP, line 51: I see a lot of the same code in two scripts for mac and lnx. Is it possible to use the common part? Because any change will be very expensive

deploy/local/vars_lnx.sh
```diff
@@ -15,77 +15,153 @@
...
+  # shellcheck disable=2128
+  vars_sourced_nm="$BASH_SOURCE" ;      # ditto
+  (return 0 2>/dev/null) && vars_sourced=1 ;
+# TODO: following needs further testing to work in dash
+elif [ "${0:-}" = "dash" ] ; then
+  vars_sourced=1 ;                      # see error messages below for outcome
+  vars_sourced_sh="dash" ;              # see error messages below for outcome
+  vars_sourced_nm="$vars_sourced_nm" ;
+# Only reliable way to detect sh source is to know the name of this script.
+# TODO: following needs further testing to work in sh
```
PetrovKP, line 137: Have you checked now what works locally on Mac and Linux?

deploy/local/vars_lnx.sh
napetrov: this is question to merging separate scripts and updating makefile to use - i don't want to handle this as we also have feebsd env file

deploy/local/vars_lnx.sh
napetrov: dash would just fail

napetrov:
/intelci: run
set infra_branch=dev/napetrov_ipldt_pi

deploy/local/vars_lnx.sh
napetrov: actually i've just commit updated version with dash support

PetrovKP (approved):

PR #1376, opened 2021-01-21
Fix stability in decision forest prediction

agorshk:
1. Fixed instability in training due to undefined order of pushing trees into a model.
2. Fixed instability in prediction due to usage of TLS with undefined order of reduction in one of code branches. + possible optimizations for this case.
3. More accurate summation for probabilities in prediction to avoid cases, when sum of probabilities > 1.0


agorshk:
/intelci: run
set infra_branch=dev/agorshk-forest_fix_model_builder_tests

PetrovKP (approved):

PR #1378, opened 2021-01-21
Common test infrastructure + PCA tests migration

rlnx:
- `test/engine` for basic test routines
  - `linalg` module to simplify linear algebra operations
  - `mkl` thin wrappers over MKL


cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_svd.cpp
```diff
@@ -61,21 +69,23 @@ static result_t call_daal_kernel(const context_cpu& ctx,
...
     const auto daal_means = interop::convert_to_daal_homogen_table(arr_means, 1, column_count);
     const auto daal_variances = interop::convert_to_daal_homogen_table(arr_vars, 1, column_count);
 
     daal_pca::internal::InputDataType dtype = daal_pca::internal::nonNormalizedDataset;
 
-    daal_pca::interface3::BatchParameter<Float, daal_pca::svdDense> parameter;
+    auto norm_alg = get_normalization_algorithm<Float>();
+    norm_alg->input.set(daal::algorithms::normalization::zscore::data, daal_data);
+    norm_alg->parameter().resultsToCompute |= daal::algorithms::normalization::zscore::mean;
+    norm_alg->parameter().resultsToCompute |= daal::algorithms::normalization::zscore::variance;
```
michael-smirnov, line 83:
```suggestion
    namespace daal_zscore = daal::algorithms::normalization::zscore;
    
    auto norm_alg = get_normalization_algorithm<Float>();
    norm_alg->input.set(daal_zscore::data, daal_data);
    norm_alg->parameter().resultsToCompute |= daal_zscore::mean;
    norm_alg->parameter().resultsToCompute |= daal_zscore::variance;
```
just a nip; optional to apply

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_svd.cpp
```diff
@@ -61,21 +69,23 @@ static result_t call_daal_kernel(const context_cpu& ctx,
...
 
-    auto normalizationAlgorithm = parameter.normalization;
-    normalizationAlgorithm->input.set(daal::algorithms::normalization::zscore::data, daal_data);
-    auto algParameter = &(normalizationAlgorithm->parameter());
-    algParameter->resultsToCompute |= daal::algorithms::normalization::zscore::mean;
-    algParameter->resultsToCompute |= daal::algorithms::normalization::zscore::variance;
+    daal_pca::interface3::BatchParameter<Float, daal_pca::svdDense> parameter;
```
michael-smirnov, line 85: why 3rd interface is explicitly used?

cpp/oneapi/dal/backend/linalg/loops.hpp
```diff
@@ -0,0 +1,69 @@
+/*******************************************************************************
```
michael-smirnov, line 1: Where the content of that file planned to be used? Tests only or somewhere else too?

cpp/oneapi/dal/backend/linalg/io.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+#include <iomanip>
+#include <ostream>
+
+#include "oneapi/dal/backend/linalg/matrix.hpp"
+
+namespace oneapi::dal::backend::linalg {
+
+template <typename T, layout lyt>
+class matrix_formatter {
```
michael-smirnov, line 27: Is it used for tests only? If yes, it is better to move it to the test-system-related folder, I guess

cpp/oneapi/dal/backend/linalg/matrix.hpp
```diff
@@ -48,27 +46,27 @@ class shape {
         }
         shape_[0] = row_count;
         shape_[1] = column_count;
-        ONEDAL_ASSERT(count() / columns() == rows(), "Shape count overflow");
+        ONEDAL_ASSERT(get_count() / get_column_count() == get_row_count(), "Shape count overflow");
```
michael-smirnov, line 49: use `ONEDAL_ASSERT_MUL_OVERFLOW`

cpp/oneapi/dal/backend/linalg/matrix.hpp
```diff
@@ -33,6 +27,10 @@ enum class layout {
     column_major,
 };
 
+inline constexpr layout transpose_layout(layout l) {
+    return (l == layout::row_major) ? layout::column_major : layout::row_major;
```
michael-smirnov, line 31: Why you cannot use the existing `data_layout` enum and created another one? You expect they can differ in future?

cpp/oneapi/dal/algo/pca/BUILD
```diff
@@ -17,46 +17,46 @@ dal_module(
 )
 
 dal_test_suite(
-    name = "common_tests",
-    dpc = False,
+    name = "cpu_tests",
```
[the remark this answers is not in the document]
rlnx: This is tests for CPU backend, no DPC++ is needed. There is another `dal_test_suite` that aggregates universal tests for `host`, `dpc-cpu`, `dpc-gpu` configurations.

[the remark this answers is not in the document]
rlnx: Need to keep `dpc = False`.

cpp/oneapi/dal/backend/linalg/io.hpp
rlnx: All linalg stuff is in `test/engine` now

cpp/oneapi/dal/backend/linalg/loops.hpp
rlnx: All linalg stuff is in test/engine now

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_svd.cpp
```diff
@@ -61,21 +69,23 @@ static result_t call_daal_kernel(const context_cpu& ctx,
     const auto daal_data =
         interop::convert_to_daal_homogen_table(arr_data, row_count, column_count);
     const auto daal_eigenvectors =
-        interop::convert_to_daal_homogen_table(arr_eigvec, column_count, component_count);
+        interop::convert_to_daal_homogen_table(arr_eigvec, component_count, column_count);
```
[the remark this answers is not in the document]
rlnx: Merged your changes from master

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_svd.cpp
rlnx: I don't know it was here from the beginning 😄

cpp/oneapi/dal/BUILD
```diff
@@ -116,6 +116,7 @@ dal_collect_test_suites(
         "table",
         "util",
         "backend/linalg",
+        "backend/mkl",
```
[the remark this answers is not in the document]
rlnx: Moved `mkl` to test engine, now `mkl` means "big" MKL

.bazelrc
```diff
@@ -1,3 +1,10 @@
 build -c opt \
       --incompatible_enable_cc_toolchain_resolution \
       --incompatible_require_linker_input_cc_api
+
+
+test:host --test_tag_filters="host"
+test:dpc-cpu --test_tag_filters="dpc" --test_arg="--device=cpu"
+test:dpc-gpu --test_tag_filters="dpc" --test_arg="--device=gpu"
+test:cpu-only --test_tag_filters="cpu-only" --test_arg="--device=cpu"
+test:gpu-only --test_tag_filters="gpu-only" --test_arg="--device=gpu"
```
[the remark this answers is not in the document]
rlnx: Updated

rlnx:
@napetrov, please have a look at `.ci/env/oneapi.sh`

cpp/oneapi/dal/backend/linalg/matrix.hpp
rlnx: There is `unknown` layout in oneDAL. I don't want to handle these cases in linalg module. I suppose the linalg layout will not be extended.

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_svd.cpp
```diff
@@ -38,13 +39,20 @@ template <typename Float, daal::CpuType Cpu>
 using daal_pca_svd_kernel_t = daal_pca::internal::
     PCASVDBatchKernel<Float, daal_pca::interface3::BatchParameter<Float, daal_pca::svdDense>, Cpu>;
 
+template <typename Float>
+inline auto get_normalization_algorithm() {
+    using normalization_alg_t = daal::algorithms::normalization::zscore::
+        Batch<Float, daal::algorithms::normalization::zscore::defaultDense>;
+    return daal::services::SharedPtr<normalization_alg_t>{ new normalization_alg_t{} };
```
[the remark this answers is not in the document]
rlnx: It cannot be deduced, because `SharedPtr<T>` constructor accepts pointer of different type `U` from which `T` cannot be deduced without deduction guide.

.ci/pipeline/ci.yml
```diff
@@ -60,12 +60,21 @@ jobs:
...
+      echo "deb [arch=amd64] https://storage.googleapis.com/bazel-apt stable jdk1.8" | \
+        sudo tee /etc/apt/sources.list.d/bazel.list
+
       sudo apt-get update
       sudo apt-get install gcc-multilib \
                            g++-multilib \
                            binutils \
                            openjdk-11-jdk \
-                           bazel-3.4.1
+                           bazel-4.0.0
```
napetrov, line 75: Might be good to get this regularly updated - for example via https://docs.renovatebot.com/modules/manager/regex/
This shouldn't block integraton

napetrov (approved):

napetrov:
> 
> 
> @napetrov, please have a look at `.ci/env/oneapi.sh`

Looks good

cpp/oneapi/dal/test/engine/dataframe_common.hpp
```diff
@@ -0,0 +1,241 @@
...
+        catch (const std::bad_any_cast&) {
+            return std::nullopt;
+        }
+    }
+
+private:
+    array<float> array_;
+    std::int64_t row_count_;
+    std::int64_t column_count_;
+    std::unordered_map<std::string, std::any> user_fields_;
```
michael-smirnov, line 110: Not clear how fields could be related to data

cpp/oneapi/dal/test/engine/dataframe_common.hpp
```diff
@@ -0,0 +1,241 @@
...
+        }
+        else {
+            const auto value = op();
+            impl_->add_field(name, value);
+            return value;
+        }
+    }
+
+private:
+    mutable dal::detail::pimpl<dataframe_impl> impl_;
```
michael-smirnov, line 181: What was the reason to mark this `mutable`?

cpp/oneapi/dal/test/engine/dataframe_common.hpp
```diff
@@ -0,0 +1,241 @@
...
+};
+
+class dataframe_builder_action {
+public:
+    virtual ~dataframe_builder_action() = default;
+    virtual std::string get_opcode() const = 0;
```
michael-smirnov, line 187: Is opcode just a serialized form of operation?

cpp/oneapi/dal/test/engine/dataframe_common.hpp
```diff
@@ -0,0 +1,241 @@
...
+};
+
+class dataframe {
+public:
+    explicit dataframe(const array<float>& data, std::int64_t row_count, std::int64_t column_count)
+            : dataframe(new dataframe_impl{ data, row_count, column_count }) {}
+
+    explicit dataframe(dataframe_impl* impl) : impl_(impl) {}
+
+    template <typename Float>
```
michael-smirnov, line 120: I don't see that template parameter is used in implementation

cpp/oneapi/dal/test/engine/gtest_main.cpp
```diff
@@ -0,0 +1,85 @@
+/*******************************************************************************
```
michael-smirnov, line 1: Do we plan to delete this when all tests will be moved to Catch2?

cpp/oneapi/dal/test/engine/macro.hpp
```diff
@@ -0,0 +1,72 @@
...
+* limitations under the License.
+*******************************************************************************/
+
+#pragma once
+
+// Disable clang-format as it dramatically
+// affects redability of macro definitions
+// clang-format off
+
+#define _TE_STRINGIFY(x) #x
```
michael-smirnov, line 23: What are the rules for macro namings? When we add underscore prefix? Why we do not add `TE` prefix for all test system -related defines?

cpp/oneapi/dal/test/engine/math.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+    if constexpr (std::is_same_v<Float, double>) {
+        return f64_tol;
+    }
+
+    return 0.0;
+}
+
+template <typename Reference, typename Actual>
+inline double l_inf_norm(const Reference& ref, const Actual& actual) {
```
michael-smirnov, line 40: Sorry, I don't understand that shortening 😄 
Is this a well-known terminology?

cpp/oneapi/dal/test/engine/mocks.hpp
```diff
@@ -0,0 +1,81 @@
...
+* limitations under the License.
+*******************************************************************************/
+
+#pragma once
+
+#include "oneapi/dal/table/common.hpp"
+
+namespace oneapi::dal::test::util {
+
+class dummy_homogen_table {
```
michael-smirnov, line 23: nip: it is just a dummy `impl` of homogen table

michael-smirnov (commented):
reviewed `dal/test/engine` folder without sub-folders

cpp/oneapi/dal/test/engine/config.cpp
```diff
@@ -14,24 +14,49 @@
...
+#include "oneapi/dal/test/engine/config.hpp"
 
 #ifdef ONEDAL_DATA_PARALLEL
-#include "oneapi/dal/test/common.hpp"
+#include "oneapi/dal/test/engine/common.hpp"
 #include "oneapi/dal/backend/interop/common_dpc.hpp"
 #endif
 
-namespace oneapi::dal::test {
+namespace oneapi::dal::test::engine {
```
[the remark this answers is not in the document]
rlnx: The idea was to have `test` namespace for each level. For example, for the functionality which is in `dal` namespace we need to have `test` namespace. At the same time, it's convenient to put all the tests to `test` dir.

```
dal/
   test/         "oneapi::dal::test"
   algo/
      pca/
         test/   "oneapi::dal::pca::test"
      svm/
         test/   "oneapi::dal::svm::test"
   table/
      test/      "oneapi::dal::test"
   io/
      test/      "oneapi::dal::test"
```

cpp/oneapi/dal/test/engine/dataframe_common.hpp
rlnx: Because of `add_field`.

`add_field` allows user to cache some precomputed value associated with dataframe, for example, mean or variance.

The function that computes mean should accept the dataframe by const ref, to not confuse the caller.
```
array<T> mean(const dataframe& df);
```

Of course, we can accept `df` via non-const reference, but it will confuse user of the `mean` function. Because it seems like, the `mean` could change the content of `df`.

Implementation of `mean` could use `df.add_field` or not, but this is implementation and shouldn't affect interface.

cpp/oneapi/dal/test/engine/dataframe_common.hpp
rlnx: Yes, it should be passed to table builder

cpp/oneapi/dal/test/engine/dataframe_common.hpp
rlnx: See explanation here, https://github.com/oneapi-src/oneDAL/pull/1378/files/073d9da0285ed0529f410a009d8c9bc5036ce9b8#r564731411

cpp/oneapi/dal/test/engine/dataframe_common.hpp
```diff
@@ -0,0 +1,241 @@
...
+
+class dataframe_impl {
+public:
+    explicit dataframe_impl(const array<float>& data,
+                            std::int64_t row_count,
+                            std::int64_t column_count)
+            : array_(data),
+              row_count_(row_count),
+              column_count_(column_count) {
+        array_.need_mutable_data();
```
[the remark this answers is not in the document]
rlnx: Probably it's artefact of early implementation. It seems lake mutable data is not needed.

The idea is to make dataframe immutable.

cpp/oneapi/dal/test/engine/gtest_main.cpp
rlnx: Yes, but it will happen in the next PR

cpp/oneapi/dal/test/engine/math.hpp
rlnx: `L_{\infty}`-norm https://en.wikipedia.org/wiki/L-infinity

cpp/oneapi/dal/test/engine/dataframe_common.hpp
rlnx: This is some identifier of `dataframe_builder_action` that is introduced for efficient caching and dataframe reuse across tests. A set of `dataframe_builder_action` builds a sequence of actions to be performed by builder (virtual processor), opcode is very popular term in this context. https://en.wikipedia.org/wiki/Opcode

cpp/oneapi/dal/test/engine/linalg/dot.hpp
```diff
@@ -0,0 +1,50 @@
...
+inline void dot(const matrix<Float, lyt_a>& a,
+                const matrix<Float, lyt_b>& b,
+                matrix<Float, lyt_c>& c,
+                Float alpha = Float(1),
+                Float beta = Float(0)) {
+    dot_op<Float, lyt_a, lyt_b, lyt_c>{}(a, b, c, alpha, beta);
+}
+
+template <typename Float, layout lyt_a, layout lyt_b, layout lyt_c = layout::row_major>
+inline matrix<Float> dot(const matrix<Float, lyt_a>& a,
```
michael-smirnov, line 42: Why resulting matrix type is not dependent on `lyt_c`?

cpp/oneapi/dal/test/engine/linalg/dot.hpp
```diff
@@ -0,0 +1,50 @@
...
+template <typename Float, layout lyt_a, layout lyt_b, layout lyt_c>
+inline void dot(const matrix<Float, lyt_a>& a,
+                const matrix<Float, lyt_b>& b,
+                matrix<Float, lyt_c>& c,
+                Float alpha = Float(1),
+                Float beta = Float(0)) {
+    dot_op<Float, lyt_a, lyt_b, lyt_c>{}(a, b, c, alpha, beta);
+}
+
+template <typename Float, layout lyt_a, layout lyt_b, layout lyt_c = layout::row_major>
```
michael-smirnov, line 41: What if user wants to change layout of `c` - does he required to specify all the template parameters before `lyt_c`?

cpp/oneapi/dal/test/engine/linalg/io.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+#include <iomanip>
+#include <ostream>
+
+#include "oneapi/dal/test/engine/linalg/matrix.hpp"
+
+namespace oneapi::dal::test::engine::linalg {
+
+template <typename T, layout lyt>
+class matrix_formatter {
```
michael-smirnov, line 27: Maybe it's ok for now, but strange to see io staff inside linalg module

cpp/oneapi/dal/test/engine/linalg/loops.hpp
```diff
@@ -0,0 +1,69 @@
...
+*******************************************************************************/
+
+#pragma once
+
+#include "oneapi/dal/test/engine/linalg/matrix.hpp"
+
+namespace oneapi::dal::test::engine::linalg {
+
+template <typename T, layout lyt, typename Op>
+inline void enumerate_row_first(const matrix<T, lyt>& m, Op&& op) {
```
michael-smirnov, line 24: Do we need to have such helpers using sycl kernels in the future?

cpp/oneapi/dal/test/engine/mkl/blas.hpp
```diff
@@ -0,0 +1,38 @@
...
+*******************************************************************************/
+
+#pragma once
+
+#include <cstdint>
+
+namespace oneapi::dal::test::engine::mkl {
+
+template <typename Float>
+void gemm(bool transa,
```
michael-smirnov, line 24: Why we use host interfaces of blas but not oneAPI interfaces?

michael-smirnov (commented):
`mkl` and `linalg` folders reviewed

cpp/oneapi/dal/test/engine/linalg/dot.hpp
rlnx: Obviously yes

cpp/oneapi/dal/test/engine/linalg/loops.hpp
rlnx: This is part of test engine, don't think we'll need to compute anything on GPU in tests. 

Initially I thought that this code will be reused for DAL kernels and tests, but, in fact, we have different requirements for these functionality. For testing, this must be as simple as possible, but for DAL kernels deliver high performance.

cpp/oneapi/dal/test/engine/mkl/blas.hpp
rlnx: This is test engine, gemm is needed for testing purposes only on CPU. It uses public MKL as reference.

cpp/oneapi/dal/test/engine/linalg/dot.hpp
rlnx: If "special" layout for `c` is needed, there is `dot` overload that accepts `c` by reference, so template parameters can be deduced.

cpp/oneapi/dal/test/engine/common.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+template <std::size_t index, typename TupleX, typename TupleY>
+struct combine_types_element {
+private:
+    static constexpr std::size_t count_y = std::tuple_size_v<TupleY>;
+    static constexpr std::size_t i = index / count_y;
+    static constexpr std::size_t j = index % count_y;
+
+public:
+    using type = std::tuple<std::tuple_element_t<i, TupleX>, std::tuple_element_t<j, TupleY>>;
```
amgrigoriev, line 181: TupleX offset is calculated from TupleY. Is it necessary to check sizes?

rlnx: I can put static assert here

cpp/oneapi/dal/test/engine/dataframe_common.hpp
rlnx: Fixed

cpp/oneapi/dal/test/engine/macro.hpp
rlnx: Rule like "add prefix `TE` for all test system-related defines" will not work. There are macros from test test frameworks such as `ASSERT`, `TEST`, `GENERATE`, etc.

Here `_TE` is added to distinguish between "normal" macros such as `ASSERT`, just because I don't want them to be used explicitly in tests.

cpp/oneapi/dal/algo/pca/detail/train_ops.hpp
```diff
@@ -39,28 +39,31 @@ struct train_ops {
...
 
-    void check_preconditions(const Descriptor& params, const input_t& input) const {
+    void check_preconditions(const Descriptor& desc, const input_t& input) const {
         using msg = dal::detail::error_messages;
 
         if (!input.get_data().has_data()) {
             throw domain_error(msg::input_data_is_empty());
         }
-        if (input.get_data().get_column_count() < params.get_component_count()) {
+        if (input.get_data().get_column_count() < desc.get_component_count()) {
```
michael-smirnov, line 48: This condition seems to be incorrect since `dest.get_component_count()` could be `0` which means we should task all the columns of input data for principal components. I think we need to add a special test case for component count equal to `0`

cpp/oneapi/dal/algo/pca/test/badarg.cpp
```diff
@@ -0,0 +1,109 @@
...
+    static constexpr std::int64_t column_count = 2;
+    static constexpr std::int64_t element_count = row_count * column_count;
+
+    auto get_descriptor() const {
+        return pca::descriptor<float, Method, pca::task::dim_reduction>{};
+    }
+
+    table get_train_data(std::int64_t override_row_count = row_count,
+                         std::int64_t override_column_count = column_count) const {
+        ONEDAL_ASSERT(override_row_count * override_column_count <= element_count);
```
michael-smirnov, line 43: ONEDAL_ASSERT_MUL_OVERFLOW ?

cpp/oneapi/dal/test/engine/dataframe_common.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+enum class table_float_type { f32, f64 };
+
+class table_id {
+public:
+    template <typename Float>
+    static table_id homogen() {
+        static_assert(dal::detail::is_floating_point<Float>());
+        if constexpr (std::is_same_v<Float, float>) {
+            return table_id{ table_kind::homogen, table_float_type::f32 };
```
michael-smirnov, line 44: Why you can't use `homogen_table::kind()` instead of creating your own enum `table_kind`?

https://github.com/rlnx/daal/blob/test-engine/cpp/oneapi/dal/table/homogen.hpp#L57
https://github.com/rlnx/daal/blob/test-engine/cpp/oneapi/dal/table/common.hpp#L144

cpp/oneapi/dal/algo/pca/test/badarg.cpp
rlnx: This is not ASSERT for overflow

rlnx: I suggest to not check overflows in tests, to me it sounds like overkill. The values in tests are controlled on our side, we can adjust values that do not lead to overflow.

cpp/oneapi/dal/algo/pca/detail/train_ops.hpp
rlnx: There is a test case for `component_count == 0`. 

Condition seems correct for `component_count == 0`, because `data.column_count < 0` is always `false`:
`table` invariant `data.column_count >= 0` => `0 < 0 == false`

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -0,0 +1,369 @@
...
+            return;
+        }
+
+        // If new entry is too large, remove all entries
+        if (new_entry_size >= max_cache_size_) {
+            return clear();
+        }
+
+        // If new_entry_size leads to overflow, remove all entries
+        if (cache_size_ + new_entry_size < cache_size_) {
```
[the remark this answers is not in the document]
rlnx: `cache_size` and `new_entry_size` are `size_t`.

```
Preconditions:
cache_size >= 0
new_entry_size > 0
```

Condition `cache_size + new_entry_size > cache_size` means that `new_entry_size > 0` which is given in preconditions, so this condition corresponds to normal algorithm flow.

Condition `cache_size + new_entry_size < cache_size` means that `new_entry_size < 0` which contradicts with preconditions, so it means that there is overflow in `cache_size + new_entry_size`.

cpp/oneapi/dal/test/engine/dataframe_common.hpp
rlnx: I expect `table::kind()` and `te::table_kind` would diverge. For example, we could create `homogen_row_major` or `homogen_column_major` or custom table type that defined in `test::engine` (e.g. mock).

michael-smirnov: Arguable, but Ok, let's keep this version for now. This is probably the indicator that `table::kind` is useless

michael-smirnov (approved):

cpp/oneapi/dal/test/engine/dataframe_common.hpp
rlnx: `table::kind` helps you to determine type of table at runtime, this is the main use case we discussed.

PR #1413, opened 2021-02-04
Add graph service functions test

Plastova:


cpp/oneapi/dal/graph/service_functions_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+		for (auto i = neigh.first; i != neigh.second; ++i) {
+            if (*i == correct_neigh[neigh_index]) correct_neigh_num++;	
+        	neigh_index++;
+        }
+    }
+    return correct_neigh_num;
+}
+
+TEST_CASE("throws if input file is empty") {
+	REQUIRE_THROWS_WITH(dal::preview::load_graph::load(desc, get_data_source("graph0.csv")),"Empty edge list");
```
rlnx, line 39: I suggest to not use `load_graph` in such simple test cases. I consider this test as **unit**, so the only one isolated component should be tests.

Right approach: Do not use `load_graph`, create graph from the data that is already in memory.

cpp/oneapi/dal/graph/BUILD
```diff
@@ -13,7 +13,20 @@ dal_module(
...
 )
 
+_DATA_DEPS = [
+    "@onedal//cpp/oneapi/dal/graph:test_data",
+]
+
 dal_test_suite(
-    name = "tests",
-    tests = [],
+    name = "graph_tests",
```
rlnx, line 21: The name should be "tests", the reason is that tests in `graph` submodule are discovered by the parent module using name "tests".

cpp/oneapi/dal/graph/BUILD
```diff
@@ -13,7 +13,20 @@ dal_module(
...
+    framework = "catch2",
+    srcs = glob([
+        "service_functions_test.cpp",
+    ]),
+    dal_deps = [
+        ":graph",
+        "@onedal//cpp/oneapi/dal/io",
```
rlnx, line 28: Ideally, we should not depend on IO here. `load_graph` functionality should be tested separately.

rlnx:
@Plastova, please use structure of tests in `dal/algo/pca` as a reference: introduce `graph/test` directory and put all graph-related test there.

cpp/oneapi/dal/graph/service_functions_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+}
+
+TEST_CASE("throws if input file does not exist") {
+	REQUIRE_THROWS_WITH(dal::preview::load_graph::load(desc, get_data_source("nonexistgraph.csv")),"File not found");
+}
+////////////////////////////////
+
+TEST_CASE("get_vertex_count for the graph with 2 vertices") {
+	const auto my_graph = dal::preview::load_graph::load(desc, get_data_source("graph1.csv"));
+	REQUIRE(dal::preview::get_vertex_count(my_graph) == 2);
```
rlnx, line 49: This test case does not test `service_functions`, but tests `load_graph`. It checks whether function `load` correctly interprets file like that:
```csv
0 1 
```

Right approach: 
- Create graph from in-memory data with specific vertex count and check if `get_vertext_count` returns expected value.
- Create separate test in `io/load_graph` module to cover `load` function

cpp/oneapi/dal/graph/service_functions_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+namespace oneapi::dal::graph::test {
+
+namespace dal = oneapi::dal;
+
+using my_graph_type = dal::preview::undirected_adjacency_vector_graph<>;
+const dal::preview::load_graph::descriptor<dal::preview::edge_list<int32_t>, my_graph_type>
+    desc;
+
+auto get_data_source (const std::string& name) {
+	const std::string name_path = "cpp/oneapi/dal/graph/test_data/" + name;
```
KulikovNikita, line 17: Here we assume that tests are called from oneDAL root directory. Is it possible that tests can be called from subdirectories?

Also it refers to @rlnx comment about `io` usage in tests

cpp/oneapi/dal/graph/service_functions_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+	const dal::preview::graph_csv_data_source ds(name_path);
+	return ds;
+}
+
+auto check_neighbors (const std::string& name, 
+	const dal::preview::vertex_size_type<my_graph_type> *correct_neigh,
```
KulikovNikita, line 23: I suppose that shorthand `neigh` is not very common or well-established. I encourage you to use full word `neighbor` instead

cpp/oneapi/dal/graph/service_functions_test.cpp
KulikovNikita: What about compile-time file loading with `#include` directive? Is it a valid option?

cpp/oneapi/dal/graph/service_functions_test.cpp
```diff
@@ -0,0 +1,166 @@
...
+	REQUIRE(check_neighbors ("graph2.csv", correct_neigh, vertices_num) == 2);
+}	
+
+TEST_CASE("get_vertex_neighbors for the complete graph") {
+	int correct_neigh_num = 0;
+	dal::preview::vertex_size_type<my_graph_type> neigh_vertex = 0;
+	const auto my_graph = dal::preview::load_graph::load(desc, get_data_source("graph3.csv"));
+	for (dal::preview::vertex_size_type<my_graph_type> j = 0; j < 7; j++) {
+		const auto neigh = dal::preview::get_vertex_neighbors(my_graph, j);
+		for (auto i = neigh.first; i != neigh.second; ++i) {
```
KulikovNikita, line 150:
```suggestion
		const auto [start, end] = dal::preview::get_vertex_neighbors(my_graph, j);
		for (auto i = start; i != end; ++i) {
```

cpp/oneapi/dal/graph/service_functions_test.cpp
rlnx: @KulikovNikita, the main problem with that test is that it does not test `graph/service_functions`, but tests `load_graph`. This test checks if empty CSV is correctly handled by data source as an empty graph. That kind of tests must be placed in `io/load_graph` submodule.

Possible approaches for data source testing:
1. Lightweight (smoke) tests on very small datasets
   - Have small datasets embedded right into the source code using C-style static array, `std::array`, etc. I don't very like `#include` idea, because the small datasets can be embedded without lose of readability.
   - Use [fixtures](https://github.com/catchorg/Catch2/blob/devel/docs/test-fixtures.md) to write the data to some file on
      fixture initialization.
   - Read that file using `load_graph` and test properties of the object loaded
2. Heavy tests
    - Don't run in the public CI
    - Have special large files stored in special location, e.g., `DATASETS_ROOT`, don't put them to repository

Plastova: Thank you for your comments. I'll edit the tests according to them.
I have to use `load_graph` in the service functions tests because `load_graph::load` is currently the only way to create a graph.

cpp/oneapi/dal/graph/BUILD
orrrrtem: The only available opportunity to create graph is load_graph. Hence, we left this dependency at least for a now, until new constructors will be arrived.

cpp/oneapi/dal/graph/test/service_functions_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+    std::string filename;
+    std::string file_content;
+    std::int64_t vertex_count;
+    std::int64_t edge_count;
+  };
+
+  class two_vertices_graph_type : public graph_base_data {
+  public:
+    two_vertices_graph_type() {
+      filename = "test_two_vertices_graph.csv";
```
orrrrtem, line 43: @rlnx Is it ok that we write file in current folder? No absolute path as it anyway will be removed. File names for each type of graph are different

cpp/oneapi/dal/graph/test/service_functions_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+  class acyclic_graph_type : public graph_base_data {
+  public:
+    acyclic_graph_type() {
+      filename = "test_acyclic_graph.csv";
+      file_content = "0 1 0 2 0 4 2 3 2 6 3 5 3 7";
+      vertex_count = 8;
+      edge_count = 7;
+    }
+
+    static const std::int64_t correct_vertex_count = 8;
```
orrrrtem, line 64: Why do you need both fields vertex_count and correct_vertex_count?

cpp/oneapi/dal/graph/test/service_functions_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+    std::array<std::int64_t, correct_vertex_count> degrees = {4, 4, 4, 4, 4};
+    std::array<std::int64_t, neighbors_count> neighbors = {
+        1, 2, 3, 4, 0, 2, 3, 4, 0, 1, 3, 4, 0, 1, 2, 4, 0, 1, 2, 3};
+  };
+
+  class service_functions_test {
+  public:
+    using my_graph_type = dal::preview::undirected_adjacency_vector_graph<>;
+
+    template <typename GraphType> auto create_graph() {
```
orrrrtem, line 92: Did you use clang format?

cpp/oneapi/dal/graph/test/service_functions_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+      const dal::preview::load_graph::descriptor<
+          dal::preview::edge_list<int32_t>, my_graph_type> desc;
+      auto my_graph = dal::preview::load_graph::load(desc, ds);
+      graph_data.delete_test_data();
+      return my_graph;
+    }
+
+    template <typename GraphType>
+    void check_vertex_degree(const GraphType &graph,
+                             std::int64_t *correct_degrees,
```
orrrrtem, line 105: If you use container, you can pass  it by const reference instead of casting to raw pointer.

cpp/oneapi/dal/graph/test/service_functions_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+          if (*i == correct_neighbors[neighbor_index])
+            correct_neighbors_num++;
+          neighbor_index++;
+        }
+      }
+      REQUIRE(correct_neighbors_num == neighbors_count);
+    }
+
+    void general_checks() {
+      two_vertices_graph_type tvg;
```
orrrrtem, line 134: While writing a code, please check oneapi coding guideline. Concretely here, not sure about using such short abbreviations.

cpp/oneapi/dal/graph/test/service_functions_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+      acyclic_graph_type ag;
+      complete_graph_type cg;
+
```
orrrrtem, line 137: Add corner graph cases like empty graph. It can be created without any service like create_graph, simple "graph_type empty_graph" and pass it to checks. Or you can adjust create graph to create empty graph with template specialization. Another use cases: edge list "0 0 1 1 2 2", edge list "0 <max_index_value - 1>". Not sure that the last should be included,

cpp/oneapi/dal/graph/test/service_functions_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+      const auto two_vertices_graph = create_graph<two_vertices_graph_type>();
+      const auto acyclic_graph = create_graph<acyclic_graph_type>();
+      const auto complete_graph = create_graph<complete_graph_type>();
+
+      SECTION("check get_vertex_count") {
```
orrrrtem, line 142: Minor, can be modified later. I see here a pattern for creating loop cycle. What if you will extend number of graphs to check up to 15. 15 times the same code with different graph? You can use some container to store graphs and iterate over it using only 2 lines of code instead of 2 * num_graphs.

cpp/oneapi/dal/graph/test/service_functions_test.cpp
rlnx: It's ok if we write to current folder, but it should be separate tests for IO, but not tests for service functions.

Plastova:
http://intel-ci.intel.com/eb7a067b-f925-f172-89be-8cdcd4b723a1

Plastova:
http://intel-ci.intel.com/eb7d7424-990b-f114-a331-8cdcd4b723a1

PR #1416, opened 2021-02-07
CSV generator

mchernov-intel:


cpp/oneapi/dal/algo/pca/test/batch.cpp
```diff
@@ -47,6 +47,20 @@ class pca_batch_test : public te::algo_fixture {
...
+        //     const auto ptr = dal::row_accessor<const float>{y}.pull().get_data();
+
+        //     std::cout << "batch" << std::endl;
+        //     for (size_t i = 0; i < y.get_row_count(); ++i) {
+        //         for (size_t j = 0; j < y.get_column_count(); ++j) {
+        //             std::cout << ptr[i * y.get_column_count() + j] << ' ';
+        //         }
+        //         std::cout << std::endl;
+        //     }
+        // }
```
KulikovNikita, line 62:
```suggestion
```
Please don't comment code. This code will be stored in GIT anyway so you'll be able to recover it in any time.

cpp/oneapi/dal/algo/pca/test/batch.cpp
```diff
@@ -188,9 +202,12 @@ class pca_batch_test : public te::algo_fixture {
...
 TEMPLATE_LIST_TEST_M(pca_batch_test, "pca common flow", "[pca][integration][batch]", pca_types) {
-    const te::dataframe data =
-        GENERATE_DATAFRAME(te::dataframe_builder{ 100, 10 }.fill_uniform(0.2, 0.5),
-                           te::dataframe_builder{ 100000, 10 }.fill_uniform(-0.2, 1.5));
+    // const te::dataframe data =
+    // GENERATE_DATAFRAME(te::dataframe_builder{ 100, 10 }.fill_uniform(0.2, 0.5),
+    //                    te::dataframe_builder{ 100000, 10 }.fill_uniform(-0.2, 1.5));
+
+    const te::dataframe data = GENERATE_DATAFRAME(te::dataframe_builder{ 3, 3 }.read_csv(
+        "/export/users/mchernov/work/DAAL/daal/cpp/oneapi/dal/algo/pca/test/test.csv"));
```
KulikovNikita, line 210: It is a very bad practice to keep absolute path hardcoded into tests.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -261,6 +263,45 @@ class dataframe_builder_action_fill_uniform : public dataframe_builder_action {
...
+            throw invalid_argument{ "Data_frame row_count should be equal to dataset row_count" };
+        }
+        if (df->get_column_count() != data_table.get_column_count()) {
+            throw invalid_argument{
+                "Data_frame column_count should be equal to dataset column_count"
+            };
+        }
+
+        const float* data_table_ptr =
+            dal::row_accessor<const float>{ data_table }.pull().get_mutable_data();
```
KulikovNikita, line 291: What is the reason to call `.get_mutable_data()` if it will be casted into `const float*`?

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -261,6 +263,45 @@ class dataframe_builder_action_fill_uniform : public dataframe_builder_action {
...
+        float* data_df_ptr = df->get_array().need_mutable_data().get_mutable_data();
+
+        std::memcpy(data_df_ptr,
+                    data_table_ptr,
+                    data_table.get_column_count() * data_table.get_row_count() * sizeof(float));
```
KulikovNikita, line 296: Shouldn't we do it using our own `memcpy` functionality?

KulikovNikita (changes requested):

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -20,6 +20,8 @@
 #include <random>
 #include <unordered_map>
 
+#include <iostream>
```
mchernov-intel, line 23: delete

cpp/oneapi/dal/algo/pca/test/batch.cpp
```diff
@@ -27,6 +27,9 @@ namespace oneapi::dal::pca::test {
 namespace te = dal::test::engine;
 namespace la = te::linalg;
 
+const std::string datasets = std::getenv("DATASETS");
```
rlnx, line 30:
```suggestion
const std::string datasets = std::getenv("DATASETSROOT");
```

rlnx: I suggest to check if variable exists. It it's not, `getenv` return `nullptr`. I'd report this to user via exception and require to set this var.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
mchernov-intel: no

cpp/oneapi/dal/algo/pca/test/batch.cpp
```diff
@@ -194,10 +194,13 @@ using pca_types = COMBINE_TYPES((float, double), (pca::method::cov, pca::method:
 
 TEMPLATE_LIST_TEST_M(pca_batch_test, "pca common flow", "[pca][integration][batch]", pca_types) {
     SKIP_IF(this->not_available_on_device());
+    
+    const std::string higgs = "/higgs/dataset/higgs_100t_train.csv";
```
rlnx, line 198:
```suggestion
    const std::string higgs = "higgs/dataset/higgs_100t_train.csv";
```

Relative dataset path must be specified without "/" in the beginning.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -273,6 +297,19 @@ dataframe_builder_impl::dataframe_builder_impl(std::int64_t row_count, std::int6
...
+    if (dataset_root_ptr == nullptr) {
+        throw invalid_argument{ "DATASETSROOT environment variable is unset" };
+    }
+    const std::string datasets_root(dataset_root_ptr);
+
+    const auto data_table = dal::read<dal::table>(dal::csv::data_source{ datasets_root + dataset });
+    program_.add<dataframe_builder_action_set>(dal::row_accessor<const float>{ data_table }.pull(),
+                                               data_table.get_row_count(),
+                                               data_table.get_column_count());
+}
```
rlnx, line 311: Do not recommend splitting logic and responsibilities between builder and action.

1. According to current design and implementation of the methods, builder has straightforward responsibilities: create action and add to the program. So move dataset reading logic to separate action.
2. This action should include dataset name as part of opcode, see `get_opcode()` method to support proper caching

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -219,6 +219,30 @@ class dataframe_builder_action_allocate : public dataframe_builder_action {
     std::int64_t column_count_;
 };
 
+class dataframe_builder_action_set : public dataframe_builder_action {
```
rlnx, line 222:
```suggestion
class dataframe_builder_action_read_external_dataset : public dataframe_builder_action {
```

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -219,6 +219,30 @@ class dataframe_builder_action_allocate : public dataframe_builder_action {
...
+public:
+    explicit dataframe_builder_action_set(const array<float>& data,
+                                          std::int64_t row_count,
+                                          std::int64_t column_count)
```
rlnx, line 226: Accept relative path to dataset, otherwise you cannot properly generate opcode

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -219,6 +219,30 @@ class dataframe_builder_action_allocate : public dataframe_builder_action {
...
+            : array_(data),
+              row_count_(row_count),
+              column_count_(column_count) {}
+
+    std::string get_opcode() const override {
+        return fmt::format("set({},{})", row_count_, column_count_);
```
rlnx, line 232:
```suggestion
        return fmt::format("read_external_dataset({})", dataset_relative_path);
```

Opcode is used as a cache key. Caching is important to speedup dataset generation, if entry with the same key is found, the already loaded dataset is used. You don't need to care about caching, it's done as part of dataframe infrastructure. The only thing you have to do is to create unique `opcode`.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -344,25 +381,75 @@ static homogen_table build_homogen_table(Policy& policy,
 }
 
 template <typename Policy>
-static table build_table(Policy& policy, const dataframe& df, const table_id& id) {
+static homogen_table build_homogen_table(Policy& policy,
+                                         const array<float>& data,
+                                         const table_id& id,
+                                         std::int64_t row_count,
+                                         std::int64_t first_column,
+                                         std::int64_t last_column) {
```
rlnx, line 389:
```suggestion
                                         std::int64_t last_column_idx) {
```

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -344,25 +381,75 @@ static homogen_table build_homogen_table(Policy& policy,
 }
 
 template <typename Policy>
-static table build_table(Policy& policy, const dataframe& df, const table_id& id) {
+static homogen_table build_homogen_table(Policy& policy,
+                                         const array<float>& data,
+                                         const table_id& id,
+                                         std::int64_t row_count,
+                                         std::int64_t first_column,
```
rlnx, line 388:
```suggestion
                                         std::int64_t first_column_idx,
```

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -344,25 +381,75 @@ static homogen_table build_homogen_table(Policy& policy,
...
+                                         std::int64_t last_column) {
+    const std::int64_t data_column_count = data.get_count() / row_count;
```
rlnx, line 390: Recommend to pass `column_count` as separate parameter

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -344,25 +381,75 @@ static homogen_table build_homogen_table(Policy& policy,
...
+    const std::int64_t column_count = last_column - first_column;
```
rlnx, line 391: Missing assertions to test precondition for  `last_column` and `first_column`

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -344,25 +381,75 @@ static homogen_table build_homogen_table(Policy& policy,
...
+
+    if (first_column == 0 && data_column_count == column_count) {
```
rlnx, line 393: It's hard to distinguish `column_count` and `data_column_count`, try to figure out more descriptive names

rlnx: For example, `required_column_count`/`requested_column_count` and `actual_column_count`

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -344,25 +381,75 @@ static homogen_table build_homogen_table(Policy& policy,
...
+        return wrap_to_homogen_table(policy, data, id, row_count, column_count);
+    }
+    else {
+        auto arr = array<float>::empty(row_count * column_count);
+        float* arr_ptr = arr.get_mutable_data();
+        const auto data_ptr = data.get_data();
```
rlnx, line 399: It's hard to distinguish `arr_ptr` and `data_ptr`, try `src_ptr` and `dst_ptr`, it gives a clue what is copied

rlnx, line 399:
```suggestion
        float* arr_ptr = arr.get_mutable_data();
        const float* data_ptr = data.get_data();
```

Sometimes it's hard to understand `auto` type from the context, it's helpful for reader to understand if the subsequent loop converts types.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -344,25 +381,75 @@ static homogen_table build_homogen_table(Policy& policy,
...
+
+template <typename Policy>
+static table build_table(Policy& policy, const dataframe& df, const table_id& id, range r) {
+    if (r.start_idx == 0 && r.end_idx == 0) {
+        r.end_idx = df.get_column_count();
+    }
+
+    if (r.end_idx < 0) {
+        r.end_idx += df.get_column_count();
+    }
```
rlnx, line 419: Seems like the can be replaced by
```c++
if (r.end_idx < 0) {
    r.end_idx += df.get_column_count();
}
```

cpp/oneapi/dal/test/engine/dataframe_common.cpp
mchernov-intel: Not required because checks are below

rlnx: Anyway, assertions are recommended it also gives a reader clues what is the expected variable range. In this case, assertion could look like `ASSERT(column_count >= 0)`.

rlnx:
I've already introduced separation on tests that uses external datasets and others. See
https://github.com/oneapi-src/oneDAL/blob/master/dev/bazel/HOWTO.md#tests-that-use-external-datasets

Use approach described there to make Bazel work in public CI.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -219,6 +219,36 @@ class dataframe_builder_action_allocate : public dataframe_builder_action {
     std::int64_t column_count_;
 };
 
+class dataframe_builder_action_read_external_dataset : public dataframe_builder_action {
+public:
+    explicit dataframe_builder_action_read_external_dataset(std::string dataset)
```
rlnx, line 224:
```suggestion
    explicit dataframe_builder_action_read_external_dataset(const std::string& dataset)
```

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -219,6 +219,36 @@ class dataframe_builder_action_allocate : public dataframe_builder_action {
...
+            : dataset_(dataset) {}
+
+    std::string get_opcode() const override {
+        return fmt::format("read_external_dataset({})", dataset_);
+    }
+
+    dataframe_impl* execute(dataframe_impl* df) const override {
+        delete df;
+
+        const char* dataset_root_ptr = std::getenv("DATASETSROOT");
```
rlnx, line 234: Not sure, but we have `DATASETS_ROOT`, but not `DATASETSROOT`, please double check.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -219,6 +219,36 @@ class dataframe_builder_action_allocate : public dataframe_builder_action {
...
+        if (dataset_root_ptr == nullptr) {
+            throw invalid_argument{ "DATASETSROOT environment variable is unset" };
+        }
+        const std::string datasets_root(dataset_root_ptr);
+
+        const auto data_table =
+            dal::read<dal::table>(dal::csv::data_source{ datasets_root + '/' + dataset_ });
```
rlnx, line 241: I suggest ti check `dataset_` extension and if it's not `csv` throw unimplemented exception. There is an option to make it in the ctor (preferred).

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -219,6 +219,36 @@ class dataframe_builder_action_allocate : public dataframe_builder_action {
     std::int64_t column_count_;
 };
 
+class dataframe_builder_action_read_external_dataset : public dataframe_builder_action {
+public:
+    explicit dataframe_builder_action_read_external_dataset(std::string dataset)
+            : dataset_(dataset) {}
```
rlnx, line 225: I suggest to introduce function `normalize_path` that removes trailing `/` in the beginning or end of file path.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
rlnx: Use `normalize_path` for `datasets_root`.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -219,6 +219,36 @@ class dataframe_builder_action_allocate : public dataframe_builder_action {
...
+    }
+
+    dataframe_impl* execute(dataframe_impl* df) const override {
+        delete df;
+
+        const char* dataset_root_ptr = std::getenv("DATASETSROOT");
+        if (dataset_root_ptr == nullptr) {
+            throw invalid_argument{ "DATASETSROOT environment variable is unset" };
+        }
+        const std::string datasets_root(dataset_root_ptr);
```
rlnx, line 238: I'd recommend to move that segment to a separate static function that "caches" result of `getenv` using Meyer's singleton approach.

cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -273,6 +303,10 @@ dataframe_builder_impl::dataframe_builder_impl(std::int64_t row_count, std::int6
     program_.add<dataframe_builder_action_allocate>(row_count, column_count);
 }
 
+dataframe_builder_impl::dataframe_builder_impl(std::string dataset) {
```
rlnx, line 306:
```suggestion
dataframe_builder_impl::dataframe_builder_impl(const std::string& dataset) {
```

cpp/oneapi/dal/test/engine/dataframe_common.cpp
rlnx: This is not addressed.

mchernov-intel: sorry

cpp/oneapi/dal/test/engine/dataframe_common.cpp
mchernov-intel: 'DAAL_DATASETS'

rlnx (approved):

KulikovNikita (approved):

PR #1421, opened 2021-02-10
[Load graph] Replace of daal atomics with std atomics

orrrrtem:
Std atomics are used only in hpp part of the lib.


averbukh (approved):
What about performance after switching to std::atomic?

PR #1427, opened 2021-02-11
Add commit hash to Azure Pipeline cache key

rlnx:
I found out that current Azure cache is not very efficient. Caches are immutable by design, so once cache is created its content cannot be changed. In fact, current Bazel jobs use the very first cache that was created a week ago.

There is a hack to force a cache update after each successful job using `restoreKeys`: https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#restore-keys


cpp/oneapi/dal/array.hpp
```diff
@@ -198,7 +198,7 @@ class array {
     /// in :literal:`other` to the new array instance
     array(array<T>&& other) : impl_(std::move(other.impl_)) {
         update_data(impl_.get());
-        other.reset_data();
+        // other.reset_data();
```
napetrov, line 201: Is this intended change for this PR?

rlnx: No, it's to test cache

napetrov (approved):

rlnx:
I realized that cache size grows because it accumulates artifacts from all previous jobs. Let's not merge it until we find solution, otherwise cache load/store could take more time than build itself.

PR #1429, opened 2021-02-13
Fix warnings that cause error on GCC 7.4

rlnx:
Tried to build bazel tests on GCC 7.4 and got some errors caused by warnings in test infrastructure.

### DIFF
--- a/cpp/oneapi/dal/test/engine/dataframe_common.cpp
+++ b/cpp/oneapi/dal/test/engine/dataframe_common.cpp
@@ -280,15 +280,15 @@ dataframe_builder& dataframe_builder::fill_uniform(double a, double b, std::int6
 
 dataframe dataframe_builder::build() const {
     const auto& program = impl_->get_program();
-    const auto [df, hit] = get_dataframe_builder_cache().lookup(program);
+    const auto df_hit = get_dataframe_builder_cache().lookup(program);
 #ifdef ONEDAL_DEBUG_DATAFRAMES_CACHE
-    const std::string hit_or_miss = hit ? "hit" : "miss";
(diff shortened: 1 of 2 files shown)


cpp/oneapi/dal/test/engine/dataframe_common.cpp
```diff
@@ -280,15 +280,15 @@ dataframe_builder& dataframe_builder::fill_uniform(double a, double b, std::int6
 
 dataframe dataframe_builder::build() const {
     const auto& program = impl_->get_program();
-    const auto [df, hit] = get_dataframe_builder_cache().lookup(program);
+    const auto df_hit = get_dataframe_builder_cache().lookup(program);
```
rlnx, line 283: GCC 7.4 complains that `hit` variable is unused, but GCC 9.x is ok

dev/bazel/flags.bzl
```diff
@@ -30,6 +30,7 @@ lnx_cc_pedantic_flags = [
     "-Wall",
     "-Wextra",
     "-Wno-unused-parameter",
+    "-Wno-unused-but-set-parameter",
```
rlnx, line 33: Sometimes leads to false-positive in `if constexpr` functions.

michael-smirnov (approved):
How long and which versions of GCC do we want to continue supporting?

rlnx:
> How long and which versions of GCC do we want to continue supporting?

Minimal supported GCC is 7.4, we test if oneAPI interfaces support it using make-based build. The warnings here are from Bazel-based tests, historically we run them publicly on GCC 9.1. GCC 7.4 will be used in private CI soon.

PR #1445, opened 2021-02-22
enabled hist method for cpu; added support of unified descriptor for infer and train

makart19:


rlnx:
I realized that decision forest has wrong order of template parameters (method and task) in dispatcher and kernels. For the kernels it can be easily fixed, but for dispatcher it will cause BC problems if we do not introduce v2 namespace. 

Recommendation: Fix order in kernels and leave dispatcher until new major release happened.

rlnx:
I found out that queue is not passed to one of the read functions here in dpc regression examples.

Moreover, seems like examples should be called df_hist_batch, rather _dense_

rlnx (approved):

PR #1465, opened 2021-03-01
CMake Common Function Fix

owerbat:


owerbat:
/intelci: run
set infra_branch=dev/ntimakin-modules_fix

PetrovKP (approved):

PR #1469, opened 2021-03-02
Sum reduction speedup with application to K-Means

amgrigoriev:


cpp/daal/src/sycl/sum_reducer.cpp
```diff
@@ -207,12 +211,9 @@ SumReducer::Result SumReducer::sum(Layout vectorsLayout, const UniversalBuffer &
...
-    Result result(context, nVectors, vectors.type(), status);
-    DAAL_CHECK_STATUS_RETURN_IF_FAIL(status, result);
-
     if (vectorsLayout == Layout::RowMajor)
     {
-        status |= sum_singlepass(context, kernelFactory, vectorsLayout, vectors, nVectors, vectorSize, maxWorkItemsPerGroup, result);
+        status |= sum_singlepass(context, kernelFactory, vectors, nVectors, vectorSize, result);
     }
     else
     {
```
makart19, line 219: After this line there is a switch for defining numDivisionsByRow.
There is similar switch in LOM and i found out that these numbers showed best performance

    nRowsBlocks = 128;
    if (nVectors < 5000)
        nRowsBlocks = 1;
    else if (nVectors < 10000)
        nRowsBlocks = 8;
    else if (nVectors < 20000)
        nRowsBlocks = 16;
    else if (nVectors < 50000)
        nRowsBlocks = 32;
    else if (nVectors < 100000)
        nRowsBlocks = 64;

you can try it to see if it provides any benefit here

makart19 (approved):

cpp/daal/src/sycl/sum_reducer.cpp
```diff
@@ -47,13 +47,12 @@ services::Status buildProgram(ClKernelFactoryIface & kernelFactory, const TypeId
...
-services::Status sum_singlepass(ExecutionContextIface & context, ClKernelFactoryIface & kernelFactory, Layout vectorsLayout,
-                                const UniversalBuffer & vectors, uint32_t nVectors, uint32_t vectorSize, uint32_t workItemsPerGroup,
-                                SumReducer::Result & result)
+services::Status sum_singlepass(ExecutionContextIface & context, ClKernelFactoryIface & kernelFactory, const UniversalBuffer & vectors,
+                                uint32_t nVectors, uint32_t vectorSize, SumReducer::Result & result)
 {
     services::Status status;
 
-    auto sum_kernel = kernelFactory.getKernel("sum_singlepass", status);
+    auto sum_kernel = kernelFactory.getKernel("sum_singlesubgroup", status);
```
michael-smirnov, line 55: Does this change affect the performance of other algorithms, except KMeans?

michael-smirnov: Discussed offline - @amgrigoriev says that every row-major case where sum-reducer is used in the library shall speed up.

michael-smirnov (approved):

cpp/daal/src/sycl/sum_reducer.cpp
amgrigoriev: Will add in a separate MR

PR #1488, opened 2021-03-09
Transfer oneAPI part to device USM

rlnx:
**What is done:**
- Create adapter `table` -> `NumericTable` in case of device memory
- Replace almost all allocations in oneAPI part by `usm::alloc::device`
- Store DPC++ queue within `array` and, thus, in tables and table builders
- Support device-device, device-host and host-device data access in `row_accessor` and `column_accessor`
- Support data conversion on the device w/o fallback to host in case of device-device data access
- Partially refactor tests in data management
- Disable double precision tests on the devices that do not support `float64` natively
- Solved all functional problems if kernels use device USM

**DAAL/oneDAL Perf. comparison**
image

**Master/branch Perf. comparison**
image


cpp/oneapi/dal/backend/common.hpp
```diff
@@ -20,7 +20,7 @@
 #include <CL/sycl.hpp>
 #endif
 
-#include "oneapi/dal/backend/dispatcher.hpp"
+#include "oneapi/dal/detail/common.hpp"
```
michael-smirnov, line 23: Does not that lead to a recursive dependency problem in the future between backend and details?

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -59,6 +59,36 @@ inline constexpr Integer up_multiple(Integer x, Integer multiple) {
     return y + z;
 }
 
+/// Finds the largest power of 2 number not larger than `x`.
+/// Return `x`, if `x` is already power of 2
+/// Example: down_pow2(10) == 8
+/// Example: down_pow2(16) == 16
+template <typename Integer>
+inline constexpr Integer down_pow2(Integer x) {
```
michael-smirnov, line 67: Do these functions really common for whole backend, or it just some math primitives?

cpp/oneapi/dal/backend/interop/table_conversion.hpp
```diff
@@ -75,51 +95,63 @@ inline daal::data_management::NumericTablePtr wrap_by_host_homogen_adapter(
...
-inline auto convert_to_daal_sycl_homogen_table(sycl::queue& queue,
-                                               array<T>& data,
-                                               std::int64_t row_count,
-                                               std::int64_t column_count) {
-    data.need_mutable_data(queue);
-    const auto daal_data =
-        daal::services::SharedPtr<T>(data.get_mutable_data(), daal_object_owner{ data });
+inline daal::data_management::NumericTablePtr convert_to_daal_table(const sycl::queue& queue,
+                                                                    const table& table) {
+    return interop::sycl_table_adapter::create(queue, table);
```
michael-smirnov, line 120: Why the logic differs from host version of that function?

cpp/oneapi/dal/backend/interop/table_conversion.hpp
```diff
@@ -75,51 +95,63 @@ inline daal::data_management::NumericTablePtr wrap_by_host_homogen_adapter(
...
     using daal::data_management::internal::SyclHomogenNumericTable;
-    return SyclHomogenNumericTable<T>::create(daal_data,
-                                              dal::detail::integral_cast<std::size_t>(column_count),
-                                              dal::detail::integral_cast<std::size_t>(row_count),
-                                              queue);
-}
+    using dal::detail::integral_cast;
 
+    ONEDAL_ASSERT(data.get_count() == row_count * column_count);
+    ONEDAL_ASSERT(data.has_mutable_data());
```
michael-smirnov, line 135: There should be an assert that array contains USM data and `queue` is compatible with `data.get_queue()` I guess

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -16,16 +16,18 @@
...
-#include "oneapi/dal/detail/common.hpp"
+#include "oneapi/dal/backend/common.hpp"
 
 namespace oneapi::dal::backend {
 
 #ifdef ONEDAL_DATA_PARALLEL
 
+inline bool is_device_accessible_usm_pointer(const sycl::queue& queue, const void* pointer) {
+    const auto pointer_type = sycl::get_pointer_type(pointer, queue.get_context());
+    return (pointer_type == sycl::usm::alloc::device) || //
```
michael-smirnov, line 27:
```suggestion
    return (pointer_type == sycl::usm::alloc::device) || 
```

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -16,16 +16,18 @@
...
+           (pointer_type == sycl::usm::alloc::shared);
```
michael-smirnov, line 28: actually, every USM pointer except `sycl::usm::alloc::unknown` is device accessible

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -80,6 +82,161 @@ class usm_deleter {
     sycl::queue queue_;
 };
 
+template <typename T>
+using unique_usm_ptr = std::unique_ptr<T, usm_deleter<T>>;
```
michael-smirnov, line 86:
```suggestion
using unique_usm_ptr = detail::unique<T, usm_deleter<T>>;
```
?

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -80,6 +82,161 @@ class usm_deleter {
...
+
+inline unique_usm_ptr<void> make_unique_usm(sycl::queue& q,
```
michael-smirnov, line 88: Why `_usm` suffix is needed here?

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -80,6 +82,161 @@ class usm_deleter {
...
+                                            std::int64_t size,
+                                            sycl::usm::alloc alloc = sycl::usm::alloc::device) {
+    const std::size_t n = dal::detail::integral_cast<std::size_t>(size);
+    return unique_usm_ptr<void>{ sycl::malloc(n, q, alloc), usm_deleter<void>{ q } };
```
michael-smirnov, line 92: Why not to use our own ``backend::malloc` in all these helpers?

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -80,6 +82,161 @@ class usm_deleter {
...
+    return unique_usm_ptr<T>{ sycl::malloc_shared<T>(n, q), usm_deleter<T>{ q } };
+}
+
+template <typename T>
+inline unique_usm_ptr<T> make_unique_usm_host(sycl::queue& q, std::int64_t count) {
+    const std::size_t n = dal::detail::integral_cast<std::size_t>(count);
+    return unique_usm_ptr<T>{ sycl::malloc_host<T>(n, q), usm_deleter<T>{ q } };
+}
+
+inline sycl::event gather_device2host(sycl::queue& q,
```
michael-smirnov, line 136: gather/scatter shall not be placed here in my opinion

cpp/oneapi/dal/backend/transfer.hpp
```diff
@@ -0,0 +1,91 @@
+/*******************************************************************************
```
michael-smirnov, line 1: data_transfer.hpp instead of transfer.hpp?

cpp/oneapi/dal/backend/transfer.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+#include "oneapi/dal/array.hpp"
+#include "oneapi/dal/backend/memory.hpp"
+
+namespace oneapi::dal::backend {
+
+#ifdef ONEDAL_DATA_PARALLEL
+template <typename T>
+inline std::tuple<array<T>, sycl::event> to_device(sycl::queue& q, const array<T>& ary) {
+    if (!ary.get_count()) {
```
michael-smirnov, line 27: Not sure it is not a mistake if we transfer empty array

cpp/oneapi/dal/backend/transfer.hpp
```diff
@@ -0,0 +1,91 @@
...
+        return { ary, sycl::event{} };
+    }
+
+    if (ary.get_queue().has_value()) {
+        // TODO: Consider replacing by exception
+        ONEDAL_ASSERT(ary.get_queue().value() == q,
+                      "We do not support data tranfer between different queues yet");
```
michael-smirnov, line 34: Why?

cpp/oneapi/dal/backend/transfer.hpp
```diff
@@ -0,0 +1,91 @@
...
+        if (ary_alloc == sycl::usm::alloc::device) {
+            return { ary, sycl::event{} };
+        }
+    }
+
+    const std::int64_t element_count = ary.get_count();
+    const auto ary_device = array<T>::empty(q, element_count, sycl::usm::alloc::device);
+    const auto event = q.memcpy(ary_device.get_mutable_data(),
+                                ary.get_data(),
+                                sizeof(T) * dal::detail::integral_cast<std::size_t>(element_count));
```
michael-smirnov, line 46: you can call just `ary.get_size()`

cpp/oneapi/dal/backend/transfer.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+    const std::int64_t element_count = ary.get_count();
+
+    ONEDAL_ASSERT(ary.get_queue().has_value());
+    auto q = ary.get_queue().value();
+
+    // TODO: Change allocation kind to normal host memory once
+    //       bug in `memcpy` with the host memory is fixed
+    const auto ary_host = array<T>::empty(q, element_count, sycl::usm::alloc::host);
+    const auto event = q.memcpy(ary_host.get_mutable_data(),
```
michael-smirnov, line 69: That is needed only if `ary` contains device data

cpp/oneapi/dal/detail/array_utils.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+#pragma once
+
+#include "oneapi/dal/array.hpp"
+
+namespace oneapi::dal::detail {
+namespace v1 {
+
+template <typename T>
+class array_via_policy {
```
michael-smirnov, line 25: Can you please point where this is used?

michael-smirnov (commented):
first part reviewed

cpp/oneapi/dal/backend/transfer.hpp
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    if (ary.get_queue().has_value()) {
+        // TODO: Consider replacing by exception
+        ONEDAL_ASSERT(ary.get_queue().value() == q,
+                      "We do not support data tranfer between different queues yet");
+
+        const auto ary_alloc = sycl::get_pointer_type(ary.get_data(), q.get_context());
+        if (ary_alloc == sycl::usm::alloc::device) {
+            return { ary, sycl::event{} };
```
amgrigoriev, line 38: No memcpy in that case?

cpp/oneapi/dal/algo/decision_forest/backend/gpu/train_kernel_reg_hist_dpc.cpp
```diff
@@ -55,13 +55,8 @@ static result_t call_daal_kernel(const context_gpu& ctx,
...
     const int64_t column_count = data.get_column_count();
 
-    auto arr_data = row_accessor<const Float>{ data }.pull(queue);
-    auto arr_label = row_accessor<const Float>{ labels }.pull(queue);
-
-    const auto daal_data =
-        interop::convert_to_daal_sycl_homogen_table(queue, arr_data, row_count, column_count);
-    const auto daal_labels =
-        interop::convert_to_daal_sycl_homogen_table(queue, arr_label, row_count, 1);
+    const auto daal_data = interop::convert_to_daal_table(queue, data);
```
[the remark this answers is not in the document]
rlnx: We'll see that once I run benchmarks and analyze perf

cpp/oneapi/dal/algo/decision_forest/test/badarg.cpp
```diff
@@ -81,10 +81,10 @@ class df_badarg_test : public te::algo_fixture {
     }
 };
 
-using df_cls_types = _TE_COMBINE_TYPES_3((float, double),
+using df_cls_types = _TE_COMBINE_TYPES_3((float),
```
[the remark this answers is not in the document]
rlnx: I think we don't need to have badargs tests for different floating point types

cpp/oneapi/dal/algo/decision_forest/test/batch.cpp
```diff
@@ -57,6 +57,11 @@ class df_batch_test : public te::algo_fixture {
         return get_policy().is_gpu() && is_dense;
     }
 
+    bool not_float64_friendly() {
```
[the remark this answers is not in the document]
rlnx: The problem is this function depends on template parameters that become available only when we extract `float_t` from the `TestType` tuple. The structure of the tuple (number of arguments and their order) is specific to the final fixture.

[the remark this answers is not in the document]
rlnx: Maybe I can create template `algo_fixture` that depends on `float_t` only and then provide this method there

cpp/oneapi/dal/algo/pca/test/batch.cpp
```diff
@@ -38,6 +38,11 @@ class pca_batch_test : public te::algo_fixture {
         return get_policy().is_gpu() && is_svd;
     }
 
+    bool not_float64_friendly() {
```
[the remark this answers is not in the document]
rlnx: Yes, see https://github.com/oneapi-src/oneDAL/pull/1488/files#r592098761

cpp/oneapi/dal/backend/common.hpp
rlnx: `backend` cannot be included to `detail` by design, it will not even compile as backend is not distributed together with detail.

cpp/oneapi/dal/backend/common.hpp
rlnx: I think this is common enough, for example it's used to determine appropriate local range of DPC++ kernel.

cpp/oneapi/dal/backend/interop/table_conversion.hpp
rlnx: Because `sycl_table_adapter` is not DPC++ equivalent of  `host_homogen_table_adapter`. 

`sycl_table_adapter` can wrap arbitrary table and extract data using `row_accessor` or `column_accessor`, but `host_homogen_table_adapter` wraps only `homogen_table`. I prefer to use old SYCL entities as little as possible to fix possible functional or perf. problems on side of new interfaces.

cpp/oneapi/dal/backend/memory.hpp
rlnx: This is hint for clang-format to split expression into two lines for better readability. Without `//` clang-format will join the lines.

cpp/oneapi/dal/backend/memory.hpp
rlnx: I see, it makes sense to rename function. It's expected to return `true` if access to pointer is efficient on device. It's efficient in case of device USM and might be efficient in case of shared USM.

cpp/oneapi/dal/backend/memory.hpp
rlnx: Now I'm thinking about removing `detail::unique`, it seems useless...

cpp/oneapi/dal/backend/memory.hpp
rlnx: I think no

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -80,6 +82,161 @@ class usm_deleter {
...
+    ONEDAL_ASSERT(dst_device);
+    ONEDAL_ASSERT(src_host);
+    ONEDAL_ASSERT(block_count > 0);
+    ONEDAL_ASSERT(dst_stride_in_bytes > 0);
+    ONEDAL_ASSERT(block_size_in_bytes > 0);
+    ONEDAL_ASSERT(dst_stride_in_bytes >= block_size_in_bytes);
+    ONEDAL_ASSERT(is_device_accessible_usm_pointer(q, dst_device));
+    ONEDAL_ASSERT_MUL_OVERFLOW(std::int64_t, block_count, block_size_in_bytes);
+
+    const auto gathered_device_unique =
```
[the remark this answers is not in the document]
rlnx: No it's gathered, because we copy from the gathered host representation to scattered device, so `gathered_device_unique` is the copy of gathered host data on the device.

cpp/oneapi/dal/backend/transfer.hpp
rlnx: Don't think the word "data" adds more clarity here

cpp/oneapi/dal/backend/transfer.hpp
rlnx: Is there use case when it leads to the problem?

cpp/oneapi/dal/backend/transfer.hpp
rlnx: Need to be replaced by context comparison, but it will complicate the logic below. It will not be enough to check alloc kind, we'll need to check weather devices are the same. Will do that.

cpp/oneapi/dal/backend/transfer.hpp
rlnx: In addition to that we need to compare devices if we remove assertion above. If we are in the same queues additional check or memcpy is not needed.

cpp/oneapi/dal/backend/transfer.hpp
rlnx: We have made decision to stick to device data in oneDAL internals.

cpp/oneapi/dal/detail/array_utils.hpp
rlnx: `homogen_table`, it will be used in other data management classes where we want to have universal code for host and DPC++.

cpp/oneapi/dal/array.hpp
```diff
@@ -229,98 +262,120 @@ class array {
...
-    /// assigns the pointer to the ownership structure of :literal:`ref` to the new instance.
-    /// Array returns :literal:`Data` pointer as its mutable or immutable block depending
-    /// on the :literal:`Data` type.
+    /// assigns the pointer to the ownership structure of :literal:`ref` to the new instance. Array
+    /// returns :literal:`Data` pointer as its mutable or immutable block depending on the
+    /// :literal:`Data` type.
     ///
-    /// @tparam ref    The type of elements in the referenced array.
-    /// @tparam data    Either :literal:`T` or $const T$ type.
+    /// @tparam Y    The type of elements in the referenced array.
```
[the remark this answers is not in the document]
rlnx: The problem is that `Y` is already in the code, but not reflected in doxygen. The old doxygen comment that was fixed by this change is just incorrect and not displayed in docs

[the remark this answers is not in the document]
rlnx: > Every doxygen change I think shall trigger spec update since they must be synchronized.

This creates overhead for each PR, I'd propose to have periodical synchronization:
- We need to keep in mind `dalapi` tool, the changes in tool need to be propagated to spec first
- After that we'll copy paste all headers to spec removing implementation, ideally this should be automated

cpp/oneapi/dal/algo/decision_forest/test/badarg.cpp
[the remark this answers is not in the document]
michael-smirnov: Yes, in general, if these tests are not dependent on the fp type. Agree

cpp/oneapi/dal/backend/memory.hpp
rlnx: I realized that if I remove `_usm`, other methods look strange:
```
make_unique_usm -> make_unique
make_unique_usm_device -> make_unique_device // Misleading, sound like I'm creating unique device
make_unique_usm_shared -> make_unique_shared // std::make_unique or std::make_shared?
make_unique_usm_host -> make_unique_host // Also not clear
```

cpp/oneapi/dal/backend/transfer.hpp
rlnx: I assume `to_device` will copy any kind of data to device

cpp/oneapi/dal/algo/kmeans/backend/gpu/train_kernel_lloyd_dense_dpc.cpp
```diff
@@ -52,17 +52,15 @@ static NumericTablePtr get_initial_centroids(const dal::backend::context_gpu& ct
...
 
     daal::data_management::NumericTablePtr daal_initial_centroids;
 
     if (!input.get_initial_centroids().has_data()) {
-        auto arr_data = row_accessor<const Float>{ data }.pull(queue);
-        const auto daal_data =
-            interop::convert_to_daal_sycl_homogen_table(queue, arr_data, row_count, column_count);
+        // We use CPU algorithm for initialization, so input data
+        // may be copied to DAAL homogen table
+        const auto daal_data = interop::copy_to_daal_homogen_table<Float>(data);
```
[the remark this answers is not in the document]
rlnx: `copy_to_daal_homogen_table` uses `row_accessor::pull`, it copies data to host if necessary.

The other case is `convert_to_daal_table`, it uses adapter, so it may require copy to host. It affects the case when we run device algorithm and then pass its result to host algorithm. There is a bunch of problems here:
- Host variant of `convert_to_daal_table` does not support device -> host conversion if it goes through adapter branch.
- There is no integration testing for that case, we need to have at least two small test cases:
   - Pass device data to host algorithm
   - Pass pure host data to device algorithm   

I can fix current version of `convert_to_daal_table` as part of this PR, but we need to have more solid solution. This requires:
- Refactoring of table/accessor implementation to simplify. The solution requires adding new methods to table iface, so it may be problematic.
- Add `to_host`/`to_device` for table, as we have for array.

cpp/oneapi/dal/algo/kmeans/test/batch.cpp
```diff
@@ -20,22 +20,19 @@
...
-#include "oneapi/dal/table/homogen.hpp"
 
 namespace oneapi::dal::kmeans::test {
 
 namespace te = dal::test::engine;
 namespace la = te::linalg;
 
 template <typename TestType>
-class kmeans_batch_test : public te::algo_fixture {
+class kmeans_batch_test : public te::float_algo_fixture<std::tuple_element_t<0, TestType>> {
```
michael-smirnov, line 35: nip: At first sight, it is hard to understand what is the `0` element in the tuple `TestType`. I can only guess it is fptype.
Can we change `TestType` from `std::tuple` to some structure with meaningful type traits?

cpp/oneapi/dal/array.hpp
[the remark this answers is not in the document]
michael-smirnov: Ok

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -16,37 +16,117 @@
...
+
+inline bool is_shared_usm_pointer(const sycl::queue& queue, const void* pointer) {
+    const auto pointer_type = sycl::get_pointer_type(pointer, queue.get_context());
+    return pointer_type == sycl::usm::alloc::shared;
+}
+
+inline bool is_host_usm_pointer(const sycl::queue& queue, const void* pointer) {
+    const auto pointer_type = sycl::get_pointer_type(pointer, queue.get_context());
+    return pointer_type == sycl::usm::alloc::host;
+}
```
michael-smirnov, line 38: The reason to create such wrappers? Even you do not use it in `is_device_friendly_usm_pointer` by yourself.

cpp/oneapi/dal/backend/memory.hpp
michael-smirnov: What about `detail::shared`?

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -80,6 +175,71 @@ class usm_deleter {
     sycl::queue queue_;
 };
 
+template <typename T>
+using unique_usm_ptr = std::unique_ptr<T, usm_deleter<T>>;
+
+inline unique_usm_ptr<void> make_unique_usm(const sycl::queue& q,
```
michael-smirnov, line 181:
```suggestion
inline unique_usm_ptr<void> make_unique(const sycl::queue& q,
```

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -80,6 +175,71 @@ class usm_deleter {
...
+    return unique_usm_ptr<T>{ malloc_shared<T>(q, count), usm_deleter<T>{ q } };
+}
+
+template <typename T>
+inline unique_usm_ptr<T> make_unique_usm_host(const sycl::queue& q, std::int64_t count) {
+    return unique_usm_ptr<T>{ malloc_host<T>(q, count), usm_deleter<T>{ q } };
+}
+
+template <typename T>
+inline bool is_device_usm(const array<T>& ary) {
```
michael-smirnov, line 222: it's related to array - probably better to move it closer to it

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -16,37 +16,117 @@
...
+}
+
+inline void check_if_same_context(const sycl::queue& q1,
+                                  const sycl::queue& q2,
+                                  const sycl::queue& q3) {
+    check_if_same_context(q1, q2);
+    check_if_same_context(q1, q3);
+}
+
+inline void check_if_same_context(const sycl::queue& q1,
```
michael-smirnov, line 94: Don't these helpers are common and not only related to memory problems? Should we place it in `common` headers?

cpp/oneapi/dal/backend/transfer.hpp
michael-smirnov: When I see the file `transfer.hpp` without opening it - I have no clue what it contains. So `data` gives me more clarity here, probably it is a personal thing.

cpp/oneapi/dal/backend/transfer.hpp
michael-smirnov: I think no - but I'd rather put an assert instead of this branch

cpp/oneapi/dal/backend/transfer_dpc.cpp
```diff
@@ -0,0 +1,126 @@
...
+                    gathered_byte[i * block_size_in_bytes + j] =
+                        src_byte[i * src_stride_in_bytes + j];
+                }
+            }
+        });
+    });
+
+    auto copy_event = q.submit([&](sycl::handler& cgh) {
+        cgh.depends_on(gather_event);
+        cgh.memcpy(dst_host, gathered_device_unique.get(), block_count * block_size_in_bytes);
```
michael-smirnov, line 64: Why `memcpy` from `details::` is not used?

cpp/oneapi/dal/table/backend/homogen_table_impl.cpp
```diff
@@ -14,7 +14,10 @@
 * limitations under the License.
 *******************************************************************************/
 
+#include <iostream>
```
michael-smirnov, line 17: remove

cpp/oneapi/dal/backend/memory.hpp
rlnx: The same

cpp/oneapi/dal/algo/kmeans/test/batch.cpp
rlnx: The order depends on how types enumeration is organized in the test itself, it's not always the first element in the tuple. 

> Can we change TestType from std::tuple to some structure with meaningful type traits?

Tests often use `COMBINE_TYPES` that returns all possible combinations of types via `std::tuple`, so it's natural to use tuple as `TestType`. Adding special type trait or struct to fix this line sounds like overkill. Anyway, there are usings for tuple elements at the [next line](https://github.com/oneapi-src/oneDAL/pull/1488/files/46f7637bc4d1ed95a66889061efcdea32a644364#diff-d1d329637178f272246390e15be6186f23eb420e2362ad466b958f1bcb678cd2L40)

cpp/oneapi/dal/backend/memory.hpp
rlnx: The reason is to not write every time something like this:
```c++
ONEDAL_ASSERT(sycl::get_pointer_type(ary.get_data(), queue.get_context()) == sycl::usm::alloc::host);
```

But write this:
```c++
ONEDAL_ASSERT(is_host_usm_pointer(queue, ary.get_data());
```

>  Even you do not use it in is_device_friendly_usm_pointer by yourself.

I'm trying to minimize number of calls to `get_pointer_type`, maybe it's overoptimization.

cpp/oneapi/dal/backend/memory.hpp
rlnx: I'm thinking to introduce such helpers to array at the `detail` level, but it should be planned and done in other PR.

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
```diff
@@ -20,13 +20,84 @@
...
+}
+#endif
+
+#ifdef ONEDAL_DATA_PARALLEL
+inline sycl::usm::alloc alloc_kind_to_sycl(alloc_kind kind) {
+    switch (kind) {
+        case alloc_kind::usm_host: return sycl::usm::alloc::host;
+        case alloc_kind::usm_device: return sycl::usm::alloc::device;
+        case alloc_kind::usm_shared: return sycl::usm::alloc::shared;
+        case alloc_kind::host: return sycl::usm::alloc::unknown;
```
michael-smirnov, line 48: Why do we need to have such a mapping from `host` to `unknown`? Why not simply remove this case and throw an exception in `default` case?

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
```diff
@@ -20,13 +20,84 @@
...
+#ifdef ONEDAL_DATA_PARALLEL
+    switch (dst_alloc_kind) {
+        case alloc_kind::host: //
+            return (src_alloc_kind == alloc_kind::usm_device);
+        case alloc_kind::usm_host: //
+            return (src_alloc_kind == alloc_kind::host) || //
+                   (src_alloc_kind == alloc_kind::usm_device);
+        case alloc_kind::usm_device: //
+            return (src_alloc_kind == alloc_kind::host) || //
+                   (src_alloc_kind == alloc_kind::usm_host);
```
michael-smirnov, line 65: Why it is **required** to copy in this case?

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
```diff
@@ -20,13 +20,84 @@
...
+        case alloc_kind::usm_shared: //
+            return (src_alloc_kind != alloc_kind::usm_shared);
```
michael-smirnov, line 67: Why it is required to copy from `usm_host` in this case?

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
```diff
@@ -20,13 +20,84 @@
...
+        default: //
+            ONEDAL_ASSERT(!"Unsupported alloc_kind");
+    }
+#else
+    ONEDAL_ASSERT(src_alloc_kind == alloc_kind::host);
+    ONEDAL_ASSERT(dst_alloc_kind == alloc_kind::host);
+    return false;
+#endif
+}
+
```
michael-smirnov, line 77: I suppose `alloc_kind` declaration and all the functions above are not homogen-table specific and are common for dpc++ memory management in oneDAL

cpp/oneapi/dal/table/backend/interop/host_homogen_table_adapter.hpp
```diff
@@ -108,12 +113,32 @@ class host_homogen_table_adapter : public daal::data_management::HomogenNumericT
                                      rw_mode_t rwflag,
                                      block_desc_t<BlockData>& block);
 
-    bool check_row_indexes_in_range(const block_info& info) const;
-    bool check_column_index_in_range(const block_info& info) const;
-
-    host_homogen_table_adapter(const homogen_table& table, status_t& stat);
+    bool check_row_indexes_in_range(const block_info& info) const {
```
michael-smirnov, line 116: nip: what was the reason to return the definition of this function to the header file?

cpp/oneapi/dal/table/backend/interop/host_homogen_table_adapter.hpp
```diff
@@ -108,12 +113,32 @@ class host_homogen_table_adapter : public daal::data_management::HomogenNumericT
...
+        return info.row_begin_index < row_count && info.row_end_index <= row_count;
+    }
+
+    bool check_column_index_in_range(const block_info& info) const {
+        const std::int64_t column_count = original_table_.get_column_count();
+        return info.single_column_requested && info.column_index < column_count;
+    }
+
+    template <typename Body>
+    daal::services::Status convert_exception_to_status(Body&& body) {
```
michael-smirnov, line 127: Can be moved to cpp and made as free-function

cpp/oneapi/dal/table/backend/interop/sycl_table_adapter_dpc.cpp
```diff
@@ -0,0 +1,289 @@
+/*******************************************************************************
```
michael-smirnov, line 1: Why header is named `sycl_table_adapter` but source file is `sycl_table_adapter_dpc`?

cpp/oneapi/dal/table/backend/interop/sycl_table_adapter_dpc.cpp
```diff
@@ -0,0 +1,289 @@
...
+    }
+}
+
+static void convert_feature_information_to_daal(const table_metadata& src,
+                                                daal_dm::NumericTableDictionary& dst) {
+    ONEDAL_ASSERT(std::size_t(src.get_feature_count()) == dst.getNumberOfFeatures());
+    for (std::int64_t i = 0; i < src.get_feature_count(); i++) {
+        auto& daal_feature = dst[i];
+        // TODO: Determine indexType
+        // daal_feature.indexType = daal_dm::features::
```
michael-smirnov, line 39: remove if unused?

cpp/oneapi/dal/table/backend/interop/sycl_table_adapter_dpc.cpp
```diff
@@ -0,0 +1,289 @@
...
+        ONEDAL_ASSERT(!"Data is accessible in read-only mode by design");
+        return daal::services::ErrorMethodNotImplemented;
+    }
+
+    const block_info info{ block, vector_idx, vector_num };
+    if (!check_row_indexes_in_range(info)) {
+        return daal::services::ErrorIncorrectIndex;
+    }
+
+    const auto [buffer, status] = pull_rows_buffer<BlockData>(info);
```
michael-smirnov, line 210: What if the `block` already contains the buffer where we can write?
I'm not sure we need to apply such optimizations here since this adapter is temporary solution

cpp/oneapi/dal/test/engine/catch.hpp
```diff
@@ -27,4 +27,6 @@
 // To make USM work, we disable signal handling in Catch2.
 #define CATCH_CONFIG_NO_POSIX_SIGNALS
 
+#define CATCH_CONFIG_DISABLE_EXCEPTIONS
```
michael-smirnov, line 30: Why this is needed?

michael-smirnov (commented):
The final part of my comments

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
rlnx: This function determines efficient copy strategy. `case alloc_kind::usm_device:` corresponds to the following scenario:
```c++
auto x = accessor.pull(q, {0, -1}, sycl::usm::alloc::device);
``` 

If we do not copy, `x` will contain USM host memory. However, user specified USM device memory in `pull` and expect efficient access to `x` on device.

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
rlnx: The same considerations mentioned [in the previous comment](https://github.com/oneapi-src/oneDAL/pull/1488#discussion_r595822344).

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
rlnx: Agree. Moreover, I'd like to move it to detail, because `array` and maybe public part of tables might benefit from this functionality. For this PR, I'd keep it as is.

cpp/oneapi/dal/table/backend/interop/host_homogen_table_adapter.hpp
rlnx: Artefact of previous design idea, will move back

cpp/oneapi/dal/test/engine/catch.hpp
rlnx: If Catch2 does not catch exceptions from test cases, it's easier to debug via GDB.

cpp/oneapi/dal/table/backend/interop/sycl_table_adapter_dpc.cpp
rlnx: `SyclHomogenTable` has similar behavior

cpp/oneapi/dal/table/backend/interop/sycl_table_adapter_dpc.cpp
rlnx: It needs to be compiled only for DPC++. It's our convention to put `_dpc` at the end.

rlnx:
/intelci: run
set infra_branch=dev/risrafil/usm-device-adapter

rlnx:
/intelci: run
set infra_branch=dev/risrafil/usm-device-adapter

rlnx:
/intelci: restart
set infra_branch=dev/risrafil/usm-device-adapter

rlnx:
/intelci: run
set infra_branch=dev/risrafil/usm-device-adapter

rlnx:
/intelci: run
set infra_branch=dev/risrafil/usm-device-adapter

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
michael-smirnov: Ok

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -59,10 +60,102 @@ inline constexpr Integer up_multiple(Integer x, Integer multiple) {
...
+/// Example: up_pow2(10) == 16
+/// Example: up_pow2(16) == 16
+template <typename Integer>
+inline constexpr Integer up_pow2(Integer x) {
+    static_assert(std::is_integral_v<Integer>);
+    ONEDAL_ASSERT(x > 0);
+    Integer power = 1;
+    while (power < x) {
+        power *= 2;
+    }
```
michael-smirnov, line 89: Are some overflow checks needed here in your opinion?

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -59,10 +60,102 @@ inline constexpr Integer up_multiple(Integer x, Integer multiple) {
...
+inline void check_if_same_context(const sycl::queue& q1,
+                                  const sycl::queue& q2,
+                                  const sycl::queue& q3,
+                                  const sycl::queue& q4) {
+    check_if_same_context(q1, q2, q3);
+    check_if_same_context(q1, q4);
+}
+
+template <typename T>
+inline bool is_same_context(const sycl::queue& q, const array<T>& ary) {
```
michael-smirnov, line 149: nip: we can put it closer to other `is_same_context` overloadings

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -16,49 +16,131 @@
...
 
 namespace oneapi::dal::backend {
 
 #ifdef ONEDAL_DATA_PARALLEL
+inline bool is_device_usm(const sycl::queue& queue, const void* pointer) {
+    const auto pointer_type = sycl::get_pointer_type(pointer, queue.get_context());
+    return pointer_type == sycl::usm::alloc::device;
+}
+
+inline bool is_shared(const sycl::queue& queue, const void* pointer) {
```
michael-smirnov, line 29: `is_shared_usm`?

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
michael-smirnov: I just trying to say that the name of the function does not clearly describe its intent. Yes, in terms of efficiency it is needed, but not required in terms of functional correctness

rlnx: Frankly speaking I don't like the current behavior. The problem is that if the user passes specific alloc kind, the function may return array allocated in different way.
```c++
auto x = accessor.pull(q, {0, -1}, sycl::usm::alloc::device);
// x could contain shared USM
```

The expected behavior to me would be if `x` is always allocated in the way that the user ask. However, it may require copy even if technically it's not needed, e.g., table contains shared but user ask for device.

rlnx:
/intelci: run
set infra_branch=dev/risrafil/usm-device-adapter

michael-smirnov (approved):

rlnx:
/intelci: restart
set infra_branch=dev/risrafil/usm-device-adapter

PR #1490, opened 2021-03-10
Remove SECTION from bazel alg tests, add test_nightly flag support and increased test timeout

vlad-nazarov:


cpp/oneapi/dal/algo/decision_forest/test/batch.cpp
```diff
@@ -170,12 +170,12 @@ class df_batch_test : public te::algo_fixture {
...
-        SECTION("infer accuracy is expected") {
-            for (auto ch : checker_list) {
-                CAPTURE(ch.name);
-                REQUIRE(ch.check(infer_result.get_labels(), y_test) < ch.required_accuracy + eps);
-            }
+        for (auto ch : checker_list) {
+            CAPTURE(ch.name);
+            REQUIRE(ch.check(infer_result.get_labels(), y_test) < ch.required_accuracy + eps);
         }
+        INFO("infer accuracy is expected")
```
rlnx, line 177: Info should go before the actual check, and it's better to rephrase like "check if infer accuracy is expected"

rlnx (approved):

PR #1492, opened 2021-03-11
Using read function to read graph from file

orazve:
Enable graph reading via read function.


cpp/oneapi/dal/detail/read_ops.hpp
```diff
@@ -17,22 +17,67 @@
...
+    typedef char yes_t[1];
+    typedef char no_t[2];
+
+    template <typename C>
+    static yes_t& test(decltype(&C::tag_t));
+    template <typename C>
+    static no_t& test(...);
+
+public:
+    enum { value = (sizeof(test<T>(0)) == sizeof(yes_t)) };
```
KulikovNikita, line 40:
```suggestion
    constexpr static bool value = sizeof(test<T>(0)) == sizeof(yes_t);
```

PR #1505, opened 2021-03-18
Serialization prototype

rlnx:
Early draft of possible serialization implementation.

Serialization example:
```c++
auto model = dal::train(svm_desc, data, lables).get_model();

dal::detail::binary_output_archive ar;
dal::detail::save(model, ar);

const void* model_bits = ar.get_data();
``` 

Source of inspiration:
https://uscilab.github.io/cereal


cpp/oneapi/dal/test/serialization.cpp
```diff
@@ -0,0 +1,108 @@
...
+    std::int64_t x4;
+    float x5;
+    double x6;
+
+    void serialize(detail::archive& ar) {
+        ar(x1, x2, x3, x4, x5, x6);
+    }
+};
+
+TEST("save POD type") {
```
rlnx, line 87: Now is very limited functionality is implemented, pay attention to this test case to understand how the things are organized

cpp/oneapi/dal/detail/serialization.hpp
```diff
@@ -0,0 +1,104 @@
+/*******************************************************************************
```
KulikovNikita, line 1: Why don't you move it using `git mv`? Now it is depicted as brand new file

rlnx: This is brand new file

PR #1506, opened 2021-03-18
Update Bazel HOWTO

rlnx:
Upstream the changes presented on the meeting.

### DIFF
--- a/dev/bazel/HOWTO.md
+++ b/dev/bazel/HOWTO.md
@@ -92,7 +92,7 @@ The most used Bazel commands are `build`, `test` and `run`.
   `dbg` and `fastbuild`. \
   Possible values:
    - `opt` _(default)_ compiles everything with optimizations `-O2`.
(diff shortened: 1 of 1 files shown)


dev/bazel/HOWTO.md
```diff
@@ -292,9 +341,21 @@ dal_test_suite(
 ```
 
 ### Run tests for the existing oneDAL build
-TBD
+1. Set `DAALROOT` env var:
```
michael-smirnov, line 344: At the moment, we have `DALROOT` variable - I think it is better to use this one. `DAALROOT` is just for compatibility I guess

averbukh (approved):

dev/bazel/HOWTO.md
rlnx: In fact Bazel uses DAALROOT now, need to fix in Bazel scripts and then in docs

PR #1511, opened 2021-03-19
Fix flaky test for row accessor

rlnx:
Fix the following case:
```sh
bazel test -c dbg //cpp/oneapi/dal/table:test_row_accessor_dpc --device=cpu
```

### DIFF
--- a/cpp/oneapi/dal/table/test/row_accessor.cpp
+++ b/cpp/oneapi/dal/table/test/row_accessor.cpp
@@ -164,9 +164,7 @@ TEST("pull as device usm from host-allocated homogen table") {
         row_accessor<const float>{ data } //
             .pull(q, { 1, 3 }, sycl::usm::alloc::device);
 
-    const auto data_arr_host =
-        la::matrix<float>::wrap(q, data_arr_device.get_data(), { row_count, column_count })
-            .to_host();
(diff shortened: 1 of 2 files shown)


KulikovNikita (approved):

cpp/oneapi/dal/table/test/row_accessor.cpp
```diff
@@ -164,9 +164,7 @@ TEST("pull as device usm from host-allocated homogen table") {
         row_accessor<const float>{ data } //
             .pull(q, { 1, 3 }, sycl::usm::alloc::device);
 
-    const auto data_arr_host =
-        la::matrix<float>::wrap(q, data_arr_device.get_data(), { row_count, column_count })
-            .to_host();
+    const auto data_arr_host = la::matrix<float>::wrap(data_arr_device).to_host();
```
michael-smirnov, line 167: Why did you decide to drop the `queue` from parameters of `wrap` function? The agreement was that all functions for DPC++ case that accept non-host-accessible memory shall accept the queue directly.

rlnx: Because queue is already in the array if it holds DPC++ memory. `data_arr_device` is `dal::array`.

PR #1518, opened 2021-03-22
Refactor array tests

rlnx:
- Migrate array tests to Catch2


KulikovNikita (approved):

cpp/oneapi/dal/test/array.cpp
```diff
@@ -0,0 +1,332 @@
...
+TEST("can make owning array from non owning readonly") {
+    array<float> arr;
+
+    float data[] = { 1.f, 2.f, 3.f };
+    arr.reset(array<float>(), (const float*)(data), 3);
+
+    REQUIRE(arr.get_count() == 3);
+    REQUIRE(arr.get_data() == data);
+    REQUIRE(!arr.has_mutable_data());
+    // ASSERT_THROW(arr.get_mutable_data(), std::bad_variant_access);
```
michael-smirnov, line 147: Uncomment or place todo here?
I think `bad_variant_access` is an implementation detail and we shall throw something like `dal::logic_error`

PR #1521, opened 2021-03-22
Update weekly and nightly tests flags

vlad-nazarov:


dev/bazel/dal.bzl
```diff
@@ -344,22 +343,13 @@ def _test_eternal_datasets_args(framework):
...
-            ],
-        })
-    return []
-
-def _test_weekly_args(framework):
+def _test_filter_args(framework):
     if framework == "catch2":
         return _select({
+            "@config//:test_nightly_enabled": ["~[weekly]"],
             "@config//:test_weekly_enabled": [],
```
rlnx, line 350: What is behavior if both options `--test_nightly` and `--test_weekly` are passed?

vlad-nazarov: The tests will not run with the error that only one of these flags is needed.

rlnx: I've tried this, error message is not very clear, but probably fine for now

rlnx (approved):

PR #1526, opened 2021-03-23
Migrate linear kernel to DPC++ interfaces

vlad-nazarov:


cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,54 @@
...
 namespace oneapi::dal::linear_kernel::backend {
 
 using dal::backend::context_gpu;
 using input_t = compute_input<task::compute>;
 using result_t = compute_result<task::compute>;
 using descriptor_t = detail::descriptor_base<task::compute>;
 
-namespace daal_linear_kernel = daal::algorithms::kernel_function::linear;
-namespace interop = dal::backend::interop;
+namespace primitives = dal::backend::primitives;
```
rlnx, line 28: It's better to stick to `pr`, so
```c++
namespace pr = dal::backend::primitives;
```

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,54 @@
...
-        dal::detail::homogen_table_builder{}.reset(arr_values, row_count_x, row_count_y).build());
-}
 
-template <typename Float>
-static result_t compute(const context_gpu& ctx, const descriptor_t& desc, const input_t& input) {
-    return call_daal_kernel<Float>(ctx, desc, input.get_x(), input.get_y());
+    const Float alpha = desc.get_scale();
+    const Float beta = desc.get_shift();
+
+    auto rows_block_x = row_accessor<const Float>(x).pull(queue, { 0, -1 });
```
rlnx, line 48: - Always work with device memory inside kernel
- It's better to stick to the naming scheme we already use in multiple places
```suggestion
    auto arr_x = row_accessor<const Float>(x).pull(queue, { 0, -1 }, sycl::usm::alloc::device);
```

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,54 @@
...
+    const auto ndarray_x =
+        primitives::ndarray<Float, 2>::wrap(rows_block_x.get_data(), { row_count_x, col_count_x });
+
+    auto rows_block_y = row_accessor<const Float>(y).pull(queue, { 0, -1 });
```
rlnx, line 52:
```suggestion
    auto arr_y = row_accessor<const Float>(y).pull(queue, { 0, -1 }, sycl::usm::alloc::device);
```

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,54 @@
...
+    const auto ndarray_y =
+        primitives::ndarray<Float, 2>::wrap(rows_block_y.get_data(), { row_count_y, col_count_y });
+
+    auto ndarray_res = primitives::ndarray<Float, 2>::empty(queue, { row_count_x, row_count_y });
```
rlnx, line 56:
```suggestion
    auto ndarray_res = primitives::ndarray<Float, 2>::empty(queue, { row_count_x, row_count_y }, sycl::usm::alloc::device);
```

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,54 @@
...
+    sycl::event event_res;
+    if (beta != 0.0)
+        event_res = ndarray_res.fill(queue, Float(1));
```
rlnx, line 59:
```suggestion
    if (beta != 0.0) {
        event_res = ndarray_res.fill(queue, Float(1));
     }
```

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,54 @@
...
+
+    gemm(queue, ndarray_x, ndarray_y.t(), ndarray_res, alpha, beta, { event_res }).wait_and_throw();
+
+    return result_t{}.set_values(dal::detail::homogen_table_builder{}
+                                     .reset(ndarray_res.flatten(queue), row_count_x, row_count_y)
+                                     .build());
```
rlnx, line 65: I'd suggest to not use builder for such simple cases. We can simplify this significantly if we slightly extend `homogen_table` by adding static `wrap` method.

```c++
const auto values = homogen_table::wrap(ndarray_res.flatten(queue), row_count_x, row_count_y);
return result_t{}.set_values(values );
```

There is already `wrap` for raw pointer in `homogen_table`. Similar method for `array` can be added.

rlnx: @vlad-nazarov, I've already implemented `wrap`, will create PR soon

vlad-nazarov: @rlnx, Added wrap method

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,55 @@
...
     dal::detail::check_mul_overflow(row_count_x, row_count_y);
-    auto arr_values =
-        array<Float>::empty(queue, row_count_x * row_count_y, sycl::usm::alloc::device);
 
-    const auto daal_x = interop::convert_to_daal_table(queue, x);
-    const auto daal_y = interop::convert_to_daal_table(queue, y);
-    const auto daal_values =
-        interop::convert_to_daal_table(queue, arr_values, row_count_x, row_count_y);
+    const Float alpha = desc.get_scale();
+    const Float beta = desc.get_shift();
```
michael-smirnov, line 46: I'd prefer to stick to one terminology - if this params named scale and shift, use this naming consistently in the linear kernel

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,55 @@
...
 
-    daal_linear_kernel::Parameter daal_parameter(desc.get_scale(), desc.get_shift());
-    daal_linear_kernel_t<Float>().compute(daal_x.get(),
-                                          daal_y.get(),
-                                          daal_values.get(),
-                                          &daal_parameter);
+    auto arr_x = row_accessor<const Float>(x).pull(queue, { 0, -1 }, sycl::usm::alloc::device);
+    const auto ndarray_x = pr::ndarray<Float, 2>::wrap(arr_x, { row_count_x, col_count_x });
```
michael-smirnov, line 49: I think this would be the common pattern in our code - can we create a helper function for this?
@rlnx what do you think?

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,55 @@
...
+    sycl::event fill_res_event;
+    if (beta != 0.0) {
+        fill_res_event = ndarray_res.fill(queue, Float(1));
+    }
+
+    gemm(queue, ndarray_x, ndarray_y.t(), ndarray_res, alpha, beta, { fill_res_event })
+        .wait_and_throw();
+
+    return result_t{}.set_values(
+        homogen_table::wrap(ndarray_res.flatten(queue), row_count_x, row_count_y));
```
michael-smirnov, line 66: The event from gemm theoretically could be passed to this wrap function if it would allow it. 
@rlnx, can we extend `homogen_table::wrap()` on array with event list parameter?

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,55 @@
...
 
-    const int64_t row_count_x = x.get_row_count();
-    const int64_t row_count_y = y.get_row_count();
+    const std::int64_t row_count_x = x.get_row_count();
+    const std::int64_t col_count_x = x.get_column_count();
+    const std::int64_t row_count_y = y.get_row_count();
+    const std::int64_t col_count_y = y.get_column_count();
 
+    ONEDAL_ASSERT(col_count_x == col_count_y);
     dal::detail::check_mul_overflow(row_count_x, row_count_y);
```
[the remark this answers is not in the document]
vlad-nazarov: To check the size of the resulting matrix for overflow

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
rlnx: Yes, makes sense, can be done as part of this PR

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
rlnx: Right, there is already such function, in my PR #1571 
```c++
#ifdef ONEDAL_DATA_PARALLEL
template <typename Float, template <typename> typename Accessor, typename Table>
inline ndarray<Float, 2> flatten_table(sycl::queue& q, const Table& table, sycl::usm::alloc alloc) {
    Accessor<const Float> accessor{ table };
    const auto data = accessor.pull(q, { 0, -1 }, alloc);
    return ndarray<Float, 2>::wrap(data, { table.get_row_count(), table.get_column_count() });
}
#endif
```

I hope this will be merged before this one.

> I think this would be the common pattern in our code 

Actually, I don't like it 🙈. We should read table by blocks to minimize memory footprint for non-homogeneous formats. However, we can postpone it until we have SOA table.

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
rlnx:
```c++
template <typename T>
homogen_table wrap(const array<T>& data, const std::vector<sycl::event>& deps) {
   sycl::event::wait_throw(deps);
   return wrap(data);
}
```

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
vlad-nazarov: Added flatten_table

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
vlad-nazarov: Added wrap with deps

cpp/oneapi/dal/table/homogen.hpp
```diff
@@ -115,6 +115,33 @@ class ONEDAL_EXPORT homogen_table : public table {
...
+    /// @param row_count    The number of rows in the table.
+    /// @param column_count The number of columns in the table.
+    /// @param dependencies Events indicating availability of the :literal:`Data` for reading or writing.
+    /// @param layout       The layout of the data. Should be :literal:`data_layout::row_major` or
+    ///                     :literal:`data_layout::column_major`.
+    template <typename Data>
+    static homogen_table wrap(const array<Data>& data,
+                              std::int64_t row_count,
+                              std::int64_t column_count,
+                              const sycl::vector_class<sycl::event>& dependencies = {},
```
rlnx, line 138:
```suggestion
                              const sycl::vector_class<sycl::event>& dependencies,
```

Otherwise, the signature is ambiguous

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,53 @@
...
-                                 const table& y) {
     auto& queue = ctx.get_queue();
-    interop::execution_context_guard guard(queue);
 
-    const int64_t row_count_x = x.get_row_count();
-    const int64_t row_count_y = y.get_row_count();
+    const std::int64_t row_count_x = x.get_row_count();
+    const std::int64_t row_count_y = y.get_row_count();
 
+    ONEDAL_ASSERT(col_count_x == col_count_y);
```
rlnx, line 40: I do not see `col_count_x` and `col_count_y` declared, does it compile with `-c dbg`?

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,53 @@
...
-                                 const descriptor_t& desc,
-                                 const table& x,
-                                 const table& y) {
     auto& queue = ctx.get_queue();
-    interop::execution_context_guard guard(queue);
 
-    const int64_t row_count_x = x.get_row_count();
-    const int64_t row_count_y = y.get_row_count();
+    const std::int64_t row_count_x = x.get_row_count();
+    const std::int64_t row_count_y = y.get_row_count();
```
rlnx, line 38: Naming pattern I saw in other algorithms/primitives `x_row_count`, `x_column_count`, let's be consistent.

cpp/oneapi/dal/algo/linear_kernel/backend/gpu/compute_kernel_dense_dpc.cpp
```diff
@@ -15,61 +15,53 @@
...
+    sycl::event fill_res_event;
+    if (shift != 0.0) {
+        fill_res_event = ndarray_res.fill(queue, Float(1));
+    }
+
+    auto gemm_event =
+        gemm(queue, ndarray_x, ndarray_y.t(), ndarray_res, scale, shift, { fill_res_event });
+
+    return result_t{}.set_values(
+        homogen_table::wrap(ndarray_res.flatten(queue), row_count_x, row_count_y, { gemm_event }));
```
rlnx, line 64: Suggest to have homogen table creation at the separate line.

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -511,4 +511,13 @@ class ndarray : public ndview<T, axis_count, order> {
     shared_t data_;
 };
 
+#ifdef ONEDAL_DATA_PARALLEL
+template <typename Float, template <typename> typename Accessor, typename Table>
+inline ndarray<Float, 2> flatten_table(sycl::queue& q, const Table& table, sycl::usm::alloc alloc) {
```
michael-smirnov, line 516: Why table is a template parameter?

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -511,4 +511,13 @@ class ndarray : public ndview<T, axis_count, order> {
...
+    Accessor<const Float> accessor{ table };
+    const auto data = accessor.pull(q, { 0, -1 }, alloc);
```
michael-smirnov, line 518: This implementation works only for `row_accessor`. To make this work with other accessors, we need to pass indexes from the user to this function directly. Otherwise, `Accessor` template parameter is not needed.

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -511,4 +511,13 @@ class ndarray : public ndview<T, axis_count, order> {
...
+    return ndarray<Float, 2>::wrap(data, { table.get_row_count(), table.get_column_count() });
```
michael-smirnov, line 519: The function is named "flatten", but it creates 2d array from the table 😕

cpp/oneapi/dal/table/homogen.hpp
```diff
@@ -115,6 +115,33 @@ class ONEDAL_EXPORT homogen_table : public table {
...
+    /// @param column_count The number of columns in the table.
+    /// @param dependencies Events indicating availability of the :literal:`Data` for reading or writing.
+    /// @param layout       The layout of the data. Should be :literal:`data_layout::row_major` or
+    ///                     :literal:`data_layout::column_major`.
+    template <typename Data>
+    static homogen_table wrap(const array<Data>& data,
+                              std::int64_t row_count,
+                              std::int64_t column_count,
+                              const sycl::vector_class<sycl::event>& dependencies,
+                              data_layout layout = data_layout::row_major) {
```
michael-smirnov, line 139: 1. Is it a proper order of these params?
2. `std::vector` instead of `sycl::vector_class`?

vlad-nazarov: I made this function consistent with another wrap
https://github.com/oneapi-src/oneDAL/blob/master/cpp/oneapi/dal/table/homogen.hpp#L80

vlad-nazarov: @michael-smirnov @rlnx, Is this wrap suitable for us or do we need changes?

rlnx: I think we need to remove this... The problem is that `dal::array` is already expected to track dependencies as it can be created from pointer and deps vector. We already have constructor in this table that accepts just an array without dependencies. I assume, if the user wants to create table with dependencies, they should create array with dependencies first and then pass to the `wrap` function. 

I suggest to think more about this function and to not introduce it in this PR as it could create a lot of problems with backward compatibility in future.

What I suggest now is to parametrize `ndarray::flatten` with additional dependencies argument, so that:
```c++
const auto array_that_encloses_dependencies = ndarr.flatten(queue, { some_event }); 
homogen_table::wrap(array_that_encloses_dependencies, row_count, column_count);
```

cpp/oneapi/dal/table/homogen.hpp
```diff
@@ -115,6 +115,33 @@ class ONEDAL_EXPORT homogen_table : public table {
...
+    ///                     type should be at least :expr:`float`, :expr:`double` or :expr:`std::int32_t`.
+    ///
+    /// @param data         The array that stores a homogeneous data block.
+    /// @param row_count    The number of rows in the table.
+    /// @param column_count The number of columns in the table.
+    /// @param dependencies Events indicating availability of the :literal:`Data` for reading or writing.
+    /// @param layout       The layout of the data. Should be :literal:`data_layout::row_major` or
+    ///                     :literal:`data_layout::column_major`.
+    template <typename Data>
+    static homogen_table wrap(const dal::array<Data>& data,
```
rlnx, line 135: We agreed to remove that function, please do it as soon as possible to close this PR

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -443,7 +443,9 @@ class ndarray : public ndview<T, axis_count, order> {
     }
 
 #ifdef ONEDAL_DATA_PARALLEL
-    array_t flatten(sycl::queue& q) const {
+    array_t flatten(sycl::queue& q,
+                    const sycl::vector_class<sycl::event>& dependencies = {}) const {
+        sycl::event::wait_and_throw(dependencies);
         ONEDAL_ASSERT(is_known_usm(q, data_.get()));
         return array_t{ q, data_, this->get_count() };
```
michael-smirnov, line 450:
```suggestion
        ONEDAL_ASSERT(is_known_usm(q, data_.get()));
        return array_t{ q, data_, this->get_count(), dependencies };
```

michael-smirnov (approved):
just a minor comment

rlnx (approved):

PR #1529, opened 2021-03-26
Scikit-learn conformance calculation for different device types

michael-smirnov:
- Created `run_tests_with_context.py` script to run scikit tests with `sycl_context` and different devices
- Created script `run_xpu_conformace.py` to calculate conformance for a given device based on scikit tests
- Created an ability to filter out failed tests from conformance results.
- Fixed a bug with wrong conformance calculation if failed tests exist in log files. Such log files contain details about failed tests, so patching information is doubled in the log files for failed tests. Previously, these doubled lines were considered twice during conformance calculation.


michael-smirnov:
Classic reports before and after my changes
[report.pdf](https://github.com/oneapi-src/oneDAL/files/6212604/report.pdf)
[report_old.pdf](https://github.com/oneapi-src/oneDAL/files/6212606/report_old.pdf)

michael-smirnov:
New reports based on context approach

[report_host.pdf](https://github.com/oneapi-src/oneDAL/files/6212615/report_host.pdf)
[report_cpu.pdf](https://github.com/oneapi-src/oneDAL/files/6212616/report_cpu.pdf)
[report_gpu.pdf](https://github.com/oneapi-src/oneDAL/files/6212617/report_gpu.pdf)

napetrov:
why we have 38% of daal4py in case of GPU run, while it's 40 for others?  I thought that this should be same number and only GPU coverage should differs?

michael-smirnov:
@napetrov because I decided to calculate conformance for GPU case differently as it is done for CPU.
Historically, all test cases are considered to calculate conformance on CPU - even failed tests.
for GPU case, I filtered out failed tests so they do not impact conformance %. This can be done for CPU case also, since conformance calculation logic is common for different device types.
But the numbers of using daal4py will not be the same in that case because in GPU case we have more failed tests comparing to CPU runs.

PetrovKP (approved):

PR #1537, opened 2021-03-31
Timeout reduce for daal4py tests

PetrovKP:


napetrov (approved):
Assuming CI will pass

PR #1538, opened 2021-03-31
Add LAPACK primitives

rlnx:
- Add LAPACK function to compute eigenvalues and eigenvectors for symmetric matrices
- Add corresponding tests
- Add some helper functions to `test::engine::linalg`


cpp/oneapi/dal/backend/primitives/lapack/eigen.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include <src/externals/service_lapack.h>
+
+#include "oneapi/dal/backend/primitives/lapack/eigen.hpp"
+#include "oneapi/dal/backend/dispatcher.hpp"
+#include "oneapi/dal/backend/interop/common.hpp"
+
+namespace oneapi::dal::backend::primitives {
+
+template <typename Float>
+static void syevd(char jobz,
```
michael-smirnov, line 26: Just to double-check: sometimes we are talking about support for different SYCL backends for oneDAL. Don't we want to create primitives in a way it will be easily to use oneMKL underneath?
For example, here we are using host-only MKL small libs, but can possibly use oneMKL interfaces.

cpp/oneapi/dal/test/engine/linalg.hpp
```diff
@@ -20,3 +20,4 @@
 #include "oneapi/dal/test/engine/linalg/loops.hpp"
 #include "oneapi/dal/test/engine/linalg/umath.hpp"
 #include "oneapi/dal/test/engine/linalg/dot.hpp"
+#include "oneapi/dal/test/engine/linalg/random.hpp"
```
michael-smirnov, line 23: Having random generators inside `linalg` confuses me. Is it a common-used practice?

rlnx: It's randomly generated matrices with specific properties

cpp/oneapi/dal/backend/primitives/lapack/eigen.cpp
rlnx: Currently we need to call LAPACK on CPU. 

> Don't we want to create primitives in a way it will be easily to use oneMKL underneath?

In general, yes. But let's move step-by-step. It's easier to add universal DPC++ versions once GPU support is needed. 

The problem is that the small version of oneMKL we are using is GPU-only, there is no CPU branch underneath, so device dispatching should be controlled on our side.

cpp/oneapi/dal/backend/primitives/lapack/eigen.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+template <typename Float>
+void sym_eigvals_impl(Float* a, std::int64_t n, std::int64_t lda, Float* w) {
+    ONEDAL_ASSERT(a);
+    ONEDAL_ASSERT(w);
+    ONEDAL_ASSERT(n > 0);
+    ONEDAL_ASSERT(lda >= n);
+
+    const std::int64_t lwork = 2 * n * n + 6 * n + 1;
+    const std::int64_t liwork = 5 * n + 3;
```
KulikovNikita, line 74: Does LAPACK provide any interfaces to get information about the required workspace size?

cpp/oneapi/dal/backend/primitives/lapack/eigen.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    ONEDAL_ASSERT(lwork > n);
+    ONEDAL_ASSERT(liwork > n);
```
KulikovNikita, line 77: Is it a check for overflow? In other case it seems to be redundant

cpp/oneapi/dal/backend/primitives/lapack/eigen.hpp
```diff
@@ -0,0 +1,67 @@
...
+/// @param[in, out] data_or_eigvecs The input parameter is interpreted as symmetric matrix of
+///                                 size [n x n]. The computed eigenvectors is written to that
+///                                 matrix. If `order == ndorder::c`, $i$-th row of the matrix
+///                                 contains $i$-th eigenvector. If `order == ndorder::f`, $i$-th
+///                                 column of the matrix contains $i$-th eigenvector.
+/// @param[out] eigvals             The output array of size [n] that stores computed eigenvalues.
+///                                 The eigenvalues are written in ascending order. $i$-th eigenvalue
+///                                 corrensponds to $i$-th eigenvector.
+template <typename Float, ndorder order>
+inline void sym_eigvals(ndview<Float, 2, order>& data_or_eigvecs, ndview<Float, 1>& eigvals) {
```
KulikovNikita, line 42: Doesn't it work only with row-major layout?

KulikovNikita: In any case it is symmetric. But I suppose that it can be useful to limit a user

cpp/oneapi/dal/backend/primitives/lapack/eigen.cpp
rlnx: Yes, it's overflow

cpp/oneapi/dal/backend/primitives/lapack/eigen.hpp
rlnx: It does work for both layouts

cpp/oneapi/dal/backend/primitives/lapack/eigen.cpp
rlnx: No, we don't have such function imported from microlibs. In old interfaces work size is calculated in PCA :smile:

cpp/oneapi/dal/detail/common.hpp
```diff
@@ -49,6 +49,28 @@
     } while (0)
 #endif
 
+#ifdef ONEDAL_DATA_PARALLEL
+#define __ONEDAL_IF_QUEUE__(optional_queue, ...) \
```
michael-smirnov, line 53: How sound the rule to create a macro with underscores?

cpp/oneapi/dal/test/engine/linalg.hpp
michael-smirnov: Matrices can allow not only linear algebra operations. Seems like you put random matrix creation here because the matrix is also defined here. I suggest separating this module on `engine/matrix`, `engine/linalg`, and `engine/random`.

cpp/oneapi/dal/detail/common.hpp
rlnx: This is to let users know, what is allowed to use. Macros with `__` are internal and should not be used. Macro w/o `__` is "safe" to use outside library.

cpp/oneapi/dal/test/engine/linalg.hpp
rlnx: Agree in general, but I'd like to stick here to numpy conventions: having data container in the root package, but linalg and random operations in submodules, such design has already proven its convenience. That's what I suggest:
1. Move `test/engine/linalg/matrix.hpp` -> `test/engine/matrix.hpp`, `test::engine::linalg::matrix` -> `test::engine::matrix`
2. Introduce `test/engine/random/*.hpp` and `test/engine/random.hpp`, move all RNG functions there
3. Remove extra namespaces like `linalg` or `random`, but keep separation on directories
    - I don't believe we need them for testing purpose, but it requires people to write `namespace la = ..`, `namespace rng = ...` in every test.

I don't want to do it as part of this PR, as it affect most of the tests. I'll create another one to refactor test engine once lapack is merged. There are other things I wanted to do, e.g., move `test/engine/dataframe_*.hpp` files to separate directory. It's good moment to do that.

KulikovNikita (approved):

cpp/daal/src/externals/core_threading_win_dll.cpp
```diff
@@ -1211,14 +1209,23 @@ CALL_VOID_FUNC_FROM_DLL(fpk_spblas_, mkl_scsrmm, (CSRMM_ARGS(float)),
...
                         (transa, m, n, k, alpha, matdescra, val, indx, pntrb, pntre, b, ldb, beta, c, ldc));
 
+typedef int IppStatus;
+typedef unsigned char Ipp8u;
+typedef unsigned short Ipp16u;
+typedef unsigned int Ipp32u;
+typedef signed short Ipp16s;
+typedef signed int Ipp32s;
+typedef float Ipp32f;
+typedef double Ipp64f;
```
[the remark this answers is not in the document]
rlnx: It's DAAL part

cpp/oneapi/dal/backend/micromkl/macro.hpp
```diff
@@ -0,0 +1,129 @@
+/*******************************************************************************
```
michael-smirnov, line 1: Macro header in micromkl module - funny moment 😄

michael-smirnov (approved):

PR #1542, opened 2021-04-01
Add array-based wrap to homogen table

rlnx:
- Add `homogen_table::wrap(array, row_count, column_count)`
- Migrate homogen table tests to Catch2
- Add more tests cases related to new `wrap` method
- Add more tests to array


cpp/oneapi/dal/table/homogen.hpp
```diff
@@ -119,18 +121,39 @@ class ONEDAL_EXPORT homogen_table : public table {
     }
 #endif
 
-public:
+    /// Creates a new ``homogen_table`` instance from externally-defined data block.
+    /// The created table stores the provided array object.
```
michael-smirnov, line 125: I'm not sure it is good to say that the table stores an array. It is up to implementation in my opinion, but the table shall store the data contained in the array.
There is no details about how it should refer to the data inside the array. `wrap()` over raw pointers does not own it, here the behavior is opposite I guess. We need to articulate this in the docs. 
Another thing is about information that the table can use from the array. Should table preserve the information about data mutability in the array? I think yes, it is needed.

cpp/oneapi/dal/table/homogen.hpp
```diff
@@ -217,6 +241,26 @@ class ONEDAL_EXPORT homogen_table : public table {
...
+    ///                      The :literal:`Data` type should be at least :expr:`float`, :expr:`double`
+    ///                      or :expr:`std::int32_t`.
+    ///
+    /// @param data          The array that stores a homogeneous data block.
+    /// @param row_count     The number of rows in the table.
+    /// @param column_count  The number of columns in the table.
+    /// @param layout        The layout of the data. Should be :literal:`data_layout::row_major` or
+    ///                      :literal:`data_layout::column_major`.
+    template <typename Data>
+    homogen_table(const array<Data> data,
```
michael-smirnov, line 257:
```suggestion
    homogen_table(const array<Data>& data,
```

cpp/oneapi/dal/table/homogen.hpp
rlnx: Agree with the point about what table stores.

> Another thing is about information that the table can use from the array. Should table preserve the information about data mutability in the array? I think yes, it is needed.

Tables are always immutable. Information on mutability can be preserved for some reason or not, this is implementation details.

cpp/oneapi/dal/table/BUILD
```diff
@@ -28,29 +28,29 @@ dal_test_suite(
...
         "test/row_accessor.cpp",
+        "test/homogen.cpp",
     ],
     dal_deps = [":table"],
     framework = "catch2",
 )
 
 dal_test_suite(
-    name = "interface_tests",
+    name = "gtest_tests",
```
amgrigoriev, line 41: I am wondering why gtest are for column_accessor  while catch2 tests are for row_accessor.

KulikovNikita (approved):

cpp/oneapi/dal/detail/array_utils.hpp
```diff
@@ -39,8 +39,67 @@ class array_via_policy {
...
+    }
+};
+
+template <typename T, typename U, typename CastOp>
+inline array<T> array_cast_generic(const array<U>& ary, const CastOp& cast_op) {
+    if (ary.get_size() % sizeof(T) > 0) {
+        throw invalid_argument{ error_messages::incompatible_array_reinterpret_cast_types() };
+    }
+
+    const std::int64_t new_count = ary.get_size() / sizeof(T);
```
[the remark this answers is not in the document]
rlnx: I already have an exception, which is thrown before this operation.

cpp/oneapi/dal/table/BUILD
rlnx: The test were initially written in gtest, I'm gradually moving them to catch2. At this iteration, I migrated homogen table tests

cpp/oneapi/dal/detail/array_utils.hpp
```diff
@@ -39,8 +39,47 @@ class array_via_policy {
 #endif
 };
 
+template <typename T, typename Body>
+inline auto dispath_by_policy(const array<T>& data, Body&& body) {
```
michael-smirnov, line 43: `data` is used just to obtain an optional on queue. Isn't it better to pass `std::optional<sycl::queue>` instead?

cpp/oneapi/dal/table/backend/homogen_table_impl.cpp
```diff
@@ -70,16 +70,8 @@ static void refer_origin_data(const array<DataSrc>& src,
...
-        // available only for builders, not for tables.
-        auto start_pointer = reinterpret_cast<DataDest*>(src.get_mutable_data() + src_start_index);
-        dst.reset(src, start_pointer, dst_count);
-    }
-    else {
-        auto start_pointer = reinterpret_cast<const DataDest*>(src.get_data() + src_start_index);
-        dst.reset(src, start_pointer, dst_count);
-    }
+    auto start_pointer = reinterpret_cast<const DataDest*>(src.get_data() + src_start_index);
+    dst.reset(src, start_pointer, dst_count);
```
michael-smirnov, line 74: Can you please check that the builder logic is not broken? I mean, pulling data from a builder does not require making a copy. At this moment, the builder shares one implementation with homogen table.

cpp/oneapi/dal/table/homogen.hpp
```diff
@@ -119,18 +121,39 @@ class ONEDAL_EXPORT homogen_table : public table {
     }
 #endif
 
-public:
+    /// Creates a new ``homogen_table`` instance from an array.
+    /// The created table shares data ownership with the given array.
```
michael-smirnov, line 125:
```suggestion
    /// The created table shares data ownership with the given array.
    ///  Information about data mutability in the array is stored in the table.
```
Not sure we need to mention it here, but definitely we need to make it in the spec.

cpp/oneapi/dal/table/homogen.hpp
```diff
@@ -119,18 +121,39 @@ class ONEDAL_EXPORT homogen_table : public table {
...
+    static homogen_table wrap(const array<Data>& data,
+                              std::int64_t row_count,
+                              std::int64_t column_count,
+                              data_layout layout = data_layout::row_major) {
+        return homogen_table{ data, row_count, column_count, layout };
+    }
+
     /// Creates a new ``homogen_table`` instance with zero number of rows and columns.
     /// The :expr:`kind` is set to``homogen_table::kind()``.
-    /// All the properties should be set to default values (see the Properties section).
```
michael-smirnov, line 125: Maybe to rephrase, but not to remove completely?

cpp/oneapi/dal/table/homogen.hpp
```diff
@@ -217,6 +241,26 @@ class ONEDAL_EXPORT homogen_table : public table {
         init_impl(std::forward<Impl>(impl));
     }
 
+    /// Creates a new ``homogen_table`` instance from externally-defined data block.
```
michael-smirnov, line 244: Why you documented protected functionality?

cpp/oneapi/dal/detail/array_utils.hpp
rlnx: The problem is that `array::get_queue()` is not available for non-DPC++ case. That's why call of this method is under macro `ONEDAL_DATA_PARALLEL`

cpp/oneapi/dal/table/backend/homogen_table_impl.cpp
rlnx: Added test case for that
```c++
TEST("pull returns mutable data from homogen_table_builder") {
    constexpr std::int64_t row_count = 3;
    constexpr std::int64_t column_count = 2;
    float data[row_count * column_count] = { 1.0f, 2.0f, 3.0f, -1.0f, -2.0f, -3.0f };

    const auto ary = array<float>::wrap(data, row_count * column_count);
    const auto builder = detail::homogen_table_builder{}.reset(ary, row_count, column_count);
    const auto block = row_accessor<const float>{ builder }.pull();

    REQUIRE(block.has_mutable_data() == true);
}
```

cpp/oneapi/dal/table/homogen.hpp
rlnx: I would not say that, this is implementation details. Moreover, it creates confusion as we say that tables are always immutable.

cpp/oneapi/dal/table/homogen.hpp
rlnx: I'd also remove the phrase about kind. It's not clear how to "set default value" since we decided stop using properties.
However, it's important to understand behavior if "empty" table is created. I suggest to mention it in the method descriptor. I'll update `get_data` description.

cpp/oneapi/dal/detail/array_utils.hpp
rlnx: I've tried `__IF_QUEUE__`, it looks worse than `dispatch_by_policy` because requires some code duplication for the `__IF_QUEUE__` and `__IF_NO_QUEUE__` cases.

cpp/oneapi/dal/table/backend/homogen_table_impl.cpp
michael-smirnov: Thanks

michael-smirnov (approved):

PR #1546, opened 2021-04-02
Fix empty table case for jaccard algo

YakhtinLeonid:
Fix the jaccard algorithm for tables with zero columns and non-zero rows

### DIFF
--- a/cpp/oneapi/dal/algo/jaccard/detail/vertex_similarity_ops.hpp
+++ b/cpp/oneapi/dal/algo/jaccard/detail/vertex_similarity_ops.hpp
@@ -33,7 +33,6 @@ struct vertex_similarity_ops_dispatcher {
         const Policy &policy,
         const Descriptor &descriptor,
         vertex_similarity_input<Graph, task_t> &input) const {
-
         const std::int64_t row_begin = descriptor.get_row_range_begin();
         const std::int64_t row_end = descriptor.get_row_range_end();
         const std::int64_t column_begin = descriptor.get_column_range_begin();
--- a/cpp/oneapi/dal/algo/jaccard/detail/vertex_similarity_ops.hpp
+++ b/cpp/oneapi/dal/algo/jaccard/detail/vertex_similarity_ops.hpp
@@ -33,16 +33,6 @@ struct vertex_similarity_ops_dispatcher {
         const Policy &policy,
         const Descriptor &descriptor,
         vertex_similarity_input<Graph, task_t> &input) const {
-        const std::int64_t row_begin = descriptor.get_row_range_begin();
-        const std::int64_t row_end = descriptor.get_row_range_end();
-        const std::int64_t column_begin = descriptor.get_column_range_begin();
-        const std::int64_t column_end = descriptor.get_column_range_end();
-        const std::int64_t number_elements_in_block =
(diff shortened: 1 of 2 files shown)


cpp/oneapi/dal/algo/jaccard/detail/vertex_similarity_ops.hpp
```diff
@@ -96,6 +96,13 @@ vertex_similarity_result vertex_similarity_ops_dispatcher<Policy, Float, Method,
     const std::int64_t max_block_size =
         get_max_block_size<Float, vertex_type<Graph>>(number_elements_in_block);
     void *result_ptr = input.get_caching_builder()(max_block_size);
+
+    if (!number_elements_in_block) {
+        homogen_table empty_table;
```
orrrrtem, line 101: Why do you need this?

cpp/oneapi/dal/algo/jaccard/detail/vertex_similarity_ops.hpp
```diff
@@ -96,6 +96,13 @@ vertex_similarity_result vertex_similarity_ops_dispatcher<Policy, Float, Method,
     const std::int64_t max_block_size =
         get_max_block_size<Float, vertex_type<Graph>>(number_elements_in_block);
     void *result_ptr = input.get_caching_builder()(max_block_size);
+
+    if (!number_elements_in_block) {
```
orrrrtem, line 100: Bad if. Make it an explicit equation.

cpp/oneapi/dal/algo/jaccard/detail/vertex_similarity_ops.hpp
```diff
@@ -96,6 +96,12 @@ vertex_similarity_result vertex_similarity_ops_dispatcher<Policy, Float, Method,
     const std::int64_t max_block_size =
         get_max_block_size<Float, vertex_type<Graph>>(number_elements_in_block);
     void *result_ptr = input.get_caching_builder()(max_block_size);
+
+    if (number_elements_in_block == 0) {
+        vertex_similarity_result res(homogen_table(), homogen_table(), 0);
```
michael-smirnov, line 101:
```suggestion
        vertex_similarity_result res(table(), table(), 0);
```
You can even stick just to the table, if you do not have a requirement that the result should always contain homogen tables.

michael-smirnov: But the better way I think is to have a default constructor for `vertex_similarity_result`

orrrrtem (commented):
What is behavior on the user side? If user will request get_coeffs?

YakhtinLeonid:
An empty table will be returned from get_coeffs. it can be printed without fails

orrrrtem:
> An empty table will be returned from get_coeffs. it can be printed without fails

Ok.  Do you measure performance differentiation after the fix?

orrrrtem:
Update Branch.

YakhtinLeonid:
> > An empty table will be returned from get_coeffs. it can be printed without fails
> 
> Ok. Do you measure performance differentiation after the fix?

The performance is almost the same (there is a small speed up, no more than 0.5%)

orrrrtem (approved):

PR #1560, opened 2021-04-10
Remove bad_variant_access

PetrovKP:

PR #1565, opened 2021-04-13
ONEDAL_ASSERT fixes in array_impl

michael-smirnov:


rlnx:
Implementation before #1560 was fine, but recent fixes are terrible. I suggest to fix it as part of this PR.

It is mixing low-level and high-level concepts. It results in non-readable code, it's difficult to follow the logic if I always need to switch my attention to low-level things. Many resources talk about that:
- https://dzone.com/articles/clean-code-dont-mix-different
- https://www.c-sharpcorner.com/article/clean-code-single-level-of-abstraction/
- http://principles-wiki.net/principles:single_level_of_abstraction

---------------------------------------------------------

If the goal is to replace `std::get` function if it's not supported by some OS, then introduce an equivalent:
```c++
cshared get_cshared() const {
   if (const auto ptr = std::get_if<cshared>(&data_owned_)) {
      return *ptr;
   }
   else {
      throw internal_error {};
   }
}

shared get_shared() const {
   if (const auto ptr = std::get_if<shared>(&data_owned_)) {
      return *ptr;
   }
   else {
      throw internal_error {};
   }
}

// Or similar template function
```

Usage:
```c++
template <typename Y>
void reset(const array_impl<Y>& ref, T* data, std::int64_t count) {
   if (ref.has_mutable_data()) {
      data_owned_ = shared { ref.get_shared(), data };
   }
   else {
      data_owned_ = cshared { ref.get_cshared(), data };
  }
}
```

PetrovKP:
/intelci: run
set daal4py_branch=dev/onedal4py
set daal4py_repo=git@github.com:PetrovKP/daal4py.git
set infra_branch=dev/onedal_tests

cpp/oneapi/dal/detail/array_impl.hpp
```diff
@@ -236,7 +214,15 @@ class array_impl : public base {
 #endif
     }
 
-    detail::shared<T> copy() {
+    template <typename Y>
+    void check_array_has_ownership_structure(const array_impl<Y>& ref) {
+        if (ref.get_if_cshared() == nullptr && ref.get_if_shared() == nullptr) {
```
rlnx, line 219: This will never happen, `variant<T, U>`  always contains either instance of `T` or `U`. There are some exceptions, see https://en.cppreference.com/w/cpp/utility/variant/valueless_by_exception

However, the described situation happen only if `T` or `U` raises exception during changing the `variant` state. This will never happen for `shared_ptr`, because all copy/move constructors/assignment are noexcept.

So this check is useless.

rlnx: I'd recommend put assertion `ONEDAL_ASSERT(!data_owned_.valueless_by_exception())`

KulikovNikita (approved):

PetrovKP (approved):

cpp/oneapi/dal/detail/array_impl.hpp
```diff
@@ -162,44 +161,25 @@ class array_impl : public base {
...
+            data_owned_ = shared(ref.get_cshared(), data);
         }
-
         count_ = count;
         reset_policy(ref);
     }
 
     template <typename Y>
     void reset(const array_impl<Y>& ref, const T* data, std::int64_t count) {
+        ONEDAL_ASSERT(!ref.data_owned_.valueless_by_exception());
```
rlnx, line 177: This should be moved to `get_shared`/`get_cshared`

PR #1566, opened 2021-04-13
Refactor internal interface of tables & accessors

rlnx:
<table style="width:100%">
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
</table>


cpp/oneapi/dal/backend/common.hpp
```diff
@@ -98,66 +98,107 @@ inline constexpr Integer up_pow2(Integer x) {
...
+        auto optional_queue = queue_like.get_queue();
+        if (optional_queue) {
+            return reference.get_context() == optional_queue->get_context();
+        }
     }
-}
 
-inline bool is_same_device(const sycl::queue& q1, const sycl::queue& q2) {
-    return q1.get_device() == q2.get_device();
+    return true;
```
michael-smirnov, line 151: Why it is true if `QueueLike` does not contain a queue?

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -98,66 +98,107 @@ inline constexpr Integer up_pow2(Integer x) {
...
-inline bool is_same_device(const sycl::queue& q1, const sycl::queue& q2, const sycl::queue& q3) {
-    return is_same_device(q1, q2) && is_same_device(q1, q3);
-}
+template <typename QueueLike>
+inline bool is_same_device_impl(const sycl::queue& reference, QueueLike&& queue_like) {
+    using queue_like_t = std::decay_t<QueueLike>;
+    constexpr bool is_queue = std::is_same_v<queue_like_t, sycl::queue>;
+    constexpr bool is_opt_queue = std::is_same_v<queue_like_t, std::optional<sycl::queue>>;
+    constexpr bool has_get_queue = has_get_queue_v<queue_like_t>;
+    constexpr bool has_get_optional_queue = has_get_optional_queue_v<queue_like_t>;
```
michael-smirnov, line 160: I think we can create a special type-trait structure for queue-like objects, which will contain all these constants - is it valuable?

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -98,66 +98,107 @@ inline constexpr Integer up_pow2(Integer x) {
...
+        return reference.get_device() == queue_like.get_queue().get_device();
     }
+    else if constexpr (has_get_optional_queue) {
+        auto optional_queue = queue_like.get_queue();
+        if (optional_queue) {
+            return reference.get_device() == optional_queue->get_device();
+        }
+    }
+
+    ONEDAL_ASSERT(!"Unreachable");
```
michael-smirnov, line 183: Its reachable if optionals do not contain a queue

cpp/oneapi/dal/detail/error_messages.cpp
```diff
@@ -69,6 +69,17 @@ MSG(pushing_rows_is_not_supported, "Pushing rows is not supported")
...
+MSG(table_builder_does_not_provide_read_access_to_rows,
+    "Given table builder does not provide read access to rows")
+MSG(table_builder_does_not_provide_write_access_to_rows,
+    "Given table builder does not provide write access to rows")
+MSG(table_does_not_provide_read_access_to_columns,
+    "Given table does not provide read access to columns")
+MSG(table_builder_does_not_provide_read_access_to_columns,
+    "Given table builder does not provide read access to columns")
+MSG(table_builder_does_not_provide_write_access_to_columns,
+    "Given table builder does not provide write access to columns")
```
michael-smirnov, line 82: Is it really needed to distinguish which object accepted by an accessor does not implement the contracts? I think "Given object does not provide ..." shall be enough

cpp/oneapi/dal/table/backend/homogen_kernels.hpp
```diff
@@ -0,0 +1,167 @@
...
+#ifdef ONEDAL_DATA_PARALLEL
+    const auto opt_queue = array.get_queue();
+    if (opt_queue.has_value()) {
+        const auto queue = opt_queue.value();
+        const auto array_alloc = sycl::get_pointer_type(array.get_data(), queue.get_context());
+        return alloc_kind_from_sycl(array_alloc);
+    }
+#endif
+    return alloc_kind::host;
+}
```
michael-smirnov, line 88: I thought it was somewhere in common functionality

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
```diff
@@ -132,118 +58,87 @@ class homogen_table_impl {
...
 
-    template <typename Data>
-    void push_column(const array<Data>& block, std::int64_t column_index, const range& rows) {
-        push_column_impl(detail::default_host_policy{}, block, column_index, rows);
+    template <typename T>
+    void pull_column(const detail::default_host_policy& policy,
+                     array<T>& block,
+                     std::int64_t column_index,
+                     const range& rows) const {
+        homogen_pull_column(policy, get_info(), data_, block, column_index, rows, alloc_kind::host);
```
michael-smirnov, line 101: What the reason was to extract implementations of pull/push as free functions?

cpp/oneapi/dal/table/common.cpp
```diff
@@ -16,24 +16,30 @@
 
 #include "oneapi/dal/table/common.hpp"
 #include "oneapi/dal/exceptions.hpp"
-#include "oneapi/dal/table/backend/empty_table_impl.hpp"
+// #include "oneapi/dal/table/backend/empty_table_impl.hpp"
```
michael-smirnov, line 19: delete

michael-smirnov (commented):
First look

cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
rlnx: To separate homogen table and homogen table builder implementations. Builder was using the `push` methods implementations from homogen table, however `push`  is not required to satisfy homogen table contract.

cpp/oneapi/dal/table/detail/table_iface.hpp
```diff
@@ -0,0 +1,159 @@
...
+using homogen_table_builder_template = table_builder_template<homogen_table_builder_iface, Derived>;
+
+} // namespace v1
+
+using v1::table_iface;
+using v1::homogen_table_iface;
+using v1::table_builder_iface;
+using v1::homogen_table_builder_iface;
+using v1::dense_table_template;
+using v1::homogen_table_template;
```
agorshk, line 151: some name of classes are repeated

cpp/oneapi/dal/table/backend/homogen_kernels.hpp
rlnx: From one hand it's general, from the other hand, is used only in homogen table impl.  Not sure if it's good way of handling differnt type of allocations to spread it to the other parts of the lib.

rlnx:
/intelci: run
set infra_branch=dev/risrafil-refactor-tables

rlnx:
/intelci: restart
set infra_branch=dev/risrafil-refactor-tables

cpp/oneapi/dal/table/backend/homogen_kernels.cpp
```diff
@@ -0,0 +1,590 @@
...
+            array.reset(policy.get_queue(), count, alloc);
+        }
+    }
+#else
+    array.reset(count);
+#endif
+}
+
+template <typename Policy, typename Data>
+ONEDAL_FORCEINLINE void make_mutable_data(const Policy& policy, array<Data>& array) {
```
michael-smirnov, line 135: Can be removed as far as it does not depend on the policy anymore

cpp/oneapi/dal/table/backend/homogen_table_builder_impl.hpp
```diff
@@ -35,8 +37,8 @@ class homogen_table_builder_impl {
         dtype_ = data_type::float32;
     }
 
-    void reset(homogen_table&& t) {
-        if (t.has_data()) {
+    void move(detail::homogen_table_iface& t) override {
+        if (t.get_row_count() > 0 || t.get_column_count() > 0) {
```
michael-smirnov, line 41:
```suggestion
        if (t.get_row_count() > 0 && t.get_column_count() > 0) {
```
maybe you meant this?

cpp/oneapi/dal/table/backend/homogen_table_builder_impl.hpp
```diff
@@ -35,8 +37,8 @@ class homogen_table_builder_impl {
         dtype_ = data_type::float32;
     }
 
-    void reset(homogen_table&& t) {
-        if (t.has_data()) {
+    void move(detail::homogen_table_iface& t) override {
```
michael-smirnov, line 40: It is strange to me that `move` accepts lvalue ref

cpp/oneapi/dal/table/column_accessor.hpp
```diff
@@ -16,40 +16,61 @@
...
+            throw invalid_argument{ msg::object_does_not_provide_read_access_to_columns() };
+        }
+
+        if (!is_readonly && !push_iface_) {
+            using msg = detail::error_messages;
+            throw invalid_argument{ msg::object_does_not_provide_write_access_to_columns() };
+        }
+    }
+
+    array<data_t> pull(std::int64_t column_index, const range& row_range = { 0, -1 }) const {
```
michael-smirnov, line 63: doxygen comments?

cpp/oneapi/dal/table/column_accessor.hpp
```diff
@@ -95,39 +109,55 @@ class column_accessor : private detail::accessor_base<T, detail::column_values_b
...
+
+private:
+    static T* get_block_data(const array<data_t>& block) {
+        if constexpr (is_readonly) {
+            return block.get_data();
+        }
+        return block.get_mutable_data();
+    }
+
+    std::shared_ptr<detail::pull_column_iface> pull_iface_;
```
michael-smirnov, line 159: `detail::shared` for consistency?

cpp/oneapi/dal/table/detail/access_iface.hpp
```diff
@@ -21,81 +21,275 @@
...
+using v1::push_rows_iface;
+using v1::push_rows_template;
+using v1::push_column_iface;
+using v1::push_column_template;
+using v1::get_pull_rows_iface;
+using v1::get_pull_column_iface;
+using v1::get_push_rows_iface;
+using v1::get_push_column_iface;
 
 } // namespace oneapi::dal::detail
```
michael-smirnov, line 295: Maybe to split this file into several files that contain just a single interface declaration?

cpp/oneapi/dal/table/detail/access_iface.hpp
```diff
@@ -21,81 +21,275 @@
...
+    DECLARE_PULL_ROWS_DPC(double)
+    DECLARE_PULL_ROWS_DPC(std::int32_t)
+#endif
 };
 
-struct column_values_block {
-    std::int64_t column_index;
-    range rows;
+template <typename Derived>
+class pull_rows_template : public base, public pull_rows_iface {
```
michael-smirnov, line 131: 1. It is not clear how it depends on `Derived`. Probably it is better to pass it as a parameter to defines `DEFINE_TEMPLATE_PULL_ROWS_*`?
2. Why do we really need to have such `*_template` helpers for pull/push and tables and cannot just inherit our implementations from interfaces directly? Is it recommended to create such a template for every new access iface or table?

cpp/oneapi/dal/table/detail/table_iface.hpp
```diff
@@ -0,0 +1,157 @@
...
+        return this;
+    }
+
+    push_column_iface* get_push_column_iface() override {
+        return this;
+    }
+};
+
+template <typename Derived>
+using dense_table_template = table_template<table_iface, Derived>;
```
michael-smirnov, line 133: Why it is named dense? Per my understanding, even sparse tables can provide access to rows/columns in a dense way

cpp/oneapi/dal/table/detail/access_iface.hpp
rlnx: > 1. It is not clear how it depends on Derived. Probably it is better to pass it as a parameter to defines DEFINE_TEMPLATE_PULL_ROWS_*?

ok

----------------------------------------
> 2. Why do we really need to have such *_template helpers for pull/push and tables and cannot just inherit our implementations from interfaces directly? Is it recommended to create such a template for every new access iface or table?

The reason is multiple pull/push overloads that accepts float/double/int. If I inherit from interface directly, I'll need to explicitly define all those overloads. `*_template` helpers allows using template functions by means of CRTP magic.

**CRTP**
```c++
class my_table : public pull_rows_template<my_table> {
public:
   template <typename Policy, typename T>
   void pull(const Policy& policy, array<T>& block) { ... }
};
``` 

**"Direct"**
```c++
class my_table : public pull_rows_iface {
public:
   void pull(const default_host_policy& policy, array<float>& block) override { ... }
   void pull(const default_host_policy& policy, array<double>& block) override { ... }
   void pull(const default_host_policy& policy, array<std::int32_t>& block) override { ... }
   void pull(const data_parallel_policy& policy, array<float>& block) override { ... }
   void pull(const data_parallel_policy& policy, array<double>& block) override { ... }
   void pull(const data_parallel_policy& policy, array<std::int32_t>& block) override { ... }

protected:
   template <typename Policy, typename T>
   void pull_impl(...) { ... }
};
``` 

I'll need to repeat this in each table impl

cpp/oneapi/dal/table/backend/homogen_table_builder_impl.hpp
rlnx: Hmm, `std::move` accepts lvalue (but not limited by lvalue)

cpp/oneapi/dal/table/detail/table_iface.hpp
rlnx: I'll remove aliases for `table_template` and will use it directly

rlnx:
/intelci: restart
set infra_branch=dev/risrafil-refactor-tables

rlnx:
/intelci: run
set infra_branch=dev/risrafil-refactor-tables

PetrovKP (approved):

cpp/oneapi/dal/table/backend/homogen_table_builder_impl.hpp
michael-smirnov: But the goal of `std::move` is different. I'm trying to say this name confuses me because it actually does not move the data of `t` inside the builder. It refers to the data of `t` if it is not empty. I'd expect that `move()` method will make `t` empty. Probably, `reset` of `refer` is a better name. What were the reasons to change the signature of that function?

cpp/oneapi/dal/table/detail/table_iface.hpp
```diff
@@ -0,0 +1,145 @@
...
+                       public pull_rows_template<Derived>,
+                       public pull_column_template<Derived> {
+public:
+    pull_rows_iface* get_pull_rows_iface() override {
+        return this;
+    }
+
+    pull_column_iface* get_pull_column_iface() override {
+        return this;
+    }
```
michael-smirnov, line 109: I suggest moving these methods to `pull_rows_iface` and `pull_column_iface` respectively

cpp/oneapi/dal/table/detail/access_iface.hpp
michael-smirnov: Thank you

cpp/oneapi/dal/table/detail/table_iface.hpp
rlnx: This is strange suggestion, because these methods are part of `table_iface` contract. Table Implements corresponding ifaces if result of those methods is not `nullptr`.

PR #1570, opened 2021-04-15
Binary SVM Labels Conversion Fix

owerbat:


cpp/oneapi/dal/algo/svm/backend/cpu/train_cls_kernel.cpp
```diff
@@ -168,7 +168,17 @@ static result_t call_binary_daal_kernel(const context_cpu& ctx,
...
+    daal::data_management::NumericTablePtr daal_labels;
+    if ((unique_label.first == Float(0.0) && unique_label.second == Float(1.0)) ||
+        (unique_label.first == Float(1.0) && unique_label.second == Float(0.0)) ||
+        (unique_label.first == Float(-1.0) && unique_label.second == Float(1.0)) ||
+        (unique_label.first == Float(1.0) && unique_label.second == Float(-1.0))) {
+        daal_labels = interop::convert_to_daal_table<Float>(labels);
+    }
+    else {
+        daal_labels = interop::convert_to_daal_homogen_table(arr_new_label, row_count, 1);
+    }
```
rlnx, line 181: Create separate function, e.g., `convert_labels`

cpp/oneapi/dal/algo/svm/backend/cpu/train_cls_kernel.cpp
```diff
@@ -131,6 +131,26 @@ static result_t call_multiclass_daal_kernel(const context_cpu& ctx,
         .set_support_indices(table_support_indices);
 }
 
+template <typename Float>
+inline auto convert_to_daal_labels(const table& labels,
```
michael-smirnov, line 135: I think it is better not to create one more function but to fix `convert_labels()` function. You can modify it the way it works:
1. Determine the values used to represent binary labels
2. If these values are in the set {(0,1), (1,0), (-1,1), (1,1)}, then use `convert_to_daal_table` function
3. Otherwise, use `convert_to_daal_homogen_table`

This way is more transparent, and do not require always to create an additional array of `arr_new_label`

michael-smirnov (changes requested):

cpp/oneapi/dal/algo/svm/backend/cpu/train_cls_kernel.cpp
```diff
@@ -167,8 +167,8 @@ static result_t call_binary_daal_kernel(const context_cpu& ctx,
 
     auto arr_label = row_accessor<const Float>{ labels }.pull();
     binary_label_t<Float> unique_label;
-    auto arr_new_label = convert_labels(arr_label, { Float(-1.0), Float(1.0) }, unique_label);
-    const auto daal_labels = interop::convert_to_daal_homogen_table(arr_new_label, row_count, 1);
+    auto daal_labels =
+        convert_labels(labels, arr_label, { Float(-1.0), Float(1.0) }, unique_label, row_count);
```
michael-smirnov, line 171: Why do you need to pass both `labels` and `arr_label`? You can pass just `labels` and made `arr_label` local inside that function

cpp/oneapi/dal/algo/svm/backend/utils.hpp
```diff
@@ -43,105 +45,142 @@ struct binary_label_t {
 };
 
 template <typename Float>
-inline array<Float> convert_labels(const array<Float>& arr_label,
-                                   const binary_label_t<Float>& in_binary_labels,
-                                   binary_label_t<Float>& out_binary_labels) {
+inline daal::data_management::NumericTablePtr convert_labels(
```
michael-smirnov, line 48: This function violates SRP now. I'd like to to make it simpler, for example (just to illustrate):

```cpp
inline dal::table convert_binary_labels(const dal::table& labels, 
                                        const binary_label_t<Float>& requested_unique_labels, 
                                        binary_label_t<Float>& previous_unique_labels) {
    previous_unique_labels = get_unique_labels(labels);
    if (previous_unique_labels == binary_label_t<Float>{0, 1} ||
        previous_unique_labels == binary_label_t<Float>{-1, 1}) {
        return labels;
    } else {
        return assign_new_labels(labels, previous_unique_labels, requested_unique_labels); 
    }
}
```
And then implement the functions from next level of abstraction. In SVM, you can use it this way:
``` cpp
  auto new_labels = convert_binary_labels(labels, {-1, 1}, input_unique_labels);
  auto daal_new_labels = convert_to_daal_table<Float>(new_labels);
```

michael-smirnov (changes requested):

cpp/oneapi/dal/algo/svm/backend/cpu/train_cls_kernel.cpp
```diff
@@ -165,10 +164,10 @@ static result_t call_binary_daal_kernel(const context_cpu& ctx,
     daal_svm_parameter.cacheSize = cache_byte;
     daal_svm_parameter.svmType = daal_svm::training::internal::SvmType::classification;
 
-    auto arr_label = row_accessor<const Float>{ labels }.pull();
     binary_label_t<Float> unique_label;
-    auto arr_new_label = convert_labels(arr_label, { Float(-1.0), Float(1.0) }, unique_label);
-    const auto daal_labels = interop::convert_to_daal_homogen_table(arr_new_label, row_count, 1);
+    const auto new_labels =
+        convert_binary_labels(labels, { Float(-1.0), Float(1.0) }, unique_label);
```
michael-smirnov, line 169: `unique_label` is not a clear name in this context, because the previous parameter does also correspond to unique labels.
Maybe `old_unique_labels`?

cpp/oneapi/dal/algo/svm/backend/utils.hpp
```diff
@@ -19,10 +19,12 @@
 #include "oneapi/dal/table/row_accessor.hpp"
 #include "oneapi/dal/detail/error_messages.hpp"
 #include "oneapi/dal/backend/transfer.hpp"
+#include "oneapi/dal/backend/interop/table_conversion.hpp"
 
 namespace oneapi::dal::svm::backend {
 
 namespace daal_svm = daal::algorithms::svm;
+namespace interop = dal::backend::interop;
```
michael-smirnov, line 27: please do not use namespace aliases in headers - at least you can do it in a function or class scope

cpp/oneapi/dal/algo/svm/backend/utils.hpp
```diff
@@ -40,108 +42,151 @@ template <typename Float>
...
-                error_msg::input_labels_contain_wrong_unique_values_count_expect_two()
-            };
+            throw invalid_argument(dal::detail::error_messages::
+                                       input_labels_contain_wrong_unique_values_count_expect_two());
         }
     }
 
-    out_binary_labels.first = value_first_class_label;
-    out_binary_labels.second = value_second_class_label;
+    return homogen_table::wrap(queue, new_label_data, count, 1);
```
michael-smirnov, line 174: Please add `ONEDAL_ASSERT` at the start of that function that checks `label` column count is equal to `1`

cpp/oneapi/dal/algo/svm/backend/utils.hpp
```diff
@@ -40,108 +42,151 @@ template <typename Float>
...
+}
 
-    return dal::backend::to_device_sync(queue, new_label_arr);
+template <typename Float>
+inline table convert_binary_labels(sycl::queue& queue,
+                                   const table& labels,
+                                   const binary_label_t<Float>& requested_unique_labels,
+                                   binary_label_t<Float>& unique_labels) {
+    unique_labels = get_unique_labels<Float>(queue, labels);
+    if (unique_labels == binary_label_t<Float>{ 0, 1 } ||
```
michael-smirnov, line 183: This if can be moved into `get_new_labels`, which can be renamed to `convert_binary_labels` while this function can be removed. We can require users to call `get_unique_labels` explicitly before the call of that function.

michael-smirnov (approved):

PR #1579, opened 2021-04-21
Fix down pow2 function

vlad-nazarov:


cpp/oneapi/dal/backend/common.hpp
```diff
@@ -71,11 +71,16 @@ template <typename Integer>
 inline constexpr Integer down_pow2(Integer x) {
     static_assert(std::is_integral_v<Integer>);
     ONEDAL_ASSERT(x > 0);
-    Integer power = 1;
-    while (power < x / 2) {
-        power *= 2;
+    Integer power = 0;
+    if (!(x & (x - 1))) {
```
michael-smirnov, line 75: Why you handle cases `x==0` and `x==1` when this function has an assert at line 73 that prohibits calling it with `x==0`?

rlnx: This is check for power of two, if `x` is already power of 2, do nothing 😅

rlnx: Suggest to have separate function for that, maybe someone needs it

rlnx:
```c++
inline constexpr bool(...)
```

vlad-nazarov: Added check for power of two

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -71,11 +81,15 @@ template <typename Integer>
 inline constexpr Integer down_pow2(Integer x) {
     static_assert(std::is_integral_v<Integer>);
     ONEDAL_ASSERT(x > 0);
-    Integer power = 1;
-    while (power < x / 2) {
-        power *= 2;
+    Integer power = 0;
+    if (is_pow2<Integer>(x))
+        return x;
```
rlnx, line 86:
```suggestion
    if (is_pow2<Integer>(x)) {
        return x;
    }
```

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -71,11 +81,16 @@ template <typename Integer>
 inline constexpr Integer down_pow2(Integer x) {
     static_assert(std::is_integral_v<Integer>);
     ONEDAL_ASSERT(x > 0);
-    Integer power = 1;
-    while (power < x / 2) {
-        power *= 2;
+    Integer power = 0;
+    if (is_pow2<Integer>(x)) {
```
michael-smirnov, line 85:
```suggestion
    if (is_pow2(x)) {
```

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -71,11 +81,16 @@ template <typename Integer>
...
+        return x;
     }
-    return power;
+
+    while (x > 1) {
+        x >>= 1;
+        power++;
+    }
+    return 1 << power;
```
michael-smirnov, line 93: Assert on overflow before returning?

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -63,6 +63,16 @@ inline constexpr Integer up_multiple(Integer x, Integer multiple) {
     return y + z;
 }
 
+/// Checks if 'x' is power of 2.
+/// Example: down_pow2(10) == false
+/// Example: down_pow2(16) == true
+template <typename Integer>
+inline constexpr bool is_pow2(Integer x) {
+    static_assert(std::is_integral_v<Integer>);
+    ONEDAL_ASSERT(x > 0);
```
michael-smirnov, line 72: Probably it's better to return `false` if `x <= 0` in this function rather than putting an assert

cpp/oneapi/dal/backend/common.hpp
vlad-nazarov: Added overflow check

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -71,11 +94,17 @@ template <typename Integer>
...
+    if (is_pow2(x)) {
+        return x;
     }
-    return power;
+
+    while (x > 1) {
+        x >>= 1;
+        power++;
+    }
+    ONEDAL_ASSERT(is_safe_bit_shift(power));
```
michael-smirnov, line 106: I'd rather make it like this:

```cpp
ONEDAL_ASSERT(power <= get_magnitude_bit_count<Integer>())
```

where `get_magnitude_bit_count()` is 

```cpp
template <typename Integer>
inline constexpr std::int64_t get_magnitude_bit_count() {
   static_assert(std::is_integral_v<Integer>);
   constexpr std::int64_t bit_count = sizeof(Integer) * 8;
   return std::is_signed_v<Integer> ? bit_count - 1 : bit_count; 
}
```

michael-smirnov (approved):
Assuming green CI

KulikovNikita (approved):

PR #1583, opened 2021-04-23
Fix homogen builder reset with rvalue ref

rlnx:
- Update `homogen_table_builder::reset` as discussed with @michael-smirnov 
- Change pointer to `array` in `table_iface`


cpp/oneapi/dal/table/backend/homogen_table_impl.hpp
```diff
@@ -71,8 +71,8 @@ class homogen_table_impl
         return meta_;
     }
 
-    const void* get_data() const override {
-        return data_.get_data();
+    array<byte_t> get_data() const override {
```
michael-smirnov, line 74: You already have `get_data_array()` on 86 line. Probably, it is needed to be removed or propagated to the base interface like `get_data()`.

rlnx: Removed

KulikovNikita (approved):

PR #1605, opened 2021-05-11
Fix passed build daal4py when there are failed

PetrovKP:
# Description

-
-
-


michael-smirnov (approved):
Don't we need to make similar changes in the daal4py repo for its own public procedures?

PetrovKP:
@michael-smirnov there it is built through a conda-build at the moment, so there are no such problems.

PR #1610, opened 2021-05-14
Explicit instantiation for integer_overflow_ops on size_t

michael-smirnov:
This is a fix of debug build on MacOS

CI (debug build): http://intel-ci.intel.com/ebb6ecd0-7382-f1d5-8066-8cdcd4b723a1

### DIFF
--- a/cpp/oneapi/dal/detail/common.cpp
+++ b/cpp/oneapi/dal/detail/common.cpp
@@ -51,7 +51,10 @@ template struct ONEDAL_EXPORT integer_overflow_ops<std::uint8_t>;
 template struct ONEDAL_EXPORT integer_overflow_ops<std::uint16_t>;
 template struct ONEDAL_EXPORT integer_overflow_ops<std::uint32_t>;
 template struct ONEDAL_EXPORT integer_overflow_ops<std::uint64_t>;
+
+#if defined(__APPLE__)
 template struct ONEDAL_EXPORT integer_overflow_ops<std::size_t>;
+#endif
 
 } // namespace v1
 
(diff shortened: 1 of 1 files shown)


rlnx:
`size_t` is a typically alias for some integer type that might be covered by `std::intXX_t`, what is `size_t` in case of MacOS?

michael-smirnov:
@rlnx it's implementation-defined what the `size_t` is according to the [standard](https://en.cppreference.com/w/cpp/types/size_t).
On MacOS, it's defined in stddef.(/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/12.0.0/include/stddef.h):

```cpp
#if defined(__need_size_t)
#if !defined(_SIZE_T) || __has_feature(modules)
/* Always define size_t when modules are available. */
#if !__has_feature(modules)
#define _SIZE_T
#endif
typedef __SIZE_TYPE__ size_t;
#endif
#undef __need_size_t
#endif /*defined(__need_size_t) */
```

cpp/oneapi/dal/detail/common.cpp
```diff
@@ -52,6 +52,10 @@ template struct ONEDAL_EXPORT integer_overflow_ops<std::uint16_t>;
 template struct ONEDAL_EXPORT integer_overflow_ops<std::uint32_t>;
 template struct ONEDAL_EXPORT integer_overflow_ops<std::uint64_t>;
 
+#if defined(__APPLE__)
+template struct ONEDAL_EXPORT integer_overflow_ops<std::size_t>;
+#endif
+
```
rlnx, line 58: Why should we define it for v1? This is preserved only for backward compatibility, all further modifications must be done in the latest vX

rlnx (approved):

PR #1612, opened 2021-05-17
Add --test_disable_fp64 option

rlnx:
- Add flag to disable double precision tests in Bazel

### DIFF
--- a/cpp/oneapi/dal/test/engine/common_dpc.cpp
+++ b/cpp/oneapi/dal/test/engine/common_dpc.cpp
@@ -37,22 +37,26 @@ static bool check_if_env_knob_is_enabled(const char* env_var) {
     }
 }
 
-static bool check_if_env_overrides_fp64_settings() {
+[[maybe_unused]] static bool check_if_env_overrides_fp64_settings() {
     return check_if_env_knob_is_enabled("OverrideDefaultFP64Settings");
 }
 
-static bool check_if_env_forces_dp_emulation() {
+[[maybe_unused]] static bool check_if_env_forces_dp_emulation() {
     return check_if_env_knob_is_enabled("IGC_EnableDPEmulation") ||
            check_if_env_knob_is_enabled("IGC_ForceDPEmulation");
 }
 
 bool device_test_policy::has_native_float64() const {
+#ifdef ONEDAL_DISABLE_FP64_TESTS
+    return false;
+#else
     const auto device = queue_.get_device();
     const auto fp_config = device.get_info<sycl::info::device::double_fp_config>();
     const bool float64_support = !fp_config.empty();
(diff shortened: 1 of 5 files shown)


dev/bazel/dal.bzl
```diff
@@ -474,7 +474,12 @@ def _dal_module(name, lib_tag="dal", is_dpc=False, features=[],
         local_defines = local_defines + ([
             "DAAL_SYCL_INTERFACE",
             "ONEDAL_DATA_PARALLEL"
-        ] if is_dpc else []),
+        ] if is_dpc else []) + select({
+            "@config//:test_fp64_enabled": [],
```
amgrigoriev, line 478: Looks a bit confusing: name of config is "test_fp64_enabled" but inside "ONEDAL_DISABLE_FP64_TESTS=1"

rlnx: See the next line. `select` with `//conditions:default` in bazel is equivalent to `if ... else ...`:
```bzl
select({ 
   "@config//:test_fp64_enabled": [ ],   # if
   "//conditions:default": [ 
      "ONEDAL_DISABLE_FP64_TESTS=1",     # else
   ],
})
```

rlnx: However, I agree this construction is confusing

cpp/oneapi/dal/test/engine/common_dpc.cpp
```diff
@@ -37,22 +37,26 @@ static bool check_if_env_knob_is_enabled(const char* env_var) {
...
-static bool check_if_env_forces_dp_emulation() {
+[[maybe_unused]] static bool check_if_env_forces_dp_emulation() {
     return check_if_env_knob_is_enabled("IGC_EnableDPEmulation") ||
            check_if_env_knob_is_enabled("IGC_ForceDPEmulation");
 }
 
 bool device_test_policy::has_native_float64() const {
+#ifdef ONEDAL_DISABLE_FP64_TESTS
+    return false;
+#else
```
michael-smirnov, line 52: Why this is embedded to `has_native_float64()` function instead of creating another one like `is_float64_enabled()`?

rlnx: Because this function is already used in almost every test.

amgrigoriev (approved):

cpp/oneapi/dal/test/engine/common_dpc.cpp
michael-smirnov: Yes, but we can easily replace `SKIP_IF(this->not_float64_friendly());` by `SKIP_IF(this->not_float64_friendly() || !this->is_float64_enabled());`. I think that having a separate function in this case is more accurate, because float64 native support and test configs are different things. But this is a minor comment.

PR #1614, opened 2021-05-18
Proposal for optional results

rlnx:
Optional results based on external constants.

### DIFF
--- a/cpp/oneapi/dal/algo/knn/common.cpp
+++ b/cpp/oneapi/dal/algo/knn/common.cpp
@@ -19,6 +19,12 @@
 #include "oneapi/dal/exceptions.hpp"
 
 namespace oneapi::dal::knn {
+
+namespace optional_results {
+const optional_result_id indices{ 0 };
+const optional_result_id distances{ 1 };
+} // namespace optional_results
+
 namespace detail {
 namespace v1 {
 
@@ -27,6 +33,7 @@ class descriptor_impl : public base {
 public:
     std::int64_t class_count = 2;
     std::int64_t neighbor_count = 1;
+    optional_result_id optional_results;
 };
 
 template <typename Task>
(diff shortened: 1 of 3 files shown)


cpp/oneapi/dal/algo/knn/common.hpp
```diff
@@ -56,6 +57,48 @@ using v1::by_default;
 
 } // namespace method
 
+class optional_result_id {
```
rlnx, line 60: This should go to the common part

cpp/oneapi/dal/algo/knn/common.hpp
```diff
@@ -56,6 +57,48 @@ using v1::by_default;
...
+    explicit optional_result_id(std::int64_t result_index) {
+        mask_.set(result_index);
+    }
+
+    const bitset_t& get_mask() const {
+        return mask_;
+    }
+
+private:
+    bitset_t mask_;
```
averbukh, line 78: Does pimpl make sense for backward compatibility?

cpp/oneapi/dal/algo/knn/common.hpp
```diff
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <bitset>
```
orrrrtem, line 19: Should it be included somewhere higher like detail common with creation of corresponding alias like it is done for shared ptr?

cpp/oneapi/dal/algo/knn/common.hpp
```diff
@@ -89,10 +132,12 @@ class descriptor_base : public base {
 
     std::int64_t get_class_count() const;
     std::int64_t get_neighbor_count() const;
+    optional_result_id get_optional_results() const;
```
orrrrtem, line 135: Any perf improvements if return as a const  ref?

cpp/oneapi/dal/algo/knn/common.cpp
```diff
@@ -19,6 +19,12 @@
 #include "oneapi/dal/exceptions.hpp"
 
 namespace oneapi::dal::knn {
+
+namespace optional_results {
```
orrrrtem, line 23: May few algorithms have a shared part of the kernel, like some common primitive in which the flow and result depend on optional_results. I am not sure about ML but in graphs, it can be a traversal case. Should we take some agreement on the order of the options?

orrrrtem: Assuming these options are each individual for each algorithm.

PR #1638, opened 2021-05-26
Fix Implicit ALS build on VS compiler

Alexsandruss:
# Description
Change operators in implicit_als code to avoid fail of build on VS compiler


Alexsandruss:
http://intel-ci.intel.com/ebbe0da8-4e7c-f11d-905e-8cdcd4b723a1

Pahandrovich (approved):

PR #1645, opened 2021-05-27
Fix backward compatibility in data management

rlnx:
- Introduce v2 for array
- Keep old `accessor_provider_iface` 
- Add adapters from new accessor ifaces to the old one
- Move some methods to cpp to prevent BC problems in the future


cpp/oneapi/dal/detail/array_impl.hpp
```diff
@@ -317,29 +327,194 @@ class array_impl : public base {
...
+    std::optional<data_parallel_policy> dp_policy_;
+#endif
+};
+
+} // namespace v2
+
+namespace v1 {
+
+template <typename T>
+class array_impl : public base {
```
michael-smirnov, line 346: Why you place the compatibility interface of `array` in `array_compat.hpp` but the compatibility interface of `array_impl` you keep here? I think it is better to group them in one file or in one directory

cpp/oneapi/dal/detail/error_messages.cpp
```diff
@@ -82,6 +82,8 @@ MSG(object_does_not_provide_read_access_to_columns,
     "Given object does not provide read access to columns")
 MSG(object_does_not_provide_write_access_to_columns,
     "Given object does not provide write access to columns")
+MSG(object_does_not_provide_access_to_rows_or_columns,
```
michael-smirnov, line 85: Place it in some "compatibility" section?

cpp/oneapi/dal/detail/array_impl.hpp
rlnx: It's not better from the user's point of view. The person is going to read code in the headers might be confused by the v1 and v2.

cpp/oneapi/dal/detail/error_messages.cpp
rlnx: I don't see strong reason. Other error messages are used in the compatibility functionality too. There is no obvious rule that decides whether an error message should be moved to this section.

michael-smirnov (approved):

PetrovKP (approved):

PR #1649, opened 2021-05-28
Some minor fixes in code

KulikovNikita:


cpp/oneapi/dal/table/backend/common_kernels.hpp
```diff
@@ -70,8 +70,8 @@ inline bool alloc_kind_requires_copy(alloc_kind src_alloc_kind, alloc_kind dst_a
 #else
     ONEDAL_ASSERT(src_alloc_kind == alloc_kind::host);
     ONEDAL_ASSERT(dst_alloc_kind == alloc_kind::host);
-    return false;
 #endif
+    return false;
```
agorshk, line 74: Does it better to throw exaptation with error like "error_msg::unsupported_alloc_kind()" or so (in line 68, and revert return in previous place)? We have similar situation with alloc_kind_from_sycl(...) or alloc_kind_to_sycl(...) and throw an exaptation.

PR #1665, opened 2021-06-01
hardcode l1, l2 and ll cache sizes

PivovarA:
We found that the getL1cache, getL2cache, and getLLcache methods cause sporadic crashes of the decisionForest algorithm on CentOs. 
It was decided to hardcode these values ​​and add their computation later.

### DIFF
--- a/cpp/daal/src/services/service_topo.cpp
+++ b/cpp/daal/src/services/service_topo.cpp
@@ -1923,29 +1923,21 @@ unsigned _internal_daal_GetStatus()
 
 unsigned getL1CacheSize()
 {
-    _internal_daal_GetSysProcessorCoreCount();
-
-    if (glbl_obj.error || !glbl_obj.maxCacheSubleaf) return 0;
-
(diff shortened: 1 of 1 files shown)


PetrovKP:
http://intel-ci.intel.com/ebc2d22f-a476-f16a-83e6-8cdcd4b723a1

PetrovKP:
perf: http://intel-ci.intel.com/ebc2e324-cb4d-f17b-a170-8cdcd4b723a1

cpp/daal/src/services/service_topo.cpp
```diff
@@ -1923,29 +1923,21 @@ unsigned _internal_daal_GetStatus()
 
 unsigned getL1CacheSize()
 {
-    _internal_daal_GetSysProcessorCoreCount();
```
SmirnovEgorRu, line 1926: Why not to delete this function at all?
To be sure, that it's not called in other places as well

SmirnovEgorRu (approved):

PR #1681, opened 2021-06-07
migrated to using float class hist, instead of double, for float prediction

makart19:
migrated to using float class hist, instead of double, for float prediction

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/oneapi/df_classification_predict_dense_oneapi_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/oneapi/df_classification_predict_dense_oneapi_impl.i
@@ -257,10 +257,12 @@ services::Status PredictKernelOneAPI<algorithmFPType, method>::predictByAllTrees
     DAAL_CHECK_STATUS_VAR(status);
 
     UniversalBuffer probasArr;
+    TArray<algorithmFPType, sse2> probasArrHost(treeBlockSize * _nClasses);
+    DAAL_CHECK_MALLOC(probasArrHost.get());
 
     if (_votingMethod == VotingMethod::weighted && pModel->getProbas(0))
     {
-        probasArr = context.allocate(TypeIds::id<double>(), treeBlockSize * _nClasses, status);
+        probasArr = context.allocate(TypeIds::id<algorithmFPType>(), treeBlockSize * _nClasses, status);
         DAAL_CHECK_STATUS_VAR(status);
         weighted = true;
     }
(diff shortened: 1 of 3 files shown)


amgrigoriev (approved):

cpp/daal/src/algorithms/dtrees/forest/classification/oneapi/df_classification_predict_dense_oneapi_impl.i
```diff
@@ -286,11 +288,32 @@ services::Status PredictKernelOneAPI<algorithmFPType, method>::predictByAllTrees
...
+            {
+                algorithmFPType * dst_ptr = probasArrHost.get() + iTree * maxTreeSize * _nClasses;
+
+                PRAGMA_IVDEP
+                PRAGMA_VECTOR_ALWAYS
+                for (size_t i = 0; i < treeSize * _nClasses; i++)
+                {
+                    dst_ptr[i] = static_cast<algorithmFPType>(probas[i]);
+                }
+            }
```
rlnx, line 307: I believe we can leave only one branch here:
```c++
algorithmFPType * dst_ptr = probasArrHost.get() + iTree * maxTreeSize * _nClasses;                
PRAGMA_IVDEP                
PRAGMA_VECTOR_ALWAYS                
for (size_t i = 0; i &lt; treeSize * _nClasses; i++)                
{                    
   dst_ptr[i] = static_cast&lt;algorithmFPType&gt;(probas[i]);                
}
```

cpp/daal/src/algorithms/dtrees/forest/classification/oneapi/df_classification_predict_dense_oneapi_impl.i
```diff
@@ -286,11 +288,32 @@ services::Status PredictKernelOneAPI<algorithmFPType, method>::predictByAllTrees
...
         }
     }
 
+    if (weighted)
+    {
+        context.copy(probasArr, 0, (void *)probasArrHost.get(), treeBlockSize * _nClasses, 0, treeBlockSize * _nClasses, status);
+        DAAL_CHECK_STATUS_VAR(status);
+    }
```
rlnx, line 315: This branch is not needed, the same condition is above, they can be combined into one

cpp/daal/src/algorithms/dtrees/forest/classification/oneapi/df_classification_predict_dense_oneapi_impl.i
makart19: Yes, it can be combined here. I wanted to use optimized copy function, but I checked that it provides the same perf as second variant, therefore used second one for both cases.

cpp/daal/src/algorithms/dtrees/forest/classification/oneapi/df_classification_predict_dense_oneapi_impl.i
makart19: This branch is needed to mitigate perf degradation caused by call of context.copy for each tree. Here it is performed only once for tree block, which results in perf improvement up to 20% for hepmass/double case for example.

rlnx: Ok

rlnx (approved):

PR #1691, opened 2021-06-08
Python oneDAL interface: pybind11

michael-smirnov:


python/oneapi/dal/common/backend/serialization_py.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include "oneapi/dal/detail/archives.hpp"
+
+namespace oneapi::dal::backend {
+
+template <typename T>
+pybind11::bytes serialize(const T& original) {
+    detail::binary_output_archive archive;
+    detail::serialize(original, archive);
+    const auto data = archive.to_array();
+    return { reinterpret_cast<const char*>(data.get_data()), archive.get_size() };
```
rlnx, line 31: Dangling pointer here! `data` will be destroyed once we return from function

python/onedal/common/serialization.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <pybind11/pybind11.h>
+
+#include "oneapi/dal/detail/serialization.hpp"
+#include "oneapi/dal/detail/archives.hpp"
+
+namespace oneapi::dal::python {
+
+template <typename T>
+pybind11::bytes serialize(const T& original) {
```
rlnx, line 27: Was scrolling through and noticed common mistake, there is no `inline` specifier for many header-defined functions. This may lead to duplicated symbols, when we add more translation units.

PR #1696, opened 2021-06-08
Add security md file

PetrovKP:


outoftardis:
note that we have a [section about security in readme](https://github.com/oneapi-src/oneDAL#security-), if you are adding this file, it'd make sense to reference it from that section

SECURITY.md
```diff
@@ -0,0 +1,12 @@
+# Security Policy
+
+## Report a Vulnerability
+
+Please report security issues or vulnerabilities to the [Intel® Security Center].
+
+For more information on how Intel® works to resolve security issues, see
```
outoftardis, line 7:
```suggestion
For more information on how Intel® resolves security issues, see
```

PetrovKP: I'm not sure if we can change the content here. @maria-Petrova It's true?

outoftardis: if it's a wording provided by legal, then it's ok to leave as it is

maria-Petrova (approved):

SECURITY.md
maria-Petrova: Yes, the wording was provided by legal, so please leave it the way it is.

PR #1698, opened 2021-06-08
Edge count fix

orrrrtem:
Fix of invalid setting of an edge count. 
Previously, edges count was equal to edges count in the unfiltered graph.


bysheva (approved):

orrrrtem:
@Mergifyio backport rls/2021.3-rls

PR #1710, opened 2021-06-09
Fix for kNN distance calculation

KalyanovD:
# Description

### DIFF
--- a/cpp/daal/src/algorithms/service_kernel_math.h
+++ b/cpp/daal/src/algorithms/service_kernel_math.h
@@ -398,10 +398,7 @@ public:
         return services::Status();
     }
 
-    virtual services::Status finalize(const size_t n, FPType * a)
-    {
-        return services::Status();
-    }
+    virtual services::Status finalize(const size_t n, FPType * a) { return services::Status(); }
 
 protected:
     FPType computeDistance(const FPType * x, const FPType * y, const size_t n)
--- a/cpp/daal/src/algorithms/service_kernel_math.h
+++ b/cpp/daal/src/algorithms/service_kernel_math.h
@@ -104,7 +104,7 @@ public:
     virtual services::Status computeBatch(const FPType * const a, const FPType * const b, size_t aOffset, size_t aSize, size_t bOffset, size_t bSize,
                                           FPType * const res) = 0;
 
-    virtual services::Status getRootBatch(const size_t n, FPType * in, FPType * out) = 0;
+    virtual services::Status finalize(const size_t n, FPType * a) = 0;
 };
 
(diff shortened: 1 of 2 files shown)


cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -378,6 +403,12 @@ class ChebyshevDistances : public PairwiseDistances<FPType, cpu>
         return services::Status();
     }
 
+    virtual services::Status getRootBatch(const size_t n, FPType * in, FPType * out)
+    {
+        out = in;
```
SmirnovEgorRu, line 408: It does nothing outside the function

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -171,6 +173,14 @@ class EuclideanDistances : public PairwiseDistances<FPType, cpu>
         return services::Status();
     }
 
+    virtual services::Status getRootBatch(const size_t n, FPType * in, FPType * out)
```
SmirnovEgorRu, line 176: better to align naming across distances
we have init(), we can call this as finalize() or so on

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -294,6 +304,21 @@ class MinkowskiDistances : public PairwiseDistances<FPType, cpu>
...
+    {
+        daal::internal::mkl::MklMath<FPType, cpu> math;
+
+        if (_p == 1.0)
+        {
+            out = in;
```
SmirnovEgorRu, line 313: The same

SmirnovEgorRu (approved):

cpp/daal/src/algorithms/service_kernel_math.h
SmirnovEgorRu: It's even a bug if ```in``` and ```out``` are different arrays.

cpp/daal/src/algorithms/service_kernel_math.h
KalyanovD: applied, thanks for advise

cpp/daal/src/algorithms/service_kernel_math.h
KalyanovD: set of arguments was changed as far as it's not necessary to have in and out arrays separate.

KalyanovD:
perf is ok: http://intel-ci.intel.com/ebc9ddbb-1172-f12e-9f24-8cdcd4b723a1

PetrovKP (approved):

PR #1734, opened 2021-06-17
Fix building with custom instruction set

PetrovKP:
### Problem
When we building of oneDAL with custom IS (Instruction Set), for example:
```bash
make onedal REQCPU=avx2
```
Kernels implementation use the selected custom IS, only on the machine that has the top cpu extension (_kernels will run with avx2 only on Haswell or Broadwell IA_), in other cases, kernels work only with sse2 code, even on SKX, CLX, etc IA.

This is because we have a similar code with the dispatcher
```c++
cpuid = avx512
switch (cpuid):
// case avx512: ... (not compiler code for custom build with avx2)
   case avx2: ... // <-avx2 kernel
   default: ...   // <-sse2 kernel

```

### Solution
Use minimal between custom IS (from make) and top cpu extension from run on IA

### DIFF
--- a/cpp/daal/include/services/internal/daal_kernel_defines.h
+++ b/cpp/daal/include/services/internal/daal_kernel_defines.h
@@ -52,7 +52,8 @@ case cpuType:                                                                   
     break;                                                                                      \
 }
 
-#define DAAL_KERNEL_BUILD_MAX_INSTRACTION_SET_ID                daal::sse2
+#undef DAAL_KERNEL_BUILD_MAX_INSTRUCTION_SET_ID
+#define DAAL_KERNEL_BUILD_MAX_INSTRUCTION_SET_ID                daal::sse2
 #define DAAL_KERNEL_SSE2_ONLY(something)                        , something
 #define DAAL_KERNEL_SSE2_ONLY_CODE(...)                         __VA_ARGS__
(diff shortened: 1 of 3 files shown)


cpp/daal/include/services/internal/daal_kernel_defines.h
```diff
@@ -142,6 +155,8 @@ case cpuType:
 #endif
 
 #if defined(DAAL_KERNEL_AVX512)
+    #undef DAAL_KERNEL_BUILD_MAX_INSTRACTION_SET_ID
```
masdevas, line 158: Please fix an errata: "instrAction" -> "instrUction". Here and in other places.

cpp/daal/include/services/internal/daal_kernel_defines.h
```diff
@@ -50,13 +52,16 @@ case cpuType:
     break;                                                                                      \
 }
 
+#define DAAL_KERNEL_BUILD_MAX_INSTRACTION_SET_ID                daal::sse2
```
masdevas, line 55: Probably we should add 
```
#undef DAAL_KERNEL_BUILD_MAX_INSTRACTION_SET_ID
```
above of this line to eliminate required order of ISA extensions block's order in this file.

rlnx (approved):

PR #1740, opened 2021-06-22
fixed check for lev zero module creating

makart19:


cpp/daal/include/services/internal/sycl/level_zero_module_sycl.h
```diff
@@ -68,8 +68,14 @@ class ZeModule : public Base
...
     {
-        const auto ptr = new ZeModule(deviceQueue, binarySize, pBinary, status);
-        if (!ptr) status |= ErrorMemoryAllocationFailed;
+        auto ptr = new ZeModule(deviceQueue, binarySize, pBinary, status);
+        if (!status)
+        {
+            if (ptr) delete ptr;
+            ptr = nullptr;
+        }
+        else if (!ptr)
```
amgrigoriev, line 77: Maybe it makes sense to check "ptr" before "status"? Just in order to ensure that ErrorMemoryAllocationFailed is added in any case.

amgrigoriev (approved):

cpp/daal/include/services/internal/sycl/level_zero_module_sycl.h
makart19: It is more probable to get incorrect status than nullptr and provides more info, therefore i set it first.

PR #1735, opened 2021-06-18
kNN model serialization

KalyanovD:
# Description
kNN serialization has been added similar to SVM, test has been added.

PR #1766, opened 2021-07-06
PCA model serialization

michael-smirnov:


cpp/oneapi/dal/algo/pca/test/serialization.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+        INFO("serialization");
+        const auto deserialized_model = te::serialize_deserialize(model);
+
+        INFO("inference");
+        const auto expected = run_inference(model);
+        const auto actual = run_inference(deserialized_model);
+        compare_infer_results(actual, expected);
+    }
+};
```
rlnx, line 91: I see a lot of almost the same code across serialization tests, see SVM, k-NN (https://github.com/oneapi-src/oneDAL/pull/1735). As a next step I'd propose to think about common fixtures for serialization tests.

rlnx (approved):

PR #1768, opened 2021-07-07
Subgraph isomorphism: error handling when target graph is smaller than pattern

orazve:
Check + test


orrrrtem (approved):

orrrrtem:
Probably, one more test should be added: sizes are equal.

orazve:
> Probably, one more test should be added: sizes are equal.

Equal sizes of graphs are correct case. It is not a bad argument.

orazve:
http://intel-ci.intel.com/ebdf3270-cc55-f105-a2f7-8cdcd4b723a1

PR #1784, opened 2021-07-27
Introduce communicator interface

rlnx:
- [x] Add `--enable_assert` flag to Bazel
- [x] Pass `-fsycl-device-code-split=per_kernel` in Bazel build scripts
- [x] Fixes in tests for `double` types to be able to run without fp64 emulation
- [x] Introduce low-level communicator interface
- [x] Introduce high-level routines based on low-level communicator
- [x] Mock communicator based on threads in test engine
- [x] Tests for high-level routines


cpp/oneapi/dal/detail/communicator.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+std::int64_t spmd_communicator::get_default_root_rank() const {
+    return impl_->get_default_root_rank();
+}
+
+void spmd_communicator::barrier() const {
+    return impl_->barrier();
+}
+
+spmd_request spmd_communicator::bcast(byte_t* send_buf,
```
makart19, line 56: why don't use simple void* instead of byte_t*?

cpp/oneapi/dal/detail/communicator.cpp
```diff
@@ -0,0 +1,172 @@
+/*******************************************************************************
+* Copyright 2020-2021 Intel Corporation
```
makart19, line 2: 2020-2021 -> 2021

cpp/oneapi/dal/detail/communicator.cpp
rlnx: `void*` is discouraged in modern C++. The interface assumes memory buffer, so it accepts array of bytes as the most convenient low-level representation. 

BTW, you will not be required to cast typed buffers to `byte_t*`, there are template versions of these function in hpp.

makart19: ok

cpp/oneapi/dal/test/communicator.cpp
```diff
@@ -0,0 +1,283 @@
...
+            const std::int64_t row_count = uniform(rng_);
+            const std::int64_t column_count = uniform(rng_);
+            tables.push_back(create_random_table(row_count, column_count));
+        }
+
+        return tables;
+    }
+
+private:
+    static constexpr int seed = 7777;
```
amgrigoriev, line 137: There are 777, 7777 and 777777 seeds ...

cpp/oneapi/dal/detail/communicator.hpp
```diff
@@ -0,0 +1,741 @@
...
+    ///                   significant only at `root`
+    /// @param displs     Entry $i$ specifies the displacement relative to
+    ///                   `recv_buf` at which to place the incoming data
+    ///                   from process $i$, must contain at least `rank_count`
+    ///                   elements, significant only at `root`
+    /// @param dtype      The type of elements in the passed buffers
+    /// @param root       The rank of the receiving process
+    ///
+    /// @return The object to track the progress of the operation
+    spmd_request gatherv(const byte_t* send_buf,
```
amgrigoriev, line 452: could we add a simplified gatherv later? I mean with recv_counts array but without explicit displacements (calculated from recv_counts

cpp/oneapi/dal/detail/communicator.hpp
```diff
@@ -0,0 +1,741 @@
...
+                         std::int64_t send_count,
+                         byte_t* recv_buf,
+                         const std::int64_t* recv_count,
```
amgrigoriev, line 455: Looks like a pointer to a single value but it's an array. Maybe recv_counts - it looks consistent with oneCCL naming.

cpp/oneapi/dal/detail/communicator.hpp
```diff
@@ -0,0 +1,741 @@
...
+        ONEDAL_ASSERT(send.get_count() > 0);
+        ONEDAL_ASSERT(recv.has_mutable_data());
+
+        spmd_request request;
+
+        __ONEDAL_IF_QUEUE__(send.get_queue(), {
+            auto q = send.get_queue().value();
+
+            ONEDAL_ASSERT(recv.get_queue().has_value());
+            ONEDAL_ASSERT(recv.get_queue().value() == q);
```
amgrigoriev, line 587: could it be another queue sharing the same context?

cpp/oneapi/dal/detail/communicator.hpp
```diff
@@ -0,0 +1,741 @@
...
+    ///                 received from each rank
+    /// @param dtype    The type of elements in the passed buffers
+    /// @param op       The reduction operation
+    ///
+    /// @return The object to track the progress of the operation
+    spmd_request allreduce(const byte_t* send_buf,
+                           byte_t* recv_buf,
+                           std::int64_t count,
+                           const data_type& dtype,
+                           const spmd_reduce_op& op = spmd_reduce_op::sum) const;
```
amgrigoriev, line 621: Is it necessary to have a default value here? Explicit value looks more informative.

amgrigoriev (approved):

cpp/oneapi/dal/detail/communicator.hpp
rlnx: No problem it can be added without losing BC

cpp/oneapi/dal/detail/communicator.hpp
rlnx: Will fix

cpp/oneapi/dal/detail/communicator.hpp
rlnx: Ok, will check contexts

cpp/oneapi/dal/detail/communicator.hpp
rlnx: It's matter of convention. To me having summation as default behavior is reasonable -- in most algorithms summation is required.

cpp/oneapi/dal/test/communicator.cpp
rlnx: What do you suggest? It could be separate refactoring task of replacing all seeds to predefined constant.

cpp/oneapi/dal/detail/communicator.hpp
rlnx: Done c0cb44a

cpp/oneapi/dal/detail/communicator.hpp
rlnx: Done, 9a95578

PR #1819, opened 2021-08-12
Fix of Decision Forest model serialization

Alexsandruss:
# Description
Since 2020.0 release decision forest serialized model grew 10x times because of probabilities tables addition. Collection of these tables has nTrees items of shape nClasses rows x nNodes columns, each item is Homogen table with dictionary property `featuresEqual = notEqual`. On serialization, numeric table dictionary saves every Feature (i.e. nNodes Features), which has 5 properties (each stored in 64 bytes block because of DataArchive implementation). In result, rough memory usage of serialized prob. tables collection: `nTrees x nNodes x 5 x 64 bytes +  nTrees x nClasses x nNodes x sizeof(double)`. 
Fix: set `featuresEqual = equal` to serialize Feature once for each table
Deserialization from previous releases is saved and checked.


agorshk (approved):
Looks good, let's wait to CI check

PR #1826, opened 2021-08-13
Introduce MPI communicator

rlnx:
- [x] Create `spmd_communicator_via_host_impl` base class for communicators
- [x] Add MPI dependencies to Bazel
- [x] Create `mpi_communicator`
- [x] Create tests for `mpi_communicator`
- [x] Create infrastructure in Bazel for writing MPI integration tests
- [ ] Create MPI integration tests for K-Means


dev/bazel/deps/mpi.bzl
```diff
@@ -0,0 +1,104 @@
...
+        "libfabric/lib/libfabric.so",
+        "libfabric/lib/libfabric.so.1",
+        "libfabric/lib/prov/libefa-fi.so",
+        "libfabric/lib/prov/libmlx-fi.so",
+        "libfabric/lib/prov/libpsmx2-fi.so",
+        "libfabric/lib/prov/librxm-fi.so",
+        "libfabric/lib/prov/libshm-fi.so",
+        "libfabric/lib/prov/libsockets-fi.so",
+        "libfabric/lib/prov/libtcp-fi.so",
+        "libfabric/lib/prov/libverbs-fi.so",
```
KulikovNikita, line 37: Can it automatized?

dev/bazel/deps/mpi.bzl
```diff
@@ -0,0 +1,104 @@
...
+    fi_dir = fi_files[0].dirname
+    for fi in fi_files:
+        if fi.dirname != fi_dir:
+            fail("All fabric interface files must reside in the same directory")
+    return fi_dir
+
+def _generate_exec_wrapper(ctx, actions, executable, fi_dir):
+    exec_wrapper = actions.declare_file(executable.basename + ".sh")
+    content = (
+        "#!/bin/bash\n" +
```
KulikovNikita, line 67: Can we use smth like Jinja (or any other templating engine)?

rlnx: No, Jinja is Python, there is no Python deps here. It looks like a python, but not Python

dev/bazel/deps/mpi.bzl
rlnx: I think yes, but requires engineering in `prebuilt_libs_repo_rule`

cpp/oneapi/dal/test/mpi_communicator.cpp
```diff
@@ -0,0 +1,27 @@
...
+* limitations under the License.
+*******************************************************************************/
+
+#include "oneapi/dal/test/engine/mpi.hpp"
+
+namespace oneapi::dal::test {
+
+namespace te = dal::test::engine;
+
+TEST("bcast") {
```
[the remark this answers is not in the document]
rlnx: Should be updated

cpp/oneapi/dal/detail/communicator.cpp
```diff
@@ -18,10 +18,283 @@
...
 namespace oneapi::dal::detail::v1 {
 
+#ifdef ONEDAL_DATA_PARALLEL
+static void check_if_pointer_matches_queue(const sycl::queue& q, const void* ptr) {
+    if (ptr) {
+        const auto alloc_kind = sycl::get_pointer_type(ptr, q.get_context());
+        if (alloc_kind == sycl::usm::alloc::unknown) {
+            throw invalid_argument{ error_messages::unknown_usm_pointer_type() };
+        }
+    }
```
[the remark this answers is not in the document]
rlnx: `alloc_kind` of `nullptr` is always `unknown`, besides for some functions, `nullptr` is valid input. These cases are handled before a call to that function

dev/bazel/deps/mpi.bzl
rlnx: I was wrong, this is easy to fix, will do

napetrov:
@amgrigoriev  - guess we don't need this PR and it can be deleted?

PR #1831, opened 2021-08-17
Train_test_split indices fix

Alexsandruss:
# Description
oneDAL train_test_split produces indices differed from sklearn because of insufficient amount of generated random numbers. This bug causes CI fails for cases where `n` is small or/and far from powers of 2.
Solution: increase RNG numbers from 150% of `n` to 200% of `n` + 32.

### DIFF
--- a/cpp/daal/src/data_management/train_test_split.cpp
+++ b/cpp/daal/src/data_management/train_test_split.cpp
@@ -95,9 +95,9 @@ services::Status generateShuffledIndicesImpl(const NumericTablePtr & idxTable, c
     const size_t nThreads  = threader_get_threads_number();
     const size_t n         = idxTable->getNumberOfRows();
     const size_t stateSize = rngStateTable->getNumberOfRows();
-    // number of generated uints: 1.5x of n for reserve
-    DAAL_OVERFLOW_CHECK_BY_MULTIPLICATION(size_t, n / 2lu, 3lu);
-    const size_t nRandomUInts = n / 2lu * 3lu;
+    // number of generated uints: 2 x n + 32 for reserve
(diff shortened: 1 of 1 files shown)


cpp/daal/src/data_management/train_test_split.cpp
```diff
@@ -95,9 +95,9 @@ services::Status generateShuffledIndicesImpl(const NumericTablePtr & idxTable, c
     const size_t nThreads  = threader_get_threads_number();
     const size_t n         = idxTable->getNumberOfRows();
     const size_t stateSize = rngStateTable->getNumberOfRows();
-    // number of generated uints: 1.5x of n for reserve
-    DAAL_OVERFLOW_CHECK_BY_MULTIPLICATION(size_t, n / 2lu, 3lu);
-    const size_t nRandomUInts = n / 2lu * 3lu;
+    // number of generated uints: 2 x n + 32 for reserve
+    DAAL_OVERFLOW_CHECK_BY_MULTIPLICATION(size_t, n + 16, 2);
+    const size_t nRandomUInts = 2 * n + 32;
```
agorshk, line 100: Am I right that in Sklearn approximately the same amount of random number were generated? Do we have understanding how much the performance will change (might be better to save current number of random number for big "n" and change for small)?

Alexsandruss: sklearn generates random numbers on-demand instead of used preemptive approach which is suitable for parallel generation. Increase of `n` is necessary since this bug is appearing on mostly all n_samples with different frequency

agorshk: Ok, if the problem is common for different n_samples (small/big), then it's reasonable

agorshk (approved):
Please check CI before merge

PR #1865, opened 2021-09-02
Dev/amk df reduce fix

makart19:
Fixed issue in reduce partial histograms kernel


makart19:
/intelci: run
set infra_branch=dev/alekseik/2021_4/fpk/gpu

PR #1898, opened 2021-09-17
OneCCL communicator + tests & sample

amgrigoriev:


makart19 (approved):

samples/oneapi/dpc/oneccl/sources/kmeans_distr_oneccl.cpp
```diff
@@ -0,0 +1,123 @@
...
+  const std::int64_t block_size_regular = row_count / split_count;
+  const std::int64_t block_size_tail = row_count % split_count;
+
+  std::vector<dal::table> result(split_count);
+
+  std::int64_t row_offset = 0;
+  for (std::int64_t i = 0; i < split_count; i++) {
+    const std::int64_t tail =
+        std::int64_t(i + 1 == split_count) * block_size_tail;
+    const std::int64_t block_size = block_size_regular + tail;
```
makart19, line 53: it would be better to think out an approach for spreading tail across all workers, instead of assigning it to the last one

amgrigoriev: Modified

PR #1948, opened 2021-10-27
Extended comments

KulikovNikita:


napetrov:
@KulikovNikita  - think it worth merging this still? There are couple comments that i think to look at, but having this in code brings more value even if it's not ideal.

PR #1960, opened 2021-10-29
Fix SVM model memory leak

Alexsandruss:
# Description
Move from table resizing to table creation due to memory leak.
Solves https://github.com/intel/scikit-learn-intelex/issues/843.


Alexsandruss:
Reason is confirmed to be on python side.

PR #1966, opened 2021-11-02
Remove unexpected dependency

amgrigoriev:


napetrov (approved):
Assuming CI would pass

PR #1983, opened 2021-11-11
added ccl samples for bs and df

makart19:
- added ccl samples for bs and df
- made formatting for corresponding mpi samples

PR #1984, opened 2021-11-11
added ccl samples for bs and df

makart19:
- added ccl samples for bs and df
- made formatting for corresponding mpi samples


Alexandr-Solovev (commented):
Please, resolve comments from https://github.com/oneapi-src/oneDAL/pull/1983

Alexandr-Solovev (approved):

PR #1998, opened 2021-11-18
Change public mklfpk version

Pahandrovich:


Alexsandruss:
What stops us from updating Linux and Mac versions too for alignment between systems?

Pahandrovich:
For lnx and mac we are already using the current versions. For win there was an update due to the move to MD/MDd

Alexsandruss (approved):

PR #2010, opened 2021-12-01
Replacing Intel clang with dpcpp in makefiles

homksei:
This PR contains: 

- Replacing clang (Intel clang) driver with dpcpp in the makefiles
- Enabling library building with icx

### DIFF
--- a/examples/daal/cpp_sycl/makefile_lnx
+++ b/examples/daal/cpp_sycl/makefile_lnx
@@ -24,8 +24,8 @@ help:
 	@echo
 	@echo "name              - example name. Please see daal.lst file."
 	@echo
-	@echo "compiler_name     - the only supported compiler is clang (default value)"
-	@echo "                    Intel(R) oneAPI DPC++ compiler"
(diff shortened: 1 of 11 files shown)


examples/daal/cpp_sycl/makefile_lnx
```diff
@@ -104,7 +104,7 @@ LOPTS := -Wl,--start-group -Wl,$(DAAL_LIB),$(EXT_LIB) $(DAAL_PATH)/libonedal_syc
 RES_DIR=_results/$(compiler)_$(_IA)_$(threading)_$(RES_EXT)
 RES = $(addprefix $(RES_DIR)/, $(if $(filter run, $(mode)), $(addsuffix .res ,$(example)), $(addsuffix .exe,$(example))))
 
-ifeq ($(compiler),clang)
```
napetrov, line 107: Let's ensure that clang still will work - we shouldn't break make compiler=clang case at least for now. Then later we can deprecate it.

vmeshche (approved):

PR #2014, opened 2021-12-07
Update oneapi random forest oob

vlad-nazarov:


makart19 (approved):

KulikovNikita:
@Alexsandruss - you fixed something related to oob score just a few days ago. Can you please check if this PR is somewhat important in this regard or no?

icfaust:
This specifically modifies some files which I am not sure are actually used.  I'm going to look into it.

I believe that the oneDAL implementation may supersede the DAAL defined TrainBatchKernelOneAPI versions.

icfaust:
This code change should be accepted. However, this only impacts running DAAL not on CPUs. We should set a clear standard about future support.

icfaust (approved):
This updates DAAL GPU code with proper checks to OOB calculations. Technically more correct, and follows the parameters available in decision_forest_training_parameter.h . Doesn't impact DAAL CPU operation.

icfaust:
Change no longer necessary.

PR #2028, opened 2021-12-21
Fix relocate empty clusters

lordoz234:
Fix the problem due to the fact that empty clusters were not removed from previous clusters. Add new array to store in which cluster is the point.

### DIFF
--- a/cpp/daal/src/algorithms/kmeans/kmeans_lloyd_batch_impl.i
+++ b/cpp/daal/src/algorithms/kmeans/kmeans_lloyd_batch_impl.i
@@ -170,7 +170,6 @@ Status KMeansBatchKernel<method, algorithmFPType, cpu>::compute(const NumericTab
 
         algorithmFPType newCentersGoalFunc = (algorithmFPType)0.0;
         algorithmFPType l2Norm             = (algorithmFPType)0.0;
-
         {
             DAAL_ITTNOTIFY_SCOPED_TASK(kmeansMergeReduceCentroids);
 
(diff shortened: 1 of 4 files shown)


cpp/daal/src/algorithms/kmeans/kmeans_lloyd_impl.i
```diff
@@ -381,16 +391,30 @@ int TaskKMeansLloyd<algorithmFPType, cpu>::kmeansUpdateCluster(int jidx, centroi
     return s0;
 }
 
+template <typename algorithmFPType, CpuType cpu>
+template <typename centroidsFPType>
+int TaskKMeansLloyd<algorithmFPType, cpu>::kmeansUpdatePoints(int jidx)
+{
+    int idx = (int)jidx;
```
Alexsandruss, line 398: why `(int)` is needed?

cpp/daal/src/algorithms/kmeans/kmeans_lloyd_batch_impl.i
```diff
@@ -174,37 +175,49 @@ Status KMeansBatchKernel<method, algorithmFPType, cpu>::compute(const NumericTab
 
             for (size_t i = 0; i < nClusters; i++)
             {
-                if (clusterS0[i] > 0)
-                {
-                    const algorithmFPType coeff = 1.0 / clusterS0[i];
+                if (clusterS0[i] == 0) {
```
Alexsandruss, line 178: `{` should be on separate row. Use `clang-format -i {files}` to apply code style:
conda install -c conda-forge clang-tools

cpp/daal/src/algorithms/kmeans/kmeans_lloyd_impl.i
lordoz234: Fixed

cpp/daal/src/algorithms/kmeans/kmeans_lloyd_batch_impl.i
lordoz234: Fixed

PR #2066, opened 2022-02-03
Fix csr computeDataSize in ovo multiclass

vlad-nazarov:


Vika-F (approved):
This solution is correct, but it is possible to allocate less memory in CSR case.
Suppose we have 3 classes and with following sizes of sub-tables:
classNonZeroValuesCount[] = {500, 400, 300}
classLabelsCount[] = {100, 200, 300}
sum of classNonZeroValuesCount[] and classLabelsCount[] = {600, 600, 600}

The algorithm implemented in this PR will end up with dataSize = 900 and nSubsetVectors = 500.
And the total storage allocated for the CSR sub-table would be 900 * sizeof(algorithmFPType) + (900 + 500 + 1) * sizeof(size_t) = 900 * sizeof(algorithFPType) + 1401 * sizeof(size_t)

At the same time to store _columnIndicesX and _rowOffsetsX of subTaskCSR we need maximum 1201 * sizeof(size_t) for any two classes in the dataset.

So, the algorithm's memory requirements can be improved.

Pahandrovich (approved):

PR #2067, opened 2022-02-03
Chebyshev KNN distance

KulikovNikita:


cpp/oneapi/dal/backend/primitives/search/test/chebyshev_search_dpc.cpp
```diff
@@ -0,0 +1,267 @@
...
+                    gtv = std::max(gtv, d);
+                }
+
+                distances_ptr[j * m + i] = gtv;
+            }
+        }
+        return de::homogen_table_builder{}.reset(distances_arr, n, m).build();
+    }
+
+    static auto argsort(const table& distances) {
```
amgrigoriev, line 222: The same name for table object and static function - it is a bit confusing

amgrigoriev, line 222: The same name for table object and static function - it is a bit confusing

amgrigoriev (approved):

PR #2068, opened 2022-02-04
SPMD communicator allgatherv method to support arbitrary displacements

VadimVoevodkin:


VadimVoevodkin:
intelci: run

VadimVoevodkin:
intelci: run

cpp/oneapi/dal/detail/ccl/communicator.hpp
```diff
@@ -318,9 +318,15 @@ class ccl_communicator_impl : public ccl_interface_selector<MemoryAccessKind>::t
         ONEDAL_ASSERT(send_buf);
         ONEDAL_ASSERT(recv_buf);
 
+        std::int64_t data_type_size = get_data_type_size(dtype);
```
amgrigoriev, line 321:
```suggestion
        const std::int64_t data_type_size = get_data_type_size(dtype);
```

amgrigoriev (changes requested):
Please add mpi/ccl tests for arbitrary displacements

cpp/oneapi/dal/test/ccl_communicator.cpp
```diff
@@ -241,6 +241,62 @@ TEST_M(ccl_comm_test, "allgatherv") {
...
+    auto comm = get_new_comm();
+    const std::int64_t granularity = 10;
+    const std::int64_t rank_count = comm.get_rank_count();
+    const std::int64_t rank = comm.get_rank();
+
+    std::vector<std::int64_t> recv_counts(rank_count);
+    std::vector<std::int64_t> displs(rank_count);
+    std::int64_t total_size = 0;
+    for (std::int64_t i = 0; i < rank_count; i++) {
+        recv_counts[i] = (i + 2) * granularity;
```
amgrigoriev, line 254: The idea is not to increase receive count for every rank but to add a gap between results obtained from different ranks, Please also add a constant for gap width. Initial array should be filled by -1 (or any other values that are different against collected values) and values inside gaps should remain unchanged after allgatherv call.

cpp/oneapi/dal/test/ccl_communicator.cpp
```diff
@@ -241,6 +241,73 @@ TEST_M(ccl_comm_test, "allgatherv") {
...
 }
 
+TEST_M(ccl_comm_test, "allgatherv_arbitrary_displacements") {
+    auto comm = get_new_comm();
+    const std::int64_t granularity = 10;
+    const std::int64_t rank_count = comm.get_rank_count();
+    const std::int64_t rank = comm.get_rank();
+    const std::int64_t arb_displc = 10;
+
+    std::vector<std::int64_t> recv_counts_final_buf(rank_count);
```
amgrigoriev, line 251: Please use shorter name: recv_counts, for instance.

PR #2074, opened 2022-02-14
Graph reader: tab-delimited data processing

Plastova:


alex-radio (approved):
It was found that with a tab-delimited (i.e. with 0x0A) edge-list is not read correctly due to difference between daal_string_to_int and expected as if stroll function behavior. With this corrections files are read correctly.

PR #2076, opened 2022-02-16
Optimize col major -> row major table transpose on host -> device conversion case

vlad-nazarov:
This optimization needed for sklearnex GPU algorithms. Block transpose on host much faster than row-by-row transpose although more RAM is required. In GPU SVM and Knn sklearnex algorithms speedup more than 5 times. This PR will remove the performance gap between the daal4py and sklearnex interfaces. 
Future plans are also to optimize device -> host transposition in the same way and device -> device transposition using MKL functionality.

### DIFF
--- a/cpp/oneapi/dal/table/backend/convert.hpp
+++ b/cpp/oneapi/dal/table/backend/convert.hpp
@@ -37,6 +37,18 @@ void convert_vector(const detail::default_host_policy& policy,
                     std::int64_t dst_stride,
                     std::int64_t element_count);
 
+void convert_matrix(const detail::default_host_policy& policy,
+                    const void* src,
+                    void* dst,
+                    data_type src_type,
+                    data_type dst_type,
+                    const std::int64_t src_row_stride,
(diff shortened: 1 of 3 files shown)


KalyanovD (approved):

cpp/oneapi/dal/table/backend/convert.hpp
```diff
@@ -55,6 +67,18 @@ void convert_vector(const detail::data_parallel_policy& policy,
                     std::int64_t dst_stride,
                     std::int64_t element_count);
 
+void convert_matrix(const detail::data_parallel_policy& policy,
```
Vika-F, line 70: It would be nice to add a description here that this function might optionally copy data to device in some cases. And to describe in which cases it does so.

Vika-F (changes requested):
The changes look good in general, but there is a lack of comments in my opinion.

Vika-F (approved):
I am Ok with the changes.

PR #2088, opened 2022-03-02
Enable seed for kmeans_init

dmitrii-kriukov:
Enable seed for kmeans_init in oneDAL interfaces


cpp/oneapi/dal/algo/kmeans_init/common.cpp
```diff
@@ -24,6 +24,7 @@ template <typename Task>
 class descriptor_impl : public base {
 public:
     std::int64_t cluster_count = 2;
+    std::int64_t seed = 777777;
```
agorshk, line 27: Does this number align with other algorithms?

agorshk (commented):
Please take into account backward compatibility

cpp/oneapi/dal/algo/kmeans_init/common.cpp
dmitrii-kriukov: This seed value was set because before commit 
https://github.com/oneapi-src/oneDAL/commit/abb65854a31795ff83f198d840927f011399b38a#diff-4589e4a10663d26c02eaf3fa0218aa75feba9df0edef43ce765a93dd5b30bb8cR40
seed for both cpu and gpu was 777777 and then changed to gpu only

But other algorithms use seed value 777
https://github.com/oneapi-src/oneDAL/blob/master/cpp/oneapi/dal/algo/decision_forest/common.cpp#L70

https://github.com/oneapi-src/oneDAL/blob/master/cpp/oneapi/dal/algo/knn/backend/cpu/train_kernel_kd_tree.cpp#L59

So it makes sense to align with them

amgrigoriev: There is no alignment in classical DAAL (777777 for kmeans_init and ALS, 777 - for others).

agorshk: My suggestion is to align at least in new oneDAL API, if it isn't impacted accuracy or other things

dmitrii-kriukov: I will check the accuracy and DBI on benchmarks

amgrigoriev: It will be ok if both CPU/GPU backed in new oneDAL use the same seed.

amgrigoriev: But it will cause a discrepancy for users after transition to new oneDAL

dmitrii-kriukov:
backward compatibility
http://intel-ci.intel.com/eca45fdb-1901-f16c-8e1f-8cdcd4b723a1

napetrov:
@KulikovNikita  - do we need this after recent merge?

PR #2121, opened 2022-08-05
Prioritize wget while downloading dependencies

Alexsandruss:
# Description
Moving of wget to 1st place in dependencies downloading options. Curl might fell while downloading because of network libraries conflict in some environments.


PivovarA (approved):

Alexsandruss:
Issue was on proxy setting side, this PR is not meaningful now.

PR #2122, opened 2022-08-08
CSR table serialization

Alexsandruss:
# Description
Add serialization functions for CSR table. Solves https://github.com/intel/scikit-learn-intelex/issues/1005


samir-nasibli (approved):
LGTM but we need more checks for merge

PR #2125, opened 2022-09-15
SYCL refocus

Arunxsk:
Updated DPC++ references in oneDAL documentation.


napetrov:
There would be conflicts with what is merged in develop. First we would cherry peak changes from develop to master and then can look on this PR.

napetrov (approved):

PR #2153, opened 2022-11-17
doc: Update build-application.rst

alexandraepan:
# Description
Update build-application.rst since dpcpp driver is deprecated in the 2023.0 release.

### DIFF
--- a/docs/source/onedal/build_app/build-application.rst
+++ b/docs/source/onedal/build_app/build-application.rst
@@ -186,12 +186,14 @@ Examples
 Dynamic linking, Multi-threaded |short_name|:
 
    * Linux* OS:
+   
    .. code-block:: text
 
      icpx -fsycl my_first_dal_program.cpp -Wl,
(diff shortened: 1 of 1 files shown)


docs/source/onedal/build_app/build-application.rst
```diff
@@ -31,7 +31,7 @@ Applications on Linux* OS
 
 #. Set environment variables by calling ``<install dir>/setvars.sh``.
 
-#. Build your application with dpcpp:
+#. Use the C++ driver with the -fsycl option to build the application:
```
napetrov, line 34: don't think that regular users would understand driver terminology - suggest to just call out executable as it was done previously.

alexandraepan: @napetrov I'm not sure what you mean by driver terminology. We cannot leave it as it is, since there will be no dcppp anymore. What do you think about: 

`` Build your application with the C++ driver: ``

Alexsandruss (changes requested):
Retarget to `develop` branch and check docbuild fails

docs/source/onedal/build_app/build-application.rst
napetrov: use of icpx /icx names instead of C++ driver. Or some other correct name describing this

Alexsandruss (approved):

napetrov (approved):

PR #2164, opened 2022-12-02
System info output in CI

Alexsandruss:


napetrov:
Let's put this in script or function - it's lot of copy pasted code

PR #2165, opened 2022-12-06
Add SONAME in Linux dynamic libraries

Alexsandruss:
# Description
Add [SONAME](https://en.wikipedia.org/wiki/Soname) in Linux dynamic libraries


Alexsandruss:
Release check CI will be reverted when library versions correctness is confirmed

napetrov (approved):

PR #2173, opened 2022-12-14
Update version to 2023.0.1

Alexsandruss:


napetrov (commented):
Think that there other places - like in Bazel?

homksei:
> > Think that there other places - like in Bazel?
> 
> I updated bazel config.
> 
> Should I add update version in docs in places like [this](https://github.com/oneapi-src/oneDAL/commit/071a10fa1ca25bd35d9924b0e87570d260d59d24#diff-470005a18e5f37436e6f99b992589dd1a8f903e8577c5a9e2e582bcbf03a99e5L41-R41)?

@maria-Petrova, could you please help with this?

napetrov:
> > Think that there other places - like in Bazel?
> 
> 
> 
> I updated bazel config.
> 
> 
> 
> Should I add update version in docs in places like [this](https://github.com/oneapi-src/oneDAL/commit/071a10fa1ca25bd35d9924b0e87570d260d59d24#diff-470005a18e5f37436e6f99b992589dd1a8f903e8577c5a9e2e582bcbf03a99e5L41-R41)?

Docs would be fine with major/minor but yes in general

PR #2174, opened 2022-12-14
Update configuration for .sln examples

homksei:


napetrov:
I think this is ok as long as examples would be passing in CI. 
Also - i think this would be good to merge after drop?

homksei:
> I think this is ok as long as examples would be passing in CI. Also - i think this would be good to merge after drop?

Agree.
Btw: https://intel-ci.intel.com/ed7c0e33-6d70-f179-a9c4-a4bf010d0e2e

napetrov (approved):

Alexsandruss (approved):

PR #2176, opened 2022-12-15
Fix cov_dpc tests

avolkov-intel:
Covariance tests were incorrect, I fixed them so that they do not produce failures


samir-nasibli (approved):

cpp/oneapi/dal/backend/primitives/stat/test/cov_dpc.cpp
```diff
@@ -289,27 +289,30 @@ TEMPLATE_TEST_M(cov_test, "correlation on one-row table", "[cor]", float) {
...
 
     INFO("check if diagonal elements are ones");
     this->check_diagonal_is_ones(corr);
 
-    INFO("check if mean is zero")
-    this->check_constant_mean(vars, 1, 0.0);
-
-    INFO("check if variance is zero")
+    INFO("check if variance is zero");
+    var_event.wait_and_throw();
```
samir-nasibli, line 313: You can wait for a vector of events to complete, and pass asynchronous errors to handlers associated with the commands. But in case of test code readability we can leave it as is

PR #2222, opened 2022-12-28
Updating oneDAL examples

napetrov:
# Description
This change brings several improvements to oneDAL examples

- cleanup obsolete code for ia32 across library
- reducing datasets for examples so they would be running within seconds
- cmake examples improvements


examples/oneapi/data/svm_reg_test_dense_label.csv
```diff
@@ -148,103 +148,3 @@
 -554.87
 -2344.25
 -93.33
-735.76
```
KulikovNikita, line 151: What is the reason to cut this files?

examples/oneapi/data/svm_reg_test_dense_data.csv
```diff
@@ -148,103 +148,3 @@
 1.1274,-1.7226,-0.8364,0.4407,-1.4097,-0.4402,-1.2004,0.1056,0.9454,-0.2825
 -0.7780,1.3920,0.4974,1.3671,0.0765,-1.2204,-1.1302,-0.4292,0.9134,1.7044
 0.0368,-0.7699,-0.5115,-1.3245,-1.1933,0.6311,-1.2972,-0.2551,0.1495,-0.5670
-0.5882,1.3870,1.4177,0.2803,1.4878,0.6297,-0.9434,0.3668,0.5495,-1.6366
```
KulikovNikita, line 151: What is the reason to cut this files?

examples/oneapi/data/svm_reg_test_dense_label.csv
napetrov: Idea was that test data should be proportionally reduced - like with SVM we have 500 rows for training and it make little sense to have large amount for testing. But probably we can keep them as is

.ci/scripts/test.sh
```diff
@@ -119,37 +116,100 @@ if [ "${interface}" == "daal/java" ]; then
...
+            export CXX=clang++
+        elif [[ ${compiler} == icx ]]; then
+            export CC=icx
+            export CXX=icpx
+        fi
+        echo "============== Configuration: =============="
+        echo Compiler:  ${compiler}
+        echo Link mode: ${link_mode}
+        echo CMAKE_C_COMPILE: ${CC}
+        echo CMAKE_CXX_COMPILER: ${CXX}
```
Alexsandruss, line 146: These echos might be misleading because CMake may use compiler not from CC/CXX in some cases

.ci/scripts/test.sh
```diff
@@ -119,37 +116,100 @@ if [ "${interface}" == "daal/java" ]; then
...
+        echo "============================================"
+
+        if [ -d "Build" ]; then
+            rm -rf ./Build/*
+        else
+            mkdir Build
+        fi
+
+        cmake -B./Build -S. -G "Unix Makefiles" -DTARGET_LINK=${link_mode} -DTBB_DIR=${TBBROOT}/lib/cmake/tbb
```
Alexsandruss, line 155: Inconsistent `Build` and `.Build` paths

docs/source/get-started/build-and-run-examples.rst
```diff
@@ -104,19 +104,22 @@ basic usage scenarios of |short_name| with DPCPP. Go to
 
            # Navigate to examples directory and build examples
            cd /examples/oneapi/dpc
-           make so example=svm_two_class_thunder_dense_batch # This will compile and run Correlation example using Intel(R) oneAPI DPC++/C++ Compiler
-           make so mode=build			   # This will compile all the examples
+           cmake -G "Unix Makefiles⁮" -DEXAMPLES=svm_two_class_thunder #This would generate makefiles for all svm examples matching passed name
+           make               # This will compile and run generated svm examples
+           cmake -G "Unix Makefiles⁮" -DTARGET_LINK=static #This wouldgenerate make for static version
```
Alexsandruss, line 109:
```suggestion
           cmake -G "Unix Makefiles⁮" -DEXAMPLES=svm_two_class_thunder # This would generate makefiles for all svm examples matching passed name
           make               # This will compile and run generated svm examples
           cmake -G "Unix Makefiles⁮" -DTARGET_LINK=static # This wouldgenerate make for static version
```

docs/source/get-started/build-and-run-examples.rst
```diff
@@ -104,19 +104,22 @@ basic usage scenarios of |short_name| with DPCPP. Go to
...
 
          .. code-block:: bash
 
             # Navigate to examples directory and build examples
             cd /examples/oneapi/dpc
-            nmake dll example=svm_two_class_thunder_dense_batch+ # This will compile and run Correlation example using Intel(R) oneAPI DPC++/C++ Compiler
-            nmake dll mode=build			     # This will compile all the examples
+           cmake  -G "NMake Makefiles" -DEXAMPLES=svm_two_class_thunder #This would generate makefiles for all svm examples matching passed name
+           nmake             # This will compile and run generated svm examples
+           cmake  -G "NMake Makefiles" -DTARGET_LINK=static #This wouldgenerate make for static version
```
Alexsandruss, line 120:
```suggestion
           cmake  -G "NMake Makefiles" -DEXAMPLES=svm_two_class_thunder # This would generate makefiles for all svm examples matching passed name
           nmake             # This will compile and run generated svm examples
           cmake  -G "NMake Makefiles" -DTARGET_LINK=static # This wouldgenerate make for static version
```

examples/cmake/setup_examples.cmake
```diff
@@ -0,0 +1,137 @@
+#===============================================================================
+# Copyright 2022 Intel Corporation
```
Alexsandruss, line 2: 2023?

.ci/scripts/test.sh
napetrov: Do you have better examples how to get relevant compilers? 
Agree on naming - would change to just CC and CXX

.ci/scripts/test.sh
napetrov: they are ./Build and Build  - same path but with statement from this directory "./"  Not sure 100% on this logic, but this is code from other scripts that we use for cmake execution.

.ci/scripts/test.sh
Alexsandruss: cmake outputs used compilers itself:
```
-- The C compiler identification is GNU 11.3.0
-- The CXX compiler identification is GNU 11.3.0
```
so, output of preset CC and CXX will be enough

.ci/scripts/test.sh
```diff
@@ -119,39 +117,101 @@ if [ "${interface}" == "daal/java" ]; then
...
+        make -C Build
+        err=$?
+        if [ ${err} -ne 0 ]; then
+            echo -e "$(date +'%H:%M:%S') BUILD FAILED\t\t"
+            TESTING_RETURN=${err}
+            continue
+        fi
+        output_result=
+        err=
+        cmake_results_dir="_cmake_results/intel_intel64_${lib_ext}"
```
Alexsandruss, line 171: Why `lib_ext` instead of `l`?

examples/daal/cpp/source/compression/compression_online.cpp
```diff
@@ -58,7 +58,7 @@ int main(int argc, char *argv[]) {
     prepareMemory();
 
     /* Create a compressor */
-    Compressor<zlib> compressor;
+    Compressor<lzo> compressor;
```
Alexsandruss, line 61: I guess same should be done for java examples

.ci/scripts/test.sh
```diff
@@ -119,39 +117,101 @@ if [ "${interface}" == "daal/java" ]; then
...
+            output_result=$(cat ${e}.res)
+            mv -f ${e}.res ${cmake_results_dir}/
+            status_ex=
+            if [ ${err} -ne 0 ]; then
+                echo "${output_result}"
+                status_ex="$(date +'%H:%M:%S') FAILED\t\t${e} with errno ${err}"
+                TESTING_RETURN=${err}
+                continue
+            else
+                echo "${output_result}" | grep -i "error\|warn"
```
Alexsandruss, line 185: Grep will output `OOB error = X` and other similar. Examples output or grep pattern should be rewritten

.ci/scripts/test.sh
Alexsandruss: It is same path but its different specification might confuse. I suggest use only `Build` and add spaces for cmake args:
```suggestion
        if [ -d "Build" ]; then
            rm -rf Build/*
        else
            mkdir Build
        fi

        cmake -B Build -S . -G "Unix Makefiles" -DTARGET_LINK=${link_mode} -DTBB_DIR=${TBBROOT}/lib/cmake/tbb
```

.ci/scripts/test.sh
napetrov: Those directories are created by cmake, so this is mapping for them. 

Not sure if it's worth updating cmake to have different directories.

examples/daal/cpp/source/compression/compression_online.cpp
napetrov: suggest to do this in separate PR.  this one intended for cmake and cpp examples were changed to remove warnings.

.ci/scripts/test.sh
napetrov: yes. this is code and logic that we already use internally.  So i think better patterns should be aligned between both. 

Here it's better way to get things printed and not skipping them at all. Might be it's better to just skipp this output.

.ci/scripts/test.sh
Alexsandruss: If it's default naming for cmake, then no change is needed

napetrov:
@Alexsandruss  @KulikovNikita  @homksei - all changes/fixes completed here. the remaining CI failures doesn't related to PR and reproducible on master

Alexsandruss (approved):

PR #2226, opened 2022-12-31
Standard atomic

KulikovNikita:


samir-nasibli:
Is the PR ready for review? Many check fails.

napetrov:
I think that introduction of std deps for DAAL should be discussed more closely

PR #2229, opened 2023-01-05
LinReg extended API

KulikovNikita:
**Major changes:**
1) Enables various result options for the `train` stage, including: `intercept`, `coefficients`
2) Eliminates unnecessary abstract class usage in Linear Regression model. Now it is implemented via simple pimpl to concrete class
3) Extends testing to ensure that `intercept`s and `coeffiicients` are computed correctly

**Minor changes:**
4) Extends debug functionality

### DIFF
--- a/cpp/oneapi/dal/algo/linear_regression/backend/cpu/train_kernel_norm_eq.cpp
+++ b/cpp/oneapi/dal/algo/linear_regression/backend/cpu/train_kernel_norm_eq.cpp
@@ -52,7 +52,7 @@ static train_result<Task> call_daal_kernel(const context_cpu& ctx,
     using dal::detail::check_mul_overflow;
 
     using model_t = model<Task>;
-    using model_impl_t = backend::norm_eq_model_impl<Task>;
+    using model_impl_t = detail::model_impl<Task>;
 
     bool intp = desc.get_compute_intercept();
 
--- a/cpp/oneapi/dal/algo/linear_regression/backend/cpu/train_kernel_norm_eq.cpp
+++ b/cpp/oneapi/dal/algo/linear_regression/backend/cpu/train_kernel_norm_eq.cpp
@@ -54,25 +54,25 @@ static train_result<Task> call_daal_kernel(const context_cpu& ctx,
     using model_t = model<Task>;
     using model_impl_t = detail::model_impl<Task>;
 
-    bool intp = desc.get_compute_intercept();
+    const bool intp = desc.get_compute_intercept();
 
-    const auto f_count = data.get_column_count();
-    const auto r_count = resp.get_column_count();
+    const auto feature_count = data.get_column_count();
+    const auto response_count = resp.get_column_count();
 
-    const auto ext_f_count = f_count + intp;
+    const auto ext_feature_count = feature_count + intp;
 
-    const auto xtx_size = check_mul_overflow(ext_f_count, ext_f_count);
+    const auto xtx_size = check_mul_overflow(ext_feature_count, ext_feature_count);
     auto xtx_arr = array<Float>::zeros(xtx_size);
 
-    const auto xty_size = check_mul_overflow(r_count, ext_f_count);
+    const auto xty_size = check_mul_overflow(response_count, ext_feature_count);
     auto xty_arr = array<Float>::zeros(xty_size);
 
-    const auto betas_size = check_mul_overflow(r_count, f_count + 1);
+    const auto betas_size = check_mul_overflow(response_count, feature_count + 1);
     auto betas_arr = array<Float>::zeros(betas_size);
 
-    auto xtx_daal_table = interop::convert_to_daal_homogen_table(xtx_arr, ext_f_count, ext_f_count);
-    auto xty_daal_table = interop::convert_to_daal_homogen_table(xty_arr, r_count, ext_f_count);
-    auto betas_daal_table = interop::convert_to_daal_homogen_table(betas_arr, r_count, f_count + 1);
+    auto xtx_daal_table = interop::convert_to_daal_homogen_table(xtx_arr, ext_feature_count, ext_feature_count);
+    auto xty_daal_table = interop::convert_to_daal_homogen_table(xty_arr, response_count, ext_feature_count);
+    auto betas_daal_table = interop::convert_to_daal_homogen_table(betas_arr, response_count, feature_count + 1);
 
     auto x_daal_table = interop::convert_to_daal_table<Float>(data);
     auto y_daal_table = interop::convert_to_daal_table<Float>(resp);
@@ -102,7 +102,7 @@ static train_result<Task> call_daal_kernel(const context_cpu& ctx,
         interop::status_to_exception(status);
     }
 
-    auto betas_table = homogen_table::wrap(betas_arr, r_count, f_count + 1);
+    auto betas_table = homogen_table::wrap(betas_arr, response_count, feature_count + 1);
 
     const auto model_impl = std::make_shared<model_impl_t>(betas_table);
     const auto model = dal::detail::make_private<model_t>(model_impl);
@@ -110,30 +110,30 @@ static train_result<Task> call_daal_kernel(const context_cpu& ctx,
     const auto options = desc.get_result_options();
     auto result = train_result<Task>().set_model(model).set_result_options(options);
 
-    const pr::ndshape<2> betas_shape{ r_count, f_count + 1 };
+    const pr::ndshape<2> betas_shape{ response_count, feature_count + 1 };
     auto betas = pr::ndarray<Float, 2>::wrap(betas_arr, betas_shape);
 
     if (options.test(result_options::intercept)) {
-        auto arr = array<Float>::zeros(r_count);
-        auto dst = pr::ndarray<Float, 2>::wrap_mutable(arr, { 1l, r_count });
+        auto arr = array<Float>::zeros(response_count);
+        auto dst = pr::ndarray<Float, 2>::wrap_mutable(arr, { 1l, response_count });
         const auto src = betas.get_col_slice(0l, 1l).t();
 
         pr::copy(dst, src);
 
-        auto intercept = homogen_table::wrap(arr, 1l, r_count);
+        auto intercept = homogen_table::wrap(arr, 1l, response_count);
         result.set_intercept(intercept);
     }
 
     if (options.test(result_options::coefficients)) {
-        const auto size = check_mul_overflow(r_count, f_count);
+        const auto size = check_mul_overflow(response_count, feature_count);
 
         auto arr = array<Float>::zeros(size);
-        const auto src = betas.get_col_slice(1l, f_count + 1);
-        auto dst = pr::ndarray<Float, 2>::wrap_mutable(arr, { r_count, f_count });
+        const auto src = betas.get_col_slice(1l, feature_count + 1);
+        auto dst = pr::ndarray<Float, 2>::wrap_mutable(arr, { response_count, feature_count });
 
         pr::copy(dst, src);
 
-        auto coefficients = homogen_table::wrap(arr, r_count, f_count);
+        auto coefficients = homogen_table::wrap(arr, response_count, feature_count);
         result.set_coefficients(coefficients);
     }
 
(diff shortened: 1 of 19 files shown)


cpp/oneapi/dal/algo/linear_regression/backend/gpu/train_kernel_norm_eq_dpc.cpp
```diff
@@ -58,7 +58,7 @@ static train_result<Task> call_dal_kernel(const context_gpu& ctx,
     using dal::detail::check_mul_overflow;
 
     using model_t = model<Task>;
-    using model_impl_t = backend::norm_eq_model_impl<Task>;
+    using model_impl_t = detail::model_impl<Task>;
```
Vika-F, line 61: I do not fully understand why norm_eq_model_impl was changed to model_impl in certain places?
As I understand norm_eq_model_impl specifies that this model was produced using a normal equations method (norm_eq). In the future we might add other methods to solve linear regression problem (QR, gradient descent, ...).
For me, all the *_model_impl classes for various compute methods will have common parts: coefficiens and intercept; but might have other differences.
What I do not understand: is the current version of the code scales well in case of adding other types of *_model_impl?

cpp/oneapi/dal/algo/linear_regression/backend/model_impl.hpp
```diff
@@ -22,43 +22,33 @@
 
 namespace oneapi::dal::linear_regression {
 
+using norm_eq_proto = ONEDAL_SERIALIZABLE(linear_regression_model_impl_id);
```
Vika-F, line 25: The same question: why the ID linear_regression_norm_eq_model_impl_id was changed to linear_regression_model_impl_id?
If we would add a model for QR compute method, for example, that model might have different internal structure than norm_eq model, so it will need a different serialization ID.

cpp/oneapi/dal/algo/linear_regression/common.cpp
```diff
@@ -60,14 +95,25 @@ namespace v1 {
...
+model<Task>::model(const std::shared_ptr<detail::model_impl<Task>>& impl) : impl_{ impl } {}
 
 template <typename Task>
 const table& model<Task>::get_betas() const {
-    return impl_->get_betas();
+    return impl_->get_packed_coefficients();
+}
+
+template <typename Task>
+const table& model<Task>::get_packed_coefficients() const {
```
Vika-F, line 109: Should we also extend the docs as new API methods were added?
BTW, I do not see the docs for linear regression at all in github.io.

cpp/oneapi/dal/algo/linear_regression/train_types.cpp
```diff
@@ -33,6 +34,11 @@ class detail::v1::train_input_impl : public base {
 template <typename Task>
 class detail::v1::train_result_impl : public base {
 public:
+    table intercept;
```
Vika-F, line 37: I see that it is possible to get an inconsistent state of train_result_impl when storing intercept and coefficients tables and model as a separate fields of the result. As they might be modified independently from each other.
Do we need set_intercept() and set_coefficients() at all? They only can lead to an object inconsistency and are not needed for compatibility with sklearn.

cpp/oneapi/dal/algo/linear_regression/train_types.hpp
```diff
@@ -98,9 +98,47 @@ class train_result {
         return *this;
     }
 
+    /// Table of Linear regression intercept
+    const table& get_intercept() const;
+
+    auto& set_intercept(const table& value) {
```
Vika-F, line 104: Do we need this?
The same question for set_coefficients, set_packed_coefficients.

cpp/oneapi/dal/backend/serialization.hpp
```diff
@@ -187,7 +187,7 @@ class serialization_ids {
     ID(6040000000, decision_forest_model_interop_impl_reg_id);
 
     // Algorithms - Linear Regression
-    ID(7010000000, linear_regression_norm_eq_model_impl_id);
+    ID(7010000000, linear_regression_model_impl_id);
```
Vika-F, line 190: See the question about serialization ID above.

Vika-F (commented):
Need some clarifications before I can approve this PR.

cpp/oneapi/dal/algo/linear_regression/backend/gpu/infer_kernel_norm_eq_dpc.cpp
```diff
@@ -109,7 +109,7 @@ static infer_result<Task> call_dal_kernel(const context_gpu& ctx,
     ONEDAL_ASSERT((f_count + 1) == betas.get_column_count());
 
     const auto resps_size = check_mul_overflow(s_count, r_count);
-    auto resps_arr = array<Float>::empty(queue, resps_size);
```
samir-nasibli, line 112: I suggest not to mix enhancement and bug fixing in one PR

cpp/oneapi/dal/algo/linear_regression/backend/gpu/infer_kernel_norm_eq_dpc.cpp
```diff
@@ -109,7 +109,7 @@ static infer_result<Task> call_dal_kernel(const context_gpu& ctx,
...
+    auto resps_arr = array<Float>::zeros(queue, resps_size, alloc);
```
samir-nasibli, line 112: I suggest not to mix enhancement and bug fixing in one PR

cpp/oneapi/dal/algo/linear_regression/common.cpp
KulikovNikita: Just because there are no such docs for all new OneAPI algorithms at all. And yes it is in todo list but not with a low priority

cpp/oneapi/dal/algo/linear_regression/backend/cpu/train_kernel_norm_eq.cpp
```diff
@@ -51,30 +52,27 @@ static train_result<Task> call_daal_kernel(const context_cpu& ctx,
     using dal::detail::check_mul_overflow;
 
     using model_t = model<Task>;
-    using model_impl_t = backend::norm_eq_model_impl<Task>;
+    using model_impl_t = detail::model_impl<Task>;
 
     bool intp = desc.get_compute_intercept();
 
-    const auto feature_count = data.get_column_count();
-    const auto response_count = resp.get_column_count();
```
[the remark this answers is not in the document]
KulikovNikita: 1) It is just too wordy
2) Naming should be consistent between CPU & GPU parts

But okay I'll fix it

cpp/oneapi/dal/algo/linear_regression/backend/gpu/train_kernel_norm_eq_dpc.cpp
KulikovNikita: As you have already said there is no many options what can be inside of the model. Only coefficients and possibly intercept values. So there is no way that another implementation of model should be added. In any way it can be restored relatively easy.

KulikovNikita: And btw you can just extend this model with empty tables if needed. Anyway the method of computations is determined based on descriptor, not on runtime info.

cpp/oneapi/dal/algo/linear_regression/backend/model_impl.hpp
KulikovNikita: Just because all models can be converted to the form: intercept + coefficients

cpp/oneapi/dal/algo/linear_regression/train_types.cpp
KulikovNikita: Yes. In some sense these tables are not necessarily required since they can be obtained by decomposition of `packed_coefficients` table.
The thing is:
1) `train_results` is a separate entity from a `model`. We do not care about this tables after them being set. The infer stage is fed only with `model` itself
2) It can be hard to split table that is stored inside the model. Our table API does not support any kind of table views
3) User algorithms should be more or less agnostic about the internal structure of the model. In example there are many algorithms like kd-tree or decision-forest when user do not have access to the guts of model at all.

cpp/oneapi/dal/backend/serialization.hpp
KulikovNikita: We need to save this ID for the sake of backward compatibility between models. Naming is changed according to a new class name

cpp/oneapi/dal/algo/linear_regression/backend/gpu/infer_kernel_norm_eq_dpc.cpp
KulikovNikita: Agreed. But in this case these fixes are pretty small

cpp/oneapi/dal/backend/serialization.hpp
KulikovNikita: Discussed

cpp/oneapi/dal/algo/linear_regression/train_types.hpp
KulikovNikita: DIscussed

cpp/oneapi/dal/algo/linear_regression/train_types.cpp
KulikovNikita: Discussed

cpp/oneapi/dal/algo/linear_regression/common.cpp
KulikovNikita: Discussed

cpp/oneapi/dal/algo/linear_regression/backend/model_impl.hpp
KulikovNikita: Discussed

cpp/oneapi/dal/algo/linear_regression/backend/gpu/train_kernel_norm_eq_dpc.cpp
KulikovNikita: Discussed offline

cpp/oneapi/dal/algo/linear_regression/backend/cpu/train_kernel_norm_eq.cpp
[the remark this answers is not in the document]
KulikovNikita: Fixed both in oneapi and daal parts

Vika-F (approved):
The changes look good to me.

cpp/oneapi/dal/algo/linear_regression/backend/model_impl.hpp
```diff
@@ -22,43 +22,33 @@
...
+
 template <typename Task>
-class detail::v1::model_impl : public base {
+class detail::v1::model_impl : public norm_eq_proto {
```
samir-nasibli, line 28: I would call it like `norm_eq_base`, but that's fine too.

cpp/oneapi/dal/algo/linear_regression/test/batch.cpp
```diff
@@ -34,10 +34,10 @@ class lr_batch_test : public lr_test<TestType, lr_batch_test<TestType>> {
...
 
     void generate_dimensions() {
-        this->t_count_ = GENERATE(327, 888, 32777);
-        this->s_count_ = GENERATE(111, 333, 16777);
-        this->f_count_ = GENERATE(2, 17, 99);
-        this->r_count_ = GENERATE(2, 15, 33);
+        this->t_count_ = GENERATE(307, 12999, 17777);
+        this->s_count_ = GENERATE(111, 777);
+        this->f_count_ = GENERATE(2, 17);
+        this->r_count_ = GENERATE(2, 15);
```
samir-nasibli, line 40: Reducing dims?

samir-nasibli (approved):
LGTM

cpp/oneapi/dal/algo/linear_regression/test/batch.cpp
KulikovNikita: Yeah. It takes just too much time to evaluate all variants especially on relatively weak hardware

PR #2234, opened 2023-01-09
KMeans init seed

KulikovNikita:


inteldimitrius:
Does pre-commit check failure make sense?

cpp/oneapi/dal/algo/kmeans_init/common.hpp
```diff
@@ -74,6 +74,12 @@ constexpr bool is_valid_method_v = dal::detail::is_one_of_v<Method,
 template <typename Task>
 constexpr bool is_valid_task_v = dal::detail::is_one_of_v<Task, task::init>;
 
+template <typename M>
+constexpr bool is_not_default_dense = !std::is_same_v<M, method::dense>;
+
+template <typename M>
+using enable_if_not_default_dense = std::enable_if_t<is_not_default_dense<M>>;
+
```
Vika-F, line 82: Those redefinitions look like an overkill for me. They do not make the code simpler, but they require additional effort from reader to understand what is hidden underneath "enable_if_not_default_dense".

Vika-F (approved):
I'd prefer the code with less number of re-definitions, but I am Ok if the author has another opinion. In general the changes look good to me.

cpp/oneapi/dal/algo/kmeans_init/common.hpp
KulikovNikita: I believe it contains everything you need t know just in naming.

Vika-F: We do not need to agree on how particularly each line of code is implemented. We can have different sense of beauty, it is Ok.
In general the code looks good to me, I approve the changes.

napetrov (approved):

PR #2243, opened 2023-01-16
makefile helpers

Alexsandruss:
# Description

- Autodetection of platform using `uname`
- Removal of platform arg. from CI and other scripts 
- Generation of cmake configs in makefile if cmake is available

### DIFF
--- a/makefile
+++ b/makefile
@@ -18,6 +18,10 @@
 # Common macros
 #===============================================================================
 
+ifeq ($(PLAT),)
+    PLAT:=$(shell bash dev/make/identify_os.sh)
+endif
+
 ifeq (help,$(MAKECMDGOALS))
     PLAT:=win32e
 endif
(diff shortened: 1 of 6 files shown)


KulikovNikita (approved):

makefile
```diff
@@ -18,6 +18,10 @@
 # Common macros
 #===============================================================================
 
+ifeq ($(PLAT),)
+    PLAT:=$(shell bash dev/make/identify_os.sh)
```
napetrov, line 22: will this work on windows?

makefile
```diff
@@ -1113,11 +1117,15 @@ endef
 $(foreach t,$(releasetbb.LIBS_Y),$(eval $(call .release.t,$t,$(RELEASEDIR.tbb.soia))))
 $(foreach t,$(releasetbb.LIBS_A),$(eval $(call .release.t,$t,$(RELEASEDIR.tbb.libia))))
 
+#----- cmake configs generation
+_release_cmake_configs:
+	$(if $(shell bash -c "command -v cmake"),cmake -DINSTALL_DIR=$(RELEASEDIR.lib)/cmake/oneDAL -P cmake/scripts/generate_config.cmake,echo 'cmake configs generation skipped')
```
napetrov, line 1122: cmake generation is part of internal infra currently - corresponding changes should be made there

makefile
Alexsandruss: yes, msys2/bash is requirement for building on Windows and it returns correct PLAT

makefile
Alexsandruss: Is there any problem if configs are generated twice? I think auto generation of configs will be more useful for users

napetrov: Agree that it should be generated here. I'm not sure what would happen on second call - if it works - then fine

Alexsandruss: It works, same configs are generated each time

PR #2248, opened 2023-01-19
fix(cmake,win32,vs2022): add missing SYCL flags to linking step

ahuber21:
Building the `daal/cpp_sycl` examples with Visual Studio 2022 and using CMake>=3.24 results in a broken linking step because the required SYCL flags are not included in the command.
The change adds the missing flags for the `daal/cpp_sycl` examples.


KulikovNikita (approved):

Alexsandruss (approved):

samir-nasibli (approved):

napetrov (approved):
Looks good. Assuming CI would pass

PR #2252, opened 2023-01-22
Disable FP64 emulation in bazel tests

napetrov:
# Description
Disable FP64 emulation variables set in case of bazel tests


Alexsandruss:
@napetrov, will you add any additional changes to this PR?

napetrov:
> @napetrov, will you add any additional changes to this PR?

yes, moving to draft for now

Alexandr-Solovev (changes requested):
We have error in bazel tests:
terminate called after throwing an instance of 'sycl::_V1::exception'
  what():  Required aspect fp64 is not supported on the device

napetrov:
> We have error in bazel tests: terminate called after throwing an instance of 'sycl::_V1::exception' what(): Required aspect fp64 is not supported on the device

this means that tests have to be updated so they can run on systems without FP64 support

PR #2256, opened 2023-01-23
Fix RF Reg example

Alexandr-Solovev:


napetrov:
@Alexandr-Solovev  - should we have bazel test added for this?

KulikovNikita (approved):

napetrov (approved):

PR #2292, opened 2023-03-10
RF: fix performance of feature sampling for node splits

ahuber21:
# Description
With merging #1725, the `chooseFeatures()` helper function has become inefficient.
It calls a helper, `rng.uniformWithoutReplacement()`, that implements a non-standard version of drawing k out of N without replacement. The algo is `O(N^2)` compute time and with the changes from #1725 `N` went from `_nFeaturesPerNode` to `maxFeatures`, where `maxFeatures` is the total number of features in the dataset and `_nFeaturesPerNode` is the number of features per node split (default value: `sqrt(maxFeatures)`).

It was reported in https://github.com/intel/scikit-learn-intelex/issues/1050 and https://github.com/intel/scikit-learn-intelex/issues/984 that our RF classification implementation is very slow and under some circumstances even stalls. I think this is related.

In an initial investigation I studied the training time for different values of `max_features`
image

We see that for large values of `max_features`, the intelex implementation outperforms the stock implementation. But for smaller values (in particular including the default value sqrt(N)), we are slower.
Profiling the code confirmed that indeed we spend most of our time sampling features
image

I have added a few changes in this PR

* Added a helper function in `service_memory.h` to initialize memory with sequential entries
* Implemented a modified Fisher Yates sampling in `service_rng.h`
  * We are using a provided buffer, therefore, no initialization of the indices is required
  * This reduces the time complexity from `O(N)` -> `O(k)`
* Implemented a modified Fisher Yates sampling in `findBestSplitSerial()`  
  * With the changes from #1725 we still generate a sample of size `N`, so we do not benefit from the optimization in the previous bullet
  * Drawing numbers only when necessary reduces the time complecity from `O(N)` to `O(k + number of constant features)`, which is the optimal solution
* Exposed the `useConstFeatures` setting to the training params - it used to be hardcoded to `false`

With the changes the picture changes
image
_Note: This second run was performed on a system with fewer cores, explaining the longer overall training times. In the comparison against the stock implementation, however, this effect cancels out._

The profiler confirms that the bottleneck of calling the random number generator disappeared
image

Most of the compute time is now spent creating the histograms.

## Statistical correctness

I have tested the statistical correctness of my Fisher Yates sampling technique numerically.
I draw 5 out of 10 numbers, meaning each number has p=0.5 of being drawn.
Running the test 10,000,000 times confirms the uniform probability of all numbers

```bash
# run with RNG seed 777
$ _results/dpcpp_intel64_a/rng.exe 10000000 777
0,1,2,3,4,5,6,7,8,9,
5000024,5000684,5000634,5002295,5000448,4999619,4998491,4998169,5000202,4999434,

# run with RNG seed 1337
$ _results/dpcpp_intel64_a/rng.exe 10000000 1337
0,1,2,3,4,5,6,7,8,9,
4998947,5002204,5000942,4997803,4999591,4999615,4999998,5000007,5001135,4999758,
```

Find the source code of the test file below

```c++
// file: rng.cpp
#include <cstdlib>
#include <iostream>
#include <vector>

#include "algorithms/engines/mt2203/mt2203.h"
#include "daal.h"
#include "src/algorithms/engines/engine_batch_impl.h"
#include "src/externals/service_rng.h"

static const int M = 5;  // draw M
static const int N = 10; // out of N

template <typename T> void printCounts(const std::vector<T> &counts) {
  if (N < 50) {
    for (int i = 0; i < N; ++i)
      std::cout << i << ",";
    std::cout << std::endl;
    for (int i = 0; i < N; ++i)
      std::cout << counts[i] << ",";
    std::cout << std::endl;
  }
}

int main(int argc, char *argv[]) {
  int EXPERIMENTS = 1000000;
  if (argc > 1) {
    EXPERIMENTS = atoi(argv[1]);
  }

  unsigned long seed = 777;
  if (argc > 2) {
    seed = atol(argv[2]);
  }

  auto engine = daal::algorithms::engines::mt2203::Batch<>::create(seed);
  auto engineImpl =
      dynamic_cast<daal::algorithms::engines::internal::BatchBaseImpl *>(
          engine.get());

  daal::internal::RNGs<int, daal::CpuType::avx512> rng;

  int result[N];
  int buffer[N];

  std::vector<unsigned int> counts(N, 0);

  for (int i = 0; i < N; ++i) {
    result[i] = -1;
    buffer[i] = i;
  }

  for (int r = 0; r < EXPERIMENTS; ++r) {
    rng.uniformWithoutReplacement(M, result, buffer, engineImpl->getState(),
                                  0, N);

    for (int i = 0; i < M; ++i) {
      counts[result[i]]++;
    }
  }

  if (EXPERIMENTS < 50) {
    std::cout << "B: ";
    for (int i = 0; i < N; ++i)
      std::cout << buffer[i] << "\t";
    std::cout << std::endl;
    std::cout << "R: ";
    for (int i = 0; i < N; ++i)
      std::cout << result[i] << "\t";
    std::cout << std::endl;
  }

  printCounts(counts);

  return 0;
}
```

### DIFF
--- a/cpp/daal/include/algorithms/decision_forest/decision_forest_training_parameter.h
+++ b/cpp/daal/include/algorithms/decision_forest/decision_forest_training_parameter.h
@@ -134,6 +134,7 @@ public:
                                                  Default is 256. Increasing the number results in higher computation costs */
     size_t minBinSize;                     /*!< Used with 'hist' split finding method only.
                                                  Minimal number of observations in a bin. Default is 5 */
+    bool useConstFeatures;                 /*!< Use or ignore constant-valued features when splitting nodes. Default is false */
 };
 /* [Parameter source code] */
 } // namespace interface2
--- a/cpp/daal/include/algorithms/decision_forest/decision_forest_training_parameter.h
+++ b/cpp/daal/include/algorithms/decision_forest/decision_forest_training_parameter.h
@@ -134,7 +134,6 @@ public:
                                                  Default is 256. Increasing the number results in higher computation costs */
     size_t minBinSize;                     /*!< Used with 'hist' split finding method only.
                                                  Minimal number of observations in a bin. Default is 5 */
-    bool useConstFeatures;                 /*!< Use or ignore constant-valued features when splitting nodes. Default is false */
 };
 /* [Parameter source code] */
 } // namespace interface2
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
@@ -117,9 +117,9 @@ public:
     bool terminateCriteria(ImpurityData & imp, algorithmFPType impurityThreshold, size_t nSamples) const { return imp.value() < impurityThreshold; }
 
     template <typename BinIndexType>
-    int findBestSplitForFeatureSorted(algorithmFPType * featureBuf, IndexType iFeature, const IndexType * aIdx, size_t n, size_t nMinSplitPart,
-                                      const ImpurityData & curImpurity, TSplitData & split, const algorithmFPType minWeightLeaf,
-                                      const algorithmFPType totalWeights, const BinIndexType * binIndex) const;
+    int64_t findBestSplitForFeatureSorted(algorithmFPType * featureBuf, IndexType iFeature, const IndexType * aIdx, size_t n, size_t nMinSplitPart,
+                                          const ImpurityData & curImpurity, TSplitData & split, const algorithmFPType minWeightLeaf,
+                                          const algorithmFPType totalWeights, const BinIndexType * binIndex) const;
     template <typename BinIndexType>
     void computeHistFewClassesWithoutWeights(IndexType iFeature, const IndexType * aIdx, const BinIndexType * binIndex, size_t n) const;
     template <typename BinIndexType>
@@ -127,16 +127,16 @@ public:
     template <typename BinIndexType>
     void computeHistManyClasses(IndexType iFeature, const IndexType * aIdx, const BinIndexType * binIndex, size_t n) const;
 
-    int findBestSplitbyHistDefault(int nDiffFeatMax, size_t n, size_t nMinSplitPart, const ImpurityData & curImpurity, TSplitData & split,
-                                   const algorithmFPType minWeightLeaf, const algorithmFPType totalWeights) const;
+    int64_t findBestSplitbyHistDefault(int nDiffFeatMax, size_t n, size_t nMinSplitPart, const ImpurityData & curImpurity, TSplitData & split,
+                                       const algorithmFPType minWeightLeaf, const algorithmFPType totalWeights) const;
 
     template <int K, bool noWeights>
-    int findBestSplitFewClasses(int nDiffFeatMax, size_t n, size_t nMinSplitPart, const ImpurityData & curImpurity, TSplitData & split,
-                                const algorithmFPType minWeightLeaf, const algorithmFPType totalWeights) const;
+    int64_t findBestSplitFewClasses(int nDiffFeatMax, size_t n, size_t nMinSplitPart, const ImpurityData & curImpurity, TSplitData & split,
+                                    const algorithmFPType minWeightLeaf, const algorithmFPType totalWeights) const;
 
     template <bool noWeights>
-    int findBestSplitFewClassesDispatch(int nDiffFeatMax, size_t n, size_t nMinSplitPart, const ImpurityData & curImpurity, TSplitData & split,
-                                        const algorithmFPType minWeightLeaf, const algorithmFPType totalWeights) const;
+    int64_t findBestSplitFewClassesDispatch(int nDiffFeatMax, size_t n, size_t nMinSplitPart, const ImpurityData & curImpurity, TSplitData & split,
+                                            const algorithmFPType minWeightLeaf, const algorithmFPType totalWeights) const;
 
     template <bool noWeights, typename BinIndexType>
     void finalizeBestSplit(const IndexType * aIdx, const BinIndexType * binIndex, size_t n, IndexType iFeature, size_t idxFeatureValueBestSplit,
@@ -613,10 +613,10 @@ void UnorderedRespHelper<algorithmFPType, cpu>::computeHistManyClasses(IndexType
 }
 
 template <typename algorithmFPType, CpuType cpu>
-int UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitbyHistDefault(int nDiffFeatMax, size_t n, size_t nMinSplitPart,
-                                                                          const ImpurityData & curImpurity, TSplitData & split,
-                                                                          const algorithmFPType minWeightLeaf,
-                                                                          const algorithmFPType totalWeights) const
+int64_t UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitbyHistDefault(int nDiffFeatMax, size_t n, size_t nMinSplitPart,
+                                                                              const ImpurityData & curImpurity, TSplitData & split,
+                                                                              const algorithmFPType minWeightLeaf,
+                                                                              const algorithmFPType totalWeights) const
 {
     auto nFeatIdx         = _idxFeatureBuf.get();
     auto featWeights      = _weightsFeatureBuf.get();
@@ -692,9 +692,10 @@ int UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitbyHistDefault(int nD
 
 template <typename algorithmFPType, CpuType cpu>
 template <int K, bool noWeights>
-int UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitFewClasses(int nDiffFeatMax, size_t n, size_t nMinSplitPart,
-                                                                       const ImpurityData & curImpurity, TSplitData & split,
-                                                                       const algorithmFPType minWeightLeaf, const algorithmFPType totalWeights) const
+int64_t UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitFewClasses(int nDiffFeatMax, size_t n, size_t nMinSplitPart,
+                                                                           const ImpurityData & curImpurity, TSplitData & split,
+                                                                           const algorithmFPType minWeightLeaf,
+                                                                           const algorithmFPType totalWeights) const
 {
     auto nSamplesPerClass = _samplesPerClassBuf.get();
     auto nFeatIdx         = _idxFeatureBuf.get();
@@ -789,10 +790,10 @@ int UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitFewClasses(int nDiff
 
 template <typename algorithmFPType, CpuType cpu>
 template <bool noWeights>
-int UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitFewClassesDispatch(int nDiffFeatMax, size_t n, size_t nMinSplitPart,
-                                                                               const ImpurityData & curImpurity, TSplitData & split,
-                                                                               const algorithmFPType minWeightLeaf,
-                                                                               const algorithmFPType totalWeights) const
+int64_t UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitFewClassesDispatch(int nDiffFeatMax, size_t n, size_t nMinSplitPart,
+                                                                                   const ImpurityData & curImpurity, TSplitData & split,
+                                                                                   const algorithmFPType minWeightLeaf,
+                                                                                   const algorithmFPType totalWeights) const
 {
     DAAL_ASSERT(_nClasses <= _nClassesThreshold);
     switch (_nClasses)
@@ -810,10 +811,9 @@ int UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitFewClassesDispatch(i
 
 template <typename algorithmFPType, CpuType cpu>
 template <typename BinIndexType>
-int UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitForFeatureSorted(algorithmFPType * featureBuf, IndexType iFeature, const IndexType * aIdx,
-                                                                             size_t n, size_t nMinSplitPart, const ImpurityData & curImpurity,
-                                                                             TSplitData & split, const algorithmFPType minWeightLeaf,
-                                                                             const algorithmFPType totalWeights, const BinIndexType * binIndex) const
+int64_t UnorderedRespHelper<algorithmFPType, cpu>::findBestSplitForFeatureSorted(
+    algorithmFPType * featureBuf, IndexType iFeature, const IndexType * aIdx, size_t n, size_t nMinSplitPart, const ImpurityData & curImpurity,
+    TSplitData & split, const algorithmFPType minWeightLeaf, const algorithmFPType totalWeights, const BinIndexType * binIndex) const
 {
     const auto nDiffFeatMax = this->indexedFeatures().numIndices(iFeature);
     _samplesPerClassBuf.setValues(nClasses() * nDiffFeatMax, 0);
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
@@ -117,9 +117,9 @@ public:
     bool terminateCriteria(ImpurityData & imp, algorithmFPType impurityThreshold, size_t nSamples) const { return imp.value() < impurityThreshold; }
 
     template <typename BinIndexType>
-    int64_t findBestSplitForFeatureSorted(algorithmFPType * featureBuf, IndexType iFeature, const IndexType * aIdx, size_t n, size_t nMinSplitPart,
-                                          const ImpurityData & curImpurity, TSplitData & split, const algorithmFPType minWeightLeaf,
-                                          const algorithmFPType totalWeights, const BinIndexType * binIndex) const;
+    int findBestSplitForFeatureSorted(algorithmFPType * featureBuf, IndexType iFeature, const IndexType * aIdx, size_t n, size_t nMinSplitPart,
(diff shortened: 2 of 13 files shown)


ahuber21:
The work is not fully finished. For instance, I do not know if / how we should escalate the error if `rng.uniform()` does not work. There is a `// TODO` comment in the latest commit.

cpp/daal/src/externals/service_memory.h
```diff
@@ -138,6 +138,17 @@ void service_memset_seq(T * const ptr, const T value, const size_t num)
     }
 }
 
+template <typename T, CpuType cpu>
+void service_memset_ser(T * const ptr, const T startValue, const size_t num)
```
Alexsandruss, line 142: Not obvious meaning of function name. Maybe, `service_memset_range`?

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -616,10 +615,10 @@ protected:
     void chooseFeatures()
     {
         const size_t n    = nFeatures();
-        const size_t nGen = (!_par.memorySavingMode && !_maxLeafNodes && !_useConstFeatures) ? n : _nFeaturesPerNode;
+        const size_t nGen = (!_par.memorySavingMode && !_maxLeafNodes && !_par.useConstFeatures) ? n : _nFeaturesPerNode;
```
icfaust, line 618: Make sure you add this into https://github.com/oneapi-src/oneDAL/blob/master/cpp/daal/include/algorithms/decision_forest/decision_forest_training_parameter.h
and
https://github.com/oneapi-src/oneDAL/blob/master/cpp/daal/src/algorithms/dtrees/forest/df_training_parameter.cpp
otherwise it will always evaluate to false.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -810,7 +813,7 @@ typename DataHelper::NodeType::Base * TrainBatchTaskBase<algorithmFPType, BinInd
         typename DataHelper::NodeType::Base * left =
             buildDepthFirst(s, iStart, split.nLeft, level + 1, split.left, bUnorderedFeaturesUsed, nClasses, split.leftWeights);
         _helper.convertLeftImpToRight(n, curImpurity, split);
-        if (!_par.memorySavingMode && !_useConstFeatures)
+        if (!_par.memorySavingMode && !_par.useConstFeatures)
```
icfaust, line 816: See comment above about parameters.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -1084,22 +1087,63 @@ bool TrainBatchTaskBase<algorithmFPType, BinIndexType, DataHelper, cpu>::findBes
...
+        /* _aFeatureIdx[swapIdx] was drawn */
+        _aFeatureIdx[i] = _aFeatureIdx[swapIdx];
+        /* swap in number at [2 * maxFeatures - 1 - i] for next draw */
+        _aFeatureIdx[swapIdx] = _aFeatureIdx[2 * maxFeatures - 1 - i];
+        /* store drawn number at end of number buffer so that no number is lost */
+        _aFeatureIdx[2 * maxFeatures - 1 - i] = _aFeatureIdx[i];
+
         const auto iFeature            = _aFeatureIdx[i];
-        const bool bUseIndexedFeatures = (!_par.memorySavingMode) && (fact > qMax * float(_helper.indexedFeatures().numIndices(iFeature)));
+        const bool bUseIndexedFeatures = (!_par.memorySavingMode) && (float(n) > qMax * float(_helper.indexedFeatures().numIndices(iFeature)));
```
icfaust, line 1144: I'm debating we set the factor to some sort of bit divide, .02 is rather arbitrary, and maybe we just left bitshift >>6 or >>5 or something.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
ahuber21: Thanks! I forgot to initialize it, but now it's included everywhere

cpp/daal/src/externals/service_memory.h
ahuber21: I went with `service_memset_incrementing`, do you agree?

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
ahuber21: Based on?
Not trying to argue, just want to emphasize that the diff is only there because I refactored something. The 0.02 was there before.

ahuber21:
Hi @Alexsandruss and @samir-nasibli could you comment on the test failures?
As far as I can tell, we are comparing against previous versions of the algorithm, not some sort of ground truth.
Since I changed to node splitting algorithm, it is expected that the resulting trees in the forest will have different numerics.
Is there any way I can confirm the results are statistically still correct?

Is it okay to update `decision_forest_regression_batch.csv` with the updated numbers?

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -1084,22 +1084,63 @@ bool TrainBatchTaskBase<algorithmFPType, BinIndexType, DataHelper, cpu>::findBes
...
-    int iBestSplit               = -1;
-    int idxFeatureValueBestSplit = -1; //when sorted feature is used
+    /* counter of the number of visited features, we visit _nFeaturesPerNode
+    *  depending on _par.useConstFeatures constant features can be skipped
+    */
+    size_t nVisitedFeature = 0;
+    /* total number of features */
+    const size_t maxFeatures = nFeatures();
+    /* (??) min fracture of observations to be handled as indexed feature values */
+    const float qMax = 0.02;
```
KulikovNikita, line 1094: This number is looking weird t me. I see that it have been in place for a quite some time already but it still looks magic. Should we consider changing this thing? (Not in this PR though)

KulikovNikita: And another question is why it is `float` not `algorithmFPType`?

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -1084,22 +1084,63 @@ bool TrainBatchTaskBase<algorithmFPType, BinIndexType, DataHelper, cpu>::findBes
...
+    /* index of the best split, initialized to first index we investigate */
+    IndexType * bestSplitIdx = featureIndexBuf(0) + iStart;
+    /* sample index */
+    IndexType * aIdx = _aSample.get() + iStart;
+    /* zero-based index of best split */
+    int iBestSplit = -1;
```
KulikovNikita, line 1100: Using integers and any size specified types (i.e. `int` instead of `std::int32_t`) is generally banned from library. At least some sanity checks and analysis should be performed to guarantee that integer overflow doesn't take place here.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
KulikovNikita: > I'm debating we set the factor to some sort of bit divide, .02 is rather arbitrary, and maybe we just right bitshift >>6 or >>5 or something.

I believe it is just a premature optimization =). 

Btw it must convert to floating point type `n` value anyway because C++ implicitly converts variables to higher types. My concern though is that we use fp32 instead of arbitrary `algorithmFPType`.

cpp/daal/src/externals/service_memory.h
```diff
@@ -138,6 +139,18 @@ void service_memset_seq(T * const ptr, const T value, const size_t num)
     }
 }
 
+/* Initialize block of memory of length num with entries [startValue, ..., startValue + num -1]*/
+template <typename T, CpuType cpu>
+void service_memset_incrementing(T * const ptr, const T startValue, const size_t num)
+{
+    PRAGMA_IVDEP
+    PRAGMA_VECTOR_ALWAYS
+    for (size_t i = 0; i < num; i++)
```
KulikovNikita, line 148:
```suggestion
    for (size_t i = 0; i < num; ++i)
```

It should not affect performance because of how smart compilers are. But it still will help to maintain consistency.

KulikovNikita: Btw we can consider using `std::iota`. I believe using standard library bits is generally not approved but in this case it is somewhat justified because it is pure header only functionality. We can think about it later as well as part of our strategy for simplification of the DAAL code maintenance.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
ahuber21: All good points. See also Ian's comments above. I'm happy to change this with this PR. But I don't have any opinion on "how" to change it.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
ahuber21: As you probably noticed, this also has been around for a long time. If we use `BinIndexType` instead, we're save. See `df_classification_train_dens_default_impl.i`, where `indexedFeatures.maxNumIndices()` is checked _or_ `dtrees::internal::IndexedFeatures::IndexType` is used.

Not, however, that `IndexType` has a todo -- not sure if that's relevant here

```c++
// found in dtrees_feature_type_helper.h:89
    typedef int IndexType; // TODO: should be unsigned int
```

cpp/daal/src/externals/service_memory.h
ahuber21: `i++` is used everywhere in `service_memory.h`. Should I replace all occurences?

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
KulikovNikita: I don't know either. Maybe it should make this code supporting larger number of samples.

ahuber21: Should I change it?

ahuber21:
I have created a PR in scikit-learn-intelex to follow up on the CI failures
https://github.com/intel/scikit-learn-intelex/pull/1213

@Alexsandruss could you sign off on the changes requested in this PR?

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
```diff
@@ -117,26 +117,26 @@ public:
     bool terminateCriteria(ImpurityData & imp, algorithmFPType impurityThreshold, size_t nSamples) const { return imp.value() < impurityThreshold; }
 
     template <typename BinIndexType>
-    int findBestSplitForFeatureSorted(algorithmFPType * featureBuf, IndexType iFeature, const IndexType * aIdx, size_t n, size_t nMinSplitPart,
-                                      const ImpurityData & curImpurity, TSplitData & split, const algorithmFPType minWeightLeaf,
-                                      const algorithmFPType totalWeights, const BinIndexType * binIndex) const;
+    int64_t findBestSplitForFeatureSorted(algorithmFPType * featureBuf, IndexType iFeature, const IndexType * aIdx, size_t n, size_t nMinSplitPart,
```
Alexsandruss, line 120: What is the need to change this type from 32 to 64 bytes?

Alexsandruss (changes requested):

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
ahuber21: As @KulikovNikita said, using vanilla `int` is banned from the library. `IndexType` as of today can be `uint8_t`, `uint16_t`, `int` (for `int` see `typedef int IndexType`). If we change `int` to `uint32_t` as the `TODO` requests, `int32_t` might overflow. Note that this is only used for local variables on the stack, no we're not talking about big chunks of memory here.

Alexsandruss: `int` is not banned in DAAL kernels. Otherwise, most part of theirs code is banned. I guess there should be separate discussion for DAAL

icfaust: There is a defined IndexType used within RF which probably should be used. Unfortunately the code uses it inconsistently.

icfaust: To be honest, the whole reason why its returning an int is poorly implemented, as it is used as a quasi-flag to trigger finalizeBestSplit. Why this roundabout logic was used is beyond me.  I think we should have a talk about to what level RF needs to be refactored vs just repaired. I think some of the issues such as the constFeatures is a symptom of too much tech debt.

ahuber21: Reverted it, but I don't think I agree with keeping the return type `int`. At the very least, we're using `size_t`, `IndexType`, and raw integer types inconsistently, and this discussion didn't explain why. If that should be refactored separately, fine.

cpp/daal/include/algorithms/decision_forest/decision_forest_training_parameter.h
```diff
@@ -134,6 +134,7 @@ class DAAL_EXPORT Parameter
                                                  Default is 256. Increasing the number results in higher computation costs */
     size_t minBinSize;                     /*!< Used with 'hist' split finding method only.
                                                  Minimal number of observations in a bin. Default is 5 */
+    bool useConstFeatures;                 /*!< Use or ignore constant-valued features when splitting nodes. Default is false */
```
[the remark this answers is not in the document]
ahuber21: Tried to add it here: https://github.com/oneapi-src/oneDAL/pull/2292/commits/6ba776fdbad071f08a0c506a7c45925101a7e402

I hope I found all relevant places. Let me know if you see something obvious missing. Thank you!

Alexsandruss:
Failures should be investigated and fixed on daal4py side:
```
FAIL: test_decision_forest_regression_default_dense_batch (test_examples.TestExNpyArray)
FAIL: test_decision_forest_regression_hist_batch (test_examples.TestExNpyArray)
FAIL: test_decision_forest_regression_default_dense_batch (test_examples.TestExPandasDF)
FAIL: test_decision_forest_regression_hist_batch (test_examples.TestExPandasDF)
```

```
=================================== FAILURES ===================================
_______________ test_classifier_class_weight_iris[class_weight5] _______________
>       assert ratio <= LOG_LOSS_RATIO, reason
E       AssertionError: Classifier class weight: scikit_log_loss=0.10065237679616006, daal4py_log_loss=0.14592264814110162
E       assert 1.4497685279367258 <= 1.4
```

Alexsandruss:
9% quality metric drop for some datasets looks unacceptable. Did you test datasets from scikit-learn_bench/configs/testing/metrics/rf_*.json?

ahuber21:
I'm running the tests right now. But regardless, could we split these issues? There is a critical underperformance in the current implementation. Compared to the stock version, the latest intelex release has a 25 % worse MSE on the airline regression dataset and a factor of 3 (!!) is reported in https://github.com/intel/scikit-learn-intelex/pull/1213
Technically, the old sampling was not uniformly random, which is why it is expected that we some changes

Alexsandruss (approved):
Merge when fix on Python side is ready

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -1084,20 +1086,61 @@ bool TrainBatchTaskBase<algorithmFPType, BinIndexType, DataHelper, cpu>::findBes
...
+        // replacement based on Fisher Yates sampling
+        // drawing uniformly from [0, maxFeatures-i] and swapping the indices
+        // assures uniform probability of all drawn numbers
+
+        /* draw the i-th index of the sample */
+        int errorcode = rng.uniform(1, &swapIdx, _engineImpl->getState(), 0, maxFeatures - i);
+        if (errorcode)
+        {
+            return false;
+        }
```
Vika-F, line 1130: The correct behavior would be to return Status (or SafeStatus if this code is running in parallel region), not bool, from findBestSplit* functions. This Status will be able to contain a description of the error. Maybe a new error code also needs to be added to handle failures in random number generators.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -619,7 +619,7 @@ protected:
         const size_t nGen = (!_par.memorySavingMode && !_maxLeafNodes && !_useConstFeatures) ? n : _nFeaturesPerNode;
         *_numElems += n;
         RNGs<IndexType, cpu> rng;
-        rng.uniformWithoutReplacement(nGen, _aFeatureIdx.get(), _aFeatureIdx.get() + nGen, _engineImpl->getState(), 0, n);
+        rng.drawKFromBufferWithoutReplacement(nGen, _aFeatureIdx.get(), _aFeatureIdx.get() + nGen, _engineImpl->getState(), n);
```
Vika-F, line 622: drawKFromBufferWithoutReplacement returns error code, but it is not checked here.
Please see the other comment in this review that describes more correct way of handling errors in DAAL.

Vika-F (commented):
The changes look good. Special thanks for adding comments.
But it seems that the error handling needs to be improved.

ahuber21:
Force-pushed after rebase. @Vika-F will add the status checks next.

ahuber21:
@Vika-F I have added status checks with minimal changes, running a few performance benchmarks, just to make sure nothing changed.

PR #2296, opened 2023-03-16
Update environmental scripts to support Toolkit changes

maria-Petrova:
- Add additional ONEAPI_ROOT check to see if the variable is defined
- Set DAALROOT if ONEAPI_ROOT is defined


deploy/local/vars_lnx.sh
```diff
@@ -216,7 +222,7 @@ my_script_path=$(get_script_path "${vars_script_name:-}")
 component_root=$(dirname -- "${my_script_path}")
 
 __daal_tmp_dir="<INSTALLDIR>"
-__daal_tmp_dir=$__daal_tmp_dir/dal
+__daal_tmp_dir=$__daal_tmp_dir
```
napetrov, line 225: guess we don't need this line

maria-Petrova: agree, let's remove it

napetrov (approved):

PR #2310, opened 2023-04-05
KNN distance voting infer fix for train sample in test sample

jmhatre-work:
# Description

-Changes to make inference for test sample when it is present in train sample , such that it should infer label as class same as train sample
-Bazel test for this issue.
-Fix for both cpu and gpu kernel implementations.


KulikovNikita (approved):

cpp/daal/src/algorithms/k_nearest_neighbors/bf_knn_impl.i
```diff
@@ -417,7 +417,7 @@ protected:
             {
                 for (size_t j = 0; j < k; ++j)
                 {
-                    if (distances[i] < epsilon)
+                    if (distances[j] < epsilon)
```
ethanglaser, line 420: Good catch

ethanglaser (approved):
Approved pending internal CI

PR #2314, opened 2023-04-16
DAAL random splitter bugfix histogram size_t underflow

icfaust:
# Description
Corrects an error with random histogram selection when the minidx =0, it causes minidx to underflow size_t.

- Swaps the order for proper logic in the ternary operator, this is in training df_regression and df_classification. The ternary checks if minidx is 0 by using the fact that (bool) of 0 is False, and then doesn't do the subtraction and instead uses other logic.
-
-

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
@@ -996,7 +996,7 @@ size_t UnorderedRespHelperRandom<algorithmFPType, cpu>::genRandomBinIdx(const In
 {
     //randomly select a histogram split index
     algorithmFPType fidx   = 0;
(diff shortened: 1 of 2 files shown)


icfaust:
To be honest, I am a bit surprised this didn't cause seg faults. I guess I need to understand underlying DAAL data arrays better.

icfaust:
Third time's the charm. Please see most recent internal CI run for valid testing. http://intel-ci.intel.com/edde1f60-ef96-f14f-a1ff-a4bf010d0e2e  CI fails for reasons outside of the PR changes.

Vika-F (approved):
The changes look good to me.

The only comment is that this getRandomBinIdx method is duplicated in regression and classification, I've missed that during the previous reviews. Please consider moving it into some common part of the implementation. This can be done in a separate PR.

inteldimitrius:
Changes looks good to me. But I have insufficient rights to approve it.

PR #2317, opened 2023-04-20
Kmeans doc update

olegkkruglov:
# Description
Kmeans doc update

- Fix stop condition description
- Add empty cluster handling description
- Add SPMD mode mentioning
- Docstring fix for K-means

### DIFF
--- a/docs/source/onedal/algorithms/clustering/kmeans.rst
+++ b/docs/source/onedal/algorithms/clustering/kmeans.rst
@@ -126,6 +126,12 @@ Programming Interface
 
 Refer to :ref:`API Reference: K-Means <api_kmeans>`.
 
+----------------
+Distributed mode
+----------------
+
+The algorithm supports distributed execution in SPMD mode (only on GPU).
+
 -------------
 Usage example
 -------------
--- a/docs/source/onedal/algorithms/clustering/kmeans.rst
+++ b/docs/source/onedal/algorithms/clustering/kmeans.rst
@@ -85,6 +85,12 @@ to each cluster.
 .. math::
    c_j^{(t + 1)} = \frac{1}{|S_j^{(t)}|} \sum_{x \in S_j^{(t)}} x, \quad 1 \leq j \leq k.
 
+If it turned out that some of :math:`S_j^{(t)}` are empty then start empty clusters handling procedure.
+In this case set :math:`c_j^{(t + 1)}` equal to the farthest point from just calculated centroids for each empty cluster.
+
+.. math::
+   c_j^{(t + 1)} = \mathrm{arg}\max_{x \in X} d^2(x, C^{(t + 1)}).
+
 The steps (1) and (2) are performed until the following **stop condition**,
 
 .. math::
--- a/docs/source/onedal/algorithms/clustering/kmeans.rst
+++ b/docs/source/onedal/algorithms/clustering/kmeans.rst
@@ -88,7 +88,7 @@ to each cluster.
 The steps (1) and (2) are performed until the following **stop condition**,
 
 .. math::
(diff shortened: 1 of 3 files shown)


docs/source/onedal/algorithms/clustering/kmeans.rst
```diff
@@ -85,10 +85,16 @@ to each cluster.
 .. math::
    c_j^{(t + 1)} = \frac{1}{|S_j^{(t)}|} \sum_{x \in S_j^{(t)}} x, \quad 1 \leq j \leq k.
 
+If it turned out that some of :math:`S_j^{(t)}` are empty then start empty clusters handling procedure.
```
alexandraepan, line 88:
```suggestion
If some of :math:`S_j^{(t)}` are empty, start the empty clusters handling procedure.
```

alexandraepan: Not clear what is meant by "some". If even one empty thing triggers the procedure, it is better to use "any". If it is more than one, we should clarify what the minimal condition is.

docs/source/onedal/algorithms/clustering/kmeans.rst
```diff
@@ -85,10 +85,16 @@ to each cluster.
...
+In this case set :math:`c_j^{(t + 1)}` equal to the farthest point from just calculated centroids for each empty cluster.
```
alexandraepan, line 89: Hmm maybe something like:

In this case, you can set the value of :math:c_j^{(t+1)} as the farthest point from the previously calculated centroids for each empty cluster. It fills in the empty clusters with new data points and makes sure that all clusters have a representative centroid.

docs/source/onedal/algorithms/clustering/kmeans.rst
```diff
@@ -120,6 +126,12 @@ Programming Interface
 
 Refer to :ref:`API Reference: K-Means <api_kmeans>`.
 
+----------------
+Distributed mode
+----------------
+
+The algorithm supports distributed execution in SPMD mode (only on GPU).
```
alexandraepan, line 133:
```suggestion
The algorithm supports distributed execution in the SPMD mode and can be run on a GPU only.
```

olegkkruglov: Moreover, I have copied this sentence from other section which has been approved. It is necessary to change there too if something is incorrect

olegkkruglov: I don't think so. The bracketed statement applies to the sentence before it and this sentence declares only that the algorithm can be run in distributed SPMD mode (as far as I understand distributed and SPMD are synonyms in this context). So, the only conclusion can be drawn is that the distributed mode is supported only for GPU.

Your last suggestion is also correct, and if we are trying to decrease brackets number in text or something like that then I can push your version :) But as for me, the current version looks more briefly and neatly.

docs/source/onedal/algorithms/clustering/kmeans.rst
olegkkruglov: done

docs/source/onedal/algorithms/clustering/kmeans.rst
olegkkruglov: I'm not really sure that it makes sense to write 'fill' here because we are not filling the new clusters at this step. 
If you want to highlight the purpose of the procedure then what about this:

In this case, you can set the value of :math:c_j^{(t+1)} as the farthest point from the previously calculated centroids for each empty cluster. This procedure makes sure that the number of cluster remains the same.

alexandraepan: I like your suggestion!

olegkkruglov: done

alexandraepan (approved):
LGTM!

PR #2321, opened 2023-04-23
Linear Regression compilation problem fix

KulikovNikita:


avolkov-intel (approved):

ethanglaser:
I have faced similar errors when running kNN bazel tests, and these changes resolve those errors. I don't think the private CI error is related. I am still a bit confused about the origin of these issues, but I suggest we merge this.

PR #2345, opened 2023-05-04
Missing values support for inference

razdoburdin:
# Description
The presented PR adding support of missing_values for GBT inference. Some minor optimizations were also made. Benchmarking for xgboost and lightgbm demonstrate only minor (<= 5%) performance degradation for some datasets. This degradation is coming from checking for nan-values in the data.

- We add additional array containing the default direction for each node. If missing values is taking place, this default direction is used instead of analyze of the split condition. Such approach is equivalent to ones used in XGBoost and LightGBM.
- Minor optimizations are related templatizations of if statements. Them allow to improve performance for small size datasets. (up to 30% speedup for 'letters' dataset in case of xgb models).


cpp/daal/include/algorithms/gradient_boosted_trees/gbt_classification_model_builder.h
```diff
@@ -123,10 +123,10 @@ class DAAL_EXPORT ModelBuilder
     *  \param[in] featureValue    Feature value for splitting
     *  \return Node identifier
     */
-    NodeId addSplitNode(TreeId treeId, NodeId parentId, size_t position, size_t featureIndex, double featureValue)
+    NodeId addSplitNode(TreeId treeId, NodeId parentId, size_t position, size_t featureIndex, double featureValue, int yes_if_missing = 0)
```
samir-nasibli, line 126: I think, these `yes_if_missing`/`yesIfMissing` naming just confusing. Doesn't be worth rename it like `any_missing_values`/`if_missing`/`any_missing`?

razdoburdin: Thanks for your feedback!
The meaning of this parameter can be described as following: "In case of missing value take place, one should go to "yes" way in the split point". For this reason I guess, `yes_if_missing` is more clear naming than `any_missing/if_missing/any_missing_values`. May be we could try to find better replacement.

ahuber21: Also, we're mixing `snake_case` and `camelCase` here. Please avoid that. Following your description. I would avoid putting details of the implementation in the variable name (what are the `yes` and `no` ways anyways? depends on our implementation!), and rather name it according to the expected effect. What about `splitIfValueIsMissing` or similar? And why is it an `int` and not a `bool` then?

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
```diff
@@ -178,8 +178,103 @@ public:
...
+    void predictByTreesVector(algorithmFPType * val, size_t iFirstTree, size_t nTrees, size_t nClasses, const algorithmFPType * x,
+                              const Dispetcher_t<hasUnorderedFeatures, hasAnyMissing>& dispetcher);
+
+    template <bool val>
+    struct BooleanConstant {
+        typedef BooleanConstant<val> type;
+    };
+
+    inline void updateResult(algorithmFPType * res, algorithmFPType * val, size_t iRow, size_t i, size_t nClasses, BooleanConstant<true> dispetcher) {
+        res[iRow + i] = algorithmFPType(getMaxClass(val + i * nClasses, nClasses));
```
ahuber21, line 196: Is the `algorithmFPType()` necessary? Wouldn't the assignment convert it anyways?

ahuber21: Also typo: `dispetcher` -> `dispatcher`.

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
```diff
@@ -178,8 +178,103 @@ public:
...
+    }
+
+    inline void updateResult(algorithmFPType * res, algorithmFPType * val, size_t iRow, size_t i, size_t nClasses, BooleanConstant<false> dispetcher) {}
```
ahuber21, line 199: Same typo (happens in even more places that I won't comment). `dispetcher` -> `dispatcher`.

Why is the implementation empty?

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
```diff
@@ -178,8 +178,103 @@ public:
...
+
+    inline algorithmFPType* updateBuffer(algorithmFPType * val, size_t buf_shift, size_t buf_size, BooleanConstant<true> dispetcher) {
+        services::internal::service_memset_seq<algorithmFPType, cpu>(val, algorithmFPType(0), buf_size);
+        return val;
+    }
+
+    inline algorithmFPType* updateBuffer(algorithmFPType * val, size_t buf_shift, size_t buf_size, BooleanConstant<false> dispetcher) {
```
ahuber21, line 206: I don't know how this is handled in the remaining part of the codebase, but seeing these two `updateBuffer` functions confuses me. We have specialization implementations for the `BooleanConstant` value and depending on that we apply the `buf_shift` or not? This is completely confusing. In the first implementation `buf_shift` isn't even used. And neither implementation uses `dispetcher` (again, typo).
Why isn't this something like

```c++
inline algorithmFPType* initAndUpdateBuffer(algorithmFPType * val, size_t buf_size);
inline algorithmFPType* updateBuffer(algorithmFPType * val, size_t buf_shift, size_t buf_size);
```

What is the use of the second function? Can't we just calculate val + buf_shift in place?

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
```diff
@@ -178,8 +178,103 @@ public:
...
+    }
+
+    template <bool hasUnorderedFeatures, bool hasAnyMissing, bool isResValidPtr, bool reuseBuffer>
+    inline void predict(size_t nTrees, size_t nClasses, size_t nRows, size_t nColumns,
+                        const algorithmFPType * x, algorithmFPType * valL, algorithmFPType * res) {
+        Dispetcher_t<hasUnorderedFeatures, hasAnyMissing> dispetcher;
+        size_t iRow = 0;
+        for (; iRow + VECTOR_BLOCK_SIZE <= nRows; iRow += VECTOR_BLOCK_SIZE)
+        {
+            algorithmFPType* val = updateBuffer(valL, iRow * nClasses, nClasses * VECTOR_BLOCK_SIZE, BooleanConstant<reuseBuffer>());
```
ahuber21, line 233: Isn't it simpler to just use a ternary expression here?

```c++
val = reuseBuffer ? valL + iRow * nClasses : services::internal::service_memset_seq<algorithmFPType, cpu>(valL, algorithmFPType(0), iRow * nClasses );
```

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
```diff
@@ -178,8 +178,103 @@ public:
...
+            for (size_t idx = 0; idx < nRows * nColumns; ++idx) {
+                if (isnan(x[idx])) return true;
+            }
+        }
+        return false;
+    }
+
+    template <bool hasUnorderedFeatures, bool hasAnyMissing, bool isResValidPtr, bool reuseBuffer>
+    inline void predict(size_t nTrees, size_t nClasses, size_t nRows, size_t nColumns,
+                        const algorithmFPType * x, algorithmFPType * valL, algorithmFPType * res) {
```
ahuber21, line 228: What is `valL` ? Seeing this in isolation, it's not clear to me. Can you spell out the name, or provide some doc lines above the function?

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
```diff
@@ -353,4 +413,4 @@ services::Status PredictMulticlassTask<algorithmFPType, cpu>::predictByAllTrees(
 } /* namespace algorithms */
 } /* namespace daal */
 
-#endif
+#endif
```
ahuber21, line 416: Missing newline

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
```diff
@@ -45,46 +45,68 @@ typedef float ModelFPType;
...
 
-template <typename algorithmFPType, typename DecisionTreeType, CpuType cpu>
-inline void predictForTreeVector(const DecisionTreeType & t, const FeatureTypes & featTypes, const algorithmFPType * x, algorithmFPType v[])
+template <bool hasUnorderedFeatures, bool hasAnyMissing>
+struct PredictDispetcher {
+    typedef PredictDispetcher<hasUnorderedFeatures, hasAnyMissing> type;
+};
+
+template<typename algorithmFPType>
+inline FeatureIndexType UpdateIndex(FeatureIndexType idx, algorithmFPType valueFromDataSet, const ModelFPType * splitPoints, const int * yesIfMissing,
```
ahuber21, line 54: Function names should start with small letters, i.e. `updateIndex()`

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -174,4 +229,4 @@ void PredictRegressionTask<algorithmFPType, cpu>::predictByTreesVector(size_t iF
 } /* namespace algorithms */
 } /* namespace daal */
 
-#endif
+#endif
```
ahuber21, line 232: Missing newline

ahuber21 (changes requested):

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
razdoburdin: > Is the `algorithmFPType()` necessary? Wouldn't the assignment convert it anyways?

you are right! Thanks.

> Also typo: `dispetcher` -> `dispatcher`.
done

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
razdoburdin: In the best of my understanding, in your realization we can't be sure in optimization of `if `statement by compiler. Am I wrong?

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
razdoburdin: done

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
razdoburdin: done

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
razdoburdin: done

cpp/daal/include/algorithms/gradient_boosted_trees/gbt_classification_model_builder.h
ahuber21: How is the `bool` vs. `int` problem solved. We could convert to `bool` right at the interface, so that at least all C++ code would be consistent.

I like `defaultLeft` !

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
razdoburdin: In fact we have 4 such comparisons. And they are in the critical section of the code. Such micro optimization become important for test cases with small number of rows. For example, inference speed of letters was increased on 30% for xgboost model builders. Unfortunately for testcases with large number of rows performance gain doesn't exceed few precents.

cpp/daal/include/algorithms/gradient_boosted_trees/gbt_classification_model_builder.h
razdoburdin: > How is the bool vs. int problem solved. We could convert to bool right at the interface, so that at least all C++ code would be consistent.

This values should be stored in `HomogenNumericTable`. As far as I see, one can't use `HomogenNumericTable<bool>`.

ahuber21 (approved):

napetrov:
There are doxygen warnings that should be addressed.

napetrov (approved):

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
```diff
@@ -45,46 +45,79 @@ typedef float ModelFPType;
...
+                                    const FeatureTypes & featTypes, FeatureIndexType splitFeature, const PredictDispatcher<false, false> & dispatcher)
+{
+    return idx * 2 + (valueFromDataSet > splitPoints[idx]);
+}
+
+template <typename algorithmFPType>
+inline FeatureIndexType updateIndex(FeatureIndexType idx, algorithmFPType valueFromDataSet, const ModelFPType * splitPoints, const int * defaultLeft,
+                                    const FeatureTypes & featTypes, FeatureIndexType splitFeature, const PredictDispatcher<true, false> & dispatcher)
+{
+    // return idx * 2 + (isUnordered ? int(valueFromDataSet) != int(splitPoints[idx]) : valueFromDataSet > splitPoints[idx]); //???///
```
samir-nasibli, line 65: Please remove it if not necessary

samir-nasibli (approved):

PR #2346, opened 2023-05-04
kmeans (feature): split select function by blocks

Alexandr-Solovev:
This splitting is necessary to support huge datasets in KMeans


cpp/oneapi/dal/algo/kmeans/backend/gpu/kernels_fp_impl.hpp
```diff
@@ -152,53 +152,65 @@ sycl::event kernels_fp<Float>::select(sycl::queue& queue,
 
     auto event = queue.submit([&](sycl::handler& cgh) {
         cgh.depends_on(deps);
+
+        const std::int64_t block_size = 4096;
```
samir-nasibli, line 156: It is better to use predefined value here instead hardcoding or just leave comment before.

samir-nasibli (approved):

PR #2350, opened 2023-05-05
CSR tests

inteldimitrius:
# Description
Tests for CSR table added.


cpp/oneapi/dal/table/test/csr_serialization.cpp
```diff
@@ -0,0 +1,166 @@
...
+    const std::int64_t row_count{ 8 };
+    const std::int64_t column_count{ 8 };
+    const std::int64_t element_count{ 9 };
+
+    csr_table init_table{ array<float>::wrap(data, element_count),
+                          array<std::int64_t>::wrap(column_indices, element_count),
+                          array<std::int64_t>::wrap(row_indices, row_count + 1),
+                          column_count,
+                          sparse_indexing::zero_based };
+    std::cout << "TABLE successfully created" << std::endl;
```
ethanglaser, line 94: I see this in just one test, I would probably remove, or add to all if its necessary.

ethanglaser (approved):
Looks good to me but will also defer to Vika and those more familiar with CSR here.

cpp/oneapi/dal/table/test/csr_serialization.cpp
samir-nasibli: please remove debug print.

PR #2354, opened 2023-05-08
Deprecate daal, java and a part of onedal examples

maria-Petrova:
The PR aims to exclude the following examples from OS and binary distributions:
- daal examples (all)
- onedal:
  -  make (all OSes)
  - MSVS solutions for windows
- java examples


examples/daal/.clang-format
```diff
@@ -1,145 +0,0 @@
----
-Language: Cpp
```
napetrov, line 2: i think we still need clang-format here

PR #2356, opened 2023-05-08
Deprecate make/nmake/sln  C++/SYCL examples

maria-Petrova:
The PR aims to exclude the following daal/onedal examples from OS and binary distributions:
-  make (all OSes)
- MSVS solutions on Windows


makefile
```diff
@@ -1049,10 +1048,10 @@ $(RELEASEDIR.daal)/$1/$2: \
 	python ./deploy/local/generate_win_solution.py $1 $(RELEASEDIR.daal)/$1 $2 --template_name $3
```
napetrov, line 1048: i think that entire release VS solution should be removed(1035 - 1055) as well as ./deploy/local/generate_win_solution.py

maria-Petrova: Thanks Nikolay. Fixed.

napetrov:
./deploy/local/generate_win_solution.py  should be removed as well

maria-Petrova:
> ./deploy/local/generate_win_solution.py should be removed as well

do we want to remove makefile generator as sell?

napetrov:
> > ./deploy/local/generate_win_solution.py should be removed as well
> 
> do we want to remove makefile generator as sell?

yes, let's also clean it

maria-Petrova:
Agreed with @homksei that he will address the issues in testing.

napetrov (approved):

PR #2358, opened 2023-05-08
constFeature implementation for DAAL RF memorySavingMode

icfaust:
# Description
DAAL RF code currently only allows for constant features to not be used when memorySavingMode = False.  This uses the splitting functions themselves to find constant Features and add them to the constant feature list.

- Adds constFeature bool to findBestSplitForFeature()
- Adds code for checking for constant data to CategoricalFeature and OrderedFeature in splitter classes in Regression and Classification
- Removes certain if statement memorySaveMode checks

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
@@ -461,6 +461,7 @@ bool UnorderedRespHelperBest<algorithmFPType, cpu>::findSplitCategoricalFeature(
     const bool bBestFromOtherFeatures = !(split.impurityDecrease < 0);
     algorithmFPType vBest             = -1;
     IndexType iBest                   = -1;
+    algorithmFPType first;
 
     const algorithmFPType vBestFromOtherFeatures = bBestFromOtherFeatures ? totalWeights * (curImpurity.var - split.impurityDecrease) : -1;
     for (size_t i = 0; i < n - nMinSplitPart;)
@@ -469,7 +470,7 @@ bool UnorderedRespHelperBest<algorithmFPType, cpu>::findSplitCategoricalFeature(
         auto weights                = this->_aWeights[aIdx[i]].val;
         size_t count                = 1;
         algorithmFPType leftWeights = weights;
-        const algorithmFPType first = featureVal[i];
+        first                       = featureVal[i];
         ClassIndexType xi           = this->_aResponse[aIdx[i]].val;
         _impLeft.hist[xi]           = weights;
         const size_t iStart         = i;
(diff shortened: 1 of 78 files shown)


cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
```diff
@@ -447,14 +452,15 @@ bool UnorderedRespHelperBest<algorithmFPType, cpu>::findSplitCategoricalFeature(
...
-                                                                                const algorithmFPType totalWeights) const
+                                                                                const algorithmFPType totalWeights, bool & constFeature) const
 {
     DAAL_ASSERT(n >= 2 * nMinSplitPart);
     _impRight.init(_nClasses);
     bool bFound                       = false;
     const bool bBestFromOtherFeatures = !(split.impurityDecrease < 0);
     algorithmFPType vBest             = -1;
     IndexType iBest                   = -1;
+    algorithmFPType first;
```
KulikovNikita, line 463:
```suggestion
    algorithmFPType first{};
```
Static analyzers don't like uninitialized variables

KulikovNikita (approved):

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
icfaust: Thanks for the suggestion, made changes related to it.

ahuber21 (changes requested):
Did you run the usual performance benchmarks? Are they unchanged?

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
```diff
@@ -500,6 +506,9 @@ bool UnorderedRespHelperBest<algorithmFPType, cpu>::findSplitCategoricalFeature(
         split.featureValue = first;
         bFound             = true;
     }
+
+    constFeature = first == featureVal[0];
```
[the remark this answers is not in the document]
icfaust: _featHelper is assigned to a value definfed by const reference meaning that a common version is used by all trees.  We would need to duplicate this object for all trees and then modify the underlying class definition to contain the necessary level and const information.

The use of _aConstFeatureIdx collects data from a separate function (https://github.com/oneapi-src/oneDAL/blob/master/cpp/daal/src/algorithms/dtrees/dtrees_train_data_helper.i#L380, called by https://github.com/oneapi-src/oneDAL/blob/master/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i#L1169 ). At worst case, hasDifferentFeatureValues unnecessarily loads almost the entire feature value vector a second time from memory just for a comparison. This accesses, like histogram, are not vectorizable (as far as I can tell) due to splitting and bootstrapping. This PR attempts to achieve results of finding constant nature of a feature without using IndexedFeatures while maintaining performance (i.e. no second data load).

icfaust:
There seems to be an seg fault issue when running it in sklearn tests. I want to hold off on this until I have more time to find out possible causes. Until then, ExtraTrees will move to the hist method due to maturity in the underlying code.

PR #2378, opened 2023-05-22
fixup: disable failing tests in batch.cpp while fix in progress

ahuber21:
This PR has to be reverted when the root cause of the issue is fixed


samir-nasibli:
Please do not remove test suits just add `SKIP_IF` with comment for corresponding one.

PR #2379, opened 2023-05-22
Fix for kernels selector sed command on macOS

Alexsandruss:
Remove '\b' symbol from sed command on macOS to fix selection of kernels.

### DIFF
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -382,10 +382,10 @@ jobs:
       .ci/scripts/describe_system.sh
     displayName: 'System info'
   - script: |
-      .ci/scripts/build.sh --compiler clang --target daal --optimizations "sse2 sse42 avx2 avx512" --conda-env ci-env
+      .ci/scripts/build.sh --compiler clang --target daal --optimizations "sse2 avx2" --conda-env ci-env
     displayName: 'make daal'
   - script: |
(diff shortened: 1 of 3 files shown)


.ci/pipeline/ci.yml
```diff
@@ -382,10 +382,10 @@ jobs:
       .ci/scripts/describe_system.sh
     displayName: 'System info'
   - script: |
-      .ci/scripts/build.sh --compiler clang --target daal --optimizations "sse2 sse42 avx2 avx512" --conda-env ci-env
```
ahuber21, line 385: I think it would be cleaner to have this in a separate commit. Why was it disabled?

Alexsandruss: @ahuber21 
Previously, subset of ISAs didn't work correctly for building on macOS and all of them were selected in CI. Now, it's changed to SSE2+AVX2

ahuber21 (approved):

ahuber21 (approved):

napetrov (approved):
Assuming CI would pass

KulikovNikita (approved):

Alexsandruss:
http://intel-ci.intel.com/edf8b235-98b2-f170-8462-a4bf010d0e2e

napetrov:
@homksei  - internal ci preview also can drop number of build targets for macOS

PR #2380, opened 2023-05-22
FIX: void function return in csr table

samir-nasibli:
# Description
Fix static scanner hints


Vika-F (approved):
Missed this one during the development. Thank you for fixing, I agree with the changes.

ahuber21 (approved):

PR #2385, opened 2023-05-23
fix: nFeatIdx[idx--] -> nFeatIdx[--idx] to avoid finding bins with ze…

ahuber21:
This PR fixes a segfault that was found in the `ijcnn` and `gisette` performance benchmarks. Without the fix, the `findSplitFewClasses()` algorithm can sometimes return feature bins with zero content, which results in a segfault downstream.

It also refactors unreadable and unnecessary `for` loops to more natural `while` loops and adds `DAAL_ASSERT` statements that were useful while debugging the issue.

Lastly, the function `findSplitbyHistDefault()` was renamed to `findSplitByHistDefault()` (proper `camelCase`).

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
@@ -82,7 +82,7 @@ public:
         : super(indexedFeatures), _nClasses(nClasses), _histLeft(nClasses), _impLeft(nClasses), _impRight(nClasses)
     {}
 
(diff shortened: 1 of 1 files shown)


ahuber21:
https://intel-ci.intel.com/edf972d8-a7b1-f1a2-a5d5-a4bf010d0e2e

Alexsandruss (approved):

Vika-F:
Overall the changes look good to me.
My only concern is about changing `for` loops to `while` loops. Yes, `while` is more readable in this case, but `for` is better recognized by the compiler and has better optimization potential.
I do not know weather those loops are on the performance-critical path, but I'd check extra trees' performance before and after the changes.

PR #2390, opened 2023-05-24
Revert "Fixed copying performance issues (#2375)"

KulikovNikita:

PR #2396, opened 2023-05-30
Initial input for backend selection

amgrigoriev:
# Description
- Single entry for backend selection
- Some reference implementations to replace MKL (naive ones)
- makefile changes - BACKEND_CONFIG was added with two possible options: "mkl" or "ref"
- bazel build system changes to reflect ref implementation
- CI for x86 openBLAS build


dev/make/cmplr.gnu.mkl.mk
```diff
@@ -0,0 +1,51 @@
...
+pedantic.opts.lnx.gnu = $(pedantic.opts.all.gnu)
+
+COMPILER.mac.gnu = $(COMPILER.all.gnu)
+link.dynamic.mac.gnu = $(link.dynamic.all.gnu)
+pedantic.opts.mac.gnu = $(pedantic.opts.all.gnu)
+
+p4_OPT.gnu   = $(-Q)march=nocona
+mc3_OPT.gnu  = $(-Q)march=corei7
+avx2_OPT.gnu = $(-Q)march=haswell
+skx_OPT.gnu  = $(-Q)march=haswell
```
KulikovNikita, line 51:
```suggestion
skx_OPT.gnu  = $(-Q)march=skylake
```

dev/make/cmplr.gnu.ref.mk
```diff
@@ -0,0 +1,51 @@
...
+pedantic.opts.lnx.gnu = $(pedantic.opts.all.gnu)
+
+COMPILER.mac.gnu = $(COMPILER.all.gnu)
+link.dynamic.mac.gnu = $(link.dynamic.all.gnu)
+pedantic.opts.mac.gnu = $(pedantic.opts.all.gnu)
+
+p4_OPT.gnu   = $(-Q)march=nocona
+mc3_OPT.gnu  = $(-Q)march=corei7
+avx2_OPT.gnu = $(-Q)march=haswell
+skx_OPT.gnu  = $(-Q)march=haswell
```
KulikovNikita, line 51:
```suggestion
skx_OPT.gnu  = $(-Q)march=skylake
```

dev/make/cmplr.gnu.mkl.mk
```diff
@@ -0,0 +1,51 @@
+#===============================================================================
+# Copyright 2012 Intel Corporation
```
KulikovNikita, line 2:
```suggestion
# Copyright 2023 Intel Corporation
```

dev/make/cmplr.gnu.ref.mk
```diff
@@ -0,0 +1,51 @@
+#===============================================================================
+# Copyright 2012 Intel Corporation
```
KulikovNikita, line 2:
```suggestion
# Copyright 2023 Intel Corporation
```

cpp/daal/src/externals/service_rng_ref.h
```diff
@@ -0,0 +1,363 @@
...
+        if (_stream != nullptr)
+        {
+            delete _stream;
+            _stream = nullptr;
+        }
+    }
+
+    int getStateSize() const { return _stream->getSize(); }
+
+    int saveState(void * dest) const
```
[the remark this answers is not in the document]
amgrigoriev: Just keeping original signature

rakgb-fujitsu:
This PR fails when `mkl` is used as backend. I get many such errors:

```
./cpp/daal/src/externals/service_math_mkl.h:146:69: error: use of undeclared identifier 'fpk_vml_dExp_L9HAynn'
```

Command used:

`make -f makefile daal oneapi_c BACKEND_CONFIG=mkl PLAT=lnx32e COMPILER=clang -j16`

cpp/daal/src/externals/config_template.h
```diff
@@ -0,0 +1,6 @@
+#ifndef __CONFIG_H__
+#define __CONFIG_H__
+
+#include "src/externals/config_{BACKEND}.h"
```
napetrov, line 4: Should we just change this approach to use defines instead of relaying on sed for replacement?

napetrov:
There are also formatting hits 
+---+---+---+---+---+---+---+---+------
|NLS|TAB|CHR|TAL|EOF|150|EXT|CRT| NAME 
+---+---+---+---+---+---+---+---+------
| U | o | o | o | X | o | o | X | cpp/daal/src/algorithms/kernel_config.h
| U | o | o | o | o | o | o | X | cpp/daal/src/externals/config_ref.h
| U | o | o | o | X | o | o | X | cpp/daal/src/externals/config_template.h
| U | o | o | o | o | o | o | X | cpp/daal/src/externals/service_service_ref.h

EOF ==> not exactly one newline before EOF
CTR ==> wrong name of file before copyright notice

rakgb-fujitsu:
Performance issues, this PR with `ref` as backend fails to run with multithreading, every algorithm seems to use only single thread. Is anything disabled in this path or could it be issues with OpenBLAS? `mkl` backend works fine (uses multithreading) and matches the performance of master.

KulikovNikita:
> Performance issues, this PR with `ref` as backend fails to run with multithreading, every algorithm seems to use only single thread. Is anything disabled in this path or could it be issues with OpenBLAS? `mkl` backend works fine (uses multithreading) and matches the performance of master.

It can probably be related to this lines of the code:
```cpp
        int old_threads = openblas_get_num_threads();
        openblas_set_num_threads(1);
        ...
        openblas_set_num_threads(old_threads);
```
I suspect that it changes the global state of OpenMP and for some reason forces it to run on a specific thread despite what TBB thread assigns it to.

I suspect that we should be able to build it without affinity: https://github.com/xianyi/OpenBLAS/blob/develop/USAGE.md#how-could-i-disable-openblas-threading-affinity-on-runtime

amgrigoriev:
BLAS/LAPACK in oneDAL contain two sets of routines - for instance, xxsyrk and xsyrk. First one is expected to run on a single thread and there is a special handling for that. But for xsyrk OpenBLAS itself is responsible to set necessary number of threads. It could be tuned by OMP_NUM_THREADS env variable. MKL backend contains functionality to detect exact number of available cores but for OpenBLAS it's questionable.

rakgb-fujitsu:
`export OMP_NUM_THREADS =<num_threads>`  at runtime has no impact. This is with openblas built with standard options and another openblas version built with threading completely disabled.  (`make USE_THREAD=0 USE_LOCKING=1 USE_OPENMP=0`)

@KulikovNikita the affinity hack does not have any impact either.

amgrigoriev:
In my experiments with OpenBLAS, performance scaling with OMP_NUM_THREADS was observed. I haven't tried more multithreading options. But it depends on build settings, too.

rakgb-fujitsu:
@amgrigoriev could you please share your build settings?

cpp/daal/src/externals/service_blas_ref.h
```diff
@@ -0,0 +1,255 @@
...
+        int old_threads = openblas_get_num_threads();
+        openblas_set_num_threads(1);
+        daxpy_(n, a, x, incx, y, incy);
+        openblas_set_num_threads(old_threads);
+    }
+
+    static double xxdot(const DAAL_INT * n, const double * x, const DAAL_INT * incx, const double * y, const DAAL_INT * incy)
+    {
+        int old_threads = openblas_get_num_threads();
+        openblas_set_num_threads(1);
```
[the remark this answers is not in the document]
amgrigoriev: It looks like it will not work properly for xxgemm which should work simultaneously on all threads. Thus, locking at the begining and unlocking at the end will result in running xxgemm on a single thread.

cpp/daal/src/externals/service_rng_ref.h
```diff
@@ -0,0 +1,363 @@
...
+        if (rngSeedSize() == 0 || rngSeed() == nullptr) return services::ErrorNullInput;
+        if (!std::is_floating_point<T>::value) return services::ErrorDataTypeNotSupported;
+        std::uniform_real_distribution<double> distrib(static_cast<double>(a), static_cast<double>(b));
+        for (size_t i = 0; i < n; i++)
+        {
+            r[i] = static_cast<T>(distrib(_gen));
+        }
+        _nSkip += n;
+        return errcode;
+    }
```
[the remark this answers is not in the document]
amgrigoriev: Internal distribution is double. Static casting of double to double is not a problem, right?

amgrigoriev:
@rakgb-fujitsu I tested with downloaded OpenBLAS binaries.
But looking into documentation, it's not quite clear how to use it properly for oneDAL (https://github.com/xianyi/OpenBLAS/wiki/Faq#multi-threaded):
"How can I use OpenBLAS in multi-threaded applications?
If your application is already multi-threaded, it will conflict with OpenBLAS multi-threading. Thus, you must set OpenBLAS to use single thread as following.

- export OPENBLAS_NUM_THREADS=1 in the environment variables. Or
- Call openblas_set_num_threads(1) in the application on runtime. Or
- Build OpenBLAS single thread version, e.g. make USE_THREAD=0 USE_LOCKING=1 (see comment below)
- If the application is parallelized by OpenMP, please build OpenBLAS with USE_OPENMP=1"

oneDAL is certaingly multi-threaded (powered by TBB). But it contains a mix of single- and multi-threaded BLAS/LAPACK calls so building as a single thread version might affect some algos.
"set_num_threads(1)" is already in place in the PR.
I guess oneDAL doesn't need OpenMP parallelization in OpenBLAS to avoid conflicts with TBB scheduler. For me, search for the best OpenBLAS settings for oneDAL looks like a big separate topic.

rakgb-fujitsu:
@amgrigoriev I tried to investigate further into this issue. I used Random Forests algorithm which does not use any of the OpenBLAS functions, through debugging its using only the random number generator calls. Despite no usage of OpenBLAS it is still running in single threaded mode. Is this some TBB related issue?

Regarding the algorithms dependant on OpenBLAS, `openblas_get_num_threads()` shows the correct number of threads but in the console htop cmd shows only 1 thread being used. What would this mean?

cpp/daal/src/externals/service_blas_ref.h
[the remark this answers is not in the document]
Vika-F: I agree with @KulikovNikita here. Using RAII approach would be better. Maybe the `_lock` in the name of the `omp_thread_lock` class is misleading.
The goal is to leave the execution logic the same as you have now in `xx` functions. I.e. set number of threads to 1 at the beginning, call OpenBLAS and restore the original number of threads after the call.
But let's hide this logic inside some helper RAII class (`omp_thread_lock` in Nikita's proposal). This is done in order not to accidently forget to restore the number of threads after the OpenBLAS call. Because it will happen automatically in the destructor of that RAII object when it goes out of the scope.

cpp/daal/src/externals/service_rng_ref.h
```diff
@@ -0,0 +1,363 @@
...
+        _seed = (unsigned int *)daal::services::daal_malloc(sizeof(unsigned int) * _seedSize);
+        if (_seedSize > 0 && seed != nullptr)
+        {
+            for (size_t i = 0; i < _seedSize; i++)
+            {
+                _seed[i] = seed[i];
+            }
+            _gen.seed(_seed[0]);
+        }
+    }
```
Vika-F, line 110: Please use the way that is described in C++ standard to serialize/deserialize C++ generators.
It says that `operator<<` and `operator>>` are defined for this purpose.

The example below shows how to save C++ RNG's state into the buffer, and then how to load it from the buffer to create an identical RNG:
```cpp
#include <iostream>
#include <sstream>
#include <cstring>
#include <random>

int main() {
    std::mt19937 gen1(777);

    std::stringstream ss_serialized;
    ss_serialized << gen1;

    const int state_lenght = ss_serialized.str().length();
    char * buffer = new char[state_lenght];

    memcpy(buffer, ss_serialized.str().c_str(), state_lenght);

    std::cout << buffer << std::endl;

    std::stringstream ss_deserialized;

    ss_deserialized << buffer;
    

    std::mt19937 gen2(0);
    ss_deserialized >> gen2;

    std::uniform_real_distribution<> dis(1.0, 2.0);
    std::cout << "Uniform distribution + gen1:" << std::endl;
    for(int i = 0; i < 10; ++i)
    {
        std::cout << dis(gen1) << ' ';
    }
    std::cout << std::endl;
    std::cout << "Uniform distribution + gen2:" << std::endl;
    for(int i = 0; i < 10; ++i)
    {
        std::cout << dis(gen2) << ' ';
    }
    std::cout << std::endl;
    delete[] buffer;
}
```

cpp/daal/src/externals/service_rng_ref.h
```diff
@@ -0,0 +1,363 @@
...
+    State(const State & other) : _gen(0), _brngId(other._brngId), _seedSize(other._seedSize), _seed(nullptr), _nSkip(other._nSkip)
```
Vika-F, line 111: Same comment as above.
And _nSkip is not needed in case the standard way of serialization/deserialization is used.

cpp/daal/src/externals/service_rng_ref.h
```diff
@@ -0,0 +1,363 @@
...
+        int errcode = 0;
+        _stream->assign(src);
+        return errcode;
+    }
+
+    // Not implemented
+    int leapfrog(size_t threadNum, size_t nThreads)
+    {
+        int errcode = 0;
+        // TODO add call for (vslLeapfrogStream, (_stream, threadNum, nThreads), errcode);
```
Vika-F, line 294: It is needed either to report an error or to make some implementation.

Vika-F (commented):
Thank you for sharing this accomplishment with us. This is a great contribution!
I have party reviewed it and the code looks good in general.

My greatest concern is the amount of changed files. I understand that this was made so because the new templates BlasInst, LapackInst, etc. were added. But for me it's better to change the already existing classes' names. For example, rename 'Blas' -> 'BlasTemplate' (or 'BlasImpl', etc.); and then rename 'BlasInst' to 'Blas'.
Then the code modifications won't be so invasive. Which is a good thing from my perspective.

amgrigoriev:
@rakgb-fujitsu Sorry for the delay, I was busy with other tasks. My suggestion is to separate two things: capability to use OpenBLAS as a backend and the best build/usage settings for OpenBLAS. The second one needs additional experiments/measurements to understand if something should be tuned in oneDAL too. There are many algorithms with different use cases for BLAS/TBB so there is a lot of thing to analyze before proposing some solution.

rakgb-fujitsu:
@amgrigoriev the issue was from this [function](https://github.com/amgrigoriev/daal/blob/2ed525f03f0c1d9c2b0c840af056e6a07be897c7/cpp/daal/src/externals/service_service_ref.h#L84), adding the correct way to detect this solved the problem for kmeans. The issue however seems to persist for RandomForests.

cpp/daal/src/externals/config_template.h
amgrigoriev: The idea is to avoid too complex rules for defines and provide a single entry point. But it might be possible to use -DDAAL_REF for that. In first case, only name of backed (existing one) should be provided for make and everything else will taken from config. Second option is to set backend_name --> define mapping in makefile and then use define macro inside header files for as a switch. In first case, mistake in a particular backend will not affect other ones.

cpp/daal/src/externals/service_rng_ref.h
amgrigoriev: _nSkip is necesssary at least to emulate SkipAhead by discard and to track a number of discarded elements

cpp/daal/src/externals/service_rng_ref.h
amgrigoriev: oneDAL implementation of the backend (externals) does not provide any capabilities for error reporting, as I can see.

cpp/daal/src/externals/service_rng_ref.h
Vika-F: There is really no need in tracking the number of discarded elements. Because it is already encoded in the state of std::mt19937. Storing nSkip as a separate field duplicates the info already available in std::mt19937 RNG's state. You can just have nSkip as an input parameter of SkipAhead method, and do gen.discard(nSkip) in the implementation.

cpp/daal/src/externals/service_service_ref.h
```diff
@@ -0,0 +1,179 @@
...
+            ;
+        if (endptr) *endptr = const_cast<char *>(cur);
+        size_t size = cur - nptr;
+        // TODO replace with static buffer
+        char * buffer = static_cast<char *>(malloc(size + 1));
+        for (size_t i = 0; i < size; ++i) buffer[i] = nptr[i];
+        buffer[size] = '\0';
+        double val   = atof(buffer);
+        free(buffer);
+        return val;
```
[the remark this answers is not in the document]
amgrigoriev: Input string may not be null-terminated so it requires copying to a separate buffer.

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -42,7 +42,10 @@
 #include "src/externals/service_lapack.h"
 #include "src/externals/service_memory.h"
 #include "src/externals/service_math.h"
-#include "immintrin.h"
+
+#if defined(__INTEL_COMPILER)
```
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/config_template.h
```diff
@@ -0,0 +1,6 @@
...
+
+#endif
```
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/service_blas_ref.h
```diff
@@ -0,0 +1,255 @@
...
+                     const DAAL_INT * lda)
+    {
+        dsyr_(uplo, n, alpha, x, incx, a, lda);
+    }
+
+    static void xxsyr(const char * uplo, const DAAL_INT * n, const double * alpha, const double * x, const DAAL_INT * incx, double * a,
+                      const DAAL_INT * lda)
+    {
+        int old_threads = openblas_get_num_threads();
+        openblas_set_num_threads(1);
```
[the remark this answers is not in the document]
amgrigoriev: Fixed with RAII (see below)

cpp/daal/src/externals/service_blas_ref.h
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/service_math_ref.h
```diff
@@ -0,0 +1,204 @@
...
+/*
+// Double precision functions definition
+*/
+
+template <CpuType cpu>
+struct RefMath<double, cpu>
+{
+    typedef size_t SizeType;
+
+    static double sFabs(double in) { return (in >= 0.0f) ? in : -in; }
```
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/service_spblas_ref.h
```diff
@@ -0,0 +1,105 @@
...
+                        const fpType * beta, fpType * c, const DAAL_INT * ldc)
+    {
+        csrmm(m, n, k, alpha, val, indx, pntrb, b, ldb, beta, c, ldc);
+    }
+    inline static fpType csracc(DAAL_INT row, DAAL_INT col, const DAAL_INT * n, const fpType * val, const DAAL_INT * indx, const DAAL_INT * pntrb)
+    {
+        DAAL_INT offset = pntrb[row] - 1;
+        DAAL_INT nnz    = pntrb[row + 1] - pntrb[row];
+        DAAL_INT csrcol = col + 1;
+        for (DAAL_INT i = 0; i < nnz; ++i)
```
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/service_spblas_ref.h
```diff
@@ -0,0 +1,105 @@
...
+        DAAL_OVERFLOW_CHECK_BY_MULTIPLICATION_THROW_IF_POSSIBLE(DAAL_INT, ldbVal, *n);
+        DAAL_OVERFLOW_CHECK_BY_MULTIPLICATION_THROW_IF_POSSIBLE(DAAL_INT, ldcVal, *n);
+        DAAL_OVERFLOW_CHECK_BY_ADDING_THROW_IF_POSSIBLE(DAAL_INT, ldbVal * ((*n) - 1), *k);
+        DAAL_OVERFLOW_CHECK_BY_ADDING_THROW_IF_POSSIBLE(DAAL_INT, ldcVal * ((*n) - 1), *m);
+
+        for (DAAL_INT mInd = 0; mInd < *m; ++mInd)
+            for (DAAL_INT nInd = 0; nInd < *n; ++nInd)
+            {
+                fpType sum = 0.0;
+                for (DAAL_INT kInd = 0; kInd < *k; ++kInd)
```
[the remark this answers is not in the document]
amgrigoriev: Done

dev/make/deps.mkl.mk
```diff
@@ -0,0 +1,78 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+#++
+#  Math backend (MKL) definitions for makefile
+#--
+
+MKLFPKDIR:= $(if $(wildcard $(DIR)/__deps/mklfpk/$(_OS)/*),$(DIR)/__deps/mklfpk,                            \
+                $(if $(wildcard $(MKLFPKROOT)/include/*),$(subst \,/,$(MKLFPKROOT)),                              \
```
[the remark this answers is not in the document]
amgrigoriev: Done

dev/make/deps.mkl.mk
```diff
@@ -0,0 +1,78 @@
...
+daaldep.fbsd32e.vml :=
+daaldep.fbsd32e.ipp := $(if $(COV.libia),$(COV.libia)/libcov.a)
+
+daaldep.vml     := $(daaldep.$(PLAT).vml)
+daaldep.ipp     := $(daaldep.$(PLAT).ipp)
+
+daaldep.math_backend.ext := $(daaldep.ipp) $(daaldep.vml) $(daaldep.mkl)
+
+
+
```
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/service_math_ref.h
```diff
@@ -0,0 +1,204 @@
...
+    static double sMax(double in1, double in2) { return (in1 < in2) ? in2 : in1; }
+
+    static double sSqrt(double in) { return sqrt(in); }
+
+    static double sPowx(double in, double in1) { return pow(in, in1); }
+
+    static double sCeil(double in) { return ceil(in); }
+
+    // Not implemented
+    static double sErfInv(double in) { return 1.0; }
```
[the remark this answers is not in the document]
amgrigoriev: It looks line an additional development. Could it be separated to another PR?

amgrigoriev:
@rakgb-fujitsu Thanks, it is clear now. I set hyperthreading flag to zero to avoid callling of the function. It should use TBB default number of threads in that case. Hope it will help.

dev/make/cmplr.clang.ref.mk
```diff
@@ -0,0 +1,50 @@
...
+
+CORE.SERV.COMPILER.clang = generic
+
+-Zl.clang =
+-DEBC.clang = -g
+
+COMPILER.mac.clang = clang++ -m64 -fgnu-runtime -stdlib=libc++ -mmacosx-version-min=10.15 -fwrapv \
+                     -DDAAL_REF -Werror -Wreturn-type
+COMPILER.lnx.clang = clang++ -m64 \
+                     -DDAAL_REF -Werror -Wreturn-type
```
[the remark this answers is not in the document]
amgrigoriev: The problem is that X86 is the simplest case where it is possible to use simple solutions. In general (having cross-compilation in mind), it is necessary to add the following:
- --target=...
- --gcc-toolchain=...
- --sysroot=...
-  -march=...
-  -mabi=... 
- -mtune=...
Setting all of this in main Makefile for different platform/compilers may result in a mess.

cpp/daal/src/externals/service_rng_ref.h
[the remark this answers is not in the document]
KulikovNikita: Please also consider using `std::generate` instead of the generic loop

cpp/daal/src/externals/service_math_ref.h
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    static void vLog(SizeType n, const double * in, double * out)
+    {
+        for (SizeType i = 0; i < n; i++) out[i] = log(in[i]);
+    }
+
+    static void vLog1p(SizeType n, const double * in, double * out)
+    {
+        for (SizeType i = 0; i < n; i++) out[i] = log1p(in[i]);
```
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/service_rng_ref.h
```diff
@@ -0,0 +1,363 @@
...
+{
+public:
+    virtual ~StateIface() {}
+    virtual int uniformRNG(const size_t n, size_t * r, const size_t a, const size_t b, const int method)      = 0;
+    virtual int uniformRNG(const size_t n, int * r, const int a, const int b, const int method)               = 0;
+    virtual int uniformRNG(const size_t n, float * r, const float a, const float b, const int method)         = 0;
+    virtual int uniformRNG(const size_t n, double * r, const double a, const double b, const int method)      = 0;
+    virtual int gaussianRNG(const size_t n, float * r, const float a, const float sigma, const int method)    = 0;
+    virtual int gaussianRNG(const size_t n, double * r, const double a, const double sigma, const int method) = 0;
+    virtual int bernoulliRNG(const size_t n, int * r, const double p, const int method)                       = 0;
```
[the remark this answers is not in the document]
amgrigoriev: added "final"

cpp/daal/src/externals/service_rng_ref.h
```diff
@@ -0,0 +1,363 @@
...
+    {
+        State * destState = static_cast<State *>(dest);
+        destState->_seedSize  = _seedSize;
+        destState->_brngId    = _brngId;
+        destState->_nSkip     = _nSkip;
+        destState->_seed      = (unsigned int *)daal::services::daal_malloc(sizeof(unsigned int) * destState->_seedSize);
+        for (size_t i = 0; i < destState->_seedSize; ++i)
+        {
+            destState->_seed[i] = _seed[i];
+        }
```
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/service_rng_ref.h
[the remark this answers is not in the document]
amgrigoriev: Done, but without generate: it accepts generator - not a distribution

cpp/daal/src/externals/service_service_ref.h
```diff
@@ -0,0 +1,179 @@
...
+    // Remove the method after that or replace with proper call from new backend library
+    static void serv_free_buffers() {}
+
+    static int serv_memcpy_s(void * dest, size_t destSize, const void * src, size_t srcSize)
+    {
+        if (destSize < srcSize) return static_cast<int>(ENOMEM);
+        memcpy(dest, src, srcSize);
+        return 0;
+        // TODO: safe funtion
+        // return memcpy_s(dest, destSize, src, srcSize);
```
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/service_rng_ref.h
Vika-F: I think it would be enough to return `ErrorLeapfrogUnsupported` status code instead of zero.

cpp/daal/src/externals/service_math_ref.h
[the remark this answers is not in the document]
Vika-F: Yes, sure. erfinv/cdfnorminv are only used in a couple of algorithms. So, most of the functionality is not affected by the absence of those functions.

cpp/daal/src/externals/service_service_ref.h
```diff
@@ -0,0 +1,179 @@
...
+    {
+        return 0;
+        // Old one - just to see what the method is for
+        // return fpk_serv_set_memory_limit(type, limit);
+    }
+
+    static size_t serv_strnlen_s(const char * src, size_t slen)
+    {
+        size_t i = 0;
+        for (; (i < slen) && (src[i] != '\0'); ++i)
```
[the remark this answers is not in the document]
amgrigoriev: replaced with strnlen_s

cpp/daal/src/externals/service_math_ref.h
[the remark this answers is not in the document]
amgrigoriev: Done

cpp/daal/src/externals/service_rng_ref.h
```diff
@@ -0,0 +1,363 @@
...
+
+    int bernoulliRNG(const size_t n, int * r, const double p, const int method) override { return iBernoulli(n, r, p, method); }
+
+    State(const int brngId, const size_t seedSize, const unsigned int * seed)
+        : _gen(0), _brngId(brngId), _seedSize(seedSize), _seed(nullptr), _nSkip(0)
+    {
+        _seed = (unsigned int *)daal::services::daal_malloc(sizeof(unsigned int) * _seedSize);
+        if (_seedSize > 0 && seed != nullptr)
+        {
+            for (size_t i = 0; i < _seedSize; i++)
```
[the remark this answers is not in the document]
amgrigoriev: Done

.ci/env/openblas.sh
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+sudo apt-get update
+sudo apt-get install libopenblas-dev
+export OPENBLASROOT=/usr/lib/x86_64-linux-gnu/openblas-pthread
```
amgrigoriev, line 20: Will not work. Expected libopenblas.xxx in $(OPENBLASDIR)/lib.Could it be fixed by a soft link?

amgrigoriev: If not, should be handled in makefile (TBD)

amgrigoriev: Includes are also expected to be located in $OPENBLASROOT/include

amgrigoriev:
It looks like build passed for OpenBLAS but there is a mistake in copy directory name: cp -fp ./__work_gnu/md/lnx32e/daal/lib/libJavaAPI.so ./**__release_lnx_gnu**
But upload is from another folder: ##[error]Path does not exist: /home/vsts/work/1/s/**__release_lnx_gnu_openblas**

amgrigoriev:
Missing dependency on fortran library in examples (injected by OpenBLAS): 
/usr/bin/ld: /home/vsts/work/1/s/__release_lnx_gnu/daal/latest/lib/intel64/libonedal_thread.so.1: undefined reference to `_gfortran_concat_string'
For makefile, it could be added like (if copy libgfortran.a manually to  $(OPENBLASROOT)/lib):
        ifdef OPENBLASROOT
            LOPTS += $(OPENBLASROOT)/lib/libgfortran.a
        endif
May be it is possible to use installed gfortran from /usr/lib.
Not clear, how to do it for cmake

cpp/daal/src/externals/service_math_ref.h
```diff
@@ -0,0 +1,240 @@
...
+    // Not implemented
+    static void vErfInv(SizeType n, const float * in, float * out)
+    {
+        for (SizeType i = 0; i < n; ++i) out[i] = std::numeric_limits<float>::quiet_NaN();
+    }
+
+    static void vErf(SizeType n, const float * in, float * out)
+    {
+#pragma omp simd
+        for (SizeType i = 0; i < n; + i) out[i] = erf(in[i]);
```
Pahandrovich, line 191:
```suggestion
        for (SizeType i = 0; i < n; ++i) out[i] = erf(in[i]);
```

cpp/daal/src/services/service_defines.h
```diff
@@ -33,9 +33,11 @@ DAAL_EXPORT int __daal_serv_cpu_detect(int);
 void run_cpuid(uint32_t eax, uint32_t ecx, uint32_t * abcd);
 bool daal_check_is_intel_cpu();
 
+#define DAAL_BASE_CPU sse2
```
keeranroth, line 36: Protect with `#ifndef DAAL_BASE_CPU`, so that it can be set at compile time?

cpp/oneapi/dal/backend/primitives/rng/rnd_seq.hpp
```diff
@@ -60,7 +60,7 @@ class rnd_seq {
         const auto count_as_size_t = dal::detail::integral_cast<std::size_t>(count);
 
         auto number_array = array<std::size_t>::empty(queue, count);
-        daal::internal::RNGs<std::size_t, daal::sse2> rng;
+        daal::internal::RNGsInst<std::size_t, daal::sse2> rng;
```
keeranroth, line 63: DAAL_BASE_CPU instead of daal::sse2? Same in rng_engine.hpp below

keeranroth (changes requested):
I assume that the code has moved on a little, but I wanted to get a feel for the code change. I've left some comments as I've gone. Looks really interesting, though.

cpp/daal/src/externals/service_spblas_ref.h
```diff
@@ -0,0 +1,111 @@
...
+
+    static void xxcsrmm(const char * transa, const DAAL_INT * m, const DAAL_INT * n, const DAAL_INT * k, const fpType * alpha, const char * matdescra,
+                        const fpType * val, const DAAL_INT * indx, const DAAL_INT * pntrb, const fpType * b, const DAAL_INT * ldb,
+                        const fpType * beta, fpType * c, const DAAL_INT * ldc)
+    {
+        csrmm(m, n, k, alpha, val, indx, pntrb, b, ldb, beta, c, ldc);
+    }
+    inline static fpType csracc(DAAL_INT row, DAAL_INT col, const fpType * val, const DAAL_INT * indx, const DAAL_INT * pntrb)
+    {
+        DAAL_INT offset = pntrb[row] - 1;
```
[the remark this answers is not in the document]
KulikovNikita: It uses one based indexing for historical reasons. 

https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-0/sparse-blas-csr-matrix-storage-format.html

cpp/daal/src/externals/service_stat_ref.h
```diff
@@ -0,0 +1,328 @@
...
+    struct ThreadingFuncs
+    {
+        threadfuncfor funcfor;
+        threadfuncfor funcforordered;
+        threadfuncsection funcsection;
+        threadfuncordered funcordered;
+        threadgetlimit getlimit;
+    };
+
+    static void _daal_mkl_threader_for_sequential(DAAL_INT n, DAAL_INT threads_request, void * a, func_type func)
```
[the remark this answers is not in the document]
KulikovNikita: I don't see these functions used anywhere. So this section can be removed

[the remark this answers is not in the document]
KulikovNikita: Maybe I was wrong. This functionality is presumably used in `core_threading_win_dll.cpp`

dev/make/cmplr.gnu.ref.mk
```diff
@@ -0,0 +1,51 @@
...
+
+CMPLRDIRSUFF.gnu = _gnu
+
+CORE.SERV.COMPILER.gnu = generic
+
+-Zl.gnu =
+-DEBC.gnu = -g
+
+COMPILER.all.gnu =  ${CXX} -m64 -fwrapv -fno-strict-overflow -fno-delete-null-pointer-checks \
+                    -DDAAL_REF -Werror -Wreturn-type
```
keeranroth, line 31: The source code uses both DAAL_REF and ONEDAL_REF. Changing all of these to be the same (either all DAAL_REF or all ONEDAL_REF) means that undefined symbol errors are not encountered.

As it stands, some of the avx512, avx2, etc. functions from micromkl get referenced from the code base

cpp/daal/src/externals/service_stat_ref.h
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    static DAAL_INT _daal_mkl_threader_get_max_threads_sequential()
+    {
+        return 1;
+    }
+
+    static void _daal_mkl_threader_for(DAAL_INT n, DAAL_INT threads_request, void * a, func_type func)
+    {
+        //fpk_vsl_serv_threader_for(n, threads_request, a, func);
```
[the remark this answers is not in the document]
amgrigoriev: According to library rules, as I remember, raising an exception is prohibited here. It would be good to separate threading changes from backend changes.

cpp/daal/src/externals/service_thread_declar_ref.h
```diff
@@ -0,0 +1,57 @@
...
+* limitations under the License.
+*******************************************************************************/
+
+/*
+//++
+//  Auxiliary class to set/restore OpenBLAS threads
+//--
+*/
+
+#ifndef __SERVICE_THREAD_DECRAL_REF_H__
```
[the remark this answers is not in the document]
amgrigoriev: It looks like the file was removed - it does not exist in my local repo.

cpp/daal/src/externals/service_stat_ref.h
```diff
@@ -0,0 +1,328 @@
...
+{
+    typedef __int64 SizeType;
+    typedef __int64 MethodType;
+    typedef int ErrorType;
+
+    static int xcp(double * data, __int64 nFeatures, __int64 nVectors, double * nPreviousObservations, double * sum, double * crossProduct,
+                   __int64 method)
+    {
+        int errcode = 0;
+
```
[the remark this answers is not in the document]
amgrigoriev: It is due to backward compatibility. OneDAL team can change it if it is possible

cpp/daal/src/externals/service_lapack_declar_ref.h
```diff
@@ -0,0 +1,95 @@
...
+*******************************************************************************/
+
+/*
+//++
+//  Declarations for Lapack OpenBLAS functions.
+//--
+*/
+
+#ifndef __SERVICE_LAPACK_DECRAL_REF_H__
+#define __SERVICE_LAPACK_DECRAL_REF_H__
```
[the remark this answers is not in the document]
amgrigoriev: Fixed

cpp/daal/src/externals/service_blas_declar_ref.h
```diff
@@ -0,0 +1,71 @@
...
+* limitations under the License.
+*******************************************************************************/
+
+/*
+//++
+//  Declarations for BLAS OpenBLAS functions.
+//--
+*/
+
+#ifndef __SERVICE_BLAS_DECRAL_REF_H__
```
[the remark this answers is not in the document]
amgrigoriev: Fixed

dev/make/cmplr.gnu.ref.mk
amgrigoriev: As an example of difference:
#ifdef ONEDAL_REF
#if (defined(DAAL_REF) && (INTPTR_MAX == INT64_MAX))
using GEMM_INT = std::int64_t;
#else
using GEMM_INT = std::int32_t;
#endif

cpp/daal/src/externals/service_thread_declar_ref.h
```diff
@@ -21,8 +21,8 @@
 //--
 */
 
-#ifndef __SERVICE_THREAD_DECRAL_REF_H__
-#define __SERVICE_THREAD_DECRAL_REF_H__
+#ifndef __SERVICE_THREAD_DECLAR_REF_H__
+    #ifndef __SERVICE_THREAD_DECLAR_REF_H__
```
keeranroth, line 25: This should be the define

makefile
```diff
@@ -539,7 +539,7 @@ ONEAPI.tmpdir_a.dpc := $(WORKDIR)/oneapi_dpc_static
 ONEAPI.tmpdir_y.dpc := $(WORKDIR)/oneapi_dpc_dynamic
 
 ONEAPI.incdirs.common := $(CPPDIR)
-ONEAPI.incdirs.thirdp := $(CORE.incdirs.common) $(MKLFPKDIR.include) $(MKLGPUFPKDIR.include) $(TBBDIR.include)
+ONEAPI.incdirs.thirdp := $(CORE.incdirs.common) $(daaldep.math_backend.incdir) $(TBBDIR.include)
```
keeranroth, line 542: It looks like there is `daaldep.math_backend_oneapi.incdir` in `deps.mkl.mk` with the appropriate values in. Should this be used here?

dev/make/deps.mkl.mk
```diff
@@ -33,7 +33,7 @@ MKLGPUFPKDIR.libia   := $(MKLGPUFPKDIR)/lib/$(_IA)
 mklgpufpk.LIBS_A := $(MKLGPUFPKDIR.libia)/$(plib)daal_sycl$d.$(a)
 mklgpufpk.HEADERS := $(MKLGPUFPKDIR.include)/mkl_dal_sycl.hpp $(MKLGPUFPKDIR.include)/mkl_dal_blas_sycl.hpp
 
-daaldep.math_backend.incdir := $(MKLFPKDIR.include)
+daaldep.math_backend.incdir := $(MKLFPKDIR.include) $(MKLGPUFPKDIR.include)
```
keeranroth, line 36: Is this required? Should the `daaldep.math_backend_oneapi.incdir` variable be used when the MKLGPUFPKDIR is needed?

cpp/daal/src/externals/service_thread_declar_ref.h
amgrigoriev: Fixed

makefile
amgrigoriev: It seems to be the same for both cases

dev/make/deps.mkl.mk
amgrigoriev: It is the question to oneDAL team too. I think it is necessary for MKL backend (to support GPUs). As I can see MKLGPUFPKDIR will be empty if specific env is not set.

amgrigoriev:
Please restart public CI. It failed due to conda loading issue.

cpp/daal/src/externals/service_service_ref.h
```diff
@@ -0,0 +1,180 @@
+/* file: service_service_refh */
```
napetrov, line 1:
```suggestion
/* file: service_service_ref.h */
```

napetrov:
Private CI have lot of library build warnings on windows for all fpk functions
"warning LNK4006: fpk_serv_strcpy already defined in daal_mkl_thread.lib(mkl_support.obj); second definition ignored"

napetrov:
> Added an option to add cpp/compression to exclude list for REF backend (examples for OpenBLAS build). But it requires a change in CI in order to be " cmake . -DREF_BACKEND=ON" for building of examples. Simplier alternative is to add "compression/cpp/*" to exclude list for all cases.

this wouldn't be an issue in longterm - compression would be removed in upcoming weeks

napetrov:
@keeranroth 
"[keeranroth](https://github.com/keeranroth) [2 weeks ago](https://github.com/oneapi-src/oneDAL/pull/2396#discussion_r1281658472)
+1 for use of standard library"

Historically DAAL minimize use of standard lib so there is less dependency on user side.  with oneDAL and move to higher C++ standards, STD use in compiler and other things use of STD is more open there. 

I would say if it's easy to adhere to this - let's keep it going.  If not, can be discussed

cpp/daal/src/externals/service_service_ref.h
```diff
@@ -0,0 +1,180 @@
...
+    {
+        size_t i = 0;
+        for (; (i < slen) && (src[i] != '\0'); ++i)
+            ;
+        return i;
+    }
+
+    static int serv_strncpy_s(char * dest, size_t dmax, const char * src, size_t slen)
+    {
+        if (dmax < slen + 1) return static_cast<int>(ENOMEM);
```
Pahandrovich, line 109: It can be root cause of some seg faults, when exception throwing
```suggestion
        if (dmax < slen) return static_cast<int>(ENOMEM);
```

cpp/daal/src/externals/service_service_ref.h
```diff
@@ -0,0 +1,180 @@
...
+        strncpy(dest, src, slen);
+        return 0;
+        // TODO: safe funtion
+        // return strncpy_s(dest, dmax, src, slen);
+    }
+
+    static int serv_strncat_s(char * dest, size_t dmax, const char * src, size_t slen)
+    {
+        if (dmax < slen + 1) return static_cast<int>(ENOMEM);
```
Pahandrovich, line 118:
```suggestion
        if (dmax < slen) return static_cast<int>(ENOMEM);
```

amgrigoriev:
I see different number of daal/cpp examples running locally and in public CI (command to run locally: .ci/scripts/test.sh --test-kind examples --build-dir __release_lnx_gnu/ --compiler gnu --interface oneapi/cpp --build_system cmake). Due to that, some missed examples are SegFaulted in public CI while everything is Ok locally.

cpp/daal/src/externals/service_service_ref.h
```diff
@@ -0,0 +1,179 @@
...
+        return val;
+    }
+
+    static float serv_string_to_float(const char * nptr, char ** endptr) { return static_cast<float>(serv_string_to_double(nptr, endptr)); }
+
+    // TODO: not a safe function - no control for the input buffer end
+    static int serv_string_to_int(const char * nptr, char ** endptr)
+    {
+        const char * cur = nptr;
+        for (; isdigit(*cur) || *cur == '-'; ++cur)
```
keeranroth, line 148: This can set endptr to past where the number was consumed, and doesn't cover the case of scientific notation, e.g. 3e-5- would have 3 as the floating point number, where it should be 3 x 10^5, and if you also allow the use of 'e', you would read the final '-', which is no longer part of the string. strtol is the function you want to use here.

amgrigoriev:
As I can see, oneapi examples and MPI samples passed - the problem is that some cpp examples throw exceptions for old interfaces. But it can not be fixed right now due to, for instance, absence of some sparse functions in OpenBLAS. My suggestion it to add those examples into EXCLUDE_LIST in cmake. To do that, it makes sense to add --backend parameter to .ci/scripts/test.sh and pass -DREF_BACKEND to cmake in case of REF backend. I can add it if nobody argues.

napetrov:
> As I can see, oneapi examples and MPI samples passed - the problem is that some cpp examples throw exceptions for old interfaces. But it can not be fixed right now due to, for instance, absence of some sparse functions in OpenBLAS. My suggestion it to add those examples into EXCLUDE_LIST in cmake. To do that, it makes sense to add --backend parameter to .ci/scripts/test.sh and pass -DREF_BACKEND to cmake in case of REF backend. I can add it if nobody argues.

Sounds good to me assuming that default would be MKL backend with no param passing

napetrov:
On compression removal - https://github.com/oneapi-src/oneDAL/pull/2462 would happen here

napetrov:
With latest private CI still have warnings on windows - 2532 warnings

daal_vmlipp_core.lib(mkl_support.obj) : warning LNK4006: fpk_serv_strcpy already defined in daal_mkl_thread.lib(mkl_support.obj); second definition ignored

daal_vmlipp_core.lib(vsl_l9_iuniformbits32.obj) : warning LNK4006: fpk_vsl_kernel_l9_iRngUniformBits32 already defined in daal_mkl_thread.lib(vsl_l9_iuniformbits32.obj); second definition ignored

@Alexsandruss - with cleanup of compression, should FPK also be rebuild to drop IPP?

.ci/pipeline/ci.yml
```diff
@@ -89,6 +89,50 @@ jobs:
...
+  steps:
+  - script: |
+      sudo apt-get update && sudo apt-get install -y gcc-multilib g++-multilib dos2unix tree
+      conda create -n ci-env -q -y -c intel -c conda-forge impi-devel=2021.8.0 openjdk=17.0.3
+    displayName: 'apt-get and conda install'
+  - script: |
+      .ci/scripts/describe_system.sh
+    displayName: 'System info'
+  - script: |
+      .ci/scripts/build.sh --compiler gnu --optimizations avx2 --target daal --backend_config ref --conda-env ci-env
```
napetrov, line 108: sse2 always build by default - forcing this config would build only 1 platform and not both sse2 and avx and should reduce build time
```suggestion
      .ci/scripts/build.sh --compiler gnu --optimizations sse2--target daal --backend_config ref --conda-env ci-env
```

.ci/pipeline/ci.yml
```diff
@@ -89,6 +89,50 @@ jobs:
...
+    displayName: 'make daal'
+  - script: |
+      .ci/scripts/build.sh --compiler gnu --optimizations avx2 --target oneapi_c --backend_config ref
```
napetrov, line 111:
```suggestion
      .ci/scripts/build.sh --compiler gnu --optimizations sse2--target oneapi_c --backend_config ref
```

amgrigoriev:
As I can see from Linux, botch ./__deps/mklfpk/lnx/lib/intel64/libdaal_mkl_thread.a and ./__deps/mklfpk/lnx/lib/intel64/libdaal_vmlipp_core.a are external libraries. From the warning, it is clear that both ones are including the same obj files and I can see the same names inside. Quick attempt to remove vmlipp resulted in failed build of cpp examples with undefined references. Maybe it is possible to rebuild MKL FPK libraries without duplication of symbols?

amgrigoriev:
Resolved conflict in cpp/daal/src/algorithms/kernel.h

KulikovNikita:
Looks good to me

cpp/daal/BUILD
```diff
@@ -128,12 +145,20 @@ daal_module(
...
+                                          ":mathbackend_thread",
+                                          "@tbb//:tbb",
+                                          "@tbb//:tbbmalloc",
+                                        ],
+            "//conditions:default": [ 
+                                      ":threading_headers",
+                                      ":mathbackend_thread",
+                                      "@tbb//:tbb",
+                                      "@tbb//:tbbmalloc",
+                                    ],
```
KulikovNikita, line 160: What is the difference?

KulikovNikita (approved):

amgrigoriev:
It looks like oneapi_c build issue is not specific for the PR. Master build failed with the same error for ".ci/scripts/build.sh --compiler gnu --optimizations sse2 --target oneapi_c". As I can see, it does not work at all for base level of optimization. My suggestion is to return back "avx2".  For "avx2" there is a command in log "echo ONEDAL_CPU_DISPATCH_AVX2" >> __work_gnu/md/lnx32e/oneapi/dal/_dal_cpu_dispatcher_gen.hpp". Something similar is absent for build with "sse2" and the error appears "_dal_cpu_dispatcher_gen.hpp: No such file or directory"

.ci/pipeline/ci.yml
```diff
@@ -89,6 +89,50 @@ jobs:
...
+      conda create -n ci-env -q -y -c intel -c conda-forge impi-devel=2021.8.0 openjdk=17.0.3
+    displayName: 'apt-get and conda install'
+  - script: |
+      .ci/scripts/describe_system.sh
+    displayName: 'System info'
+  - script: |
+      .ci/scripts/build.sh --compiler gnu --optimizations sse2 --target daal --backend_config ref --conda-env ci-env
+    displayName: 'make daal'
+  - script: |
+      .ci/scripts/build.sh --compiler gnu --optimizations sse2 --target oneapi_c --backend_config ref
```
napetrov, line 111:
```suggestion
      .ci/scripts/build.sh --compiler gnu --optimizations avx2 --target daal --backend_config ref --conda-env ci-env
    displayName: 'make daal'
  - script: |
      .ci/scripts/build.sh --compiler gnu --optimizations avx2 --target oneapi_c --backend_config ref
```

napetrov:
> It looks like oneapi_c build issue is not specific for the PR. Master build failed with the same error for ".ci/scripts/build.sh --compiler gnu --optimizations sse2 --target oneapi_c". As I can see, it does not work at all for base level of optimization. My suggestion is to return back "avx2". For "avx2" there is a command in log "echo ONEDAL_CPU_DISPATCH_AVX2" >> __work_gnu/md/lnx32e/oneapi/dal/_dal_cpu_dispatcher_gen.hpp". Something similar is absent for build with "sse2" and the error appears "_dal_cpu_dispatcher_gen.hpp: No such file or directory"

yes, looks like there is problem in make for oneapi_c that doesn't allow building for lowest opt. I guess it can be fixed fairly easy but we can do this in separate PR.  Returned AVX2 in code

napetrov:
> As I can see from Linux, botch ./__deps/mklfpk/lnx/lib/intel64/libdaal_mkl_thread.a and ./__deps/mklfpk/lnx/lib/intel64/libdaal_vmlipp_core.a are external libraries. From the warning, it is clear that both ones are including the same obj files and I can see the same names inside. Quick attempt to remove vmlipp resulted in failed build of cpp examples with undefined references. Maybe it is possible to rebuild MKL FPK libraries without duplication of symbols?

yes, work in progress not to drop IPP from FPK. Should have results today/tomorrow - @Alexsandruss

.ci/pipeline/ci.yml
```diff
@@ -89,6 +89,50 @@ jobs:
     condition: failed()
     continueOnError: true
 
+- job: 'LinuxMakeGNU_OpenBLAS'
```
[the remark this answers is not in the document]
napetrov: The plan here is to move OpenBLAS job to non x86 with next iteration of PRs, but i'm fine with renaming this to _MKL for now

.ci/env/openblas.sh
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+sudo apt-get update
+sudo apt-get install build-essential gcc gfortran
+git clone https://github.com/xianyi/OpenBLAS.git
+pushd OpenBLAS
+  make clean
+  make -j4 NO_FORTRAN=1
```
napetrov, line 23:
```suggestion
CpuCount=$(lscpu -p | egrep -v '^#' | wc -l)
pushd OpenBLAS
  make clean
  make -j${CpuCount} NO_FORTRAN=1
```

.ci/env/openblas.sh
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+sudo apt-get update
+sudo apt-get install build-essential gcc gfortran
+git clone https://github.com/xianyi/OpenBLAS.git
+CoreCount=$(lscpu -p | egrep -v '^#' | wc -l)
```
napetrov, line 21:
```suggestion
CoreCount=$(lscpu -p | grep -Ev '^#' | wc -l)
```

napetrov (changes requested):
Just blocking merge as safety measure until win build warnings would be fixed.

napetrov:
So, on warning fixes
plan A check on results of dropping IPP from FPK
plan B on Java removal - https://github.com/oneapi-src/oneDAL/pull/2469

napetrov:
FPK update didn't help - duplication happens even without IPP lib

amgrigoriev:
Could you expose the warning?

amgrigoriev:
Is this that one: daal_vmlipp_core.lib(vml_dLog1p_L9HAynn.obj) : warning LNK4006: fpk_vml_dLog1pI_L9HAynn already defined in daal_mkl_thread.lib(vml_dLog1p_L9HAynn.obj)? I am wondering if another library contains duplicated symbols after IPP removal?

napetrov:
> Is this that one: daal_vmlipp_core.lib(vml_dLog1p_L9HAynn.obj) : warning LNK4006: fpk_vml_dLog1pI_L9HAynn already defined in daal_mkl_thread.lib(vml_dLog1p_L9HAynn.obj)? I am wondering if another library contains duplicated symbols after IPP removal?

daal_vmlipp_core.lib(mkl_support.obj) : warning LNK4006: fpk_serv_strcat2 already defined in daal_mkl_thread.lib(mkl_support.obj); second definition ignored

It's same obj as it still contain vml/vsl part although ipp was removed. Now there is 2100 warning lines vs 2500 with previous FPK. All warnings are only happen for vml/vsl functions now.

napetrov:
@amgrigoriev  - thank you for fix. Warnings are gone now. Waiting for CI checks to finish

napetrov (approved):
Thanks to everyone involved!

Alexsandruss:
@amgrigoriev, this PR reverted https://github.com/oneapi-src/oneDAL/pull/2392. Are you sure there are no additional rebasing mistakes in PR?

napetrov:
+ at least this one - https://github.com/oneapi-src/oneDAL/pull/2390

PR #2398, opened 2023-05-30
Add RNG engine support to oneDAL Decision_forest descriptor

icfaust:
# Add RNG engine support to DF in oneDAL
This allows for modification of the decision forest descriptor to change aspects of the RNG engine.  This will enable the use of random state by properly setting the seed on DF for CPU, and for other engines to be used for CPU and GPU (like MT19937). This will enable better determinism for testing and operation.

- Add desc.engine using the oneapi::dal::backend::primitive::engine class, as using daal engines explicitly is bad abstraction
- Modify CPU and GPU backends to work with oneapi::dal::backend::primitive::engine
- Add another constructor for engine_collection to take an engine class, which allows for better interoperability.

THIS IS A DRAFT, I knowingly see that there is a conflict between setting desc.seed and desc.engine.  I will need to modify the get_seed and set_seed to regenerate the desc.engine with the new seed, but I need to talk to someone first as to the correct way to do this, especially when the state has been updated.

Secondly, I need to make sure that by returning the enginePtr in some cases that maybe I need to be returning the engine state instead since it updates. I need someone to walk through this with me as to the correct procedure. I may be misusing the engines.

### DIFF
--- a/cpp/oneapi/dal/algo/decision_forest/common.hpp
+++ b/cpp/oneapi/dal/algo/decision_forest/common.hpp
@@ -251,7 +251,6 @@ public:
 
     std::int64_t get_seed() const;
     oneapi::dal::backend::primitives::engine get_engine() const;
-    
 
 protected:
     void set_observations_per_tree_fraction_impl(double value);
@@ -607,7 +606,7 @@ public:
         base_t::set_seed_impl(value);
         return *this;
     }
-    
+
     oneapi::dal::backend::primitives::engine get_engine() const {
         return base_t::get_engine();
     }
--- a/cpp/oneapi/dal/algo/decision_forest/common.hpp
+++ b/cpp/oneapi/dal/algo/decision_forest/common.hpp
@@ -616,7 +616,6 @@ public:
         base_t::set_engine_impl(value);
         return *this;
     }
-    
 };
 
 /// @tparam Task   Tag-type that specifies the type of the problem to solve. Can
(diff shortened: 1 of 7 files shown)


cpp/oneapi/dal/algo/decision_forest/common.hpp
```diff
@@ -603,6 +606,15 @@ class descriptor : public detail::descriptor_base<Task> {
...
     }
+
+    oneapi::dal::backend::primitives::engine get_engine() const {
+        return base_t::get_engine();
+    }
+
+    auto& set_engine(oneapi::dal::backend::primitives::engine value) {
+        base_t::set_engine_impl(value);
+        return *this;
+    }
```
KulikovNikita, line 617: You **should not** use internal functionality in public interfaces. This thing is twofold:
1) headers from `primitives` are not exposed to user at all,
2) this API can be changed at any moment, we have no commitment to preserve it 

Thus I have 2 proposals:
1) Move it to the `detail` section, but you need to move away from `primitives` entities anyway
2) You can create a separate `random_number_generator` algorithms for different types of RNGs. Please take a look on combination of `svm` + `rbf_kernel`, `linear_kernel`

icfaust: Darn. I think either option would probably require a review of architecture before implementation, and would impact all of dal... Unfortunately its beyond the scope/time allocation, I'll probably close.

PR #2400, opened 2023-05-31
Add RNG seed to oneDAL Decision Forest on CPU

icfaust:
# Pass RNG seed to DAAL RNG for DAAL CPU Decision_Forest in oneDAL
For whatever reason, the seed is not passed to the RNG constructor for decision forest on CPU in oneDAL. While this doesn't enable engines other than MT2203, at least this can enable various random state starting points.

- Use seed from descriptor and pass to the DAAL RNG constructor in classifier
- Use seed from descriptor and pass to the DAAL RNG constructor in regressor


icfaust:
CI failure is an unrelated test, specifically a failure on cpp/oneapi/dal/backend/primitives/blas:test_gemv_dpc

Alexsandruss (approved):

PR #2401, opened 2023-05-31
icx compiler and avx2 optimization level for CI

Alexsandruss:
Changes:
- Set compiler to `icx` for `LinuxMakeDPCPP` job
- Set optimization level to avx2 for `LinuxMake*` jobs
- Fix `icx` compilation errors:
  - Force inline for some of decision forest classification `PredictClassificationTask` methods
  - Replace std::isnan with `x != x` comparison (orevents icx error: `explicit comparison with NaN in fast floating point mode`)
  - Explicit ifdef(AVX512F) in intersection_tc.hpp

### DIFF
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -131,15 +131,19 @@ jobs:
     displayName: 'Upload build artifacts'
     continueOnError: true
   - script: |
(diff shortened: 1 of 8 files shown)


razdoburdin (approved):
looks nice for me

PR #2407, opened 2023-06-06
doc: provide maxBins/minBinSize/binningStrategy description

ahuber21:
I think `maxBins` and `minBinSize` descriptions should have been added before. Would be great if someone could double-check.
The `binningStrategy` description is new and should certainly be added.


Alexsandruss (approved):

napetrov (approved):

ahuber21:
Moved common options to the common description. I have also sorted the parameters alphabetically to make the table easier to navigate.

PR #2420, opened 2023-06-15
Add DAAL examples for Extra Trees algorithm

icfaust:
# Description
This adds examples for extratrees operation via decision forest (specifically for regression and classification). This is important to clarify for DAAL users how to enable the algorithm (via changing two parameters), and more vitally allows for testing the different objects utilized with the random splitter.  Since bootstrap is false for extratrees, any operation of oob calculations is disabled.

- Adds df_cls_default_dense_extratrees_batch.cpp
- Adds df_reg_default_dense_extratrees_batch.cpp
- Adds df_cls_hist_dense_extratrees_batch.cpp
- Adds df_reg_hist_dense_extratrees_batch.cpp


examples/daal/cpp/source/decision_forest/df_reg_hist_dense_extratrees_batch.cpp
```diff
@@ -0,0 +1,137 @@
+/* file: df_reg_hist_dense_extratrees_Ebatch.cpp */
```
Alexsandruss, line 1:
```suggestion
/* file: df_reg_hist_dense_extratrees_batch.cpp */
```

examples/daal/cpp/source/decision_forest/df_cls_default_dense_extratrees_batch.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    algorithm.parameter().nTrees = nTrees;
+    algorithm.parameter().featuresPerNode = nFeatures;
+    algorithm.parameter().minObservationsInLeafNode = minObservationsInLeafNode;
+    algorithm.parameter().minObservationsInSplitNode = minObservationsInSplitNode;
+    algorithm.parameter().minWeightFractionInLeafNode = minWeightFractionInLeafNode;
+    algorithm.parameter().minImpurityDecreaseInSplitNode = minImpurityDecreaseInSplitNode;
+    algorithm.parameter().varImportance = algorithms::decision_forest::training::MDI;
+    /* Enable ExtraTrees classification algorithm with bootstrap=false and random splitter*/
+    algorithm.parameter().splitter = algorithms::decision_forest::training::random;
```
Alexsandruss, line 90: I don't think it's meaningful to create separate decision forest examples if the only difference is in this row.
Might be better to set `splitter` as argument in `trainModel` function and run both splitters in same file

examples/daal/cpp/source/decision_forest/df_cls_default_dense_batch.cpp
```diff
@@ -85,8 +90,15 @@ training::ResultPtr trainModel() {
...
-        algorithms::decision_forest::training::computeOutOfBagError;
+    if (extraTrees) {
+        /* Extra Trees is set through two parameters which are different by default */
+        algorithm.parameter().bootstrap = false;
+        algorithm.parameter().splitter = algorithms::decision_forest::training::random;
+    }
+    else {
+        algorithm.parameter().resultsToCompute =
+            algorithms::decision_forest::training::computeOutOfBagError;
+    }
```
KulikovNikita, line 101: Please just move ExtraTrees examples into a different file

KulikovNikita (commented):
Generally, I don't think it is a good idea to mix RandomForest and ExtraTrees examples into one.

examples/daal/cpp/source/decision_forest/df_cls_default_dense_batch.cpp
icfaust: @Alexsandruss, I now have two others (including @Alexandr-Solovev ) who have recommended separate files, is there a standard for this or should we make one? I'd side with Nikita on this, and I've reverted my changes back to what he recommends.

KulikovNikita (approved):

examples/daal/cpp/source/decision_forest/df_cls_default_dense_batch.cpp
samir-nasibli: @icfaust Examples should be simple as much as possible and for separate algo separate one.

Alexsandruss (approved):

PR #2436, opened 2023-07-12
WIP: Distributed LOM optimization

Alexandr-Solovev:


inteldimitrius (approved):
Looks good to me

ethanglaser (approved):

cpp/oneapi/dal/algo/basic_statistics/backend/gpu/compute_kernel_dense_impl_dpc.cpp
```diff
@@ -1209,7 +1081,7 @@ compute_kernel_dense_impl<Float, List>::compute_by_blocks(const pr::ndview<Float
     DECLSET_IF(Float*, asum2cent_ptr, sum2cent_based_stat, ndbuf.get_sum2cent().get_mutable_data())
 
     const auto* data_ptr = data.get_data();
-    const auto wg_size = be::device_max_wg_size(this->q_);
```
ahuber21, line 1212: I agree with removing `this->`, but I have question about the variable name. 
http://oneapi-src.github.io/oneDAL/contribution/coding_guide.html

`q_` is wrong for many reasons. First, the rule about private members says
> `private` class members and methods should have an additional underscore (_) at the beginning

... and not at the end.

And then there are also rules about variable names 

* Give descriptive and meaningful names.
* Avoid ambiguity.
* Do not abbriviate (sic) words.
* Do not remove letters from words

`q_` violates all of these rules.

I see that this is a more global problem because it's also done in other files. Would it make sense to refactor this?

cpp/oneapi/dal/algo/basic_statistics/backend/gpu/compute_kernel_dense_impl_dpc.cpp
```diff
@@ -193,8 +192,7 @@ struct singlepass_processor_kernel {
         (DefferedFin && check_mask_flag(bs_list::varc | bs_list::stdev | bs_list::vart, params));
     std::conditional_t<output_sum2cent, Float*, empty> sum2cent_ptr;
 
-    constexpr static inline bool output_mean =
```
ahuber21, line 196: Why can `!DeferredFin` be removed?

ahuber21 (changes requested):
Could you describe in a few words how your changes optimize LOM so that I can properly review them?

cpp/oneapi/dal/algo/basic_statistics/backend/gpu/compute_kernel_dense_impl_dpc.cpp
Alexandr-Solovev: I also agree with this point, that in different algorithms we have different names for common variables, like queues or communicators, and it is a global problem. But, my view is that we need to do it consistency in additional pr for each algo. For LOM I will fix it in new commit. Thanks!

cpp/oneapi/dal/algo/basic_statistics/backend/gpu/compute_kernel_dense_impl_dpc.cpp
Alexandr-Solovev: I've removed `!DeferredFin` because I've changed logic for merging data from different gpus. In this implementation I do use local means not only for single gpu, where its final result, also do it for distributed as a subproduct for sum2cent computation.
The idea of optimizations was based on simplifying merging function. I changed logic from comm.allgather and loop(column count x rank_count) which compute a lot of statistics on data after allgather, to comm.allreduce and small loop(column count) for computing data which based only on merged data. Also, bigger loop computed sum+means(as a current loop's sum divided by current mean's scale)+sum2cent, new one just sum2cent+means(as a allreduced sum + number of rows)

ahuber21 (approved):

PR #2437, opened 2023-07-14
Implement Newton-CG solver

avolkov-intel:
# Description
Add new optimization method newton-cg (or hessian free) as primitive.
This method is an analogue for this optimization method from scipy https://docs.scipy.org/doc/scipy/reference/optimize.minimize-newtoncg.html

Changes proposed in this pull request:
-Implement conjugate gradients solver
-Add tests for cg-solver
-Develop function wrapper interface
-Develop newton-cg optimization primitive
-Develop line search method for 1d-optimization https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.line_search.html
-Add tests for newton-cg solver

PR #2451, opened 2023-08-03
fix: fix deduction of min template in newton cg

razdoburdin:
# Description
Replacing of https://github.com/oneapi-src/oneDAL/pull/2449

https://github.com/oneapi-src/oneDAL/pull/2437 can lead to compile error here https://github.com/oneapi-src/oneDAL/blob/199d4a56ab1913d5f39b561212d5952b06e7a844/cpp/oneapi/dal/backend/primitives/optimizers/newton_cg_dpc.cpp#L63


razdoburdin:
CI failure will be fixed by https://github.com/intel/scikit-learn-intelex/pull/1364

avolkov-intel (approved):

razdoburdin:
GPU related CI failures are caused by the decision tree tests, and are not related to the current PR.

PR #2458, opened 2023-08-09
Fix the include in service_hash_table.h

Vika-F:
* Change the #include "service_utils.h" to the #include "service_arrays.h" in service_hash_table.h. The change is necessary due to the use of TArray template defined in service_arrays.h
* Fix build errors that appeared after the include change


ethanglaser (approved):
LGTM

PR #2460, opened 2023-08-11
Add SHAP calculation to GBT regression

ahuber21:
This PR explores SHAP values (originally added in XGBoost here: https://github.com/dmlc/xgboost/pull/2438). I have taken the same algorithm and modified it to work with oneDAL datastructures. Most performance bottlenecks were removed. However, there is still room for improvement. (For instance a more efficient use of vector instructions in the `unwoundPathSum()` and `unwindPath()` helper functions.

I am creating this as a draft because I am aware of a few things that must be changed. Feel free to review regardless. All comments are helpful at this point.

Things to be addressed (incomplete list, likely to be extended)

- [x] Remove std entities (like std::vector in treeShap).
- [x] Align model builder classification and regression APIs. The property `cover` was added for regression, since it is needed to calculate SHAP values. I have added a dummy `cover = 0` in some places to fix compilation errors, but that obviously needs improvement. Moving `treeShap()` to a common base class and enabling it for both classification and regression is probably the best solution.
- [x] Right now, treeShap execution is hardcoded, even though `predShapContributions` and `predShapInteractions` model parameters are available. The code in `PredictRegressionTask::run()` must be updated accordingly. Additionally, the fitting size of result memory must be allocated in gbt_regression_predict_result_fpt.cpp (also hardcoded to accommodate SHAP contributions).
- [x] It is only possible to calculate **either** SHAP contributions **or** interactions - this must be addressed in the constructor of the model parameters.
- [x] Validate changes don't affect performance / provide benchmarks
- [x] Rebase to current master to fix conflicts.


cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -220,26 +291,320 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::run(const gbt::reg
...
+            uniquePath[i].partialWeight = nextOnePortion * (uniqueDepth + 1) / static_cast<float>((i + 1) * oneFraction);
+            nextOnePortion              = tmp - uniquePath[i].partialWeight * zeroFraction * (uniqueDepth - i) / static_cast<float>(uniqueDepth + 1);
+        }
+        else
+        {
+            uniquePath[i].partialWeight = (uniquePath[i].partialWeight * (uniqueDepth + 1)) / static_cast<float>(zeroFraction * (uniqueDepth - i));
+        }
+    }
+
+    for (auto i = pathIndex; i < uniqueDepth; ++i)
```
KulikovNikita, line 342:
```suggestion
    for (int i = pathIndex; i < uniqueDepth; ++i)
```

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -220,26 +291,320 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::run(const gbt::reg
...
+// extend our decision path with a fraction of one and zero extensions
+template <typename algorithmFPType, CpuType cpu>
+void PredictRegressionTask<algorithmFPType, cpu>::extendPath(PathElement * uniquePath, std::uint32_t uniqueDepth, float zeroFraction,
+                                                             float oneFraction, gbt::prediction::internal::FeatureIndexType featureIndex) const
+{
+    uniquePath[uniqueDepth].featureIndex  = featureIndex;
+    uniquePath[uniqueDepth].zeroFraction  = zeroFraction;
+    uniquePath[uniqueDepth].oneFraction   = oneFraction;
+    uniquePath[uniqueDepth].partialWeight = (uniqueDepth == 0 ? 1.0f : 0.0f);
+    for (int i = uniqueDepth - 1; i >= 0; i--)
```
KulikovNikita, line 312:
```suggestion
    for (int i = uniqueDepth - 1; i >= 0; --i)
```

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -220,26 +291,320 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::run(const gbt::reg
...
+        uniquePath[i].zeroFraction = uniquePath[i + 1].zeroFraction;
+        uniquePath[i].oneFraction  = uniquePath[i + 1].oneFraction;
+    }
+}
+
+// !!! 100% taken from XGBoost !!! WIP
+// determine what the total permutation weight would be if
+// we unwound a previous extension in the decision path
+template <typename algorithmFPType, CpuType cpu>
+float PredictRegressionTask<algorithmFPType, cpu>::unwoundPathSum(const PathElement * uniquePath, std::uint32_t uniqueDepth,
```
KulikovNikita, line 354: Looks like a code duplication with the `unwindPath` function

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -220,26 +291,320 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::run(const gbt::reg
...
+                                                                  std::uint32_t pathIndex) const
```
KulikovNikita, line 355: Why some kernels use `std::uint32_t` while others use `int` instead?

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -220,26 +291,320 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::run(const gbt::reg
...
+        const algorithmFPType * currentX = x + (iRow * nColumnsData);
+        algorithmFPType * phi            = res + (iRow * nColumnsPhi);
+        for (size_t currentTree = iTree; currentTree < iTree + nTrees; ++currentTree)
+        {
+            // regression model builder tree 0 contains only the base_score and must be skipped
+            if (currentTree == 0) continue;
+
+            // prepare memory for unique path data
+            const int depth = _aTree[currentTree]->getMaxLvl() + 2;
+            std::vector<PathElement> uniquePathData((depth * (depth + 1)) / 2);
```
KulikovNikita, line 437: We have an internal agreement not to use entities from STL

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -220,26 +291,320 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::run(const gbt::reg
...
+
+            // TODO: Not using a separate variable (test) for the phi values causes a lot of cache misses and the code
+            //       runs 10x slower - why?
+            std::vector<algorithmFPType> test(nColumnsData, 0);
+            treeShap<hasUnorderedFeatures, hasAnyMissing>(currentTree, currentX, test.data(), nColumnsData, 1, 0, 0, uniquePathData.data(), 1, 1, -1,
```
KulikovNikita, line 442: Too many parameters. What about passing them as structures?

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -220,26 +291,320 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::run(const gbt::reg
...
+    //   - the feature value that is used for the split for internal nodes
+    //   - the tree prediction for leaf nodes
+    // we are accounting for the fact that nodeIndex is counted from 1 (not from 0) as required by helper functions
+    const gbt::prediction::internal::ModelFPType * const splitValues     = tree->getSplitPoints() - 1;
+    const gbt::prediction::internal::FeatureIndexType * const fIndexes   = tree->getFeatureIndexesForSplit() - 1;
+    const gbt::prediction::internal::ModelFPType * const nodeCoverValues = tree->getNodeCoverValues() - 1;
+    const int * const defaultLeft                                        = tree->getDefaultLeftForSplit() - 1;
+
+    // stop if we have no weight coming down to us
+    if (conditionFraction == 0) return;
```
KulikovNikita, line 502: 1) Not a great idea to compare floating point variable with zero. Please compare with machine epsilon
2) You can check it right at the beginning of the function. It can bring (very) minor performance improvement

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -220,26 +291,320 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::run(const gbt::reg
...
+
+    DAAL_ASSERT(hotZeroFraction < 1.0f);
+    DAAL_ASSERT(coldZeroFraction < 1.0f);
+
+    // see if we have already split on this feature,
+    // if so we undo that split so we can redo it for this node
+    std::uint32_t pathIndex = 0;
+    for (; pathIndex <= uniqueDepth; ++pathIndex)
+    {
+        if (uniquePath[pathIndex].featureIndex == splitFeature) break;
```
KulikovNikita, line 555: You can check this thing right in the `for(;...;)` continuation condition

cpp/daal/include/algorithms/decision_forest/decision_forest_classification_model_builder.h
```diff
@@ -112,7 +112,8 @@ class DAAL_EXPORT ModelBuilder
     NodeId addLeafNode(const TreeId treeId, const NodeId parentId, const size_t position, const size_t classLabel)
     {
         NodeId resId;
-        _status |= addLeafNodeInternal(treeId, parentId, position, classLabel, resId);
+        const double cover = 0.0; // TODO: Add cover
```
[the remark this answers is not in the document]
ahuber21: Cover is needed for treeShap. We are using it in regression, not in in classification. This will be added. For now it's just there to have symmetric classification & regression APIs

cpp/daal/src/algorithms/dtrees/dtrees_model_impl.h
```diff
@@ -53,19 +53,21 @@ struct DecisionTreeNode
...
 {
 public:
-    DecisionTreeTable(size_t rowCount = 0) : data_management::AOSNumericTable(sizeof(DecisionTreeNode), 3, rowCount)
+    DecisionTreeTable(size_t rowCount = 0) : data_management::AOSNumericTable(sizeof(DecisionTreeNode), 5, rowCount)
     {
         setFeature<int>(0, DAAL_STRUCT_MEMBER_OFFSET(DecisionTreeNode, featureIndex));
         setFeature<ClassIndexType>(1, DAAL_STRUCT_MEMBER_OFFSET(DecisionTreeNode, leftIndexOrClass));
         setFeature<ModelFPType>(2, DAAL_STRUCT_MEMBER_OFFSET(DecisionTreeNode, featureValueOrResponse));
         setFeature<int>(3, DAAL_STRUCT_MEMBER_OFFSET(DecisionTreeNode, defaultLeft));
+        setFeature<double>(4, DAAL_STRUCT_MEMBER_OFFSET(DecisionTreeNode, cover));
```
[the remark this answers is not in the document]
ahuber21: That is because there was a bug here. In master it's already fixed. After rebase there will be only one new line and an increase by 1.

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
ahuber21: I've simplified the user-facing function. The internal (recursive) part is keeping its many params. I don't want to deviate too much from other TreeSHAP implementations.

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
ahuber21: It has become a bit more complex in some cases (added DAAL_ASSERTS). Would like to keep it consistent across different versions of TreeSHAP.

ahuber21:
@Alexsandruss please use this CI http://intel-ci.intel.com/ee5dea85-3976-f168-8976-a4bf010d0e2e

napetrov:
@ahuber21  - please rebase/merge from master - currently  pr shows changes that already in master as a diff , hard to review

ahuber21:
http://intel-ci.intel.com/ee674475-2b8f-f13d-9e94-a4bf010d0e2e

ahuber21:
@napetrov I have rebased the branch.
@KulikovNikita @razdoburdin @Alexsandruss the latest CI run (see previous comment) is from just before the rebase and I expect all checks to succeed (the previous run[*] only failed one Python test due to a missing dependency, which I resolved).
Note, that we can not do an `intelci: run` here, as changes from 3 repositories are required.

[*] http://intel-ci.intel.com/ee66c989-27cb-f119-9855-a4bf010d0e2e

ahuber21:
Updated CI 
http://intel-ci.intel.com/ee691f19-524a-f1d7-8419-a4bf010d0e2e

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -109,44 +114,120 @@ protected:
...
+        }
+    }
+
+    // TODO: Add vectorBlockSize templates, similar to predict
+    // template <size_t vectorBlockSize>
+    inline services::Status predictContributions(size_t iTree, size_t nTrees, size_t nRowsData, const algorithmFPType * x, algorithmFPType * res,
+                                                 int condition, FeatureIndexType conditionFeature, const DimType & dim)
+    {
+        const size_t nColumnsData = dim.nCols;
+        const bool hasAnyMissing  = checkForMissing(x, nTrees, nRowsData, nColumnsData);
```
KulikovNikita, line 188:
```suggestion
        const bool hasAnyMissing  = checkForMissing(x, nTrees, nRowsData, dim.nCols);
```

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -109,44 +114,120 @@ protected:
...
+        }
+    }
+
+    // TODO: Add vectorBlockSize templates, similar to predict
+    // template <size_t vectorBlockSize>
+    inline services::Status predictContributionInteractions(size_t iTree, size_t nTrees, size_t nRowsData, const algorithmFPType * x,
+                                                            const algorithmFPType * nominal, algorithmFPType * res, const DimType & dim)
+    {
+        const size_t nColumnsData = dim.nCols;
+        const bool hasAnyMissing  = checkForMissing(x, nTrees, nRowsData, nColumnsData);
```
KulikovNikita, line 223:
```suggestion
        const bool hasAnyMissing  = checkForMissing(x, nTrees, nRowsData, dim.nCols);
```

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -156,42 +237,45 @@ protected:
...
-                        const DimType & dim, BooleanConstant<true> keepLooking)
+    inline void predict(size_t iTree, size_t nTrees, size_t nRows, const algorithmFPType * x, algorithmFPType * res, const DimType & dim,
+                        BooleanConstant<true> keepLooking, size_t resIncrement)
     {
+        const size_t nColumns                = dim.nCols;
         constexpr size_t vectorBlockSizeStep = DimType::vectorBlockSizeStep;
         if (dim.vectorBlockSizeFactor == vectorBlockSizeFactor)
         {
-            predict<vectorBlockSizeFactor * vectorBlockSizeStep>(iTree, nTrees, nRows, nColumns, x, res);
+            predict<vectorBlockSizeFactor * vectorBlockSizeStep>(iTree, nTrees, nRows, nColumns, x, res, resIncrement);
```
KulikovNikita, line 249:
```suggestion
            predict<vectorBlockSizeFactor * vectorBlockSizeStep>(iTree, nTrees, nRows, dim.nCols, x, res, resIncrement);
```

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -207,39 +291,182 @@ protected:
...
-    DAAL_CHECK_MALLOC(this->_featHelper.init(*this->_data));
+    DAAL_CHECK_MALLOC(_featHelper.init(*_data));
     const auto nTreesTotal = (nIterations ? nIterations : m->size());
-    this->_aTree.reset(nTreesTotal);
-    DAAL_CHECK_MALLOC(this->_aTree.get());
-    for (size_t i = 0; i < nTreesTotal; ++i) this->_aTree[i] = m->at(i);
-    return runInternal(pHostApp, this->_res);
+    _aTree.reset(nTreesTotal);
+    DAAL_CHECK_MALLOC(_aTree.get());
+    for (size_t i = 0; i < nTreesTotal; ++i) _aTree[i] = m->at(i);
```
KulikovNikita, line 313: Can it be vectorized?

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -207,39 +291,182 @@ protected:
...
+    Status st;
+
+    const size_t nColumnsData  = dim.nCols;
+    const size_t nColumnsPhi   = nColumnsData + 1;
+    const size_t biasTermIndex = nColumnsPhi - 1;
+
+    for (size_t iRow = 0; iRow < nRowsData; ++iRow)
+    {
+        const algorithmFPType * currentX = x + (iRow * nColumnsData);
+        algorithmFPType * phi            = res + (iRow * nColumnsPhi);
```
KulikovNikita, line 344:
```suggestion
        const algorithmFPType * const currentX = x + (iRow * nColumnsData);
        algorithmFPType * phi     const        = res + (iRow * nColumnsPhi);
```

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -207,39 +291,182 @@ protected:
...
+                                                                                                conditionFeature);
+        }
+
+        if (condition == 0)
+        {
+            // find bias term by leveraging bias = nominal - sum_i phi_i
+            for (int iFeature = 0; iFeature < nColumnsData; ++iFeature)
+            {
+                phi[biasTermIndex] -= phi[iFeature];
+            }
```
KulikovNikita, line 358: It can be vectorized (not a lot of performance boost though).

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -207,39 +291,182 @@ protected:
...
+    // Copy nominal values (for bias term) to the condition = 0 buffer
+    PRAGMA_IVDEP
+    PRAGMA_VECTOR_ALWAYS
+    for (size_t i = 0; i < nRowsData; ++i)
+    {
+        contribsDiag[i * nColumnsPhi + biasTermIndex] = nominal[i];
+    }
+
+    predictContributions(iTree, nTrees, nRowsData, x, contribsDiag, 0, 0, dim);
+    for (size_t i = 0; i < nColumnsPhi; ++i)
```
KulikovNikita, line 414: I highly recommend to use `0ul` when initializing variable supposed to be `size_t` aka `unsigned long long int`

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -207,39 +291,182 @@ protected:
...
+
+        predictContributions(iTree, nTrees, nRowsData, x, contribsOff, -1, i, dim);
+        predictContributions(iTree, nTrees, nRowsData, x, contribsOn, 1, i, dim);
+
+        for (size_t j = 0; j < nRowsData; ++j)
+        {
+            const unsigned dataRowOffset = j * interactionMatrixSize + i * nColumnsPhi;
+            const unsigned columnOffset  = j * nColumnsPhi;
+            res[dataRowOffset + i]       = 0;
+            for (size_t k = 0; k < nColumnsPhi; ++k)
```
KulikovNikita, line 427: Should it really have 3 nested `for` loops?

KulikovNikita: I believe it can be vectorized.

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -207,39 +291,182 @@ protected:
...
+            {
+                // fill in the diagonal with additive effects, and off-diagonal with the interactions
+                if (k == i)
+                {
+                    res[dataRowOffset + i] += contribsDiag[columnOffset + k];
+                }
+                else
+                {
+                    res[dataRowOffset + k] = (contribsOn[columnOffset + k] - contribsOff[columnOffset + k]) / 2.0f;
```
KulikovNikita, line 436: Please never divide by constant value. Use explicit optimization:
```suggestion
                    constexpr algorithmFPType half(0.5);
                    res[dataRowOffset + k] = (contribsOn[columnOffset + k] - contribsOff[columnOffset + k]) * half;
```

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -207,39 +291,182 @@ protected:
...
+    const size_t nColumnsData  = dim.nCols;
+    const size_t nColumnsPhi   = nColumnsData + 1;
+    const size_t biasTermIndex = nColumnsPhi - 1;
+
+    const size_t interactionMatrixSize = nColumnsPhi * nColumnsPhi;
+
+    // Allocate buffer for 3 matrices for algorithmFPType of size (nRowsData, nColumnsData)
+    const size_t elementsInMatrix = nRowsData * nColumnsPhi;
+    const size_t bufferSize       = 3 * sizeof(algorithmFPType) * elementsInMatrix;
+    algorithmFPType * buffer      = static_cast<algorithmFPType *>(daal_calloc(bufferSize));
```
KulikovNikita, line 393: Please consider using data management entities for that. It can be just plain `SharedPtr` or `TArray` by choice.

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -248,11 +475,68 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::runInternal(servic
...
+                safeStat |= predictContributions(iTree, nTreesToUse, nRowsToProcess, xBD.get(), resRowPtr, 0, 0, dim);
+            }
+            else if (predShapInteractions)
+            {
+                // thread-local write rows into global result buffer
+                WriteOnlyRows<algorithmFPType, cpu> resRow(result, iStartRow, nRowsToProcess);
+                DAAL_CHECK_BLOCK_STATUS_THR(resRow);
+
+                // nominal values are required to calculate the correct bias term
+                algorithmFPType * nominal = static_cast<algorithmFPType *>(daal_malloc(nRowsToProcess * sizeof(algorithmFPType)));
```
KulikovNikita, line 509: Please use already existing entities

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -248,11 +475,68 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::runInternal(servic
...
+                if (!nominal)
+                {
+                    safeStat.add(ErrorMemoryAllocationFailed);
+                    return;
+                }
+                for (size_t i = 0; i < nRowsToProcess; ++i)
+                {
+                    nominal[i] = predictionBias;
+                }
```
KulikovNikita, line 518: It can be optimized. Please consider using `fill` function

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -248,11 +475,68 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::runInternal(servic
...
+
+                // TODO: support tree weights
+                safeStat |= predictContributionInteractions(iTree, nTreesToUse, nRowsToProcess, xBD.get(), nominal, resRow.get(), dim);
+
+                daal_free(nominal);
+            }
+            else
+            {
+                algorithmFPType * res = resMatrix.get() + iStartRow;
+                if ((predictionBias * predictionBias) > (DBL_EPSILON * DBL_EPSILON))
```
KulikovNikita, line 529: 1) Comparison operands order
2) Please use `constexpr` variable to compare with
3) (Important) `DBL_EPSILON` is extremely small. Square of this value is even smaller. There is no point in comparison with it. Please just use something like comparison with an absolute value.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.cpp
```diff
@@ -0,0 +1,214 @@
...
+    return fallback;
+}
+namespace internal
+{
+
+namespace v0
+{
+
+// extend our decision path with a fraction of one and zero extensions
+void extendPath(PathElement * uniquePath, size_t uniqueDepth, float zeroFraction, float oneFraction, int featureIndex)
```
KulikovNikita, line 44: Should we use different `fp` types?

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    const bool isLeaf = gbt::internal::ModelImpl::nodeIsLeaf(nodeIndex, *tree, depth);
+
+    // leaf node
+    if (isLeaf)
+    {
+        for (size_t i = 1; i <= uniqueDepth; ++i)
+        {
+            const float w          = unwoundPathSum(uniquePath, uniqueDepth, i);
```
KulikovNikita, line 133: `algorithmFPType`

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
```diff
@@ -0,0 +1,476 @@
...
+ * \param depth how deep are we in the tree
+ * \param uniqueDepth how many unique features are above the current node in the tree
+ * \param parentUniquePath a vector of statistics about our current path through the tree
+ * \param parentZeroFraction what fraction of the parent path weight is coming as 0 (integrated)
+ * \param parentOneFraction what fraction of the parent path weight is coming as 1 (fixed)
+ * \param parentFeatureIndex what feature the parent node used to split
+ * \param conditionFraction what fraction of the current weight matches our conditioning feature
+ */
+template <typename algorithmFPType, bool hasUnorderedFeatures, bool hasAnyMissing>
+inline void treeShap(const gbt::internal::GbtDecisionTree * tree, const algorithmFPType * x, algorithmFPType * phi,
```
KulikovNikita, line 101: Do we want to return status?

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    const bool isLeaf = gbt::internal::ModelImpl::nodeIsLeaf(nodeIndex, *tree, depth);
+
+    if (isLeaf)
+    {
+        // +1 to account for -1 in splitValues array
+        const unsigned valuesOffset = nodeIndex * numOutputs;
+        unsigned valuesNonZeroInd   = 0;
+        unsigned valuesNonZeroCount = 0;
```
KulikovNikita, line 295: Why don't you use `std::uint32_t`?

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
```diff
@@ -0,0 +1,476 @@
...
+            if (valuesNonZeroCount == 1)
+            {
+                phi[phiOffset + valuesNonZeroInd] += scale * splitValues[valuesOffset + valuesNonZeroInd];
+            }
+            else
+            {
+                for (unsigned j = 0; j < numOutputs; ++j)
+                {
+                    phi[phiOffset + j] += scale * splitValues[valuesOffset + j];
+                }
```
KulikovNikita, line 331: It can beoptimized

cpp/daal/src/algorithms/dtrees/gbt/treeshap.cpp
```diff
@@ -0,0 +1,214 @@
...
+        }
+        else if (zeroFraction != 0)
+        {
+            total += (uniquePath[i].partialWeight / zeroFraction) / ((uniqueDepth - i) / static_cast<float>(uniqueDepth + 1));
+        }
+        else
+        {
+            DAAL_ASSERT(uniquePath[i].partialWeight == 0);
+        }
+    }
```
razdoburdin, line 113:
```suggestion
  if (oneFraction != 0)
  {
    float fraq = zeroFraction / oneFraction;
    for (int i = uniqueDepth - 1; i >= 0; --i)
    {
      const float tmp = nextOnePortion / (i + 1);
      total += tmp;
      nextOnePortion = uniquePath[i].partialWeight - tmp * fraq * (uniqueDepth - i);
    }
    total *= (uniqueDepth + 1) / oneFraction;
  }
  else if (zeroFraction != 0)
  {
    for (int i = uniqueDepth - 1; i >= 0; --i)
    {
      total += uniquePath[i].partialWeight  / (uniqueDepth - i);
    }
    total *= (uniqueDepth + 1) / zeroFraction;
  }
  else
  {
    DAAL_ASSERT(uniquePath[i].partialWeight == 0);
  }
```

razdoburdin (commented):
I guess some optimization can be applied.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
```diff
@@ -0,0 +1,482 @@
...
+ * \param conditionFeature the index of the feature to fix
+ */
+template <typename algorithmFPType, bool hasUnorderedFeatures, bool hasAnyMissing>
+inline services::Status treeShap(const gbt::internal::GbtDecisionTree * tree, const algorithmFPType * x, algorithmFPType * phi,
+                                 const FeatureTypes * featureHelper, int condition, FeatureIndexType conditionFeature)
+{
+    services::Status st;
+    const int depth              = tree->getMaxLvl() + 2;
+    const size_t nUniquePath     = ((depth * (depth + 1)) / 2);
+    PathElement * uniquePathData = static_cast<PathElement *>(daal_malloc(sizeof(PathElement) * nUniquePath));
```
Vika-F, line 220: Please use TArray instead of the raw pointer.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
```diff
@@ -0,0 +1,482 @@
...
+        for (unsigned i = 1; i <= uniqueDepth; ++i)
+        {
+            const PathElement & el   = uniquePath[i];
+            const unsigned phiOffset = el.featureIndex * numOutputs;
+            // update contributions to SHAP values for features satisfying the thresholds and not satisfying the thresholds separately
+            if (el.oneFraction != 0)
+            {
+                const algorithmFPType w = unwoundPathSum(uniquePath, pWeights, uniqueDepth, uniqueDepthPWeights, i);
+                scale                   = w * pWeightsResidual * (1 - el.zeroFraction) * conditionFraction;
+            }
```
Vika-F, line 317: Vectorization might bring additional performance to this part.

Vika-F (commented):
It's great to see that theeSHAP algorithm is integrated into the codebase!
But I am concerned with the use of environment variable to switch across the two implementations.
Please also see the other comments below.

cpp/daal/include/algorithms/gradient_boosted_trees/gbt_classification_model_builder.h
```diff
@@ -127,16 +136,25 @@ class DAAL_EXPORT ModelBuilder
...
-        _status |= addSplitNodeInternal(treeId, parentId, position, featureIndex, featureValue, resId, defaultLeft);
+        _status |= addSplitNodeInternal(treeId, parentId, position, featureIndex, featureValue, defaultLeft, cover, resId);
         services::throwIfPossible(_status);
         return resId;
     }
 
+    /**
+    *  \DAAL_DEPRECATED
+    */
+    DAAL_DEPRECATED NodeId addSplitNode(TreeId treeId, NodeId parentId, size_t position, size_t featureIndex, double featureValue)
```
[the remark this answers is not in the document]
ahuber21: This is the original `addSplitNode` API. I have decided against adding default values to the other version to better separate deprecated from recommended usage.

cpp/daal/include/algorithms/gradient_boosted_trees/gbt_regression_model.h
```diff
@@ -123,8 +123,26 @@ class DAAL_EXPORT Model : public algorithms::regression::Model
      */
     virtual size_t getNumberOfTrees() const = 0;
 
+    /**
+     * \brief Set the Prediction Bias term
+     *
+     * \param value global prediction bias
+     */
+    void setPredictionBias(double value) { _predictionBias = value; }
```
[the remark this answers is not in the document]
ahuber21: 1. The virtual functions are deleted and defined in the regression/classification specifications - but setting/getting the prediction bias is common to both model types and therefore implemented in this base class (do you think this breaks anything?)
2. Why?

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_model_builder.cpp
```diff
@@ -83,19 +83,22 @@ services::Status ModelBuilder::createTreeInternal(size_t nNodes, size_t clasLabe
         {
             return Status(ErrorID::ErrorIncorrectParameter);
         }
-        if (clasLabel > (_nClasses - 1))
+        if (classLabel > (_nClasses - 1))
```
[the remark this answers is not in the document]
ahuber21: I only fixed a typo here, but sure, changed it :)

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
```diff
@@ -66,6 +66,13 @@ public:
     PredictBinaryClassificationTask(const NumericTable * x, NumericTable * y, NumericTable * prob) : super(x, y), _prob(prob) {}
     services::Status run(const gbt::classification::internal::ModelImpl * m, size_t nIterations, services::HostAppIface * pHostApp)
     {
+        double bias = m->getPredictionBias();
+        DAAL_ASSERT((bias >= 0) && (bias < 1));
```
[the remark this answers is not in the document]
ahuber21: I thought 0 would be useful for the `logitraw` objective, but turns out we're using 0.5 here. Removed the possibility for 0 bias.

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
```diff
@@ -66,6 +66,13 @@ public:
     PredictBinaryClassificationTask(const NumericTable * x, NumericTable * y, NumericTable * prob) : super(x, y), _prob(prob) {}
     services::Status run(const gbt::classification::internal::ModelImpl * m, size_t nIterations, services::HostAppIface * pHostApp)
     {
+        double bias = m->getPredictionBias();
```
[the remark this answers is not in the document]
ahuber21: This comment triggered a complete refactoring of `gbt_classification_predict_dense_default_batch_impl.i` where I have separated class / function definitions from implementations for the binary and multiclass classes.

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
```diff
@@ -433,7 +440,7 @@ void PredictMulticlassTask<algorithmFPType, cpu>::predictByTreesVector(algorithm
 }
 
 template <typename algorithmFPType, CpuType cpu>
-services::Status PredictMulticlassTask<algorithmFPType, cpu>::predictByAllTrees(size_t nTreesTotal, size_t nClasses, const DimType & dim)
+services::Status PredictMulticlassTask<algorithmFPType, cpu>::predictByAllTrees(size_t nTreesTotal, size_t nClasses, double bias, const DimType & dim)
```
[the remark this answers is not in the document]
ahuber21: The `bias` is a member of `ModelImpl`, which is not templated. Suboptimal, but the only solution I found. Do you have a better idea?

cpp/daal/src/algorithms/dtrees/gbt/gbt_model.cpp
```diff
@@ -108,10 +107,10 @@ void ModelImpl::traverseBFS(size_t iTree, tree_utils::regression::TreeNodeVisito
...
     const GbtDecisionTree & gbtTree = *at(iTree);
 
-    const gbt::prediction::internal::ModelFPType * splitPoints        = gbtTree.getSplitPoints();
-    const gbt::prediction::internal::FeatureIndexType * splitFeatures = gbtTree.getFeatureIndexesForSplit();
-    const int * nodeSamplesCount                                      = getNodeSampleCount(iTree);
-    const double * imp                                                = getImpVals(iTree);
+    const ModelFPType * splitPoints        = gbtTree.getSplitPoints();
+    const FeatureIndexType * splitFeatures = gbtTree.getFeatureIndexesForSplit();
+    const int * nodeSamplesCount           = getNodeSampleCount(iTree);
+    const double * imp                     = getImpVals(iTree);
```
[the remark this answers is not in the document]
ahuber21: I only touched this code because I moved it from regression model to the base model. I'm hesitant to change anything in the implementation.

cpp/daal/src/algorithms/dtrees/gbt/gbt_model_impl.h
```diff
@@ -264,6 +282,29 @@ using TreeImpRegression = GbtTreeImpl<dtrees::internal::TreeNodeRegression<Regre
...
+};
+
+class DecisionTreeTable : public data_management::AOSNumericTable
+{
+public:
+    DecisionTreeTable(size_t rowCount, services::Status & st) : data_management::AOSNumericTable(sizeof(DecisionTreeNode), 3, rowCount, st)
+    {
+        setFeature<size_t>(0, DAAL_STRUCT_MEMBER_OFFSET(DecisionTreeNode, dimension));
+        setFeature<size_t>(1, DAAL_STRUCT_MEMBER_OFFSET(DecisionTreeNode, leftIndexOrClass));
+        setFeature<double>(2, DAAL_STRUCT_MEMBER_OFFSET(DecisionTreeNode, cutPointOrDependantVariable));
```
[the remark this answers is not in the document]
ahuber21: As I said, I'm just refactoring code. Can you please open an issue / ticket for this request?

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -62,40 +64,43 @@ public:
...
+
+    services::Status run(const gbt::regression::internal::ModelImpl * m, size_t nIterations, services::HostAppIface * pHostApp,
+                         bool predShapContributions, bool predShapInteractions);
 
 protected:
     template <bool hasUnorderedFeatures, bool hasAnyMissing>
     using dispatcher_t = gbt::prediction::internal::PredictDispatcher<hasUnorderedFeatures, hasAnyMissing>;
 
-    services::Status runInternal(services::HostAppIface * pHostApp, NumericTable * result);
+    services::Status runInternal(services::HostAppIface * pHostApp, NumericTable * result, double predictionBias, bool predShapContributions,
```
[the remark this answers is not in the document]
ahuber21: See above, let's move the discussion there.

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
ahuber21: Likely, I'm adding `PRAGMA_VECTOR_ALWAYS` - the tests will show if it worked

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
ahuber21: Changed where I was quickly able to find it

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
ahuber21: Yes. IIRC they have 4 in XGBoost (due to chunking)

ahuber21: About vectorization: I'd say story for another day.

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
ahuber21: Yeah, doing DBL_EPSILON * DBL_EPSILON is completely pointless, how could I miss this.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.cpp
ahuber21: Could be an upgrade, the reference implementations all use `float` though

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
ahuber21: I return a status for all functions that allocate memory. The recursive parts of `treeShap` are kept `void`.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
ahuber21: Can it be a follow-up task?

cpp/daal/src/algorithms/dtrees/gbt/treeshap.cpp
```diff
@@ -0,0 +1,214 @@
...
+    const float oneFraction  = uniquePath[pathIndex].oneFraction;
+    const float zeroFraction = uniquePath[pathIndex].zeroFraction;
+    float nextOnePortion     = uniquePath[uniqueDepth].partialWeight;
+
+    for (int i = uniqueDepth - 1; i >= 0; --i)
+    {
+        if (oneFraction != 0)
+        {
+            const float tmp             = uniquePath[i].partialWeight;
+            uniquePath[i].partialWeight = nextOnePortion * (uniqueDepth + 1) / static_cast<float>((i + 1) * oneFraction);
```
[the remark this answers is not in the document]
ahuber21: Taken from the reference implementation, and I think they added it to be explicit and to avoid implicit casting.

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
razdoburdin: If the result is symmetric, it should be possible to reduce the inner loop to something like that:
```
res[dataRowOffset + i] = contribsDiag[columnOffset + k];
for (size_t k = i + 1; k < nColumnsPhi; ++k) {
   auto val = (contribsOn[columnOffset + k] - contribsOff[columnOffset + k]) / 2.0f;
   res[dataRowOffset + k] = val;
   res[j * interactionMatrixSize + k * nColumnsPhi + i] = val;
   res[dataRowOffset + i] -= 2 * val;
}
```

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -248,11 +475,68 @@ services::Status PredictRegressionTask<algorithmFPType, cpu>::runInternal(servic
...
-            algorithmFPType * res = resBD.get() + iStartRow;
 
-            predict(iTree, nTreesTotal, nRowsToProcess, dim.nCols, xBD.get(), res, dim);
+            if (predShapContributions)
+            {
+                // thread-local write rows into global result buffer
+                WriteOnlyRows<algorithmFPType, cpu> resRow(result, iStartRow, nRowsToProcess);
+                DAAL_CHECK_BLOCK_STATUS_THR(resRow);
+                auto resRowPtr = resRow.get();
+                if ((predictionBias * predictionBias) > (DBL_EPSILON * DBL_EPSILON))
```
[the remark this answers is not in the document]
ahuber21: Yes, Nikita pointed that out, too. Thanks for catching it. Brain fart on my end :)

cpp/daal/src/algorithms/dtrees/gbt/treeshap.cpp
```diff
@@ -0,0 +1,214 @@
...
+    }
+    return fallback;
+}
+namespace internal
+{
+
+namespace v0
+{
+
+// extend our decision path with a fraction of one and zero extensions
```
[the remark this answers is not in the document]
ahuber21: It is happening in the header and currently waiting for approval.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.cpp
```diff
@@ -0,0 +1,214 @@
...
+    uniquePath[uniqueDepth].oneFraction  = oneFraction;
+    if (oneFraction != 0)
+    {
+        // extend partialWeights iff the feature of the last split satisfies the threshold
+        partialWeights[uniqueDepthPartialWeights] = (uniqueDepthPartialWeights == 0 ? 1.0f : 0.0f);
+        for (int i = uniqueDepthPartialWeights - 1; i >= 0; i--)
+        {
+            partialWeights[i + 1] += partialWeights[i] * (i + 1) / static_cast<float>(uniqueDepth + 1);
+            partialWeights[i] *= zeroFraction * (uniqueDepth - i) / static_cast<float>(uniqueDepth + 1);
+        }
```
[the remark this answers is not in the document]
ahuber21: We can go through this once more, hunting down small inefficiencies. I tested such changes in various places and my (naive) benchmarks didn't see a speedup. So I didn't want to spend too much time on "premature optimization" and rather come back to this later. The changes are already big enough, and deviating too much from the reference implementations could risk the overall correctness.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.cpp
```diff
@@ -0,0 +1,214 @@
...
+    }
+}
+
+void unwindPath(PathElement * uniquePath, float * partialWeights, unsigned uniqueDepth, unsigned uniqueDepthPartialWeights, unsigned pathIndex)
+{
+    const float oneFraction  = uniquePath[pathIndex].oneFraction;
+    const float zeroFraction = uniquePath[pathIndex].zeroFraction;
+    float nextOnePortion     = partialWeights[uniqueDepthPartialWeights];
+
+    if (oneFraction != 0)
```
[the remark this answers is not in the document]
ahuber21: Completely dependent on the data

ahuber21:
Pushed updates after applying review comments. Only the usage of `TArray` / `service_memset` and the dispatching logic is missing.

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_predict_dense_default_batch_impl.i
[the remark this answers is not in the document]
ahuber21: Update: It is required for proper LightGBM model support

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
```diff
@@ -109,44 +114,120 @@ protected:
...
+            return predictContributions<false>(iTree, nTrees, nRowsData, x, res, condition, conditionFeature, dim);
+        }
+    }
+
+    template <bool hasUnorderedFeatures, bool hasAnyMissing>
+    inline services::Status predictContributionInteractions(size_t iTree, size_t nTrees, size_t nRowsData, const algorithmFPType * x,
+                                                            const algorithmFPType * nominal, algorithmFPType * res, const DimType & dim);
+
+    template <bool hasAnyMissing>
+    inline services::Status predictContributionInteractions(size_t iTree, size_t nTrees, size_t nRowsData, const algorithmFPType * x,
```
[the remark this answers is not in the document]
ahuber21: I would need your help with that - haven't used it before. But aren't std primitives disallowed in daal? @Vika-F

cpp/daal/src/algorithms/dtrees/gbt/treeshap.cpp
```diff
@@ -0,0 +1,214 @@
...
+    const float zeroFraction = uniquePath[pathIndex].zeroFraction;
+    float nextOnePortion     = uniquePath[uniqueDepth].partialWeight;
+
+    for (int i = uniqueDepth - 1; i >= 0; --i)
+    {
+        if (oneFraction != 0)
+        {
+            const float tmp             = uniquePath[i].partialWeight;
+            uniquePath[i].partialWeight = nextOnePortion * (uniqueDepth + 1) / static_cast<float>((i + 1) * oneFraction);
+            nextOnePortion              = tmp - uniquePath[i].partialWeight * zeroFraction * (uniqueDepth - i) / static_cast<float>(uniqueDepth + 1);
```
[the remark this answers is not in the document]
ahuber21: Will take another pass through the code for optimization - changes in these loops tend to break things.

cpp/daal/include/algorithms/gradient_boosted_trees/gbt_classification_model.h
```diff
@@ -123,8 +123,26 @@ class DAAL_EXPORT Model : public classifier::Model
...
+     * \return double prediction bias
+     */
+    double getPredictionBias() const { return _predictionBias; }
+
 protected:
     Model() : classifier::Model() {}
+
+private:
+    /* global bias applied to predictions*/
+    double _predictionBias;
```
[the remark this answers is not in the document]
Vika-F: I think this is not fixed yet. The _predictionBias field and implementations of the respective set/get methods need to be moved into internal .cpp files.
Otherwise it will break the binary compatibility: the layout of the Model class had changed with the addition of _predictionBias field.

cpp/daal/include/algorithms/gradient_boosted_trees/gbt_regression_model.h
```diff
@@ -123,8 +123,26 @@ class DAAL_EXPORT Model : public algorithms::regression::Model
...
+     * \return double prediction bias
+     */
+    double getPredictionBias() const { return _predictionBias; }
+
 protected:
     Model();
+
+private:
+    /* global bias applied to predictions*/
+    double _predictionBias;
```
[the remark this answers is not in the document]
Vika-F: Still needs to be fixed.

cpp/daal/include/algorithms/gradient_boosted_trees/gbt_regression_predict_types.h
```diff
@@ -104,9 +104,16 @@ namespace interface1
...
+    Parameter() : daal::algorithms::Parameter(), nIterations(0), predShapContributions(false), predShapInteractions(false) {}
+    Parameter(const Parameter & o)
+        : daal::algorithms::Parameter(o),
+          nIterations(o.nIterations),
+          predShapContributions(o.predShapContributions),
+          predShapInteractions(o.predShapInteractions)
+    {}
+    size_t nIterations;         /*!< Number of iterations of the trained model to be uses for prediction*/
+    bool predShapContributions; /*!< Predict SHAP contributions */
+    bool predShapInteractions;  /*!< Predict SHAP interactions */
```
Vika-F: Please apply the changes similar to classification parameter here:
- resultToCompute instead of boolean flags,
- `baseClass` to shorten the implementation.

cpp/daal/include/algorithms/tree_utils/tree_utils.h
```diff
@@ -59,6 +59,7 @@ struct DAAL_EXPORT SplitNodeDescriptor : public NodeDescriptor
 {
     size_t featureIndex; /*!< Feature used for splitting the node */
     double featureValue; /*!< Threshold value at the node */
+    double coverValue;   /*!< Cover (sum_hess) for the node */
```
Vika-F, line 62: let's try to make our public headers more understandable
```suggestion
    double coverValue;   /*!< Cover, a sum of the Hessian values of the loss function evaluated at the points flowing through the node */
```

cpp/daal/src/algorithms/dtrees/gbt/regression/gbt_regression_predict_dense_default_batch_impl.i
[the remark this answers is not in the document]
Vika-F: Yes, I think we cannot use std::visit here because of STL's templates that do not match with DAAL's building scheme.
I can only think of adding something like a C-style functions' table. Which is also not an elegant solution. A pseudocode:
```cpp

/// define a pointer to the predictContributions function
typedef Status (*predictContributionsFuncPtr) (size_t, size_t, size_t, const algorithmFPType *, algorithmFPType *, int, FeatureIndexType, const DimType &);

struct predictContributionsFunctor {
    /// call the corresponding variant from the functions's table
    Status call(bool hasUnorderedFeatures, bool hasAnyMissing, size_t iTree, ..., const DimType & dim) {
       
        return ptr_[int(hasUnorderedFeatures)][int(hasAnyMissing)](iTree, ..., dim);
    }
private:
    /// table that contains all the template variants of the predictContributions function
    const predictContributionsFuncPtr ptr_[2][2] = {{
            predictContributions<false, false>,
            predictContributions<false, true>
        }, {
            predictContributions<true, false>,
            predictContributions<true, true>
        }
    };
};
```
then it can be used like:
```cpp
Status st = predictContributionsFunctor().call(_featHelper.hasUnorderedFeatures(), checkForMissing(x, nTrees, nRowsData, dim.nCols), iTree, ..., dim);
```

cpp/daal/include/algorithms/gradient_boosted_trees/gbt_classification_model.h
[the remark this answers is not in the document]
ahuber21: My bad and thanks for the explanation!

ahuber21:
http://intel-ci.intel.com/ee6f459b-ec5f-f1b4-98cd-a4bf010d0e2e

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
```diff
@@ -0,0 +1,482 @@
...
+            uniqueDepthPWeights -= 1;
+        }
+    }
+
+    const bool isLeaf = gbt::internal::ModelImpl::nodeIsLeaf(nodeIndex, *tree, depth);
+
+    if (isLeaf)
+    {
+        // +1 to account for -1 in splitValues array
+        const unsigned valuesOffset = nodeIndex * numOutputs;
```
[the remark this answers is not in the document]
Vika-F: In DAAL we are mostly using `size_t` for unsigned integers. Which is 64-bit when compiled with 64-bit compiler.
But in this code `unsigned` is used as an integer type, which is 32-bit.
Here `nodeIndex` is `size_t`. So, I think a clarification is needed to make sure that this code does not break anything when assigning a 64-bit `nodeIndex` to a 32-bit `valueOffset`.
Please white a comment that says that `nodeIndex` is typically varying in the range [0, ..., (for example) 2^20], so we can assume that we can safely assign it to `unsigned`.

Also, it is better to replace `unsigned` with `uint32_t` to explicitly specify the size of the type and to be aligned with other algorithm's codes.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
Vika-F: Agree here with Nikita.
Even if the code is copied from some third party source, we need to align it with our [coding guidelines](https://oneapi-src.github.io/oneDAL/contribution/coding_guide.html).
Otherwise our codebase will evolve into a patchwork someday.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.h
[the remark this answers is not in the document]
ahuber21: I think I wasn't being consistent here. I changed `unsigned` to `uint32_t` and then I went through the code again, making node indices `size_t`. We accept a `size_t nNodes` when constructing the `GbtDecisionTree`, so this is what should be used to index the split values, feature indices, cover values, etc.

ahuber21:
New CI http://intel-ci.intel.com/ee7349e5-0f83-f1d0-84c0-a4bf010d0e2e

ahuber21:
Evaluation of CI failures: LinuxDaal4py is expected to fail, we depend on changes from https://github.com/intel/scikit-learn-intelex/pull/1399

LinuxMakeDPCPP: Happens in other pipelines as well. Needs to be investigated separately.

Vika-F (approved):
As an initial implementation of treeSHAP in oneDAL this code looks good.
LGTM
And the other changes requested by the reviewers can be done in the subsequent PRs.

PR #2475, opened 2023-08-25
Coverity hits fixes

inteldimitrius:
# Description
Fixes coverity hits.

### DIFF
--- a/cpp/daal/include/services/collection.h
+++ b/cpp/daal/include/services/collection.h
@@ -234,9 +234,12 @@ public:
         }
 
         size_t minSize = newCapacity < _size ? newCapacity : _size;
-        for (size_t i = 0; i < minSize; i++)
+        if (_array)
         {
-            newArray[i] = _array[i];
+            for (size_t i = 0; i < minSize; i++)
+            {
+                newArray[i] = _array[i];
+            }
         }
 
         for (size_t i = 0; i < _capacity; i++)
--- a/cpp/daal/src/algorithms/dbscan/dbscan_result_types.cpp
+++ b/cpp/daal/src/algorithms/dbscan/dbscan_result_types.cpp
@@ -70,6 +70,11 @@ services::Status Result::check(const daal::algorithms::Input * input, const daal
     const Input * algInput   = dynamic_cast<Input *>(const_cast<daal::algorithms::Input *>(input));
     const Parameter * algPar = static_cast<const Parameter *>(par);
 
+    if (!algInput)
+    {
+        return services::Status(services::ErrorIncorrectParameter);
+    }
+
     const size_t nRows     = algInput->get(data)->getNumberOfRows();
     const size_t nFeatures = algInput->get(data)->getNumberOfColumns();
 
--- a/cpp/daal/src/algorithms/distributions/uniform/uniform_impl.i
+++ b/cpp/daal/src/algorithms/distributions/uniform/uniform_impl.i
@@ -69,7 +69,11 @@ Status UniformKernel<algorithmFPType, method, cpu>::compute(algorithmFPType a, a
                                                             algorithmFPType * resultArray)
 {
     auto engineImpl = dynamic_cast<daal::algorithms::engines::internal::BatchBaseImpl *>(&engine);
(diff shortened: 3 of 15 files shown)


Alexandr-Solovev (approved):
LGTM

cpp/daal/include/services/collection.h
avolkov-intel, line 1: I see that after that code there is another cycle that goes over _array https://github.com/oneapi-src/oneDAL/blob/38ecb335aa5f72297abfba7920a809788d33add7/cpp/daal/include/services/collection.h#L245
I think it should also be moved inside if-statement

cpp/daal/src/algorithms/dbscan/dbscan_result_types.cpp
avolkov-intel, line 1: I suggest to create new status ErrorIncorrectInput (or ErrorDynamicCastFailed) because status ErrorIncorrectParameter is misleading.

cpp/daal/src/algorithms/distributions/uniform/uniform_impl.i
avolkov-intel, line 1: Maybe we should also change the name of status here

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_container.h
avolkov-intel, line 1: DAAL_ASSERT seems redundant

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
avolkov-intel, line 1: I think we should throw exception if condition is not satisfied

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_training_types.cpp
avolkov-intel, line 1: Add throw exception if algParameter3 is NULL. Moreover, this check should be right after the dynamic cast (above line 112)

cpp/daal/src/algorithms/dtrees/forest/regression/df_regression_train_dense_default_impl.i
avolkov-intel, line 1: Return error status if condition is not satisfied

cpp/daal/src/algorithms/pca/pca_result.cpp
avolkov-intel, line 1: throw exception or return error status if it  is not NULL

cpp/oneapi/dal/detail/common.hpp
avolkov-intel, line 1: I think the original code was correct, and the change will break it

avolkov-intel (changes requested):

cpp/daal/include/services/collection.h
inteldimitrius: It is okay, becuase down in the _array is reassigned. 
In fact it is resize() method for collection. So it creates new array and tries to copy (previously unsafe) existing array.

cpp/daal/src/algorithms/dtrees/forest/regression/df_regression_train_dense_default_impl.i
inteldimitrius: It is okay for splitter not to have best split. It is normal scenario when node will just remain a leaf.

cpp/daal/src/algorithms/pca/pca_result.cpp
inteldimitrius: I added else case with empty constructor in case if previous result is not available.

ethanglaser:
Looks like one test failure - //cpp/oneapi/dal/algo/decision_forest:test_spmd_dpc - is it reproducible locally? I think its a new failure.

ethanglaser:
> 

Never mind, its from master

ethanglaser:
> Looks like one test failure - //cpp/oneapi/dal/algo/decision_forest:test_spmd_dpc - is it reproducible locally? I think its a new failure.

Never mind, seems its from master

avolkov-intel (approved):

PR #2482, opened 2023-08-29
Fmt 10.1.1 update

Alexandr-Solovev:
# Description
Update fmt dependency
Make a workaround for one fmt::format call.


Alexandr-Solovev:
Reopen https://github.com/oneapi-src/oneDAL/pull/2366 with small fix

Alexandr-Solovev:
Its necessary for support icpx in bazel

PR #2499, opened 2023-09-07
Fix coverity hits

md-shafiul-alam:
# Description
- Removed unreachable return statements
- Refactored unreachable branches
- Added checks for non-zero division
- Removed redundant comparisons


cpp/daal/src/algorithms/pca/transform/pca_transform_dense_default_batch_impl.i
```diff
@@ -105,10 +105,6 @@ services::Status TransformKernel<algorithmFPType, method, cpu>::compute(NumericT
     const algorithmFPType * pBasis = basis.get();
 
     size_t numRowsInBlock = _numRowsInBlock;
-    if (numRowsInBlock < 1)
```
ethanglaser, line 108: would we want to replace this with an assertion/check?

ethanglaser:
Please run clang script

ethanglaser (approved):
Looks good to me, will need manager approval as well.

cpp/daal/src/algorithms/pca/transform/pca_transform_dense_default_batch_impl.i
md-shafiul-alam: In this case, _numRowsInBlock is assigned internally in a header file. We do not expect it to be assigned to zero or negatives.

PR #2510, opened 2023-09-12
Update env script according to the latest comments

maria-Petrova:
The PR aims to address the most recent requirements related to the env script on Linux


napetrov (approved):
assuming CI would work

PR #2511, opened 2023-09-12
Fix remaining coverity hits

avolkov-intel:
# Description
Substitute integral_cast_debug check with integral_cast check in constructor to validate parameter and throw exception if it is incorrect


samir-nasibli:
@avolkov-intel could you please provide more detailed description:
what, why and how fixed.

PR #2514, opened 2023-09-13
Revised namespace cleanup

ethanglaser:
- Cleanup of classes, etc. that have since been replaced in newer namespaces, both in oneapi and daal interfaces
- Cleanup of array functionality marked as removal for 2022.1
- Mostly straightforward deletion of affected classes and corresponding cleanup for build/test - some things to more closely review would be weak_learner replacement of interface1 with interface2 and v1 array_impl removal


inteldimitrius (approved):
LGTM

Vika-F (approved):
I had focused on reviewing DAAL's 'classifier' and 'weak_learner' functionality. The changes look good to me. I assume that everything else is Ok as well.

The only thing that can be further refactored is the use of `interface2::Parameter` in *.cpp files. Because now we don't need to use `interface2` in those codes, as there is the only one variant of the 'Parameter' class. I've left one comment about this, but there are the cases in several other algorithms as well.

The refactoring part can be done in a separate PR (if considered necessary).

PR #2524, opened 2023-09-26
fix: fix clang format step

Alexandr-Solovev:


napetrov:
@mergify copy mergify/bp/rls/2024.0.0-rls/pr-2522

PR #88, opened 2019-09-24
Exception messages have been made secure

masdevas:

PR #2527, opened 2023-09-26
Introduce dispatching in oneAPI covariance algorithm  

Vika-F:
Only batch algorithm was modified for now. Online part of the covariance algorithm will be modified further in a separate PR.

- Added computre_parameters_cpu and computre_parameters_gpu classes in covariance algorithm. Those classes encapsulate the dispatching functionality for CPU and GPU respectively.
- Added test for dispatching functionality in covariance algorithm.


inteldimitrius (approved):
LGTM

cpp/oneapi/dal/algo/covariance/detail/compute_ops.cpp
```diff
@@ -23,12 +24,36 @@ namespace v1 {
 
 template <typename Policy, typename Float, typename Method, typename Task>
 struct compute_ops_dispatcher<Policy, Float, Method, Task> {
-    compute_result<Task> operator()(const Policy& policy,
+    compute_result<Task> operator()(const Policy& ctx,
```
ahuber21, line 27: Why are you renaming the `policy` to `ctx`? I see both names make sense, but for most algos we seem to be using `Policy& policy` (even though `Policy& ctx` is used frequently, too). Can't we just pick one and be consistent? Or is there a deeper meaning in the name?

cpp/oneapi/dal/algo/covariance/detail/compute_ops.cpp
```diff
@@ -23,12 +24,36 @@ namespace v1 {
...
+        const auto params = select_parameters(ctx, desc, input);
+        return implementation(ctx, desc, params, input);
+    }
+
+private:
+    inline auto implementation(const Policy& ctx,
+                               const descriptor_base<Task>& desc,
+                               const compute_parameters<Task>& params,
+                               const compute_input<Task>& input) const {
         using kernel_dispatcher_t = dal::backend::kernel_dispatcher< //
```
ahuber21, line 54: I see the same `//` to achieve a line break was already in the code ... hmm, is this really the best way to do it?

cpp/oneapi/dal/algo/covariance/detail/compute_ops.hpp
```diff
@@ -24,6 +24,13 @@ namespace v1 {
 
 template <typename Context, typename Float, typename Method, typename Task, typename... Options>
 struct compute_ops_dispatcher {
+    compute_result<Task> operator()(const Context&,
```
ahuber21, line 27: Now here it makes obviously sense to use `Context& ctx`. But now I'm confused that a `Policy&` can be a `ctx`, too.

cpp/oneapi/dal/algo/covariance/parameters/cpu/compute_parameters.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "oneapi/dal/algo/covariance/parameters/cpu/compute_parameters.hpp"
+
+namespace oneapi::dal::covariance::parameters {
+
+using dal::backend::context_cpu;
+
+template <typename Float>
+std::int64_t propose_block_size(const context_cpu& ctx, const std::int64_t row_count) {
+    std::int64_t block_size = 140l;
```
ahuber21, line 36: Again, I think excessive use of comments to explain the values doesn't hurt here. Even if it seems trivial to you

cpp/oneapi/dal/algo/covariance/parameters/cpu/compute_parameters.cpp
```diff
@@ -0,0 +1,64 @@
...
+    if (ctx.get_enabled_cpu_extensions() == dal::detail::cpu_extension::avx512) {
+        if (5000l < row_count && row_count <= 50000l) {
+            block_size = 1024l;
+        }
+    }
+    return block_size;
```
ahuber21, line 42: Return values should be tested in a unit test.

cpp/oneapi/dal/algo/covariance/parameters/cpu/compute_parameters.cpp
```diff
@@ -0,0 +1,64 @@
...
+    params_t operator()(const context_cpu& ctx,
+                        const detail::descriptor_base<Task>& desc,
+                        const compute_input<Task>& input) const {
+        const auto& x = input.get_data();
+
+        const auto row_count = x.get_row_count();
+
+        const auto block = propose_block_size<Float>(ctx, row_count);
+
+        return params_t{}.set_cpu_macro_block(block);
```
ahuber21, line 57: Return values should be tested in a unit test.

ahuber21 (commented):
Just a few points for discussion and a request for more tests. Very interesting work!

cpp/daal/include/algorithms/algorithm_base_mode.h
```diff
@@ -111,11 +111,11 @@ class Algorithm : public AlgorithmIfaceImpl
 
     AlgorithmContainerImpl<mode> * _ac;
 
+    const Hyperparameter * _hpar;
```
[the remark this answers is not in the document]
Vika-F: No. `Parameter` structure is what represents algorithm's parameters in DAAL. Performance hyperparameters are much different.

For example, `Hyperparameter`s cannot be set by a user in DAAL (though in oneDAL API it is Ok to set them). Another difference is that `Parameter`s can be the collections of various data types: simple types, numeric tables, etc. And Hyperparameters can be of only 2 types: int64_t and double.
There are other differences as well.

That's why a separate data structure is required for storing hyperparameters, and I cannot use already existing `Parameter` structure.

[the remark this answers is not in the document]
Vika-F: Sorry for misunderstanding.
Again no. As `Algorithm.compute()` can be called multiple times for the same instance of `Algorithm` with various parameters. So, `Parameter` should not be `const`.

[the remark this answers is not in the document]
ahuber21: Good point!

cpp/oneapi/dal/algo/covariance/detail/compute_ops.hpp
Vika-F: I understand your concerns. For now I am just aligning with the rest of the code base. But yes, we'll need to define more clear rules regarding 'ctx' and 'policy' naming - where to use what.

cpp/oneapi/dal/algo/covariance/compute_types.cpp
```diff
@@ -38,16 +40,39 @@ class detail::v1::compute_result_impl : public base {
...
+class partial_compute_result_impl : public base {
 public:
     table nobs;
     table crossproduct;
     table sums;
 };
 
+template <typename Task>
+struct compute_parameters_impl : public base {
+    std::int64_t cpu_macro_block = 140l;
```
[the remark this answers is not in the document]
Vika-F: Done.

cpp/oneapi/dal/algo/covariance/compute_types.cpp
```diff
@@ -38,16 +40,39 @@ class detail::v1::compute_result_impl : public base {
...
+};
+
+template <typename Task>
+compute_parameters<Task>::compute_parameters() : impl_(new compute_parameters_impl<Task>{}) {}
+
+template <typename Task>
+std::int64_t compute_parameters<Task>::get_cpu_macro_block() const {
```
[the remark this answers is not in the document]
Vika-F: Done.

cpp/oneapi/dal/algo/covariance/parameters/cpu/compute_parameters.cpp
Vika-F: Done.

cpp/oneapi/dal/algo/covariance/parameters/cpu/compute_parameters.cpp
Vika-F: There is no much sense in testing those values. Because they are hard-coded only for now. Further this logic will be replaced by some ML model that will compute those return values on the fly.
So, only the ranges of those values, but not the exact values, can be checked.
I will add the ranges test.

cpp/oneapi/dal/algo/covariance/backend/cpu/BUILD
```diff
@@ -0,0 +1,18 @@
+package(default_visibility = ["//visibility:public"])
+load("@onedal//dev/bazel:dal.bzl",
+    "dal_module",
+    "dal_test_suite",
+)
+
+dal_module(
```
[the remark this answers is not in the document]
Vika-F: I am not planning to add a test on this functionality. Because it is not exposed to users. And I do not see a reason to expose this functionality.

So, the potential test should be a black-box test. And for now there is no black box testing in oneDAL. This is not good, but this is the current state. We are not testing every internal kernel that we have, only the APIs that are exposed to the user.

cpp/oneapi/dal/algo/covariance/parameters/cpu/compute_parameters.cpp
ahuber21: I disagree, because unit tests also serve as a MWE of using a function, and as such they help the code to be self-documenting.
But you are right in the sense that the benefit is limited, so don't worry about it if it's high effort. Thank you for the explanation!

cpp/oneapi/dal/algo/covariance/detail/compute_ops.hpp
ahuber21: Yep, definitely something to keep in mind!

cpp/oneapi/dal/algo/covariance/backend/cpu/BUILD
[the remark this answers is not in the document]
ahuber21: We have to start somewhere :) I just added a test to an internal utility function which revealed that we were using it incorrectly before.

ahuber21 (approved):
Thanks for addressing my comments! I don't want to block progress here any longer.

Alexandr-Solovev (approved):

cpp/oneapi/dal/algo/covariance/parameters/gpu/compute_parameters_dpc.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+using dal::backend::context_gpu;
+
+template <typename Float, typename Task>
+struct compute_parameters_gpu<Float, method::dense, Task> {
+    using params_t = detail::compute_parameters<Task>;
+    params_t operator()(const context_gpu& ctx,
+                        const detail::descriptor_base<Task>& desc,
+                        const compute_input<Task>& input) const {
+        return params_t{};
```
KulikovNikita, line 39: Do we have blocking here?

cpp/oneapi/dal/algo/covariance/test/compute_parameters.cpp
```diff
@@ -0,0 +1,92 @@
...
+                     "[covariance][params]",
+                     covariance_types) {
+    SKIP_IF(this->not_float64_friendly());
+
+    const te::dataframe input =
+        GENERATE_DATAFRAME(te::dataframe_builder{ 1000, 20 }.fill_uniform(-30, 30, 7777),
+                           te::dataframe_builder{ 100, 10 }.fill_uniform(0, 1, 7777),
+                           te::dataframe_builder{ 100, 10 }.fill_uniform(-10, 10, 7777),
+                           te::dataframe_builder{ 500, 40 }.fill_uniform(-100, 100, 7777),
+                           te::dataframe_builder{ 500, 250 }.fill_uniform(0, 1, 7777));
```
KulikovNikita, line 83: I think you want to have some non-trivial dispatch results. Please remove really small tests and extend with tests designed for wide and/or long datasets

cpp/oneapi/dal/algo/covariance/parameters/gpu/compute_parameters_dpc.cpp
Vika-F: I think no, at least for now. I do not see any blocking here:
https://github.com/oneapi-src/oneDAL/blob/master/cpp/oneapi/dal/backend/primitives/stat/cov_dpc.cpp#L55 
and here:
https://github.com/oneapi-src/oneDAL/blob/master/cpp/oneapi/dal/algo/covariance/backend/gpu/compute_kernel_dense_impl_dpc.cpp#L120

cpp/oneapi/dal/algo/covariance/parameters/cpu/compute_parameters.cpp
```diff
@@ -0,0 +1,77 @@
...
+template <typename Float>
+std::int64_t propose_block_size(const context_cpu& ctx, const std::int64_t row_count) {
+    /// The constants are defined as the values that show the best performance results
+    /// in the series of performance measurements with the varying block sizes and dataset sizes.
+    std::int64_t block_size = 140l;
+    if (ctx.get_enabled_cpu_extensions() == dal::detail::cpu_extension::avx512) {
+        /// Here if AVX512 extensions are available on CPU
+        if (5000l < row_count && row_count <= 50000l) {
+            block_size = 1024l;
+        }
```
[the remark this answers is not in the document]
Vika-F: This logic is replicated from DAAL. I think performance testing needed here to be sure that the change gives some benefit on AVX2.

I think this is outside of the scope of this PR.

cpp/oneapi/dal/algo/covariance/test/compute_parameters.cpp
Vika-F: Done.

cpp/oneapi/dal/algo/covariance/backend/cpu/compute_kernel_dense.cpp
```diff
@@ -27,17 +27,34 @@ namespace oneapi::dal::covariance::backend {
...
+static daal_hyperparameters_t convert_parameters(const detail::compute_parameters<Task>& params) {
+    using daal_covariance::internal::HyperparameterId;
+
+    const std::int64_t block = params.get_cpu_macro_block();
+
+    daal_hyperparameters_t daal_hyperparameter;
+    auto status = daal_hyperparameter.set(HyperparameterId::denseUpdateStepBlockSize, block);
+    interop::status_to_exception(status);
+
+    return daal_hyperparameter;
```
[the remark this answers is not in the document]
Vika-F: I think `shared_ptr` is too heavy for this case.
It's better to just assign the result to const reference -> no passing by value and the lifetime is acceptable.
The full description of the trick is in [GotW #88](https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/)

PR #2534, opened 2023-09-29
doc: Add RELEASE_PLANS.md

ahuber21:
I propose to add a new file: `RELEASE_PLANS.md` in which we collect actionable items and other TODOs for upcoming releases. My other PR #2460 deprecates some model builder functions in favor of new versions with more parameters. This file will help us to keep track of such changes.

### DIFF
--- a/RELEASE_PLANS.md
+++ b/RELEASE_PLANS.md
@@ -1,0 +1,6 @@
+# Upcoming releases: Plans and required actions
+
+## 2025.0.0
+
+* **[AR]** Remove deprecated functions
+  * Marked via `\DAAL_DEPRECATED { will be removed in r2025.0.0 }`


KulikovNikita:
Looks interesting to me

RELEASE_PLANS.md
```diff
@@ -0,0 +1,6 @@
+# Upcoming releases: Plans and required actions
```
aepanchi, line 1:
```suggestion
# Upcoming Releases: Plans and Required Actions
```

aepanchi (approved):
small suggestion, the rest looks good

napetrov:
Talked offline but also would put this here.
The intent of this change is to have forcing function for cleaning up deprecated functionality - but i don't see how this implementation would help, other than adding another place with deprecation mention.

Also from  terminology perspective
deprecation - is fact of declaring deprecation that something will be removed
removal - is actually fact of removing functionality/api

So we have to declare deprecation in:
1. release notes
2. deprecated functionality section in documentation
3. potentially add deprecated mechanics in code

From removal perspective this have to be an activity planned for next major release to complete cleanup.

Also important considerations - deprecating functionality in code and comments should be done without implying versions. i.e. deprecated functionality is not enforcing timelines or targets.

ahuber21:
Thank you Nikolay, these points make sense to me. Closing this PR as we'll be taking a different route.

PR #2540, opened 2023-10-05
IPP defs cleanup

Alexsandruss:


KulikovNikita (approved):

ethanglaser:
What's the status for this PR?

Alexsandruss:
These symbols are part of FPK, therefore sorting functionality of FPK should be removed before this PR, but it's used by DAAL's sorting currently.

PR #2541, opened 2023-10-06
Fix online lom performance

Alexandr-Solovev:


inteldimitrius (approved):
LGTM

cpp/oneapi/dal/algo/basic_statistics/backend/cpu/finalize_compute_kernel_dense.cpp
```diff
@@ -50,11 +50,23 @@ static compute_result<Task> call_daal_kernel_finalize_compute(
...
+    std::int64_t column_count = 10;
 
+    const auto res_op = desc.get_result_options();
+
+    if (res_op.test(res_mean_varc)) {
+        column_count = input.get_partial_sum().get_column_count();
+    }
+    if (res_op.test(res_min_max)) {
+        column_count = input.get_partial_min().get_column_count();
+    }
```
Vika-F, line 67: Maybe I do not understand something. But this is a very suspicious change. What if the user does not compute any of {min, max, mean, variance}, and computes, for example, 'sum_squares' only. Then column_count will remain == 10. Is it Ok?

Vika-F: Also the use of a magic constant '10' is not a good practice anyway. Better change to zero + adding the assert after the actual column_count is defined in the code .

cpp/oneapi/dal/algo/basic_statistics/backend/gpu/partial_compute_kernel_dense_dpc.cpp
```diff
@@ -164,42 +167,52 @@ auto init_computation(sycl::queue& q,
...
-
+    const auto res_op = desc.get_result_options();
+    if (res_op.test(result_options::min)) {
+        auto reduce_event_min = pr::reduce_by_columns(q,
+                                                      data,
+                                                      result_min,
+                                                      pr::min<Float>{},
+                                                      pr::identity<Float>{},
+                                                      { nobs_update_event });
+        reduce_event_min.wait_and_throw();
```
KulikovNikita, line 178: Why don't we just use a standard kernel?

cpp/oneapi/dal/algo/basic_statistics/backend/gpu/partial_compute_kernel_dense_dpc.cpp
```diff
@@ -164,42 +167,52 @@ auto init_computation(sycl::queue& q,
...
+    }
+    if (res_op.test(result_options::sum) || res_op.test(result_options::sum_squares_centered) ||
+        res_op.test(res_mean_varc)) {
+        auto reduce_event_sums = pr::reduce_by_columns(q,
+                                                       data,
+                                                       result_sums,
+                                                       pr::sum<Float>{},
+                                                       pr::identity<Float>{},
+                                                       { nobs_update_event });
+        reduce_event_sums.wait_and_throw();
```
KulikovNikita, line 197: Do we really need to wait?

I believe you can do as following:
1) Create an empty `std::vector<sycl::event>`
2) Push events to this vector in each branch
3) Wait for this vector once

cpp/oneapi/dal/algo/basic_statistics/backend/cpu/finalize_compute_kernel_dense.cpp
Alexandr-Solovev: Good point! Will fix in next commit!

PR #2547, opened 2023-10-12
feature: dbscan optimizations

Alexandr-Solovev:
Dbscan optimization bazed on reducing loop of counting distances.
The intermediate check has been added for partially computed distances.


Alexandr-Solovev:
https://github.com/oneapi-src/oneDAL/pull/2547

cpp/oneapi/dal/algo/dbscan/backend/gpu/kernel_fp_impl.hpp
```diff
@@ -138,14 +153,19 @@ struct get_core_narrow_kernel {
                                     data_ptr[j * column_count + i];
                         sum += val * val;
                     }
-                    if constexpr (use_weights) {
-                        count += sum <= epsilon ? weights_ptr[j] : count_type(0);
+                    if (sum > epsilon) {
```
ethanglaser, line 156: would if(epsilon > sum) { break;} have the same effect? might be cleaner to have a breaking condition rather than a continue condition.

cpp/oneapi/dal/algo/dbscan/backend/gpu/kernel_fp_impl.hpp
```diff
@@ -77,22 +77,37 @@ struct get_core_wide_kernel {
...
                     for (std::int64_t j = 0; j < row_count; j++) {
                         Float sum = Float(0);
-                        for (std::int64_t i = local_id; i < column_count; i += local_size) {
+                        for (std::int64_t i = local_id; i < column_count / 2; i += local_size) {
+                            Float val = data_ptr[(block_start + wg_id) * column_count + i] -
+                                        data_ptr[j * column_count + i];
+                            sum += val * val;
+                        }
+                        Float distance_check =
+                            sycl::reduce_over_group(sg, sum, sycl::ext::oneapi::plus<Float>());
```
ethanglaser, line 86: how expensive is this distance check? would it make sense to add inside the for loop and break immediately once threshold is achieved instead of break for loop into 2?

Alexandr-Solovev: Its a good point. I did it mostly based on results on real data, and looks like its expensive to call this reduce each loop iteration. May be its better to make some mechanism to count number of splits based on column count

inteldimitrius: Additional question to this line: you are reducing over subgroup (passing sg as first arg), however computations are over group. Does it affect on accuracy or  everything is fine?

Alexandr-Solovev: All tests are passed, and DBI is the same as in master, but thanks for comment, I will doublecheck it!

cpp/oneapi/dal/algo/dbscan/backend/gpu/kernel_fp_impl.hpp
Alexandr-Solovev: Its a good point overall, but for narrow_kernel(when column count <= 4), we are using default for loop, and this continue means transition to next iteration, not breaking the loop.

inteldimitrius (approved):
LGTM

cpp/oneapi/dal/algo/dbscan/backend/gpu/kernel_fp_impl.hpp
```diff
@@ -60,40 +60,44 @@ struct get_core_wide_kernel {
         std::int32_t* cores_ptr = cores.get_mutable_data();
         auto event = queue.submit([&](sycl::handler& cgh) {
             cgh.depends_on(deps);
-            std::int64_t wg_size = get_recommended_sg_size(queue, column_count);
+            std::int64_t wg_size = get_recommended_wg_size(queue, column_count);
```
samir-nasibli, line 63: better to add `const`

cpp/oneapi/dal/algo/dbscan/backend/gpu/kernel_fp_impl.hpp
```diff
@@ -60,40 +60,44 @@ struct get_core_wide_kernel {
...
+                    const std::uint32_t local_id = item.get_local_id(0);
 
                     count_type count = 0;
                     for (std::int64_t j = 0; j < row_count; j++) {
                         Float sum = Float(0);
-                        for (std::int64_t i = local_id; i < column_count; i += local_size) {
-                            Float val = data_ptr[(block_start + wg_id) * column_count + i] -
+                        Float distance = Float(0);
+                        for (std::int64_t i = local_id; i < column_count; i += wg_size) {
+                            Float val = data_ptr[(block_start + row_id) * column_count + i] -
```
samir-nasibli, line 78: be carefully with different types of int multiplication

cpp/oneapi/dal/algo/dbscan/backend/gpu/kernel_fp_impl.hpp
```diff
@@ -60,40 +60,44 @@ struct get_core_wide_kernel {
...
+                            count += use_weights ? weights_ptr[j] : count_type(1);
+                            if (count >= min_observations) {
+                                if (local_id == 0) {
+                                    cores_ptr[row_id] = count_type(1);
+                                }
+                                break;
+                            }
+                            if (row_count - j + count < min_observations) {
+                                break;
+                            }
```
avolkov-intel, line 98: I think we can apply this break only if use_weights is set to False. If it is set to true then potentially an instance could be a core sample if the sum of w_j,...,w_count plus count is greater than min_observation. Considering that weights could be more than 1 using this if statement we may miss some core camples.

cpp/oneapi/dal/algo/dbscan/backend/gpu/kernel_fp_impl.hpp
```diff
@@ -138,14 +142,19 @@ struct get_core_narrow_kernel {
...
+                    count += use_weights ? weights_ptr[j] : count_type(1);
+
+                    if (count >= min_observations) {
+                        cores_ptr[idx] = count_type(1);
+                        break;
                     }
-                    else {
-                        count += sum <= epsilon ? count_type(1) : count_type(0);
+                    if (row_count - j + count < min_observations) {
+                        break;
```
avolkov-intel, line 155: The same comment as above, we should add !use_weights in this if statement

cpp/oneapi/dal/algo/dbscan/backend/gpu/kernel_fp_impl.hpp
```diff
@@ -77,23 +86,38 @@ struct get_core_wide_kernel {
...
+                        if (distance <= epsilon) {
+                            count += use_weights ? weights_ptr[j] : count_type(1);
+                            if (count >= min_observations) {
+                                if (local_id == 0) {
+                                    cores_ptr[wg_id] = count_type(1);
+                                }
+                                break;
+                            }
+                            if (!use_weights && (row_count - j + count < min_observations)) {
+                                break;
```
ethanglaser, line 117: what does this break out of?

PR #2550, opened 2023-10-16
feature: online pca algorithm

Alexandr-Solovev:


cpp/oneapi/dal/algo/covariance/backend/cpu/finalize_compute_kernel_dense.cpp
```diff
@@ -57,7 +57,20 @@ static compute_result<Task> call_daal_kernel_finalize(const context_cpu& ctx,
     const auto daal_nobs_matrix = interop::convert_to_daal_table<Float>(input.get_partial_n_rows());
 
     auto result = compute_result<Task>{}.set_result_options(desc.get_result_options());
-
+    auto rows_count_global =
+        row_accessor<const Float>(input.get_partial_n_rows()).pull({ 0, -1 })[0];
+    daal_covariance::internal::Hyperparameter daal_hyperparameter;
+    /// the logic of block size calculation is copied from DAAL,
+    /// to be changed to passing the values from the performance model
+    std::int64_t blockSize = 140;
```
ethanglaser, line 65: why 140? and would it be 140 when row_count > 50000?

cpp/oneapi/dal/algo/pca/detail/finalize_train_ops.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+    void check_preconditions(const Descriptor& params, const input_t& input) const {
+        ONEDAL_ASSERT(input.get_partial_n_rows().has_data());
+        ONEDAL_ASSERT(input.get_partial_n_rows().get_column_count() == 1);
+        ONEDAL_ASSERT(input.get_partial_n_rows().get_row_count() == 1);
+        ONEDAL_ASSERT(input.get_partial_crossproduct().has_data());
+        ONEDAL_ASSERT(input.get_partial_sum().has_data());
+    }
+
+    void check_postconditions(const Descriptor& params,
```
ethanglaser, line 49: is anything done here?

cpp/oneapi/dal/algo/pca/detail/finalize_train_ops_dpc.cpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+};
+
+#define INSTANTIATE(F, M, T)      \
+    template struct ONEDAL_EXPORT \
+        finalize_train_ops_dispatcher<dal::detail::data_parallel_policy, F, M, T>;
+
+INSTANTIATE(float, method::cov, task::dim_reduction)
+INSTANTIATE(double, method::cov, task::dim_reduction)
+INSTANTIATE(float, method::svd, task::dim_reduction)
```
ethanglaser, line 43: what happens here if svd not implemented for gpu? should it be removed?

cpp/oneapi/dal/algo/pca/detail/partial_train_ops_dpc.cpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+};
+
+#define INSTANTIATE(F, M, T)      \
+    template struct ONEDAL_EXPORT \
+        partial_train_ops_dispatcher<dal::detail::data_parallel_policy, F, M, T>;
+
+INSTANTIATE(float, method::cov, task::dim_reduction)
+INSTANTIATE(double, method::cov, task::dim_reduction)
+INSTANTIATE(float, method::svd, task::dim_reduction)
```
ethanglaser, line 43: same comment on svd gpu

cpp/oneapi/dal/algo/covariance/backend/cpu/finalize_compute_kernel_dense.cpp
Alexandr-Solovev: I've taken the `blockSize` from batch kernel. May be @Vika-F can clarify this value?

cpp/oneapi/dal/algo/pca/detail/finalize_train_ops.hpp
Alexandr-Solovev: Good point, will fix!

examples/oneapi/cpp/source/pca/pca_dense_online.cpp
```diff
@@ -0,0 +1,78 @@
...
+*******************************************************************************/
+
+#include "oneapi/dal/algo/pca.hpp"
+#include "oneapi/dal/io/csv.hpp"
+
+#include "example_util/utils.hpp"
+
+namespace dal = oneapi::dal;
+
+/*******************************************************************************
```
ethanglaser, line 24: double copyrights/includes

cpp/oneapi/dal/algo/pca/detail/finalize_train_ops_dpc.cpp
Alexandr-Solovev: We have exception in placeholder for gpu svd online. I did it similarly as in batch implementation.

cpp/oneapi/dal/algo/pca/test/fixture.hpp
```diff
@@ -141,10 +196,42 @@ class pca_test : public te::crtp_algo_fixture<TestType, Derived> {
...
+        check_nans(result);
+
+        INFO("check if eigenvectors order is descending")
+        this->check_eigenvalues_order(eigenvalues);
+
+        INFO("check if eigenvectors matrix is orthogonal")
+        check_eigenvectors_orthogonality(eigenvectors);
+    }
+
     void check_infer_result(const pca::descriptor<Float, Method>& desc,
```
ethanglaser, line 214: why is this function empty? shouldnt it be validating the results? or does that happen elsewhere

cpp/oneapi/dal/algo/pca/test/online.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace te = dal::test::engine;
+namespace la = te::linalg;
+namespace pca = oneapi::dal::pca;
+using pca_types_cov = COMBINE_TYPES((float, double), (pca::method::cov));
+using pca_types_svd = COMBINE_TYPES((float, double), (pca::method::svd));
+
+template <typename TestType>
+class pca_online_test : public pca_test<TestType, pca_online_test<TestType>> {};
+
```
ethanglaser, line 29: will any tests be ran on devices that are not float64 friendly? currently our CI validates on hardware that does not support float64, so all of these will be skipped

Alexandr-Solovev: Fixed

cpp/oneapi/dal/algo/pca/test/fixture.hpp
Alexandr-Solovev: We dont have accuracy check for results in batch. Will be added in separate pr

examples/oneapi/cpp/source/pca/pca_dense_online.cpp
```diff
@@ -0,0 +1,78 @@
...
+template <typename Method>
+void run(const dal::table& x_train, const std::string& method_name) {
+    const auto pca_desc =
+        dal::pca::descriptor<float, Method>().set_component_count(5).set_deterministic(true);
+    const std::int64_t nBlocks = 10;
+
+    dal::pca::partial_train_result<> partial_result;
+    std::cout << method_name << "\n" << std::endl;
+    auto input_table = split_table_by_rows<double>(x_train, nBlocks);
+
```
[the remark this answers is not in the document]
Alexandr-Solovev: Good point, will be added in separate pr

cpp/oneapi/dal/algo/pca/test/fixture.hpp
ethanglaser: would it ever fail?

cpp/oneapi/dal/algo/pca/test/online.cpp
Alexandr-Solovev: And yes, it will be skipped if device doesnt support double, and the test type is double, so, it will be launched only with float

ethanglaser: I didn't necessarily mean to remove all checks (looks like its failing now) but instead just ensuring we are having some tests run on devices that aren't float64 friendly.

cpp/oneapi/dal/algo/covariance/backend/cpu/finalize_compute_kernel_dense.cpp
Vika-F: This logic was copied from DAAL. The constants were defined in DAAL as the result of a series of performance measurements performed in the past. In DAAL, batch and online implementations share large amount of code, including those constants which are actually performance related hyperparameters of the algorithm.  
In the future this logic will be moved into `hyperparameter` classes, this was already implemented in batch oneDAL algorithm, and will be replaced further by more intelligent solution (not just hard-coded constants).

cpp/oneapi/dal/algo/pca/detail/partial_train_ops_dpc.cpp
Alexandr-Solovev: We have exception in placeholder for gpu svd online. I did it similarly as in batch implementation.

cpp/oneapi/dal/algo/pca/backend/gpu/finalize_train_kernel_cov_dpc.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+using bk::context_gpu;
+using model_t = model<task::dim_reduction>;
+using task_t = task::dim_reduction;
+using input_t = partial_train_result<task_t>;
+using result_t = train_result<task_t>;
+using descriptor_t = detail::descriptor_base<task::dim_reduction>;
+
+template <typename Float>
+auto compute_eigenvectors_on_host(sycl::queue& q,
```
Vika-F, line 40: It is possible to compute eigenvectors on device in three steps using oneMKL LAPACK functions:
1. Reduce symmetric correlation matrix to tridiagonal form $A = Q \cdot T \cdot Q^T$ with [?sytrd](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-fortran/2023-2/sytrd.html) or ?syrdb
2. Compute eigenvectors $u$ of a tridiagonal matrix T using [?stemr](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-fortran/2023-2/stemr.html)
3. Compute eigenvectors $v$ of the original matrix $A$ from the eigenvectors of the matrix $T$ as $v = Q \cdot u$

Eigenvalues of $A$ and $T$ are the same.

cpp/oneapi/dal/algo/pca/backend/gpu/partial_train_kernel_cov_dpc.cpp
```diff
@@ -0,0 +1,229 @@
...
+    ONEDAL_PROFILER_TASK(compute_sums, q);
+    ONEDAL_ASSERT(data.has_data());
+
+    const std::int64_t column_count = data.get_dimension(1);
+    auto sums = pr::ndarray<Float, 1>::empty(q, { column_count }, sycl::usm::alloc::device);
+    auto reduce_event =
+        pr::reduce_by_columns(q, data, sums, pr::sum<Float>{}, pr::identity<Float>{}, deps);
+
+    return std::make_tuple(sums, reduce_event);
+}
```
Vika-F, line 57: This code is duplicated at least in batch covariance algorithm, maybe somewhere else as well:
https://github.com/oneapi-src/oneDAL/blob/master/cpp/oneapi/dal/algo/covariance/backend/gpu/compute_kernel_dense_impl_dpc.cpp#L45
Please consider moving this into some common place and share the implementation between Covariance and PCA_Cov algorithms.

cpp/oneapi/dal/algo/pca/backend/gpu/partial_train_kernel_cov_dpc.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+template <typename Float>
+auto compute_crossproduct(sycl::queue& q,
```
Vika-F, line 60: This code is also duplicated here and in Covariance. Please consider removing the duplication.

examples/oneapi/cpp/source/pca/pca_dense_online.cpp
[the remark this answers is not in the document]
Vika-F: I also think that it would be better not to split the data table which was read from a single file into 10 blocks. More naturally would be to have several  (maybe not 10, but rather 3) input files that contain the input blocks of data. And compute the result from those files.

Vika-F (commented):
The API looks good to me, but I have a couple of comments to the internal implementation.

* In CPU part of the batch PCA-covariance algorithms the DAAL covariance algorithm is just passed to PCA kernel, but in the online PCA you decided to call Covariance explicitly from oneDAL. Why it is implemented like this? Maybe passing the Covariance as an object to PCA can simplify the implementation.
* I think it is not a question to online algorithm, but to PCA implementation in general. Currently PCA + covariance implementation differs in oneDAL and sklearn. oneDAL uses correlation matrix to compute PCA, and sklearn uses variance-covariance matrix. The results of those two approaches are very different in case the input dataset is not normalized. I see that this implementation also use correlation matrix to compute PCA and this choice is hard-coded. I think it is better to have an option to choose which kind of matrix: variance-covariance or correlation to use in PCA.
*  The code related to performance hyperparameters is duplicated 3+ times in this PR. I think it is important to have a follow up task of adding `hyperparameter` classes into online Covariance algorithm in oneDAL.

Please see the other comments also in the PR.

cpp/oneapi/dal/algo/pca/backend/gpu/partial_train_kernel_cov_dpc.cpp
Alexandr-Solovev: Looks like removing duplication in this pr requires a lot of changes in batch implementations also. Is it ok for you, if I do it in a separate pr?

Alexandr-Solovev:
@Vika-F Thanks for review!
About difference in CPU part with calling PCA and Cov kernels. So, DAAL implementation of PcaOnline kernel contains additional variable. Its shared pointer to DAAL object OnlineCovariance. It contains all methods and all variables from OnlineCovariance class. And it works when we call finalize and partial from the same place, but with oneDAL online design we had two ways:
1)Contain DAAL OnlineCovariance as part of input of finalize
2)Avoid it and using daal Cov kernel, because in general online PcaCov also calls Cov kernel.

cpp/oneapi/dal/algo/covariance/backend/cpu/finalize_compute_kernel_dense.cpp
ethanglaser: @Vika-F thank you for clarification, seems reasonable for now. Based on this logic block_size would be 140 when number of rows is > 50k, is that intended as well? Seems a bit odd to set to 1024 for larger data but then back to 140 for very large data.

examples/oneapi/cpp/source/pca/pca_dense_online.cpp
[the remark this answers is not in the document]
Vika-F: Yes, this comment is related to examples only.
In Bazel tests it is more convenient to have the data generated on the fly.

cpp/oneapi/dal/algo/covariance/backend/cpu/finalize_compute_kernel_dense.cpp
Vika-F: It might be that in case of large number of rows the particular size of the block is not that sufficient - as there is enough work available for each thread. But this is only my guess. The original data from the performance experiments based on which those constants were defined is already unavailable.
And that's why we are moving towards more intelligent solution than hard-coded constants.

Vika-F (approved):
LGTM.
But let's have the comments regarding code duplications addressed in a separate PR the sooner the better.

PR #2562, opened 2023-10-27
Fix 'no type named 'uint32_t'' failure in ICX build

Vika-F:
Removed unnecessary #include <cstdint> from "algorithm_types.h" file.

By design we are trying not to include additional standard or third-party files into DAAL user-facing headers.
This is done so as not to clutter up end user's code with additional declarations.

Following changes were made in the code of `algorithms::Hyperparameter` and the derived classes to get rid of data types defined in <cstdint>:
- `std::unit32_t` type was changed to `unsigned int` in interface part of the code and to `uint32_t` in kernel part of the code
- `std::int64_t` types was changed to `DAAL_INT64`

### DIFF
--- a/cpp/daal/src/algorithms/covariance/covariance_hyperparameter.cpp
+++ b/cpp/daal/src/algorithms/covariance/covariance_hyperparameter.cpp
@@ -21,6 +21,7 @@
 //--
 */
 
+#include <stdint.h>
 #include "src/algorithms/covariance/covariance_hyperparameter_impl.h"
 
 namespace daal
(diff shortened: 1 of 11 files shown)


Vika-F:
Previous CI was green instead of ClangFromat step that fails due to the files I hadn't touched in this PR.

Link to the internal CI:
[http://intel-ci.intel.com/ee797994-62a3-f193-8987-a4bf010d0e2e](http://intel-ci.intel.com/ee797994-62a3-f193-8987-a4bf010d0e2e)

Alexandr-Solovev (approved):
LGTM

cpp/daal/src/algorithms/covariance/covariance_hyperparameter.cpp
```diff
@@ -21,6 +21,7 @@
 //--
 */
 
+#include <stdint.h>
```
KulikovNikita, line 24: It contradicts usage of `DAAL_INT64`

KulikovNikita (approved):

cpp/daal/src/algorithms/covariance/covariance_hyperparameter.cpp
Vika-F: It is there for uint32_t - a type of the keys in hash tables. I do not see the contradiction.

PR #2563, opened 2023-10-27
Update TPP file according to the recent changes

maria-Petrova:
# Description
The PR reflects the most recent changes in the oneDAL 3rd parties.

### DIFF
--- a/third-party-programs.txt
+++ b/third-party-programs.txt
@@ -10,7 +10,7 @@ party programs as set forth in the “third-party-programs.txt” file.
 
 The third party programs and their corresponding required notices and/or license terms are listed below.
 
--------------------------------------------------------------
+=====================================================================================
 
 
 1.  Intel® Math Kernel Library (Intel® MKL)
@@ -82,7 +82,7 @@ each party agrees that it submits to the personal jurisdiction and venue of thos
 waives any objections. The United Nations Convention on Contracts for the International
 Sale of Goods (1980) is specifically excluded and will not apply to the Software.
 
--------------------------------------------------------------
+=====================================================================================
 
 2.  Sphinx
 Copyright (c) 2007-2019 by the Sphinx team (see AUTHORS file). All rights reserved.
@@ -123,7 +123,7 @@ Licenses for incorporated software
 The included smartypants module, included as sphinx.util.smartypants,
 is available under the following license:
 
-----------------------------------------------------------------------
+------------------------------------------------------------------------
 SmartyPants_ license::
 
     Copyright (c) 2003 John Gruber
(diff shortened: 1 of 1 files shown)


third-party-programs.txt
```diff
@@ -312,9 +312,201 @@ INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT
 STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
 WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
--------------------------------------------------------------
+=====================================================================================
 
 4. CuML
+   Copyright 2018 NVIDIA CORPORATION
+   
+   xgboost
```
ahuber21, line 320: I'm confused. Is `xgboost` to be understood as part of `CuML`? Or should it be

```
5. xgboost
```

In this case, what's the `CuML` license? Latest branch mentions Apache 2.0
https://github.com/rapidsai/cuml/blob/branch-23.12/LICENSE

third-party-programs.txt
```diff
@@ -312,9 +312,201 @@ INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
 
    Copyright 2018 NVIDIA CORPORATION
```
ahuber21, line 511: This line now appears twice (also in l. 318). Mistake?

ahuber21 (changes requested):
I may be mistaken, but to me it looks like the `xgboost` part is not added in the correct place.

third-party-programs.txt
maria-Petrova: That file comes from legal. Point #4 only groups the components that obtain the same license in order to save space in the file. In the case of XGBoost and cuML it's Apache 2.0.

third-party-programs.txt
maria-Petrova: Line 318 just refers to the general copyright of cuML, while l.511 is just a part of the cuML license, so that should be fine.

ahuber21 (approved):
Thanks for the clarification

napetrov (approved):

PR #2566, opened 2023-10-31
feature: Pca interfaces change

Alexandr-Solovev:
Aligning PcaSVD results with scikit-learn results


samir-nasibli:
Please provide some description for this iface changes.
It would be fine also update examples and tests in this case in one PR

PR #2572, opened 2023-11-08
[bug] fix numerical instability in right-most node in ensemble algorithms

icfaust:
# Description
DAAL random forest and extra trees implementations use the depthFirstSearch function to help build the trees. It generates right nodes of the trees via left nodes, parents nodes and math tricks of the cost functions. This causes float type error accumulation in the value of the right-most deepest node. When a right node has a single element, these tricks aren't necessary.  By directly assigning the value of the single element, it removes the numerical error and allows for (in some cases) perfect recall of the data (which matches sklearn).  This is the cause of some test failures moving code out of preview in scikit-learn-intelex.

- Add simpleSwap function for classification, and a check for a single node element


Alexsandruss:
@mergifyio rebase

icfaust:
@Mergifyio rebase

Vika-F (commented):
The changes are perfectly described in the PR's description section. Please add the similar description into the code.
Because the logic behind the added 'if' statement is not obvious from the code.

ahuber21 (changes requested):
Agree with Vica, please add a comment in the code that explains the rationale

ahuber21 (approved):

PR #2577, opened 2023-11-14
MPI GPU interface refactoring

ethanglaser:
# Description
- Add virtual get_mpi_offload_support function to base communicator - defaults to false in nearly all cases
- Add logic to get_mpi_offload_support function in mpi/communicator.h to check mpi libs for correct symbol and determine if level zero is supported
- Add conditional in detail/communicator.cpp that uses result of get_mpi_offload_support to determine whether to convert data to host (previous default) or leave as is (yields performance improvements if GPU offload support in MPI)
- Modify sendrecv_replace args to include optional additional buffer to accommodate MPICH workaround to call sendrecv with 2 GPU buffers


cpp/oneapi/dal/detail/communicator.cpp
```diff
@@ -84,54 +75,15 @@ spmd::request_iface* spmd_communicator_via_host_impl::allgatherv(
...
-    ONEDAL_ASSERT(recv_counts_host);
-
-    for (std::int64_t i = 0; i < rank_count; i++) {
-        const std::int64_t src_offset = check_mul_overflow(dtype_size, displs_host_root_ptr[i]);
-        const std::int64_t dst_offset = check_mul_overflow(dtype_size, displs_host[i]);
-        const std::int64_t copy_size = check_mul_overflow(dtype_size, recv_counts_host[i]);
-        if (copy_size > 0) {
-            memcpy_host2usm(q, recv_buf + dst_offset, recv_buf_host_ptr + src_offset, copy_size);
-        }
+        ONEDAL_PROFILER_TASK(comm.allgatherv_gpu, q);
```
Alexandr-Solovev, line 80: for allgatherv maybe it's necessary to add check_mul_overglow(num_ranks*one_rank_memory_size)

cpp/oneapi/dal/test/engine/thread_communicator.hpp
```diff
@@ -358,11 +415,11 @@ class thread_communicator_impl
...
               ctx_(thread_count),
               barrier_(ctx_),
-              bcast_(ctx_),
+              bcast_(ctx_, queue),
               gather_(ctx_),
-              allgatherv_(ctx_),
-              sendrecv_replace_(ctx_),
-              allreduce_(ctx_),
+              allgatherv_(ctx_, queue),
+              sendrecv_replace_(ctx_, queue),
```
Alexandr-Solovev, line 421: as a potential approach, can we have 2 comm function definitions:
sendrecv_replace(ctx)
sendrecv_replace(ctx, queue)
for dispatching between usm and host memory

Alexandr-Solovev (commented):
Looks as a great opportunity to get more speedup across all algorithms

ethanglaser:
> Looks as a great opportunity to get more speedup across all algorithms

Thanks! Yeah its pretty ugly right now, working towards functional first then will clean things up. But good points.

cpp/oneapi/dal/detail/communicator.cpp
ethanglaser: actually looks like recv_counts_host is already allreduced counts - this function is dependent on a previous allreduce?

ethanglaser: should be resolved now - some unnecessary logic before conditional in case of real mpi but generally should work

ethanglaser:
Nightly combined with infra branch: http://intel-ci.intel.com/eedcda75-51a4-f11e-8bab-a4bf010d0e2e

cpp/oneapi/dal/detail/communicator.cpp
```diff
@@ -52,15 +53,22 @@ spmd::request_iface* spmd_communicator_via_host_impl::bcast(sycl::queue& q,
     const std::int64_t dtype_size = get_data_type_size(dtype);
     const std::int64_t size = check_mul_overflow(dtype_size, count);
 
-    const auto send_buff_host = array<byte_t>::empty(size);
-    if (get_rank() == root) {
-        memcpy_usm2host(q, send_buff_host.get_mutable_data(), send_buf, size);
+    if (get_mpi_gpu_support()) {
```
Alexandr-Solovev, line 56: does the adding of `const bool gpu_offloading = get_mpi_gpu_support() ` and after that reuse this variable instead of calling the function make sense?

cpp/oneapi/dal/detail/ccl/communicator.hpp
```diff
@@ -325,6 +325,14 @@ class ccl_communicator_impl : public ccl_interface_selector<MemoryAccessKind>::t
         return default_root_;
     }
 
+    bool get_mpi_gpu_support() override {
```
Alexandr-Solovev, line 328: may be it makes sense to rename it and add smth related to offloading, due to its little bit confusing

cpp/oneapi/dal/detail/communicator.cpp
```diff
@@ -102,35 +110,47 @@ spmd::request_iface* spmd_communicator_via_host_impl::allgatherv(
     const std::int64_t dtype_size = get_data_type_size(dtype);
     const std::int64_t send_size = check_mul_overflow(dtype_size, send_count);
     const std::int64_t total_recv_size = check_mul_overflow(dtype_size, total_recv_count);
-    // Workaround for zero send_size
-    const auto send_buff_host = array<byte_t>::empty(send_size > 0 ? send_size : 1);
-    if (send_size > 0) {
-        memcpy_usm2host(q, send_buff_host.get_mutable_data(), send_buf, send_size);
+
+    if (get_mpi_gpu_support()) {
```
Alexandr-Solovev, line 114: the same comment about usage the variable instead of calling function

cpp/oneapi/dal/detail/mpi/communicator.hpp
```diff
@@ -146,6 +149,53 @@ class mpi_communicator_impl : public via_host_interface_selector<MemoryAccessKin
         return default_root_;
     }
 
+    bool get_mpi_gpu_support() override {
+        // TODO: explore additional conditions (i.e. GPU type, data parallel, etc.)
+        // TODO: make this function bool convert() and instead determine if conversion necessary?
+
+        // Check libmpi.so for symbol
+        void* handle = dlopen("libmpi.so", RTLD_LAZY);
```
Alexandr-Solovev, line 157: may be it makes sense to ask MPI team about the adding public method for getting this info through a mpi function? Due to its little bit hardcoded and could be not flexible if MPI change smth

ethanglaser: This was their recommended approach, although I agree that it may be worth seeing if they can add something as easy to use as with CCL

Alexandr-Solovev: @Vika-F Hi! Can you please review this part of the code?

Alexandr-Solovev: Hi @irozanova, thanks for the answer! In terms of backward compatibility, I guess that we can add versioning and for previous versions don't use GPU offloading. The question is, is it possible to somehow get the information about "MPIX_Query_ze_support" via some of MPI function?

irozanova: > The question is, is it possible to somehow get the information about "MPIX_Query_ze_support" via some of MPI function?

No, there are no any functions, which provide information about other functions.

ethanglaser: This approach is also compatible with MPICH and other MPI backends, as they have the same symbol to indicate if offloading is supported

Vika-F: Taking into account the fact that `MPIX_Query_ze_support` is only supported in Intel MPI and only starting from 2021.12, I think it would be better to leave the code as is.

irozanova: MPIX_Query_ze_support is supported also by MPICH implementation
https://www.mpich.org/static/docs/latest/www3/MPIX_Query_ze_support.html

ethanglaser:
Final steps are to check MPICH scalability with alternative approach and determine whether its necessary to add any additional conditions to use offloading

ethanglaser:
Job with infra branch: http://intel-ci.intel.com/ef1f7d20-65f3-f16f-89f4-a4bf010d0e2e

ethanglaser:
Job with infra branch: http://intel-ci.intel.com/ef2396dd-0148-f19e-b1e2-a4bf010d0e2e

ethanglaser:
> Job with infra branch: http://intel-ci.intel.com/ef2396dd-0148-f19e-b1e2-a4bf010d0e2e

Updated job: http://intel-ci.intel.com/ef2db06a-7b63-f187-9c4d-a4bf010d0e2e

cpp/oneapi/dal/detail/mpi/communicator.hpp
```diff
@@ -282,15 +336,34 @@ class mpi_communicator_impl : public via_host_interface_selector<MemoryAccessKin
...
+            mpi_call(MPI_Sendrecv_replace(buf,
+                                          integral_cast<int>(count),
+                                          make_mpi_data_type(dtype),
+                                          integral_cast<int>(destination_rank),
+                                          zero_tag,
+                                          integral_cast<int>(source_rank),
+                                          zero_tag,
+                                          mpi_comm_,
+                                          &status));
+        }
```
ethanglaser, line 366:
```suggestion
        /*
        if (recv_buf) {
            // MPICH-specific workaround for GPU performance
            mpi_call(MPI_Sendrecv(buf,
                                  integral_cast<int>(count),
                                  make_mpi_data_type(dtype),
                                  integral_cast<int>(destination_rank),
                                  zero_tag,
                                  recv_buf,
                                  integral_cast<int>(count),
                                  make_mpi_data_type(dtype),
                                  integral_cast<int>(source_rank),
                                  zero_tag,
                                  mpi_comm_,
                                  &status));
        }
        else {
        */
        // Standard call to sendrecv_replace of designated mpi backend
        mpi_call(MPI_Sendrecv_replace(buf,
                                      integral_cast<int>(count),
                                      make_mpi_data_type(dtype),
                                      integral_cast<int>(destination_rank),
                                      zero_tag,
                                      integral_cast<int>(source_rank),
                                      zero_tag,
                                      mpi_comm_,
                                      &status));
        }
```

cpp/oneapi/dal/detail/mpi/communicator.hpp
```diff
@@ -291,6 +365,7 @@ class mpi_communicator_impl : public via_host_interface_selector<MemoryAccessKin
                                       zero_tag,
                                       mpi_comm_,
                                       &status));
+        }
```
ethanglaser, line 368:
```suggestion
        //}
```

PR #2580, opened 2023-11-16
Fix: Add -flink-huge-device-code to debug build to fix linking issue

ahuber21:
This fixes the PC-relative offset error when linking `libonedal_dpc.so`.

My test build [here](http://intel-ci.intel.com/ee83f896-9c7c-f1a5-ab82-a4bf010d0e2e) passed build for linux.
On Windows there was an error: `warning LNK4019: corrupt string table (table end); new end assumed`. This was observed sporadically in [chromium debug builds](https://bugs.chromium.org/p/chromium/issues/detail?id=743069) and might vanish in future builds. Note that is a different error than previously before adding the linking flag. So the situation was at least improved. I am continuing to investigate the Windows debug build.

~~I suggest to merge this definite fix for linux and potential fix for Windows.~~

Investigating the Windows error & `failed to convert GOTPCREL relocation against '_ZTVN6oneapi3fpk4ngen24invalid_object_exceptionE'; relink with --no-relax` on one of the LNX builds


ahuber21:
Additional debug build for Win here: http://intel-ci.intel.com/ee8456dd-1b94-f18d-aff0-a4bf010d0e2e

ahuber21:
I am working on adding the flag also for statically linked examples. But this work can be continued independently of this PR. It seems there is an actual bug in the examples right now, which will prevent the CI from going green.
Would suggest to merge this now.

examples/oneapi/dpc/CMakeLists.txt
```diff
@@ -16,7 +16,6 @@
 
 cmake_minimum_required(VERSION 3.1)
 
-set(CMAKE_BUILD_TYPE Release)
```
ahuber21, line 19: Should be provided via command line `-DCMAKE_BUILD_TYPE={Release,Debug}`

Alexandr-Solovev: Is it really necessary to remove it, because, for me, looks that with `set(CMAKE_BUILD_TYPE Release)` it will be set by default, and from command line you can -DCMAKE_BUILD_TYPE=Debug

ahuber21: For me, this line overrode what I specified with `-DCMAKE_BUILD_TYPE`. That's why I removed it. With this line, the link flag was not added correctly.

ahuber21: What do you think about

```
if(NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE Release CACHE STRING "Build type not specified, defaulting to Release" FORCE)
endif()
```

makefile
```diff
@@ -830,6 +830,7 @@ $(WORKDIR.lib)/$(oneapi_y.dpc): \
     $(ONEAPI.tmpdir_y.dpc)/$(oneapi_y.dpc:%.$y=%_link.txt) ; $(DPC.LINK.DYNAMIC) ; $(LINK.DYNAMIC.POST)
 $(WORKDIR.lib)/$(oneapi_y.dpc): LOPT += $(-fPIC)
 $(WORKDIR.lib)/$(oneapi_y.dpc): LOPT += $(daaldep.rt.dpc)
+$(WORKDIR.lib)/$(oneapi_y.dpc): LOPT += $(if $(REQDBG),-flink-huge-device-code,)
```
napetrov, line 833: wow, interesting and thanks for finding it

ahuber21: Side note: It used to be `-fsycl-link-huge-device-code`, but this flag is now deprecated and superseded by `-flink-huge-device-code`.

examples/oneapi/dpc/CMakeLists.txt
Alexandr-Solovev: I think that its worse to remove default value, as a basic example:
if(NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build (Debug or Release)")
endif()

napetrov: but it works for same syntax for DPCPP/oneDAL vars below? 

I think that you have problems with cashing your variables and clean runs wouldn't be impacted

https://cmake.org/cmake/help/book/mastering-cmake/chapter/CMake%20Cache.html

ahuber21: Don't think caching can be a factor here, as I deleted the build directory between every run. I can demonstrate

```
# WITH set(CMAKE_BUILD_TYPE Release) in place
$ rm -rf Build/ && mkdir Build
$ CC=icx CXX=icpx cmake -B./Build -S. -G "Unix Makefiles" -DONEDAL_LINK=static -DCMAKE_BUILD_TYPE=Debug
# ...
$ grep -r flink-huge-device-code Build/ | wc -l
0

# WITHOUT set(CMAKE_BUILD_TYPE Release) (original version from this PR)
$ rm -rf Build/ && mkdir Build
$ CC=icx CXX=icpx cmake -B./Build -S. -G "Unix Makefiles" -DONEDAL_LINK=static -DCMAKE_BUILD_TYPE=Debug
# ...
$ grep -r flink-huge-device-code Build/ | wc -l                                                                                                                                                                                                                                                                                                                                                                                                                     
30 

# WITH the if() statement (current version of this PR)
$ rm -rf Build/ && mkdir Build
$ CC=icx CXX=icpx cmake -B./Build -S. -G "Unix Makefiles" -DONEDAL_LINK=static -DCMAKE_BUILD_TYPE=Debug
# ...
$ grep -r flink-huge-device-code Build/ | wc -l                                                                                                                                                                                                                                                                                                                                                                                                                     
30 

# For completeness, without -DCMAKE_BUILD_TYPE to illustrate defaul behavior
$ rm -rf Build/ && mkdir Build
$ CC=icx CXX=icpx cmake -B./Build -S. -G "Unix Makefiles" -DONEDAL_LINK=static 
# ...
$ grep -r flink-huge-device-code Build/ | wc -l                                                                                                                                                                                                                                                                                                                                                                                                                     
0 
```

napetrov: I think it then worth moving this to https://github.com/oneapi-src/oneDAL/blob/master/cmake/templates/oneDALConfig.cmake.in

So we would handle this param passing for all cases, not only for DPC examples.

napetrov (approved):
Changes looks good - let's validate that CIs are working but hope everything would be ok

ahuber21:
It appears on Windows we bump into exactly this issue: Build mode defaults to `Debug` which is causing some issues downstream. I still think we should remove the hardcoded `Release` from the examples

```
C:/.../cmake-3.16/Modules/CMakeInitializeConfigs.cmake(21):  if(NOT CMAKE_NO_BUILD_TYPE )
C:/.../cmake-3.16/Modules/CMakeInitializeConfigs.cmake(22):  set(CMAKE_BUILD_TYPE Debug CACHE STRING Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel ... )
C:/.../cmake-3.16/Modules/CMakeInitializeConfigs.cmake(25):  list(APPEND _CONFIGS Debug )
```

PR #2583, opened 2023-11-17
feature: adding blocking in reduction primitives

Alexandr-Solovev:
# Description
Adding new method for reduction primitive with support blocking for extremely large datasets.
It helps to avoid the out of integer limit error on big datasets, due to 

```
    const std::int64_t max_loop = std::numeric_limits<std::int32_t>::max();
    const auto range = width * height;
    if (range >= max_loop) {
        return reduction_method::blocking;
```
In case that our reduction range more than numeric limit, we will call the blocking reduction method, which works fine on any dimensions until device bad_alloc.


ethanglaser:
Can you add to description what the result of these changes is?

Alexandr-Solovev:
@ethanglaser Yes, sure. So, a parallel_for loop has a restriction that multiplications of range/ranges should be less that the numeric_limit<int_32t>, otherwise you will get the out of integer limit error, and it often exists on big datasets. With this new method for reduction, if we have      
```
    const std::int64_t max_loop = std::numeric_limits<std::int32_t>::max();
    const auto range = width * height;
    if (range >= max_loop) {
        return reduction_method::blocking;
```
   range more than numeric limit, we will call the blocking reduction method, which works fine on any dimensions until device bad_alloc. In terms of accuracy it returns exactly the same numbers as other methods. In terms of performance it works little bit slower on small cases, its expected.

inteldimitrius (approved):
Looks good to me, let's wait CI checks

cpp/oneapi/dal/backend/primitives/reduction/reduction_rm_cw_blocking_dpc.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+#include "oneapi/dal/backend/primitives/reduction/common.hpp"
+#include "oneapi/dal/backend/primitives/reduction/reduction_rm_cw_dpc.hpp"
+
+namespace oneapi::dal::backend::primitives {
+namespace bk = dal::backend;
+template <typename Float>
+std::int64_t propose_block_size(const sycl::queue& q, const std::int64_t r) {
+    constexpr std::int64_t fsize = sizeof(Float);
+    return 0x10000l * (8 / fsize);
```
ethanglaser, line 25: How was this value chosen?

cpp/oneapi/dal/backend/primitives/reduction/reduction_rm_cw_blocking_dpc.cpp
```diff
@@ -0,0 +1,125 @@
...
+            cgh.parallel_for(bk::make_multiple_nd_range_2d({ curr_block, wg_ }, { 1, wg_ }),
+                             [=](sycl::nd_item<2> it) {
+                                 // Common for whole WG
+                                 const auto col_idx = it.get_global_id(0) + first_column;
+                                 const auto loc_idx = it.get_local_id(1);
+                                 const auto range = it.get_global_range(1);
+                                 // It should be converted to upper type by default
+                                 Float acc = (override_init || (loc_idx != 0)) ? //
+                                                 binary.init_value
+                                                                               : output[col_idx];
```
ethanglaser, line 71: I guess clang passes but this formatting is crazy

ethanglaser (approved):

PR #2584, opened 2023-11-17
Add --no-banner flag to covc

ahuber21:
Currently the bullseye banner is printed for every single source file, resulting in unnecessarily large log.


homksei:
It might also be advisable not to invoke `cov01 -1` in the `COMPILER.lnx.icc` line every time. According to the [documentation](https://www.bullseye.com/help/ref-cov01.html), calling it just once before build should be enough.

napetrov:
@ahuber21  @homksei  - will we integrate this version or would add more changes?

napetrov (approved):

ahuber21:
Removing `cov01 -1` seems more difficult to me, as we're just overloading the compiler command. I would keep it like this and merge. @napetrov feel free to merge if you agree.

PR #2587, opened 2023-11-19
Fix return type of get_class_count_method

avolkov-intel:
# Description
The return type was incorrect, we need to fix it from double to std::int64_t


avolkov-intel:
@aepanchi @maria-Petrova we would need to merge it into current release

Alexandr-Solovev (approved):

avolkov-intel:
@Mergifyio copy rls/2024.0.1-rls

PR #2592, opened 2023-11-21
[bug] fix MT2203 RNG non-uniformity and random bin indices in decision forest training

icfaust:
# Description
Each MT2203 RNG engine is independently uniform when taking samples. However, when two or more engines are compared, the initial aggregated random numbers are not uniform.  Because the randomness between trees needs to be guaranteed for the decision forest algorithm (each with its own RNG engine), a imperceptible performance loss is introduced to burn RNG values to where the engine collection is empirically uniform. 

The second issue is a problem with the binary search associated with finding a split for ExtraTrees (regressor and classifier). The search failed to find the largest bin left edge in its current orientation, and so it has been switched to always guarantee a valid split. This change comes from the ambiguity of using a binning approach with the Extra Trees algorithm definition.  All use of the .min parameter are removed, and so it is completely removed from IndexedFeatures and initial binning scripts.

This will fix the following deselected_tests from sklearnex:
 tests/test_multioutput.py::test_classifier_chain_tuple_order
 ensemble/tests/test_forest.py::test_distribution

However, this is changing the determinism of the trees used in the sklearnex tests, which means some tests which passed by chance could now fail. 

This non-uniformity negatively impacts both the random forest in the bootstrapping process, and in extra trees in the initial chosen splits.

- Check for a family engine (only MT2203)
- Burn a magic number of samples for every engine (400)
- remove .min() from IndexedFeatures
- change binary search in genRandomBinIdx for classification and regression

### DIFF
--- a/cpp/daal/include/algorithms/pca/pca_explained_variance_batch.h
+++ b/cpp/daal/include/algorithms/pca/pca_explained_variance_batch.h
@@ -109,6 +109,8 @@ public:
         input.set(eigenvalues, other.input.get(eigenvalues));
     }
 
+    ~Batch() = default;
+
     /**
      * Returns the method of the algorithm
      * \return Method of the algorithm
--- a/cpp/daal/include/algorithms/pca/pca_types.h
+++ b/cpp/daal/include/algorithms/pca/pca_types.h
@@ -200,6 +200,7 @@ class DAAL_EXPORT Input : public InputIface
 public:
     Input();
     Input(const Input & other);
+    Input & operator=(const Input & other);
 
     virtual ~Input() {};
 
@@ -556,6 +557,7 @@ class DistributedInput<correlationDense> : public InputIface
 public:
     DAAL_EXPORT DistributedInput();
     DAAL_EXPORT DistributedInput(const DistributedInput & other);
+    DAAL_EXPORT DistributedInput operator=(const DistributedInput & other);
 
     /**
         * Sets input objects for the PCA on the second step in the distributed processing mode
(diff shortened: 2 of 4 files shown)


icfaust:
special private CI run with mentioned tests enabled: http://intel-ci.intel.com/ee886edb-4adb-f114-a7aa-a4bf010d0e2e

icfaust:
Performance comparisons are available upon request

icfaust:
The test_distribution for ExtraTreesRegressor is not fixed. This requires further analysis.

icfaust:
special private CI run with mentioned tests enabled: http://intel-ci.intel.com/ee891bc7-4b31-f12d-a430-a4bf010d0e2e

icfaust:
Private CI shows this causes an issue with ExtraTreesClassification now, will need to investigate

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -438,6 +439,16 @@ services::Status computeImpl(HostAppIface * pHostApp, const NumericTable * x, co
...
         numElems[i]                    = 0;
         auto engineImpl                = dynamic_cast<engines::internal::BatchBaseImpl *>(engines[i].get());
+        if (technique == engines::internal::family)
+        {
+            // The MT2203 engine is the only family engine in oneDAL.
+            // The family of MT2203 engines have correlated random
+            // numbers in the first 400. Thus, 400 samples are burned
+            // to guarantee uniformity.
+            RNGsInst<unsigned int, cpu> rng;
+            services::internal::TArray<unsigned int, cpu> temp(burn);
```
KulikovNikita, line 449:
```suggestion
            RNGsInst<uint32_t, cpu> rng;
            services::internal::TArray<uint32_t, cpu> temp(burn);
```

KulikovNikita (approved):
LGTM

PR #2607, opened 2023-12-01
chore(deps): update GPU FPK

homksei:


napetrov:
i think we would need cpu fpk lib to be updated as well

homksei:
@Mergifyio backport rls/2024.0.1-rls

PR #2614, opened 2023-12-15
Enable ARM(SVE) CPU support with reference backend

ajay-fuji:
## Proposed changes in this PR
- Enabled support for build on ARM CPUs (SVE).
- Arm build supports DAL Interface (daal) and oneAPI C++ interface (oneapi_c).
- ci/env/tbb.sh helps to build tbb for ARM (SVE) cpu and installs it in the right folder structure for use during compile time.
- Make command for arm(sve):
`make -f makefile daal oneapi_c BACKEND_CONFIG=ref PLAT=lnxarm -j$(nproc)`

**Open Issues/Limitations:**
- oneAPI DPC++ interface (oneapi) is not supported since Intel's DPC++ compiler is not supported on arm.
- Bazel build is yet not supported.


makefile
```diff
@@ -514,10 +553,14 @@ $(CORE.objs_a): COPT += -D__TBB_NO_IMPLICIT_LINKAGE -DDAAL_NOTHROW_EXCEPTIONS \
                         $(if $(CHECK_DLL_SIG),-DDAAL_CHECK_DLL_SIG)
 $(CORE.objs_a): COPT += @$(CORE.tmpdir_a)/inc_a_folders.txt
 $(filter %threading.$o, $(CORE.objs_a)): COPT += -D__DO_TBB_LAYER__
+
+ifneq ($(ARCH_is_arm),yes)
```
keeranroth, line 557: Avoid the negation here, and do something like:

```
ifeq ($(ARCH_is_x86_64),yes)
```

cmake/scripts/generate_config.cmake
```diff
@@ -54,4 +55,14 @@ file(MAKE_DIRECTORY ${config_install_dir})
 configure_file(${CMAKE_CURRENT_LIST_DIR}/../templates/oneDALConfig.cmake.in ${config_install_dir}/oneDALConfig.cmake @ONLY)
 configure_file(${CMAKE_CURRENT_LIST_DIR}/../templates/oneDALConfigVersion.cmake.in ${config_install_dir}/oneDALConfigVersion.cmake @ONLY)
 
+set(FILE_PATH ${config_install_dir}/oneDALConfig.cmake)
+# Read the content of the input file
+file(READ "${FILE_PATH}" INPUT_CONTENT)
+
+# Perform string substitution
+string(REPLACE "lib/intel64" "lib/${SUB_DIR}" MODIFIED_CONTENT "${INPUT_CONTENT}")
```
keeranroth, line 63: It feels like the string substitution should not be necessary, and the right subdirectory should be written into the string, from the oneDALConfig.cmake already

cpp/daal/include/algorithms/algorithm_container_base_batch.h
```diff
@@ -143,10 +144,15 @@ class AlgorithmContainerImpl<batch> : public AlgorithmContainer<batch>
  * \tparam avx2Container        Implementation for Intel(R) Advanced Vector Extensions 2 (Intel(R) AVX2)
  * \tparam avx512Container      Implementation for Intel(R) Xeon(R) processors based on Intel AVX-512
  */
+#ifdef __ARM_ARCH
```
keeranroth, line 147: Start with the x86 check, and have an error condition for unknown architectures. This is a general comment, where a check is done for the (micro-)architecture

cpp/daal/include/services/env_detect.h
```diff
@@ -91,7 +97,12 @@ class DAAL_EXPORT Environment : public Base
     enum CpuTypeEnable
     {
         cpu_default = 0, /*!< Default processor type */
-        avx512      = 2  /*!< Intel(R) Xeon(R) processors based on Intel(R) Advanced Vector Extensions 512 (Intel(R) AVX-512) \DAAL_DEPRECATED */
+        
+        #ifdef __ARM_ARCH
+            sve         = 2, /*!< ARM processors based on Arm's Scalable Vector Extension (SVE) */
```
keeranroth, line 102: Should this have a value of 2? As this is the only CPU arch for Arm, this should be 1?

cpp/daal/include/services/internal/daal_kernel_defines.h
```diff
@@ -35,6 +36,7 @@
 #define DAAL_KERNEL_SSE42
 #define DAAL_KERNEL_AVX2
 #define DAAL_KERNEL_AVX512
+#define DAAL_KERNEL_SVE
```
keeranroth, line 39: The defines should be protected by an ifdef. No need for the x86 defines to exist in an Arm build, and vice versa

cpp/daal/src/algorithms/algorithm_hyperparameter.cpp
```diff
@@ -65,11 +66,17 @@ struct HyperparameterImpl : public HyperparameterBaseImpl
     }
 
 protected:
+    
+    #ifdef __ARM_ARCH
+        #define CPU_ARG sve
```
keeranroth, line 71: This is a default CPU, or minimum CPU? The name should be more specific than just CPU_ARG. @napetrov any suggestions for a good name?

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -135,8 +136,14 @@ static inline size_t getBlockSize(size_t nrows)
...
+#define CPU_TYPE sve
+#else
+#define CPU_TYPE avx512
+#endif
+
 template <>
-inline size_t getBlockSize<avx512>(size_t nrows)
+inline size_t getBlockSize<CPU_TYPE>(size_t nrows)
 {
     return (nrows > 5000 && nrows <= 50000) ? 1024 : 140;
```
keeranroth, line 148: Is this a good block size for SVE as well? There is already the default block size in the implementation above this. Maybe `#ifdef` this and include it only if AVX512 can be present, e.g. `#ifdef DAAL_KERNEL_AVX512`

cpp/daal/src/algorithms/kernel_config.h
```diff
@@ -24,6 +25,9 @@
 #ifndef __KERNEL_CONFIG_H__
 #define __KERNEL_CONFIG_H__
 
-#include "src/algorithms/kernel_inst_x86.h"
-
+#ifdef __ARM_ARCH
+    #include "src/algorithms/kernel_inst_arm.h"
```
keeranroth, line 29: I would prefer new files for an arch to go into a sub-folder. Otherwise, if more architectures are supported, the directory will get cluttered.

cpp/daal/src/algorithms/kernel_inst_arm.h
```diff
@@ -0,0 +1,131 @@
...
+* limitations under the License.
+*******************************************************************************/
+
+/*
+//++
+//  Defines used for kernel allocation, deallocation and calling kernel methods
+//--
+*/
+
+#ifndef __KERNEL_INST_X86_H__
```
keeranroth, line 24: `__KERNEL_INST_ARM_H__`

cpp/daal/src/services/compiler/generic/env_detect_features.cpp
```diff
@@ -41,6 +46,28 @@
 void __daal_serv_CPUHasAVX512f_enable_it_mac();
 #endif
 
+#if defined(__ARM_ARCH)
```
keeranroth, line 49: Split the file up into different arch implementations, and put these in appropriate sub-directories

cpp/daal/src/services/service_topo.h
```diff
@@ -60,6 +61,9 @@ typedef cpuset_t cpu_set_t;
         #ifdef __x86_64__
             #define LNX_PTR2INT unsigned long long
             #define LNX_MY1CON  1LL
+        #elif __aarch64__
+using LNX_PTR2INT                = unsigned long long;
```
keeranroth, line 65: Keep the existing style, and use the defines here as well

cpp/oneapi/dal/algo/subgraph_isomorphism/backend/cpu/compiler_adapt.hpp
```diff
@@ -16,7 +17,11 @@
 
 #pragma once
 #include <cstdint>
-#include <immintrin.h>
+
+#ifndef __ARM_ARCH
```
keeranroth, line 21: Something like `#ifdef TARGET_X86_64` - it shouldn't be included if Arm isn't defined. It should be included if x86 is defined.

cpp/oneapi/dal/algo/triangle_counting/backend/cpu/intersection_tc.hpp
```diff
@@ -68,6 +71,34 @@ ONEDAL_FORCEINLINE std::int32_t _popcnt32_redef(const std::int32_t& x) {
     {}
 #endif
 
+#ifdef __ARM_ARCH
```
keeranroth, line 74: This implementation looks the same as the default one above, so the specialisation isn't required at the moment.

cpp/oneapi/dal/backend/primitives/intersection/intersection.hpp
```diff
@@ -62,6 +65,30 @@ ONEDAL_FORCEINLINE std::int32_t _popcnt32_redef(const std::int32_t &x) {
     {}
 #endif
 
+#ifdef __ARM_ARCH
```
keeranroth, line 68: The Arm implementation looks like the default implementation above, so this specialization isn't required at the moment.

cpp/oneapi/dal/detail/policy.hpp
```diff
@@ -61,10 +62,14 @@ inline constexpr bool is_data_parallel_policy_v = is_data_parallel_policy<T>::va
 
 enum class cpu_extension : uint64_t {
     none = 0U,
+    #ifdef __ARM_ARCH
+    sve = 1U << 1,
```
keeranroth, line 66: As this is the only architecture, should this be a shift by zero?

makefile
```diff
@@ -672,9 +742,13 @@ $1: $(1:%.$a=%_link.txt) | $(dir $1)/. ; $(value LINK.STATIC)
 endef
 
 $(ONEAPI.dispatcher_cpu): | $(dir $(ONEAPI.dispatcher_cpu))/.
+ifeq ($(ARCH_is_arm),yes)
+	$(if $(filter sve,$(USECPUS)),echo "#define ONEDAL_CPU_DISPATCH_SSE42" >> $@)
```
keeranroth, line 746: Should be ONEDAL_CPU_DISPATCH_SVE rather than ONEDAL_CPU_DISPATCH_SSE42

azure-pipelines.yml
```diff
@@ -0,0 +1,19 @@
+# Starter pipeline
```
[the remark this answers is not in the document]
ajay-fuji: Yes you are right, this file should not be part of this. Will be deleted.

rakshithgb-fujitsu:
> For the most part, the structure looks good, and a change for building on a generic (non vendor-specific) CPU looks very similar to what is here. One of the most common comments I have (I didn't put all of them in), is that it's probably good to have x86 branches first, as this is the most likely code path at the moment.
> 
> There are a few things to discuss still, which can be quite big in themselves:
> 
> 1. Do we target building with Bazel for new architectures? I don't have an opinion yet, as I'm not familiar with Bazel
> 2. How do we structure the code for new architectures? At the moment, there are multiple `#ifdef` in the files, where if this is done for each (micro-)architecture, the files can become cluttered. I would advocate for having target-specific sub-directories, and put target specific code in those, keeping as much common code in the files they are currently in.

@keeranroth we din't really change any structure and stuck to what x86 follows so we just added arm compatibility on top of it. My thoughts on further points:

1. @napetrov prefers the new architectures to support bazel, however x86 itself isn't fully ready yet so we chose to support the make file approach. We are now actively working towards supporting bazel on arm. Bazel is new for us as well and might take time to support it.
2. This would require a massive code re-architecture since only external APIs are isolated (BLAS/LAPACK etc.), ideally we would need to replicate similar methods for other parts of the algorithms as well. I agree with you on this to isolate the micro-architecture components.

cpp/oneapi/dal/detail/policy.hpp
rakshithgb-fujitsu: since NEON would be supported in the future, SVE was assigned 2 on the list.

dev/make/cmplr.clang.mkl.32e.mk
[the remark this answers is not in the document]
rakshithgb-fujitsu: I would recommend to support make file for now and start bazel support immediately after that. @napetrov thoughts on this?

cpp/daal/include/services/env_detect.h
rakshithgb-fujitsu: assumption was neon would be 1 in the future.

.ci/scripts/build.sh
```diff
@@ -47,7 +48,7 @@ PLATFORM=$(bash dev/make/identify_os.sh)
 OS=${PLATFORM::3}
 ARCH=${PLATFORM:3:3}
 
-optimizations=${optimizations:-avx2}
+optimizations=$(if [ "${ARCH}" == "arm" ]; then echo "${optimizations:-sve}"; else echo "${optimizations:-avx2}"; fi)
```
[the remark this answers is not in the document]
napetrov: yes, would like expanded variant more

cmake/scripts/generate_config.cmake
napetrov: yes, it seems it's better to setup oneDALConfig properly and not do updates on the fly

cpp/daal/include/services/env_detect.h
```diff
@@ -42,11 +43,16 @@ namespace daal
  */
 enum CpuType
 {
+#ifdef __ARM_ARCH
+    sve         = 0, /*!< ARM processors based on Arm's Scalable Vector Extension (SVE) */
```
napetrov, line 47: are there plans to bring NEON instructions? Per my understanding sve is not the oldest one - if there plans to bring older ISA then it worth set SVE higher number

cpp/daal/src/algorithms/kernel_config.h
napetrov: yes, and also we have to have rules that can guide ownership rules. So we can't use flat files schema

amgrigoriev:
The approach above involves too many mk files like CMPLR.BACKEND.ARCH.mk. Having "n" backends and "m" archs, it will give [nxm] matrix. For every new arch it will be necessary to add all backends and vise versa and repeat it for all supported compilers. My suggestion is to separate backend and arch definitions to have a single file at least for every backend. For instance, defs.ref.mk for "ref" backend could contain **COMPILER.defs += -DDAAL_REF -DONEDAL_REF -DONEAPI_DAAL_NO_MKL_GPU_FUNC** while cmplr.clang.x86.mk could use COMPILER.defs above (both are included in main makefile): ### COMPILER.mac.clang = clang++ -m64 -fgnu-runtime -stdlib=libc++ -mmacosx-version-min=10.15 -fwrapv \
                     -Werror -Wreturn-type $(COMPILER.defs)

COMPILER.lnx.clang = clang++ -m64 \
                     -Werror -Wreturn-type \
		     $(COMPILER.defs)

cpp/daal/include/services/env_detect.h
aepanchi, line 47:
```suggestion
    sve         = 0, /*!< ARM(R) processors based on Arm(R) Scalable Vector Extension (SVE) */
```

cpp/daal/include/services/env_detect.h
```diff
@@ -91,7 +97,12 @@ class DAAL_EXPORT Environment : public Base
     enum CpuTypeEnable
     {
         cpu_default = 0, /*!< Default processor type */
-        avx512      = 2  /*!< Intel(R) Xeon(R) processors based on Intel(R) Advanced Vector Extensions 512 (Intel(R) AVX-512) \DAAL_DEPRECATED */
+
+#ifdef __ARM_ARCH
+        sve = 2, /*!< ARM processors based on Arm's Scalable Vector Extension (SVE) */
```
aepanchi, line 102:
```suggestion
        sve = 2, /*!< ARM(R) processors based on Arm(R) Scalable Vector Extension (SVE) */
```

cpp/daal/src/algorithms/kernel_inst_arm.h
```diff
@@ -0,0 +1,131 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+/*
+//++
+//  Defines used for kernel allocation, deallocation and calling kernel methods
```
aepanchi, line 20:
```suggestion
//  The defines used for kernel allocation, deallocation, and calling kernel methods
```

dev/make/cmplr.clang.ref.arm.mk
```diff
@@ -0,0 +1,45 @@
+# file: cmplr.clang.ref.arm.mk
+#===============================================================================
+# Copyright 2023-24 FUJITSU LIMITED
```
aepanchi, line 3: @napetrov do we need the Intel copyright here as well?

cpp/oneapi/dal/detail/policy.hpp
Vika-F: @keeranroth is correct. SVE can have a zero shift and NEON - a one shift. No need to do gaps between the shifts. Newer ARM architectures will have bigger shifts.
In previous versions of oneDAL we had more architectures supported: sse2 - 0, ssse3 - 1, sse42 - 2, ...
But recently some of them were removed to reduce the build time and binary size. That's why there are those gaps between the shifts now.

dev/download_tbb.sh
```diff
@@ -18,65 +19,96 @@
...
+  pushd __deps/tbb
+      mkdir -p lnx
+      mv lib/ lnx/
+      mv include/ lnx/ 
+      pushd lnx
+          mkdir -p lib/arm/gcc4.8
+          mv lib/libtbb* lib/arm/gcc4.8
+      popd
+  popd 
+fi
```
Vika-F, line 114: Please add a new line here.

dev/make/cmplr.clang.ref.arm.mk
```diff
@@ -0,0 +1,45 @@
...
+pedantic.opts.clang = -pedantic \
+                      -Wall \
+                      -Wextra \
+                      -Wno-unused-parameter
+
+pedantic.opts.mac.clang = $(pedantic.opts.clang)
+pedantic.opts.lnx.clang = $(pedantic.opts.clang)
+
+# For SVE
+a8sve_OPT.clang = $(-Q)march=armv8-a+sve
```
Vika-F, line 45: Please add a new line here.

Vika-F (commented):
Thank you for adding ARM support to oneDAL! This is a tremendous work.
I have reviewed the code in ./cpp folder and it looks good and mostly correct. Please see the additional comments below.

But my biggest concern is that the Azure pipelines are not updated to test those changes.
.ci/pipelines/ci.yml file needs to be updated to support a new job for ARM.
Can you please update the pipelines so we can see if the changes are working as expected?

azure-pipelines.yml
```diff
@@ -0,0 +1,19 @@
...
+  vmImage: ubuntu-latest
+
+steps:
+- script: echo Hello, world!
+  displayName: 'Run a one-line script'
+
+- script: |
+    echo Add other tasks to build, test, and deploy your project.
+    echo See https://aka.ms/yaml
+  displayName: 'Run a multi-line script'
```
KulikovNikita, line 19: Looks just like a placeholder from Azure tutorials =) Probably should be updated later

cpp/daal/include/algorithms/algorithm_container_base_batch.h
```diff
@@ -1,6 +1,7 @@
 /* file: algorithm_container_base_batch.h */
 /*******************************************************************************
 * Copyright 2014 Intel Corporation
+* Copyright 2023-24 FUJITSU LIMITED
```
KulikovNikita, line 4:
```suggestion
* Copyright 2024 FUJITSU LIMITED
```

I believe we have an internal agreement that year in the file header reflect the year of initial file creation or modification (in your case)

cpp/daal/include/algorithms/algorithm_container_base_batch.h
KulikovNikita: What about other macros? I believe I've seen at least `__arch64__`, `_M_ARM64` somewhere in ARM specific code (in other repos)

cpp/daal/include/algorithms/algorithm_container_base_batch.h
```diff
@@ -143,10 +144,15 @@ class AlgorithmContainerImpl<batch> : public AlgorithmContainer<batch>
...
  */
+#ifdef __ARM_ARCH
+template <typename SVEContainer DAAL_KERNEL_SVE_ONLY(typename sveContainer)>
+class DAAL_EXPORT AlgorithmDispatchContainer<batch, SVEContainer DAAL_KERNEL_SVE_ONLY(sveContainer)> : public AlgorithmContainerImpl<batch>
+#else
 template <typename sse2Container DAAL_KERNEL_SSE42_ONLY(typename sse42Container) DAAL_KERNEL_AVX2_ONLY(typename avx2Container)
               DAAL_KERNEL_AVX512_ONLY(typename avx512Container)>
 class DAAL_EXPORT AlgorithmDispatchContainer<batch, sse2Container DAAL_KERNEL_SSE42_ONLY(sse42Container) DAAL_KERNEL_AVX2_ONLY(avx2Container)
                                                         DAAL_KERNEL_AVX512_ONLY(avx512Container)> : public AlgorithmContainerImpl<batch>
+#endif
```
KulikovNikita, line 155:
```suggestion
#if defined(__ARM_ARCH) || defined(__aarch64__)
template <typename SVEContainer DAAL_KERNEL_SVE_ONLY(typename sveContainer)>
class DAAL_EXPORT AlgorithmDispatchContainer<batch, SVEContainer DAAL_KERNEL_SVE_ONLY(sveContainer)> : public AlgorithmContainerImpl<batch>
#elif defined(_WIN64) || defined(__x86_64__)
template <typename sse2Container DAAL_KERNEL_SSE42_ONLY(typename sse42Container) DAAL_KERNEL_AVX2_ONLY(typename avx2Container)
              DAAL_KERNEL_AVX512_ONLY(typename avx512Container)>
class DAAL_EXPORT AlgorithmDispatchContainer<batch, sse2Container DAAL_KERNEL_SSE42_ONLY(sse42Container) DAAL_KERNEL_AVX2_ONLY(avx2Container)
                                                        DAAL_KERNEL_AVX512_ONLY(avx512Container)> : public AlgorithmContainerImpl<batch>
#else
    #error "Compiling for the unknown architecture"
#endif
```

cpp/daal/include/algorithms/algorithm_container_base_batch.h
```diff
@@ -143,10 +144,15 @@ class AlgorithmContainerImpl<batch> : public AlgorithmContainer<batch>
  * \tparam avx2Container        Implementation for Intel(R) Advanced Vector Extensions 2 (Intel(R) AVX2)
  * \tparam avx512Container      Implementation for Intel(R) Xeon(R) processors based on Intel AVX-512
```
KulikovNikita, line 145: Please update documentation as well.

cpp/daal/include/algorithms/algorithm_container_base_common.h
```diff
@@ -1,6 +1,7 @@
 /* file: algorithm_container_base_common.h */
 /*******************************************************************************
 * Copyright 2014 Intel Corporation
+* Copyright 2023-24 FUJITSU LIMITED
```
KulikovNikita, line 4:
```suggestion
* Copyright 2024 FUJITSU LIMITED
```

cpp/daal/include/services/daal_defines.h
```diff
@@ -65,6 +66,8 @@
 #if !defined(DAAL_INT)
     #if defined(_WIN64) || defined(__x86_64__)
         #define DAAL_INT __int64
+    #elif defined(__aarch64__)
```
KulikovNikita, line 69: I see inonsistency in macro usage. Like `defined(__aarch64__)` vs `defined(__ARM_ARCH)`

cpp/daal/src/services/compiler/generic/env_detect_features.cpp
```diff
@@ -1,6 +1,7 @@
 /* file: env_detect_features.cpp */
```
KulikovNikita, line 1: I would actually suggest to split this particular file into `env_detct_x86_features.cpp` and ``env_detect_arm_features.cpp` in order to not create the `#ifdef` hell

cpp/daal/src/services/service_topo.h
KulikovNikita: I believe we can just use C's native `uintptr_t` instead

cpp/daal/src/threading/export_lnxarm.ref.def
```diff
@@ -0,0 +1,64 @@
+;===============================================================================
+; Copyright 2023 Intel Corporation
```
KulikovNikita, line 2:
```suggestion
```

KulikovNikita: Since it is an ARM specific file don't think that it woth keeping Intel in the header

cpp/oneapi/dal/algo/subgraph_isomorphism/backend/cpu/compiler_adapt.hpp
```diff
@@ -83,6 +88,47 @@ ONEDAL_FORCEINLINE std::int32_t ONEDAL_popcnt64(std::uint64_t a) {
...
+ONEDAL_FORCEINLINE std::int32_t ONEDAL_lzcnt_u32<dal::backend::cpu_dispatch_sve>(std::uint32_t a) {
+    if (a == 0)
+        return 32;
+    std::uint32_t one_bit = 0x80000000; // binary: 1000 0000 0000 0000 0000 0000 0000 0000
+    std::int32_t bit_pos = 0;
+    while ((a & one_bit) == 0) {
+        bit_pos++;
+        one_bit >>= 1;
+    }
+    return bit_pos;
```
KulikovNikita, line 102: I would consider using [compiler specific builtins](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) wherever it is possible. In case of GCC/Clang it would be [`__builtin_clz`](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005fclz)

cpp/oneapi/dal/algo/subgraph_isomorphism/backend/cpu/compiler_adapt.hpp
```diff
@@ -83,6 +88,47 @@ ONEDAL_FORCEINLINE std::int32_t ONEDAL_popcnt64(std::uint64_t a) {
...
+}
+
+template <>
+ONEDAL_FORCEINLINE std::int32_t ONEDAL_popcnt64<dal::backend::cpu_dispatch_sve>(std::uint64_t a) {
```
KulikovNikita, line 118: Please consider using [`__builtin_lclz`](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005fclzl)

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -135,8 +136,14 @@ static inline size_t getBlockSize(size_t nrows)
     return 140;
 }
 
+#ifdef __ARM_ARCH
+    #define CPU_TYPE sve
+#else
+    #define CPU_TYPE avx512
+#endif
+
```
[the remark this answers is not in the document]
chandan-fujitsu-research: Thank you for suggestion

.ci/scripts/build.sh
```diff
@@ -1,6 +1,7 @@
 #! /bin/bash
 #===============================================================================
 # Copyright 2019 Intel Corporation
+# Copyright 2023-24 FUJITSU LIMITED
```
napetrov, line 4: Per discussion with UXL - we can start adding "Copyright Contributors to the oneDAL project", while keeping existing Intel copyright that already exist. 

There should be some checks fixes done internally to reflect this + we will be starting using new copyright as well

cpp/oneapi/dal/algo/covariance/backend/cpu/partial_compute_kernel_dense.cpp
```diff
@@ -53,7 +60,7 @@ static partial_compute_result<Task> call_daal_kernel_partial_compute(
     /// the logic of block size calculation is copied from DAAL,
     /// to be changed to passing the values from the performance model
     std::int64_t blockSize = 140;
-    if (ctx.get_enabled_cpu_extensions() == dal::detail::cpu_extension::avx512) {
+    if (ctx.get_enabled_cpu_extensions() == CPU_EXTENSION) {
```
napetrov, line 63: not sure if this is right way to handle this - the idea of this block is that this is avx512 specialized code here  and as well as in some other algorithms. It assume avx512 isa and i'm not sire that it would work that easily on ARM. We probably can have a better approach defined for architecture/ISA specific implementation in algorithms.

deploy/local/vars_lnx.sh
```diff
@@ -239,8 +249,8 @@ if [ "$(basename "${my_script_path}")" = "env" ] ; then   # assume stand-alone
       export LD_LIBRARY_PATH="$__daal_tmp_dir/lib${LD_LIBRARY_PATH+:${LD_LIBRARY_PATH}}"
     else
       export CPATH="$__daal_tmp_dir/include${CPATH+:${CPATH}}"
-      export LIBRARY_PATH="$__daal_tmp_dir/lib/intel64${LIBRARY_PATH+:${LIBRARY_PATH}}"
-      export LD_LIBRARY_PATH="$__daal_tmp_dir/lib/intel64${LD_LIBRARY_PATH+:${LD_LIBRARY_PATH}}"
+      export LIBRARY_PATH="$__daal_tmp_dir/lib/$SUB_DIR${LIBRARY_PATH+:${LIBRARY_PATH}}"
```
napetrov, line 252: Should we name this as arch here instead of sub_dir? so variable name would be meaningful

dev/download_tbb.sh
```diff
@@ -18,65 +19,96 @@
...
+    echo "Downloaded and unpacked oneTBB to ${DST}/${OS}"
   else
-    echo tar -xvf "${DST}/${TBB_PACKAGE}" -C "${DST}"
-    tar -C "${DST}/${OS}" --strip-components=1 -xvf "${DST}/${TBB_PACKAGE}"
+    echo "oneTBB is already installed in ${DST}/${OS}"
   fi
-  ls -al "${DST}/${OS}/"
-  echo "Downloaded and unpacked oneTBB to ${DST}/${OS}"
+
 else
```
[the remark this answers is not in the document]
napetrov: +1 for separating download vs build scripts. 
Build scripts can be put in to ci scripts.

dev/download_tbb.sh
```diff
@@ -18,65 +19,96 @@
...
-  echo "oneTBB is already installed in ${DST}/${OS}"
-fi
+  sudo apt-get update
+  sudo apt-get install build-essential gcc gfortran cmake -y
+  git clone https://github.com/oneapi-src/oneTBB.git
+  CoreCount=$(lscpu -p | grep -Ev '^#' | wc -l)
+
+  rm -rf __deps/tbb
+  pushd oneTBB
+    git checkout v2021.11.0
```
napetrov, line 95: even once this would be in separate script it would worth placing dependency version at least to the header of the file

makefile
```diff
@@ -1122,8 +1212,15 @@ $(foreach t,$(releasetbb.LIBS_Y),$(eval $(call .release.t,$t,$(RELEASEDIR.tbb.so
 $(foreach t,$(releasetbb.LIBS_A),$(eval $(call .release.t,$t,$(RELEASEDIR.tbb.libia))))
 
 #----- cmake configs generation
+SUB_DIR=
```
napetrov, line 1215: as mentioned above - worth to name this as architecture and not just subdir?

.ci/scripts/build.sh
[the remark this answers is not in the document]
ajay-fuji:
```suggestion
if [[ "${ARCH}" == "32e" ]]
then
optimizations=${optimizations:-avx2}
elif [[ "${ARCH}" == "arm" ]]
then
optimizations=${optimizations:-sve}
else
echo "Unknown architecture '${ARCH}'"
exit 1
fi
```
Looking at return value of `identify_os.sh`, `ARCH=32e` for x86_64

dev/make/cmplr.clang.ref.arm.mk
napetrov: This should be UXL copyright that we all will have to migrate

cpp/daal/src/algorithms/kernel_inst_arm.h
```diff
@@ -0,0 +1,131 @@
...
+    template class ClassName<Mode, ContainerTemplate<__VA_ARGS__, sve> DAAL_KERNEL_SVE_CONTAINER(ContainerTemplate, __VA_ARGS__)>; \
+    }
+
+#define __DAAL_INSTANTIATE_DISPATCH_CONTAINER_SYCL(ContainerTemplate, Mode, ...)                                                               \
+    __DAAL_INSTANTIATE_DISPATCH_SYCL_IMPL(ContainerTemplate, Mode, AlgorithmDispatchContainer, AlgorithmContainerImpl<Mode>, __DAAL_GET_CPUID, \
+                                          __VA_ARGS__)
+
+#define __DAAL_INSTANTIATE_DISPATCH_CONTAINER_SYCL_SAFE(ContainerTemplate, Mode, ...)                                                               \
+    __DAAL_INSTANTIATE_DISPATCH_SYCL_IMPL(ContainerTemplate, Mode, AlgorithmDispatchContainer, AlgorithmContainerImpl<Mode>, __DAAL_GET_CPUID_SAFE, \
+                                          __VA_ARGS__)
```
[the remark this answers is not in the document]
rakshithgb-fujitsu: @Vika-F how do we remove this? By removing its breaking ` covariance_dense_default_batch_fpt_dispatcher.cpp ` which is calling it like this:

```
namespace algorithms
{
__DAAL_INSTANTIATE_DISPATCH_CONTAINER_SYCL(covariance::BatchContainer, batch, DAAL_FPTYPE, covariance::defaultDense)
}
```

makefile
```diff
@@ -1,5 +1,6 @@
 #===============================================================================
```
keeranroth, line 1: It would be nice to have this file with architecture-specific parts split out into their own files. That requires quite a bit of a refactor of this file so that we can keep as much common code in here as possible. It should be possible to just split out the setting of variables into architecture specific parts at the start of the file, and then use these later on in the makefile. I'm happy to have a look at this, if required.

cpp/daal/include/services/internal/daal_kernel_defines.h
```diff
@@ -1,6 +1,7 @@
 /* file: daal_kernel_defines.h */
```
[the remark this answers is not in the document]
rakshithgb-fujitsu: I'm okay with this suggestion, if intel team agrees will commit those changes accordingly.

makefile
rakshithgb-fujitsu: @keeranroth yup please have a look at this and do make suggestions as needed, this seems a bit tricky to isolate the parts.

cpp/daal/src/algorithms/kernel_inst_arm.h
```diff
@@ -0,0 +1,131 @@
+/* file: kernel_inst_arm.h */
```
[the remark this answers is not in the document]
rakshithgb-fujitsu: @Vika-F that PR does not show any SYCL related modifications, could you please help point it out?

cpp/daal/src/algorithms/covariance/covariance_impl.i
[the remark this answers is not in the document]
0chandansharma: Hello @Vika-F ,

Can you please share the python equivalent code to generate block size, and what "input_file_name" should be use to get optimal block size.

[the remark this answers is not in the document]
chandan-fujitsu-research: Hello @Vika-F, 

Can you please share the python equivalent code to generate block size, and what "input_file_name" should be use to get optimal block size ?

.ci/env/tbb.sh
```diff
@@ -0,0 +1,26 @@
+sudo apt-get update
+sudo apt-get install build-essential gcc gfortran cmake -y
+git clone https://github.com/oneapi-src/oneTBB.git
+CoreCount=$(lscpu -p | grep -Ev '^#' | wc -l)
+
+rm -rf __deps/tbb
+pushd oneTBB
```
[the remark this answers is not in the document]
ajay-fuji: @keeranroth thank you for suggestion. 
Currently makefile expect TBB dependency to reside inside `__deps/tbb/`. Changing name of directory would require making changes in other makefiles as well, which at this point I think is not required. @napetrov any comments.

.ci/env/tbb.sh
```diff
@@ -0,0 +1,26 @@
+sudo apt-get update
+sudo apt-get install build-essential gcc gfortran cmake -y
+git clone https://github.com/oneapi-src/oneTBB.git
+CoreCount=$(lscpu -p | grep -Ev '^#' | wc -l)
```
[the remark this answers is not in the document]
ajay-fuji: since `CoreCount=$(lscpu -p | grep -Ev '^#' | wc -l)` code is taken from `.ci/env/openblas.sh` so taking suggestion from intel team would be better. @napetrov please suggest if changes are required?

.ci/env/tbb.sh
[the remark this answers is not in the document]
keeranroth: I was referring to the `oneTBB` directory name, which is put into the root directory at the moment. This can probably be onetbb-src or something similar, so that the clone at the top of this file would be:
```
git clone https://github.com/oneapi-src/oneTBB.git onetbb-src
```

and it's probably worth doing a shallow clone so that the switch of the branch isn't necessary

```
git clone --depth 1 --branch v2021.11.0 https://github.com/oneapi-src/oneTBB.git onetbb-src
```

makefile
```diff
@@ -53,19 +55,36 @@ $(error Building with the parameters library is not available on Windows OS)
...
+CPUs := sve
+CPUs.files := a8sve
+else
 CPUs := sse2 sse42 avx2 avx512
 CPUs.files := nrh neh hsw skx
+endif
 
 USERREQCPU := $(filter-out $(filter $(CPUs),$(REQCPU)),$(REQCPU))
 USECPUS := $(if $(REQCPU),$(if $(USERREQCPU),$(error Unsupported value/s in REQCPU: $(USERREQCPU). List of supported CPUs: $(CPUs)),$(REQCPU)),$(CPUs))
+ifeq ($(ARCH_is_arm),yes)
```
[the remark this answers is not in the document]
ajay-fuji: Check for unknown architecture is already present in makefile above. 
```ifeq ($(PLAT),lnx32e)
  BACKEND_CONFIG ?= mkl
  ARCH = 32e
else ifeq ($(PLAT),mac32e)
  BACKEND_CONFIG ?= mkl
  ARCH = 32e
else ifeq ($(PLAT),lnxarm)
  BACKEND_CONFIG ?= ref
  ARCH = arm
else
  $(error Unsupported platform: $(PLAT))
endif

File: makefile [hunk 49]
```diff
@@ -133,21 +152,36 @@ y      := $(notdir $(filter $(_OS)/%,lnx/so win/dll mac/dylib))
...
+else
 p4_OPT   := $(p4_OPT.$(COMPILER))
 mc3_OPT  := $(mc3_OPT.$(COMPILER))
 avx2_OPT := $(avx2_OPT.$(COMPILER))
 skx_OPT  := $(skx_OPT.$(COMPILER))
+endif
 
 _OSr := $(if $(OS_is_win),win,$(if $(OS_is_lnx),lin,))
 
+ifeq ($(ARCH_is_arm),yes)
```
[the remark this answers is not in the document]
ajay-fuji: Hi @keeranroth thanks for suggestion
I feel code is already conditioned as per the compilation flow. If you think any branch can be merged, please suggest we will do it. (Since `makefile` is deprecating in future, does it require this effort?)

makefile
keeranroth: Ack. I'll start having a look at this tomorrow.

dev/bazel/config/cpudetect.cpp
```diff
@@ -14,7 +15,9 @@
 * limitations under the License.
 *******************************************************************************/
 
+#ifndef __ARM_ARCH
```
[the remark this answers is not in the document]
rakshithgb-fujitsu: This can be added in existing file - `cpp/daal/services/daal_defines.h`, thoughts @napetrov ?

cpp/daal/src/algorithms/covariance/covariance_impl.i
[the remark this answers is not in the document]
Vika-F: Covariance algorithms' hyperparameters can be tuned using PCA, as the Covariance algorithm is used as part of PCA.
Here is the code:

```
    pca = PCA(n_components=2, svd_solver="full")
    pca.get_hyperparameters("fit").cpu_macro_block = 140 # tested block size
    pca.fit(X)
```

The `input_file_name` (or ndarray X in Python code) should point to the file with the input data for Covariance/PCA algorithm.
You can place there any data which you think is relevant for performance tuning.

.ci/env/tbb.sh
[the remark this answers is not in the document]
napetrov: There is no strong preference. We can try nproc, but align across other scripts.

.ci/env/tbb.sh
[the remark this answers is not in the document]
napetrov: i think downloading sources archive is a better option that gitclones here. 

In terms of names - i guess the point is to have build directory properly named as build directory and compiler tbb can be put in to dependencies directory. Point is that it would be easier to debug things with clear naming

cpp/daal/include/services/internal/daal_kernel_defines.h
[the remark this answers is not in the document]
napetrov: yes, this looks as good approach to state with. 

One question - would we like to have uniformity in files?  i.e. having single kernel_defines.h enforced for both(all) architectures vs  having different sets? 

To enforce first one we probably should do something along of lines having single include with arch variable as part of the name?

cpp/daal/include/services/env_detect.h
```diff
@@ -91,7 +97,12 @@ class DAAL_EXPORT Environment : public Base
     enum CpuTypeEnable
```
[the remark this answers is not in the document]
napetrov: @Vika-F might be remember more, why there is CpuType and CpuTypeEnable. 

I think this is along lines of - we might use not matching codepath on cpu. i.e. cpu would be avx512 but we will be forcing sse2 ISA.

.ci/env/tbb.sh
```diff
@@ -0,0 +1,60 @@
+# Function to display help
```
napetrov, line 1: Currently location and naming assumes that it's widely applicable for all platforms, while it's hardcoded for ARM. + there is following download script that takes prebuild tbb versions- https://github.com/oneapi-src/oneDAL/blob/main/dev/download_tbb.sh

Might eb it worth aligning or might be merging them?

cpp/daal/src/algorithms/kernel_inst_arm.h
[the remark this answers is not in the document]
napetrov: @rakshithgb-fujitsu  - this is not part of this PR, but a comment for longerterm plans.  all sycl related finctionality in DAAL library is deprecated and will be removed.  so there is no point in copying DAAL SYCL related defines.

cpp/daal/src/threading/export_lnxarm.ref.def
napetrov: as mentioned above - this should be UXL copyright for new files.

deploy/nuget/prepare_dal_nuget.sh
```diff
@@ -101,6 +110,9 @@ create_package() {
         if [ "${platform}" = "linux-x64" ]; then
             dl_postfix=.so.${major_binary_version}.${minor_binary_version}
             sl_postfix=.a
+        elif [ "${platform}" = "linux-aarch64" ]; then
```
napetrov, line 113: Do we need this? or we should generalize to just linux? as per my understanding this would be arch agnostic postfixes?

dev/bazel/config/cpudetect.cpp
```diff
@@ -154,20 +157,25 @@ int check_sse42_features() {
...
-    else if (check_avx2_features()) {
-        return "avx2";
-    }
-    else if (check_sse42_features()) {
-        return "sse42";
-    }
-    else {
-        return "sse2";
-    }
+    #ifdef __ARM_ARCH
```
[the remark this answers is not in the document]
napetrov: Agree

deploy/local/vars_lnx.sh
```diff
@@ -223,6 +224,19 @@ if [ ! -d $__daal_tmp_dir ]; then
     __daal_tmp_dir=${component_root}
 fi
 
+
+PLATFORM=$(bash dev/make/identify_os.sh)
```
napetrov, line 228: this will not work. vars scripts are executed for configured build already.  i.e. one that get packages and provided to users - there is no dev/make/identify_os.sh script that will be available there. 

i think arch identification have to be implemented here instead

.ci/env/tbb.sh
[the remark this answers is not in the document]
ajay-fuji: @napetrov, what if we go with shallow clone as suggested by @keeranroth so that same task can be done in less lines of code as compared to archive download. 
```bash
git clone --depth 1 --branch v2021.11.0 https://github.com/oneapi-src/oneTBB.git onetbb-src
```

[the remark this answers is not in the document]
napetrov: have nothing against git clone. The only point - it would be good to have single place for versions modification. 

Also there is open question on if env/tbb.sh should be merged with download_tbb script from dev folder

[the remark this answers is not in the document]
keeranroth: The original pull request had both in the same file, which I commented I thought looked weird, in that the download script was also an Arm build recipe. I think that it's better with two different scripts, but happy to go with the majority on that one.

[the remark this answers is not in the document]
napetrov: 2 different scipts is option, but then tbb build script for arm should be named properly. For now it's just tbb.sh that would mean that this is universal script. So might be just rename it it to reflect that it's intended to work on arm only and potentially add platform check there.

deploy/local/dal
```diff
@@ -60,7 +61,14 @@ proc ModulesHelp { } {
...
 set dalroot "$componentroot"
-set daal_target_arch "intel64"
+set daalroot "$componentroot/$modulefilever"
+if {[string equal [info machine] "aarch64"]} {
+    set daal_target_arch "arm"
+} else {
+    set daal_target_arch "intel64"
+}
+
+module-whatis "Intel(R) oneAPI Data Analytics Library for $daal_target_arch."
```
napetrov, line 71: I think we can use just generic name here

```suggestion
module-whatis "oneAPI Data Analytics Library for $daal_target_arch."
```

.ci/env/tbb.sh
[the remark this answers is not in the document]
ajay-fuji: I also think `env/tbb.sh` should be generic. Only catch is for intel and arm, we know dependency path are `__deps/tbb/lnx/lib/intel64/gcc4.8/` and `__deps/tbb/lnx/lib/arm/gcc4.8/` respectively. How do we take care for other architecture?
Our suggestion for deciding architecture directory -
```bash
arch=$(uname -m)
if [ "${arch}" == "x86_64" ]; then
    arch_dir="intel64"
elif [ "${arch}" == "aarch64" ]; then
    arch_dir="arm"
else
    arch_dir=${arch}
fi
```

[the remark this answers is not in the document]
napetrov: yes, we can put things in this way.  Although - i thought that arm support would be cross compiled ?  i.e. build would happen on x86 machine for arm target. so uname will not necessary return arm platform

Also i would say we can move download_tbb script logic here for Intel and just use prebuild tbb instead of building it from sources.

napetrov:
on top of clang format, there are some other internal checks that should eb fixed 
 +---+---+---+---+---+---+---+---+------
 |NLS|TAB|CHR|TAL|EOF|150|EXT|CRT| NAME 
 +---+---+---+---+---+---+---+---+------
 | U | X | o | X | o | o | o | o | /lnx32e/src/cpp/daal/include/services/daal_defines.h
 | U | o | o | o | X | o | o | X | /lnx32e/src/cpp/daal/include/services/internal/aarch64/aarch64_kernel_defines.h
 | U | o | o | o | X | o | o | X | /lnx32e/src/cpp/daal/include/services/internal/x86_64/x86_64_kernel_defines.h
 +---+---+---+---+---+---+---+---+------
  Files tested:           3506
 |NLS|TAB|CHR|TAL|EOF|150|EXT|CRT| NAME 
 +---+---+---+---+---+---+---+---+------
  Files with errors:      3
  Fixes applied:          0
  Non-fixed errors found: 6
  Warnings found:         1361
 [legend] NLS ==> mixed dos-unix new line styles
 [legend] TAB ==> tabs in files
 [legend] CHR ==> chars outside of ASCII 127
 [legend] TAL ==> tail spaces
 [legend] EOF ==> not exactly one newline before EOF
 [legend] 150 ==> more than 150 symbols in a line
 [legend] EXT ==> exit label found
 [legend] CRT ==> wrong name of file before copyright notice
 [legend] YER ==> wrong year in copyright line

napetrov:
Internal build passing, but attempts to build daal4py on top of it failing with 
__release_lnx/daal/latest/include/oneapi/dal/detail/policy.hpp:20:10: fatal error: daal/include/services/daal_defines.h: No such file or directory

So seems first public build should be fixed + there are bazel build failures

ajay-fuji:
> on top of clang format, there are some other internal checks that should eb fixed +---+---+---+---+---+---+---+---+------ |NLS|TAB|CHR|TAL|EOF|150|EXT|CRT| NAME +---+---+---+---+---+---+---+---+------ | U | X | o | X | o | o | o | o | /lnx32e/src/cpp/daal/include/services/daal_defines.h | U | o | o | o | X | o | o | X | /lnx32e/src/cpp/daal/include/services/internal/aarch64/aarch64_kernel_defines.h | U | o | o | o | X | o | o | X | /lnx32e/src/cpp/daal/include/services/internal/x86_64/x86_64_kernel_defines.h +---+---+---+---+---+---+---+---+------ Files tested: 3506 |NLS|TAB|CHR|TAL|EOF|150|EXT|CRT| NAME +---+---+---+---+---+---+---+---+------ Files with errors: 3 Fixes applied: 0 Non-fixed errors found: 6 Warnings found: 1361 [legend] NLS ==> mixed dos-unix new line styles [legend] TAB ==> tabs in files [legend] CHR ==> chars outside of ASCII 127 [legend] TAL ==> tail spaces [legend] EOF ==> not exactly one newline before EOF [legend] 150 ==> more than 150 symbols in a line [legend] EXT ==> exit label found [legend] CRT ==> wrong name of file before copyright notice [legend] YER ==> wrong year in copyright line

Thank you @napetrov for running checks. 
Few Questions - 
1. What does `U` mean in the table? `X` - means fail and `o` mean pass, is our understanding correct?
2. Erros and warnings are with respect to clang code format only? Do we need to fix warnings also? Currently we run `.ci/scripts/clang-format.sh` to do code formatting.

cpp/daal/src/algorithms/kernel_inst_arm.h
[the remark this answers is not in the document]
Vika-F: @rakshithgb-fujitsu Yes, my point was about the deprecation of SYCL-related functionality in DAAL. Since it is deprecated, no need to replicate those defines for ARM. They will be removed in near future anyway.

Now I also see that it is not easy to remove those defines at all right now as they are used in various files.

So, let's leave them as is for now in order not to block the development anymore.

cpp/daal/src/algorithms/kernel_inst_arm.h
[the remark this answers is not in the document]
Vika-F: Please leave it as is for now. I see that it is not an easy task to have them removed at this stage.

napetrov:
> > on top of clang format, there are some other internal checks that should eb fixed +---+---+---+---+---+---+---+---+------ |NLS|TAB|CHR|TAL|EOF|150|EXT|CRT| NAME +---+---+---+---+---+---+---+---+------ | U | X | o | X | o | o | o | o | /lnx32e/src/cpp/daal/include/services/daal_defines.h | U | o | o | o | X | o | o | X | /lnx32e/src/cpp/daal/include/services/internal/aarch64/aarch64_kernel_defines.h | U | o | o | o | X | o | o | X | /lnx32e/src/cpp/daal/include/services/internal/x86_64/x86_64_kernel_defines.h +---+---+---+---+---+---+---+---+------ Files tested: 3506 |NLS|TAB|CHR|TAL|EOF|150|EXT|CRT| NAME +---+---+---+---+---+---+---+---+------ Files with errors: 3 Fixes applied: 0 Non-fixed errors found: 6 Warnings found: 1361 [legend] NLS ==> mixed dos-unix new line styles [legend] TAB ==> tabs in files [legend] CHR ==> chars outside of ASCII 127 [legend] TAL ==> tail spaces [legend] EOF ==> not exactly one newline before EOF [legend] 150 ==> more than 150 symbols in a line [legend] EXT ==> exit label found [legend] CRT ==> wrong name of file before copyright notice [legend] YER ==> wrong year in copyright line
> 
> Thank you @napetrov for running checks. Few Questions -
> 
> 1. What does `U` mean in the table? `X` - means fail and `o` mean pass, is our understanding correct?
> 2. Erros and warnings are with respect to clang code format only? Do we need to fix warnings also? Currently we run `.ci/scripts/clang-format.sh` to do code formatting.

1. Yes, correct.  U - unix. This one just checks for new line symbols.
We heve set of internal source code checks and some of them can be moved to opensource. Although they are not ideal. This is old script that was used back in the days for SVN. But it's still relevant in terms of keeping some of code consistency.

2. Errors - i mean that clang format check was failing previously. thanks for fixing

makefile
```diff
@@ -53,19 +55,50 @@ $(error Building with the parameters library is not available on Windows OS)
...
+  BACKEND_CONFIG ?= mkl
+  ARCH = 32e
+else ifeq ($(PLAT),mac32e)
+  BACKEND_CONFIG ?= mkl
+  ARCH = 32e
+else ifeq ($(PLAT),lnxarm)
+  BACKEND_CONFIG ?= ref
+  ARCH = arm
+else
+  $(error Unsupported platform: $(PLAT))
```
napetrov, line 68: for call make oneapi BUILDREV= "CHECK_DLL_SIG=yes"  -j24 REQCPU=avx2 
on windows platform. 
makefile:68: *** Unsupported platform: win32e.  Stop.

This have to be fixed to allow windows build as well . not sure what happening in public CI though

.ci/scripts/test.sh
```diff
@@ -127,12 +134,6 @@ for link_mode in ${link_modes}; do
             export CC=icx
             export CXX=icpx
         fi
-        echo "============== Configuration: =============="
```
napetrov, line 130: Why this is removed? just printing state of configs

ajay-fuji: Thanks for pointing out @napetrov. 
This is removed unintentionally, will revert changes.

napetrov:
env/lnxx86_64.sh: No such file or directory 

Failures in internal CI. Not sure what line is causing this. but expectation should be that env files would remain the same

ajay-fuji:
> env/lnxx86_64.sh: No such file or directory

Hi @napetrov,
image
This is screenshot of oneDAL's `main` branch directory tree. I don't see `lnxx86_64.sh` file there. 
Are you referring to some other path?

.ci/scripts/test.sh
```diff
@@ -54,7 +55,17 @@ TESTING_RETURN=0
...
 ARCH=${PLATFORM:3:3}
-full_arch=intel64
+if [ "$ARCH" == "32e" ]; then
+    full_arch=intel64
+    arch_dir=intel_intel64
+elif [ "$ARCH" == "arm" ]; then
+    full_arch=arm
+    arch_dir=arm_aarch64
+else
+    echo "Uknown architecture ${ARCH} detected for platform ${PLATFORM}"
```
ethanglaser, line 65:
```suggestion
    echo "Unknown architecture ${ARCH} detected for platform ${PLATFORM}"
```

ethanglaser:
> env/lnxx86_64.sh: No such file or directory
> 
> Failures in internal CI. Not sure what line is causing this. but expectation should be that env files would remain the same

Seems like somewhere ARCH is being set to x86_64 (instead of 32e)

ajay-fuji:
> > env/lnxx86_64.sh: No such file or directory
> > Failures in internal CI. Not sure what line is causing this. but expectation should be that env files would remain the same
> 
> Seems like somewhere ARCH is being set to x86_64 (instead of 32e)

Is it possible to provide more information about the error like from which command this error is generated? Which file is throwing this error? These things will help us investigate issue further.

deploy/local/vars_lnx.sh
```diff
@@ -223,6 +224,17 @@ if [ ! -d $__daal_tmp_dir ]; then
     __daal_tmp_dir=${component_root}
 fi
 
+ARCH=$(uname -m)
```
ethanglaser, line 227: Since this function is sourced, any use of ARCH/ARCH_DIR in environments that call this script would be overwritten. Would prefer changing ARCH and ARCH_DIR to ARCH_ONEDAL and ARCH_DIR_ONEDAL to avoid, as ARCH is a common env variable.

ethanglaser, line 227:
```suggestion
ARCH_ONEDAL=$(uname -m)
```

deploy/local/vars_lnx.sh
```diff
@@ -239,8 +251,8 @@ if [ "$(basename "${my_script_path}")" = "env" ] ; then   # assume stand-alone
       export LD_LIBRARY_PATH="$__daal_tmp_dir/lib${LD_LIBRARY_PATH+:${LD_LIBRARY_PATH}}"
     else
       export CPATH="$__daal_tmp_dir/include${CPATH+:${CPATH}}"
-      export LIBRARY_PATH="$__daal_tmp_dir/lib/intel64${LIBRARY_PATH+:${LIBRARY_PATH}}"
-      export LD_LIBRARY_PATH="$__daal_tmp_dir/lib/intel64${LD_LIBRARY_PATH+:${LD_LIBRARY_PATH}}"
+      export LIBRARY_PATH="$__daal_tmp_dir/lib/$ARCH_DIR${LIBRARY_PATH+:${LIBRARY_PATH}}"
+      export LD_LIBRARY_PATH="$__daal_tmp_dir/lib/$ARCH_DIR${LD_LIBRARY_PATH+:${LD_LIBRARY_PATH}}"
```
ethanglaser, line 255:
```suggestion
      export LIBRARY_PATH="$__daal_tmp_dir/lib/$ARCH_DIR_ONEDAL${LIBRARY_PATH+:${LIBRARY_PATH}}"
      export LD_LIBRARY_PATH="$__daal_tmp_dir/lib/$ARCH_DIR_ONEDAL${LD_LIBRARY_PATH+:${LD_LIBRARY_PATH}}"
```

cpp/daal/include/services/internal/x86_64/x86_64_kernel_defines.h
```diff
@@ -0,0 +1,79 @@
...
+#else
+    #define DAAL_KERNEL_AVX512_ONLY(something)
+    #define DAAL_KERNEL_AVX512_ONLY_CODE(...)
+    #define DAAL_KERNEL_AVX512_CONTAINER(ContainerTemplate, ...)
+    #define DAAL_KERNEL_AVX512_CONTAINER1(ContainerTemplate, ...)
+    #define DAAL_KERNEL_AVX512_CONTAINER_CASE(ContainerTemplate, ...)
+    #define DAAL_KERNEL_AVX512_CONTAINER_CASE_SYCL(ContainerTemplate, ...)
+#endif
+
+#endif
```
ethanglaser, line 79:
```suggestion
#endif

```
Please add new line at eof

dev/make/cmplr.clang.ref.arm.mk
ethanglaser, line 45:
```suggestion
a8sve_OPT.clang = $(-Q)march=armv8-a+sve

```
Please add new line at eof

ethanglaser:
Looks like 1 new CI fail based on updates:
`In file included from cpp/daal/include/algorithms/algorithm_container_base.h:28,
                 from cpp/daal/include/algorithms/algorithm_container_base_common.h:30,
                 from cpp/daal/include/algorithms/algorithm_base.h:28,
                 from cpp/daal/include/algorithms/algorithm.h:34,
                 from cpp/daal/include/algorithms/sorting/sorting_batch.h:28,
                 from cpp/daal/src/algorithms/sorting/sorting_batch_container.h:27,
                 from cpp/daal/src/algorithms/sorting/sorting_dense_default_batch_fpt_dispatcher.cpp:24:
bazel-out/k8-opt/bin/cpp/daal/include/services/internal/daal_kernel_defines.h:42:14: fatal error: x86_64/x86_64_kernel_defines.h: No such file or directory
   42 |     #include "x86_64/x86_64_kernel_defines.h`

Also clang formatting check failed.

cpp/daal/include/services/internal/daal_kernel_defines.h
```diff
@@ -31,11 +32,18 @@
...
+
 #define DAAL_KERNEL_SSE2
 #define DAAL_KERNEL_SSE42
 #define DAAL_KERNEL_AVX2
 #define DAAL_KERNEL_AVX512
 
+#if defined(TARGET_X86_64)
+    #include "x86_64/x86_64_kernel_defines.h"
+#elif defined(TARGET_ARM)
+    #include "aarch64/aarch64_kernel_defines.h"
```
ethanglaser, line 44:
```suggestion
    #include "services/internal/x86_64/x86_64_kernel_defines.h"
#elif defined(TARGET_ARM)
    #include "services/internal/aarch64/aarch64_kernel_defines.h"
```
@rakshithgb-fujitsu this should resolve the error. As far as bazel changes go I don't believe they are necessary.

cpp/daal/include/services/internal/daal_kernel_defines.h
```diff
@@ -50,82 +58,7 @@ case cpuType:
...
-    #define DAAL_KERNEL_AVX512_CONTAINER1(ContainerTemplate, ...)
-    #define DAAL_KERNEL_AVX512_CONTAINER_CASE(ContainerTemplate, ...)
-    #define DAAL_KERNEL_AVX512_CONTAINER_CASE_SYCL(ContainerTemplate, ...)
-#endif
-
 #define DAAL_EXPAND(...) __VA_ARGS__
 /** @} */
 
-#endif
+#endif
```
ethanglaser, line 64:
```suggestion
#endif

```
eof here as well

cpp/daal/include/services/internal/daal_kernel_defines.h
rakshithgb-fujitsu: Thanks!

cpp/daal/include/services/internal/x86_64/x86_64_kernel_defines.h
ethanglaser, line 1: needs header

cpp/daal/include/services/internal/aarch64/aarch64_kernel_defines.h
ethanglaser, line 1: needs header and newline at eof

ethanglaser:
Please sync branch to main, as there is a change which will fix a problem with internal CI build.

dev/bazel/config/cpudetect.cpp
```diff
@@ -14,7 +15,11 @@
 * limitations under the License.
 *******************************************************************************/
 
+#include <daal/include/services/daal_defines.h>
```
ethanglaser, line 18:
```suggestion
#include "cpp/daal/include/services/daal_defines.h"
```

ethanglaser: Getting a file not found error with this - use of quotes should help

dev/bazel/config/cpudetect.cpp
```diff
@@ -14,7 +15,11 @@
 * limitations under the License.
 *******************************************************************************/
 
+#include "cpp/daal/include/services/daal_defines.h"
```
ethanglaser, line 18: This line still is causing some trouble in our CI - lets try redefining the imported variables instead (add [this logic](https://github.com/oneapi-src/oneDAL/pull/2614/files#diff-830416203aa5bbb105889fc222326f2b0057677eb29abca3727a56e83c1d0b4bR32) instead of including daal_defines)

rakshithgb-fujitsu: okay, made those changes now

ethanglaser:
Please sync to main again - getting build error with deprecated abs that should be corrected by rebasing branch

cpp/daal/include/services/internal/aarch64/aarch64_kernel_defines.h
```diff
@@ -0,0 +1,24 @@
...
+#else
+    #define DAAL_KERNEL_SVE_ONLY(something)
+    #define DAAL_KERNEL_SVE_ONLY_CODE(...)
+    #define DAAL_KERNEL_SVE_CONTAINER(ContainerTemplate, ...)
+    #define DAAL_KERNEL_SVE_CONTAINER1(ContainerTemplate, ...)
+    #define DAAL_KERNEL_SVE_CONTAINER_CASE(ContainerTemplate, ...)
+    #define DAAL_KERNEL_SVE_CONTAINER_CASE_SYCL(ContainerTemplate, ...)
+#endif
+
+#endif
```
ethanglaser, line 24: add new line

cpp/daal/include/services/internal/aarch64/aarch64_kernel_defines.h
ethanglaser, line 1: needs copyright

ethanglaser:
Looks like bazel tests are running better now. @homksei could you check those example fails? looks like may require updating path in infra. http://intel-ci.intel.com/eec4ff3c-da06-f16f-b641-a4bf010d0e2e

cpp/daal/include/services/internal/aarch64/aarch64_kernel_defines.h
rakshithgb-fujitsu: added

cpp/daal/include/services/internal/aarch64/aarch64_kernel_defines.h
rakshithgb-fujitsu: added

cmake/scripts/generate_config.cmake
```diff
@@ -19,7 +20,7 @@ set(DAL_ROOT_REL_PATH "../../..")
 set(INC_REL_PATH "include")
 set(LIB_REL_PATH "lib")
 set(DLL_REL_PATH "redist")
-set(SUB_DIR "intel64")
+set(ARCH_DIR "${ARCH_DIR}")
```
ethanglaser, line 23: Let's use ARCH_DIR_ONEDAL here as well (and in makefile and oneDALConfig.cmake.in)

rageshhajela16: Thanks for the review. We have incorporated this comment in the latest commit pushed. Thanks

ethanglaser:
The internal CI failure with our examples are still unresolved - anyone else on the intel side able to take a peek at the fails?

rageshhajela16:
> The internal CI failure with our examples are still unresolved - anyone else on the intel side able to take a peek at the fails?

@ethanglaser @napetrov  can you please help to check if the CI is still failing with the recent rebase! Thanks

Alexandr-Solovev:
Hi @rageshhajela16!
I have checked the last commit. Unfortunately, the red ci steps still exist. After the investigation of internal ci issues, I found the issue.
The true issue log in examples is 
image

I doublecheck how __release_lnx/daal/latest/lib/cmake/oneDAL/oneDALConfig.cmake looks like and it looks 

image

 So instead of lib/"arch_name" it uses just lib, and I would like to say that its incorrectly. 
 In comparison with master(it has green ci) it looks like
 
image

I also doublechecked your changes and these changes looks good for me https://github.com/oneapi-src/oneDAL/pull/2614/files#diff-a711276942a2647ec3ee01db428047af7e0efcd9ed25a93f600bc9bf53b70c2dR212
But may be its possible to replace it with ${ARCH_DIR_ONEDAL}, what do you think?

ajay-fuji:
> Hi @rageshhajela16! I have checked the last commit. Unfortunately, the red ci steps still exist. After the investigation of internal ci issues, I found the issue. The true issue log in examples is image
> 
> I doublecheck how __release_lnx/daal/latest/lib/cmake/oneDAL/oneDALConfig.cmake looks like and it looks
> 
> image
> 
> So instead of lib/"arch_name" it uses just lib, and I would like to say that its incorrectly. In comparison with master(it has green ci) it looks like
> 
> image
> 
> I also doublechecked your changes and these changes looks good for me https://github.com/oneapi-src/oneDAL/pull/2614/files#diff-a711276942a2647ec3ee01db428047af7e0efcd9ed25a93f600bc9bf53b70c2dR212 But may be its possible to replace it with ${ARCH_DIR_ONEDAL}, what do you think? Or may be you have some more ideas about what could be the reason. I checked env variables and ARCH_DIR_ONEDAL is set correctly. I also will work on the investigation

From the build directory name it looks like, onedal is built using `icc` compiler (now depracated in version `2024.0`, please confirm!). 
We have run tried building onedal and examples using below commands - 
`make -f makefile onedal_c -j32`
`.ci/scripts/test.sh --test-kind examples --build-dir__release_lnx --compiler gnu --interface daal/cpp --build_system cmake`
`.ci/scripts/test.sh --test-kind examples --build-dir__release_lnx --compiler gnu --interface oneapi/cpp --build_system cmake`

Everything works fine. All test examples building and passing successfully. 

here compiler is automatically taken as `icc` (`/opt/intel/oneapi/compiler/2023.2.1/linux/bin/intel64/icc`) (noticed from logs). 
`icc` installed via `sudo apt install intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic`

oneDALConfig.cmake contain correct prefix - 
image

Just to cross verify, is it possible to share configuration used to generate binaries? More precisely build command.  
Thanks!

napetrov:
> > Hi @rageshhajela16! I have checked the last commit. Unfortunately, the red ci steps still exist. After the investigation of internal ci issues, I found the issue. The true issue log in examples is image
> > I doublecheck how __release_lnx/daal/latest/lib/cmake/oneDAL/oneDALConfig.cmake looks like and it looks
> > image
> > So instead of lib/"arch_name" it uses just lib, and I would like to say that its incorrectly. In comparison with master(it has green ci) it looks like
> > image
> > I also doublechecked your changes and these changes looks good for me https://github.com/oneapi-src/oneDAL/pull/2614/files#diff-a711276942a2647ec3ee01db428047af7e0efcd9ed25a93f600bc9bf53b70c2dR212 But may be its possible to replace it with ${ARCH_DIR_ONEDAL}, what do you think? Or may be you have some more ideas about what could be the reason. I checked env variables and ARCH_DIR_ONEDAL is set correctly. I also will work on the investigation
> 
> From the build directory name it looks like, onedal is built using `icc` compiler (now depracated in version `2024.0`, please confirm!). We have run tried building onedal and examples using below commands - `make -f makefile onedal_c -j32` `.ci/scripts/test.sh --test-kind examples --build-dir__release_lnx --compiler gnu --interface daal/cpp --build_system cmake` `.ci/scripts/test.sh --test-kind examples --build-dir__release_lnx --compiler gnu --interface oneapi/cpp --build_system cmake`
> 
> Everything works fine. All test examples building and passing successfully.
> 
> here compiler is automatically taken as `icc` (`/opt/intel/oneapi/compiler/2023.2.1/linux/bin/intel64/icc`) (noticed from logs). `icc` installed via `sudo apt install intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic`
> 
> oneDALConfig.cmake contain correct prefix - image
> 
> Just to cross verify, is it possible to share configuration used to generate binaries? More precisely build command. Thanks!

on icc usage question - yes, we use it for build currently with plan for migration to icx. 
But for icc it's public build, one of latest versions.

Alexandr-Solovev:
@ethanglaser  here the job with infra fixes https://intel-ci.intel.com/eecfd78a-97b3-f181-afe8-a4bf010d0e2e

napetrov:
> @ethanglaser here the job with infra fixes https://intel-ci.intel.com/eecfd78a-97b3-f181-afe8-a4bf010d0e2e Examples look good now.

would suggest running nightly validation to check on broader scope, but so far things looks good and ready for integration

Alexandr-Solovev:
> > @ethanglaser here the job with infra fixes https://intel-ci.intel.com/eecfd78a-97b3-f181-afe8-a4bf010d0e2e Examples look good now.
> 
> would suggest running nightly validation to check on broader scope, but so far things looks good and ready for integration

Previous regular job has successfully finished!
Nightly job with this pr:  https://intel-ci.intel.com/eecfef9e-97dc-f1b8-9360-a4bf010d0e2e

ethanglaser (approved):
Good from CI perspective. Thanks for taking a look @Alexandr-Solovev!

napetrov:
Thanks everyone for working on this PR!

rageshhajela16:
> Thanks everyone for working on this PR!

Thanks everyone contributing to this PR!

PR #2626, opened 2024-01-10
fix: bazel heterogen serialization

Alexandr-Solovev:
# Description
Its a fix for our ci test: `test_heterogen_serialization_dpc`, which sporadically fails. Changes contain removing `parallel_for` loop on cpu, and changing it to default `for` loop due to sporadic memory issue.
In terms of performance I extended list of generated datasets especially in `test_heterogen_serialization_host` , due to 
impossibility of getting stable performance results in dpc, but for both of these cases `host` and `dpc` these functions are callable. 
On the test with command line bazel test //cpp/oneapi/dal/table:test_heterogen_serialization_host --test_output=all, on the same hardware. `Random heterogen table - host` with parameters
`const std::int64_t column_count = GENERATE(1, 9, 99, 999, 10000, 100000, 1000000);`
`const std::int64_t row_count = GENERATE(1, 9, 99, 999, 9'999, 10000, 80000, 800000);`

this branch takes: 1st run **136.7** seconds, 2nd run **136.9** seconds.
main branch takes: 1st run **143.3** seconds, 2nd run **141.3** seconds.


Vika-F (approved):
The changes look good to me.

The only request is to add more details into the description: name of the bazel test that was fixed, the failure's log; the info that the fix does not affect the performance, etc.

ethanglaser (approved):
LGTM as well. Maybe worth checking with Nikita about the original loop intention if possible? 2 fails in CI are unrelated to these changes.

PR #2629, opened 2024-01-12
Add GPU support in detail::csr_table_builder class

Vika-F:
Add ability to build CSR tables from USM-allocated data.

### DIFF
--- a/cpp/oneapi/dal/table/backend/csr_table_builder_impl.hpp
+++ b/cpp/oneapi/dal/table/backend/csr_table_builder_impl.hpp
@@ -25,7 +25,6 @@ namespace oneapi::dal::backend {
 
 class csr_table_builder_impl : public detail::csr_table_builder_template<csr_table_builder_impl> {
 public:
-
 #ifdef ONEDAL_DATA_PARALLEL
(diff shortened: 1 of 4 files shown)


cpp/oneapi/dal/table/backend/csr_table_builder_impl.hpp
```diff
@@ -25,9 +25,15 @@ namespace oneapi::dal::backend {
 
 class csr_table_builder_impl : public detail::csr_table_builder_template<csr_table_builder_impl> {
 public:
+#ifdef ONEDAL_DATA_PARALLEL
+    csr_table_builder_impl() : dependencies_(nullptr) {
```
Alexandr-Solovev, line 29: What's the necessity of usage dependencies here? In comparison with homogen_table, there is no dependencies in homogen_table_builder

Alexandr-Solovev (commented):
LGTM, but the addition of dependencies vector is little bit unclear to me, due to I didn't find any depends_on/wait_and_throw/other functions which call smth with this dependencies

Alexandr-Solovev (approved):

PR #2635, opened 2024-01-19
feat: SHAP values for binary classification [target: r2025.0]

ahuber21:
# Description
Enable calculation of SHAP values for binary classification models

- Allocate result similar to GBT regression
- Rely on GBT regression SHAP dispatching, analogous to calculation of probabilities
- Correctness is tested in a matching update in sklearnex, where we compare against SHAP values from XGBoost

ToDos:

- [ ] Potentially add version checks, depending on CI results


cpp/daal/include/algorithms/gradient_boosted_trees/gbt_classification_predict_types.h
```diff
@@ -68,7 +90,7 @@ enum ResultToComputeId
 };
 
 /**
- * \brief Contains version 2.0 of the Intel(R) oneAPI Data Analytics Library interface.
+ * \brief Contains version 2 .0 of the Intel(R) oneAPI Data Analytics Library interface.
```
samir-nasibli, line 93:
```suggestion
 * \brief Contains version 2.0 of the Intel(R) oneAPI Data Analytics Library interface.
```

ahuber21:
There is now way around the API-breaking change. I will finish the work as much as possible, but merging needs to be postponed to a major release.

icfaust (approved):
LGTM

PR #2638, opened 2024-01-23
Add RF clsf inference hyperparameters

ahuber21:
# Description
Adds hyperparameters to random forest classification. Adds all the necessary glue logic in oneDAL to set the parameters in the DAAL kernel, to which the oneDAL implementation falls back.

- Replace hard-coded `#define` statements in `df_classification_predict_dense_default_batch_impl.i` with oneDALs hyperparameter framework
- Per default, the non-AVX512 values are configured, AVX512 will be configured through the Python interface (to be added)

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
@@ -1026,9 +1026,9 @@ Status PredictClassificationTask<algorithmFPType, cpu>::predictAllPointsByAllTre
             parallelPredict(aX, aNode, treeSize, nBlocks, nCols, _blockSize, residualSize, tlsData.local(tid), iTree);
         });
 
-        const size_t nThreads       = tlsData.nthreads();
+        const size_t nThreads  = tlsData.nthreads();
         const size_t blockSize = 256; // TODO: Why can't this be the class value _blockSize?
-        const size_t nBlocks        = nRowsOfRes / blockSize + !!(nRowsOfRes % blockSize);
+        const size_t nBlocks   = nRowsOfRes / blockSize + !!(nRowsOfRes % blockSize);
 
         daal::threader_for(nBlocks, nBlocks, [&](const size_t iBlock) {
             const size_t begin = iBlock * blockSize;
@@ -1092,7 +1092,7 @@ Status PredictClassificationTask<algorithmFPType, cpu>::predictAllPointsByAllTre
         if (prob != nullptr || res != nullptr)
         {
             const size_t blockSize = 256;
-            const size_t nBlocks    = nRowsOfRes / blockSize + !!(nRowsOfRes % blockSize);
+            const size_t nBlocks   = nRowsOfRes / blockSize + !!(nRowsOfRes % blockSize);
 
             daal::threader_for(nBlocks, nBlocks, [&, nCols](const size_t iBlock) {
                 const size_t begin = iBlock * blockSize;
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
@@ -1133,10 +1133,7 @@ Status PredictClassificationTask<algorithmFPType, cpu>::predictAllPointsByAllTre
 template <typename algorithmFPType, CpuType cpu>
 Status PredictClassificationTask<algorithmFPType, cpu>::run(services::HostAppIface * const pHostApp)
 {
-    if (!assertHyperparameters())
-    {
-        return services::Status(services::ErrorIncorrectDataRange);
-    }
+    DAAL_CHECK(assertHyperparameters(), services::ErrorIncorrectDataRange);
 
     const auto nTreesTotal = _model->size();
     if (_cachedData != _data)
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
@@ -998,13 +998,8 @@ Status PredictClassificationTask<algorithmFPType, cpu>::predictAllPointsByAllTre
     algorithmFPType * const res  = resBD.get();
     algorithmFPType * const prob = probBD.get();
     daal::SafeStatus safeStat;
-#if defined(TARGET_X86_64)
-    const size_t blockSize = cpu == avx512 ? _defaultBlockSize : _defaultBlockSizeCommon;
-#elif defined(TARGET_ARM)
-    const size_t blockSize = cpu == avx512 ? _defaultBlockSize : _defaultBlockSizeCommon;
-#endif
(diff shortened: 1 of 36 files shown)


ahuber21:
I think the Pre-commit step shouldn't have failed, restarted it here http://intel-ci.intel.com/eee09227-f812-f180-afe8-a4bf010d0e2e

ahuber21:
The combination of the two runs shows all green checks.

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
```diff
@@ -213,6 +232,10 @@ protected:
     size_t _nClasses;
     size_t _cachedNClasses;
     VotingMethod _votingMethod;
+    size_t _blockSize;
+    size_t _minTreesForThreading;
+    size_t _minNumberOfRowsForVectSeqCompute;
+    float _scaleFactorForVectParallelCompute;
     static const size_t s_cMaxClassesBufSize = 32;
```
Vika-F, line 239: This also looks like a hyperparameter. Can be target for a future work in this direction.

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
```diff
@@ -523,7 +593,7 @@ DAAL_FORCEINLINE void PredictClassificationTask<double, avx512>::predictByTree(c
         {
             checkMask = 0;
             size_t i  = 0;
-            for (size_t i = 0; i < _DEFAULT_BLOCK_SIZE; i += 8)
+            for (size_t i = 0; i < _blockSize; i += 8)
```
Vika-F, line 596: It looks like there is a restriction: A `_blockSize` should actually be a multiple of 8. But the code does not checks this; and it can be easily broken if `_blockSize % 8 != 0`.
I think it would be better to have a hyperparameter defined as `_blockSizeMultiplier` and define the `_blockSize` as  `_blockSize = 8 * _blockSizeMultiplier`.
Then `_blockSizeMultiplier` won't have additional restrictions, it should only be positive and that's it.

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
```diff
@@ -969,23 +1035,23 @@ Status PredictClassificationTask<algorithmFPType, cpu>::predictAllPointsByAllTre
...
             const DecisionTreeNode * const aNode = (const DecisionTreeNode *)(*_aTree[iTree]).getArray();
-            parallelPredict(aX, aNode, treeSize, nBlocks, nCols, blockSize, residualSize, tlsData.local(tid), iTree);
+            parallelPredict(aX, aNode, treeSize, nBlocks, nCols, _blockSize, residualSize, tlsData.local(tid), iTree);
         });
 
-        const size_t nThreads  = tlsData.nthreads();
-        const size_t blockSize = 256;
-        const size_t nBlocks   = nRowsOfRes / blockSize + !!(nRowsOfRes % blockSize);
+        const size_t nThreads       = tlsData.nthreads();
+        const size_t localBlockSize = 256; // TODO: Why can't this be the class value _blockSize?
```
Vika-F, line 1049: This also looks like another one hyperparameter. As this value differs from the value of `_blockSize` and is used differently.
I think it can also be a target for a future extension of the hyperparameters set in RF.

Vika-F (commented):
Thank you for adding hyperparameters functionality into random forest inference. That is a huge work and hopefully it will help us improve RF's performance.

I have following major comments:

1. Copyright messages need to be updated in new files in accordance with the new guidelines.
2. `_blockSize` hyperparameter has to be a multiplier of 8 (or 16 in case of float data type) on AVX512. As now it is possible to provide this value from the user code, it is needed at least to add checks that the value is divisible by 8 (or 16) where needed.
3. It is necessary to add a test that the values of hyperparameters can actually be set from oneDAL's API.

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
ahuber21: This assumption contradicts with the value `_DEFAULT_BLOCK_SIZE_COMMON = 22`. Previously, `_DEFAULT_BLOCK_SIZE` and `_DEFAULT_BLOCK_SIZE_COMMON` were independent. But now we're  setting both via the `blockSize` hyperparam. I strongly assume this is fine, because I don't see any reason why the `_DEFAULT_BLOCK_SIZE_COMMON` value would have to be `22`. So I will just always initialize to `32` and require `blockSize % blockSizeMultiplier == 0`. This will allow us to benchmark all scenarios later, even the `22` case.

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
ahuber21: Yes, I would add it now, but I think in order to give it an appropriate name, we should first understand its exact role. So I'll leave it for future work.

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
ahuber21: Good catch.

ahuber21:
@Vika-F I have addressed your first two points, let's hope the CI is still green. For the third point, I will create separate tasks!

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_predict_dense_default_batch_impl.i
Vika-F: It actually does not contradict with `_DEFAULT_BLOCK_SIZE_COMMON = 22`.
Because there are 2 separate versions of `predictByTree` method:

1. `predictByTreeCommon`. This one can accept any value of `_blockSize`, even 22.
2. PredictClassificationTask<float[double], **avx512**>::predictByTree. This one can accept only `_blockSize` which is a multiple of 16 (8 for double), because it involves SIMD computations written with intrinsics. This part of the code just can't handle any value of `_blockSize`, unfortunately.

I see that `_DEFAULT_BLOCK_SIZE` and `_DEFAULT_BLOCK_SIZE_COMMON` are now merged into a single hyperparameter. I think it is a good decision to do so. But now some additional logic needs to be implemented in order not to break the code accidentally by passing incorrect (i.e. not divisible by 8[16]) value of `_blockSize` into  `PredictClassificationTask<float[double], avx512>::predictByTree`.

ethanglaser:
CI is in good shape. @Vika-F if changes look good feel free to merge it

Vika-F:
@ethanglaser Yes, 90+% of the work is done here, need to implement the test and add a couple of changes. I've moved the work into https://github.com/oneapi-src/oneDAL/pull/2692. Will merge that one as soon as it is ready.

ethanglaser:
@Vika-F can this one be closed now?

Vika-F:
Closing this PR as https://github.com/oneapi-src/oneDAL/pull/2692 was already merged.

PR #2642, opened 2024-01-24
Align Newton-CG and CG-solver stopping conditions with sklearn.utils

avolkov-intel:
# Description
Aligning implementation of cg-solver and newton-cg with https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/optimize.py#L148 and https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/optimize.py#L84

- Change newton-cg and cg-solver primitives stopping conditions
- Change the default number of maxinner iterations for newton-cg
- Add profiler kernels

### DIFF
--- a/cpp/oneapi/dal/algo/logistic_regression/detail/optimizer.cpp
+++ b/cpp/oneapi/dal/algo/logistic_regression/detail/optimizer.cpp
@@ -52,7 +52,7 @@ public:
                                                        pr::base_function<Float>& f,
                                                        pr::ndview<Float, 1>& x,
                                                        const be::event_vector& deps = {}) {
-        return pr::newton_cg(q, f, x, Float(tol_), max_iter_, deps);
+        return pr::newton_cg(q, f, x, Float(tol_), max_iter_, 200l, deps);
     }
 
     std::pair<sycl::event, std::int64_t> minimize(sycl::queue& q,
--- a/cpp/oneapi/dal/backend/primitives/optimizers/cg_solver_dpc.cpp
+++ b/cpp/oneapi/dal/backend/primitives/optimizers/cg_solver_dpc.cpp
@@ -19,21 +19,23 @@
 #include "oneapi/dal/backend/primitives/blas/gemv.hpp"
 #include "oneapi/dal/backend/primitives/element_wise.hpp"
 #include "oneapi/dal/detail/error_messages.hpp"
+#include "oneapi/dal/detail/profiler.hpp"
 
 namespace oneapi::dal::backend::primitives {
 
 template <typename Float>
-sycl::event cg_solve(sycl::queue& queue,
-                     base_matrix_operator<Float>& mul_operator,
-                     const ndview<Float, 1>& b,
-                     ndview<Float, 1>& x,
-                     ndview<Float, 1>& residual,
-                     ndview<Float, 1>& conj_vector,
-                     ndview<Float, 1>& buffer,
-                     Float tol,
-                     Float atol,
-                     std::int64_t maxiter,
-                     const event_vector& deps) {
+std::pair<sycl::event, std::int64_t> cg_solve(sycl::queue& queue,
+                                              base_matrix_operator<Float>& mul_operator,
+                                              const ndview<Float, 1>& b,
+                                              ndview<Float, 1>& x,
+                                              ndview<Float, 1>& residual,
+                                              ndview<Float, 1>& conj_vector,
+                                              ndview<Float, 1>& buffer,
+                                              Float tol,
+                                              Float atol,
+                                              std::int64_t maxiter,
+                                              const event_vector& deps) {
+    ONEDAL_PROFILER_TASK(cg_solve, queue);
     // Solving the equation mul_operator(x) = b
     const std::int64_t p = b.get_dimension(0);
     ONEDAL_ASSERT(b.has_data());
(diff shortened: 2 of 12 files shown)


cpp/oneapi/dal/backend/primitives/optimizers/common_dpc.cpp
```diff
@@ -90,6 +90,33 @@ sycl::event l1_norm(sycl::queue& queue,
     });
 }
 
+template <typename Float>
+sycl::event max_abs(sycl::queue& queue,
```
ethanglaser, line 94: There are some existing primitive functions that could possibly be equivalent to this. Not sure if you've checked those out. Source: https://github.com/oneapi-src/oneDAL/blob/main/cpp/oneapi/dal/backend/primitives/reduction/reduction_dpc.cpp, example call: https://github.com/oneapi-src/oneDAL/blob/main/cpp/oneapi/dal/algo/svm/backend/gpu/train_results.hpp#L114-L119

cpp/oneapi/dal/backend/primitives/optimizers/line_search_dpc.cpp
```diff
@@ -42,7 +42,8 @@ Float backtracking(sycl::queue queue,
         .wait_and_throw();
     std::int64_t iter_num = 0;
     Float cur_val = 0;
-    while ((iter_num == 0 || cur_val > f0 + c1 * alpha * df0) && iter_num < 100) {
+    constexpr Float eps = std::numeric_limits<Float>::epsilon();
+    while ((iter_num == 0 || cur_val > f0 + c1 * alpha * df0) && alpha > eps) {
```
ethanglaser, line 46: would iter_num condition be completely removed?

cpp/oneapi/dal/backend/primitives/optimizers/newton_cg.hpp
```diff
@@ -30,6 +30,7 @@ std::pair<sycl::event, std::int64_t> newton_cg(sycl::queue& queue,
                                                ndview<Float, 1>& x,
                                                Float tol = 1.0e-5,
                                                std::int64_t maxiter = 100l,
+                                               std::int64_t maxinner = 200l,
```
ethanglaser, line 33: how was 200 chosen?

avolkov-intel: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/optimize.py#L156

cpp/oneapi/dal/backend/primitives/optimizers/line_search_dpc.cpp
avolkov-intel: We still need to track if it is the first iteration or not. We could have supported a bool flag instead but for me there is no difference in that case

avolkov-intel: Also, I'm thinking about returning not only sycl::event but also iter_num as a result, so I think it is fine to leave it like this.

cpp/oneapi/dal/backend/primitives/optimizers/common_dpc.cpp
avolkov-intel: These primitives work take 2d-data as an input while we need to reduce a 1d-vector. I think it is unreasonable to cast data to 2d ndview in order to use these functions.

cpp/oneapi/dal/backend/primitives/optimizers/cg_solver_dpc.cpp
```diff
@@ -19,21 +19,23 @@
...
-                     const ndview<Float, 1>& b,
-                     ndview<Float, 1>& x,
-                     ndview<Float, 1>& residual,
-                     ndview<Float, 1>& conj_vector,
-                     ndview<Float, 1>& buffer,
-                     Float tol,
-                     Float atol,
-                     std::int64_t maxiter,
-                     const event_vector& deps) {
+std::pair<sycl::event, std::int64_t> cg_solve(sycl::queue& queue,
```
Alexandr-Solovev, line 27: its not critical, but for the most part of oneDAL functions we firstly return values, and after that events

Alexandr-Solovev (approved):
LGTM, with minor comments

cpp/oneapi/dal/backend/primitives/optimizers/cg_solver_dpc.cpp
avolkov-intel: At the beginning this function was only returning sycl::event, then num_iter appeared as additional return result

cpp/oneapi/dal/algo/logistic_regression/detail/optimizer.cpp
```diff
@@ -52,7 +52,7 @@ class newton_cg_optimizer_impl : public optimizer_impl {
                                                        pr::base_function<Float>& f,
                                                        pr::ndview<Float, 1>& x,
                                                        const be::event_vector& deps = {}) {
-        return pr::newton_cg(q, f, x, Float(tol_), max_iter_, deps);
+        return pr::newton_cg(q, f, x, Float(tol_), max_iter_, 200l, deps);
```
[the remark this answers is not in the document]
avolkov-intel: In scikit-learn it is impossible to change max_inner parameter, so I think for now we don't need to do this.

PR #2644, opened 2024-01-25
[CI] disable docker-validation-nightly for forks

icfaust:
# Description
It seems like docker validation is working on personal forks of oneDAL. This seems to be a waste of resources

- Add if statement to .github/workflows/docker-validation-nightly.yml

if statement will cause a skip (https://github.com/icfaust/oneDAL/actions/runs/7654417941). (github.repository == 'oneapi-src/oneDAL')
Here is the if statement if true (https://github.com/icfaust/oneDAL/actions/runs/7654476528). (github.repository == 'icfaust/oneDAL'
 Slightly modified version so I could manually trigger it (using workflow_dispatch).

### DIFF
--- a/.github/workflows/docker-validation-nightly.yml
+++ b/.github/workflows/docker-validation-nightly.yml
@@ -26,6 +26,7 @@ permissions:
 jobs:
   validate:
     name: Docker validation
+    if: github.repository == 'oneapi-src/oneDAL'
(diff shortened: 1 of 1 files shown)


ahuber21 (approved):
Let's give this a try. Please monitor the CI some days after merging this and watch out for unexpected side effects.

PR #2631, opened 2024-01-16
fix: replace deprecated `sycl::abs` to `sycl::fabs`

homksei:

PR #2648, opened 2024-01-26
[fix] update float sycl::abs calls to sycl::fabs due to depreciation

icfaust:
# Description
Future oneapi sycl compilers (icpx) will raise 
```C++
error: 'abs<float>' is deprecated: abs for floating point types is non-standard and has been deprecated. Please use fabs instead. [-Werror,-Wdeprecated-declarations]
```
In the standard build setup. Cases which use sycl::abs with floats need to switch to using sycl::fabs.
- sycl::abs -> sycl::fabs


icfaust:
Accidental copy of #2631 will close now

PR #2657, opened 2024-02-06
fix: forest coverity issues

Alexandr-Solovev:
# Description
Pr contains fixes for coverity in Decision Forest algorithm

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/dtrees_train_data_helper.i
+++ b/cpp/daal/src/algorithms/dtrees/dtrees_train_data_helper.i
@@ -119,7 +119,8 @@ struct SplitData
           nLeft(0),
           iStart(0),
           totalWeights(0.0),
-          leftWeights(0.0)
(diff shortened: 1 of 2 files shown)


cpp/daal/src/algorithms/dtrees/dtrees_train_data_helper.i
```diff
@@ -119,7 +119,8 @@ struct SplitData
           nLeft(0),
           iStart(0),
           totalWeights(0.0),
-          leftWeights(0.0)
+          leftWeights(0.0),
+          left {}
```
ahuber21, line 123: Isn't it best practice to instantiate in the same order we declare in the class?

Alexsandruss:
@mergifyio rebase

Alexsandruss (changes requested):
Correct args order for SplitData

Alexsandruss (approved):

PR #2658, opened 2024-02-06
feature: adding assume_centered flag in Covariance

Alexandr-Solovev:
# Description
Adding assume centering flag to reduce computation time and increase sklearn conformance.

-
-
-


cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -35,7 +35,7 @@
 #include "src/algorithms/service_error_handling.h"
 #include "src/threading/threading.h"
 #include "src/externals/service_profiler.h"
-
+#include <iostream>
```
olegkkruglov, line 38: probably need to remove

cpp/oneapi/dal/algo/covariance/backend/gpu/compute_kernel_dense_impl_dpc.cpp
```diff
@@ -59,11 +59,16 @@ result_t compute_kernel_dense_impl<Float>::operator()(const descriptor_t& desc,
...
+    auto assume_centered = desc.get_assume_centered();
+
     auto result = compute_result<task_t>{}.set_result_options(desc.get_result_options());
+    // pr::ndarray<Float, 1> sums;
 
     const auto data_nd = pr::table2ndarray<Float>(q_, data, alloc::device);
-
-    auto [sums, sums_event] = compute_sums(q_, data_nd);
+    // sycl::event sums_event;
+    // if (!assume_centered) {
```
olegkkruglov, line 69: why is it commented?

cpp/oneapi/dal/algo/covariance/backend/gpu/misc.hpp
```diff
@@ -43,17 +43,25 @@ namespace pr = dal::backend::primitives;
...
 
     const std::int64_t column_count = data.get_dimension(1);
-    auto sums = pr::ndarray<Float, 1>::empty(q, { column_count }, alloc::device);
-    constexpr pr::sum<Float> binary{};
-    constexpr pr::identity<Float> unary{};
-    auto sums_event = pr::reduce_by_columns(q, data, sums, binary, unary, deps);
-    return std::make_tuple(sums, sums_event);
+    if (assume_centered) {
+        auto sums = pr::ndarray<Float, 1>::empty(q, { column_count }, alloc::device);
+        auto fill_event = pr::fill(q, sums, Float(0), deps);
```
icfaust, line 55: Dumb question: above in compute_kernel_dense_impl_dpc.cpp you used pr::ndarray<Float, 1>::zeros to fill with zeros, and here you start with an empty and fill with zeros. Is there a difference/ which is superior? Or are there two cases here that require different approaches?

icfaust (commented):
Maybe extend the testing to run with assumed_centered False and True just to make sure that the code gets exercised (in tests/batch). I hope is wouldn't be too much work. I would wait for mathematical verification on sklearnex side (which should be pretty straighforward/quick)

icfaust (approved):
After discussion, with a couple tiny changes I think it is good to go.

samir-nasibli (commented):
Look good to me. But I also expecting tests extension for this param enabled

cpp/oneapi/dal/algo/covariance/backend/gpu/misc.hpp
Alexandr-Solovev: here its necessary due to the function returns event+ndarray

icfaust:
rerun of private CI based off last commit (due to infrastructure issues): http://intel-ci.intel.com/eef7bc59-abd5-f1d9-8842-a4bf010d0e2e

cpp/oneapi/dal/algo/covariance/backend/gpu/misc.hpp
Vika-F: +1 to Ian's question. `ndarray::zeros` also returns ndarray+event.
It seems you can just do `return pr::ndarray<Float, 1>::zeros(...);`.

cpp/oneapi/dal/backend/primitives/stat/cov_dpc.cpp
```diff
@@ -79,13 +80,41 @@ inline sycl::event compute_covariance(sycl::queue& q,
...
-                cov_ptr[gi] *= multiplier;
-            }
+            cov_ptr[gi] -= inv_n * sums_ptr[id.get_id(0)] * sums_ptr[id.get_id(1)];
+            cov_ptr[gi] *= multiplier;
+        });
+    });
+}
+
+template <typename Float>
+inline sycl::event compute_covariance_centered(sycl::queue& q,
```
Vika-F, line 91: Please add the description. It is not obvious that on input `cov` is a matrix of sum of squares and on output it is the covariance matrix.
The same applies to `compute_covariance`.

cpp/oneapi/dal/backend/primitives/stat/cov_dpc.cpp
```diff
@@ -79,13 +80,41 @@ inline sycl::event compute_covariance(sycl::queue& q,
...
+
+    Float* cov_ptr = cov.get_mutable_data();
+
+    return q.submit([&](sycl::handler& cgh) {
+        const auto range = make_range_2d(p, p);
+
+        cgh.depends_on(deps);
+        cgh.parallel_for(range, [=](sycl::item<2> id) {
+            const std::int64_t gi = id.get_linear_id();
+            cov_ptr[gi] *= multiplier;
```
Vika-F, line 117: Wouldn't it be better to have 1d range of size p * p instead?

examples/oneapi/cpp/source/covariance/cov_dense_biased_batch.cpp
```diff
@@ -27,7 +27,8 @@ int main(int argc, char const *argv[]) {
     const auto input = dal::read<dal::table>(dal::csv::data_source{ input_file_name });
     auto cov_desc = dal::covariance::descriptor{}
                         .set_result_options(dal::covariance::result_options::cov_matrix)
-                        .set_bias(true);
+                        .set_bias(true)
+                        .set_assume_centered(true);
```
Vika-F, line 31: Shouldn't it be `false`? I think `assume_centered` can be true only when the input data is centered.

cpp/oneapi/dal/backend/primitives/stat/cov_dpc.cpp
Alexandr-Solovev: It makes sense! Thanks!

PR #2662, opened 2024-02-12
Fix csr_accessor crash when moving data from USM or host memory to host memory

Vika-F:
Backend function `copy_all2all` that unifies the data copying across USM and non-USM allocated memory was working incorrectly:
The wrong number of bytes were copied using `copy` function in the case when both source and destination data buffers are _not_ device friendly.
Copying the increased number of bytes caused out-of-bound memory access which led to crash.

`copy` was replaces with `memcpy` that copies exactly the requested number of bytes.
`copy` was multiplying the requested number of bytes to the type size which produces the wrong size of the copied array.

 csr_accessor test was also extended to cover the cases that were failing previously.


inteldimitrius (approved):
LGTM, let's wait CI checks and private CI

PR #2669, opened 2024-02-20
[enh] Add avx2 support in finiteness_checker

icfaust:
# Description
Duplicate AVX512 functions for AVX2 by switching certain numbers to half size and changing instructions from 512 to 256 bit width.  Due to the hardcoded nature of the functions, it is not easily templated out without performance loss. This implementation should improve sklearnex performance on standard benchmarks.

- Add avx2 finite sum check
- Add avx2 finiteness per element check
- Add avx2 SOA supports

Tasks
- [x] Implement AVX2
- [ ] Get it to compile
- [ ] Green CI
- [ ] Run sklearnex Benchmarks

### DIFF
--- a/cpp/daal/src/data_management/finiteness_checker.cpp
+++ b/cpp/daal/src/data_management/finiteness_checker.cpp
@@ -168,8 +168,6 @@ bool checkFinitenessSOA(NumericTable & table, bool allowNaN, services::Status & 
 const size_t BLOCK_SIZE       = 8192;
 const size_t THREADING_BORDER = 262144;
 
-////////// AVX512 Implementation ///////////
-
 template <typename DataType>
 DataType sumWithAVX512(size_t n, const DataType * dataPtr)
 {
@@ -445,7 +443,7 @@ double computeSumSOA<avx2>(NumericTable & table, bool & sumIsFinite, services::S
 }
 
 services::Status checkFinitenessInBlocks512(const float ** dataPtrs, bool inParallel, size_t nTotalBlocks, size_t nBlocksPerPtr, size_t nPerBlock,
-                                         size_t nSurplus, bool allowNaN, bool & finiteness)
+                                            size_t nSurplus, bool allowNaN, bool & finiteness)
 {
     services::Status s;
     const size_t nPerInstr = 16;
@@ -501,7 +499,7 @@ services::Status checkFinitenessInBlocks512(const float ** dataPtrs, bool inPara
 }
 
 services::Status checkFinitenessInBlocks512(const double ** dataPtrs, bool inParallel, size_t nTotalBlocks, size_t nBlocksPerPtr, size_t nPerBlock,
-                                         size_t nSurplus, bool allowNaN, bool & finiteness)
+                                            size_t nSurplus, bool allowNaN, bool & finiteness)
 {
     services::Status s;
     const size_t nPerInstr = 8;
@@ -557,7 +555,7 @@ services::Status checkFinitenessInBlocks512(const double ** dataPtrs, bool inPar
 }
 
 services::Status checkFinitenessInBlocks256(const float ** dataPtrs, bool inParallel, size_t nTotalBlocks, size_t nBlocksPerPtr, size_t nPerBlock,
-                                         size_t nSurplus, bool allowNaN, bool & finiteness)
+                                            size_t nSurplus, bool allowNaN, bool & finiteness)
 {
     services::Status s;
     const size_t nPerInstr = 8;
@@ -613,7 +611,7 @@ services::Status checkFinitenessInBlocks256(const float ** dataPtrs, bool inPara
 }
 
 services::Status checkFinitenessInBlocks256(const double ** dataPtrs, bool inParallel, size_t nTotalBlocks, size_t nBlocksPerPtr, size_t nPerBlock,
-                                         size_t nSurplus, bool allowNaN, bool & finiteness)
+                                            size_t nSurplus, bool allowNaN, bool & finiteness)
 {
     services::Status s;
     const size_t nPerInstr = 4;
--- a/cpp/daal/src/data_management/finiteness_checker.cpp
+++ b/cpp/daal/src/data_management/finiteness_checker.cpp
@@ -168,6 +168,8 @@ bool checkFinitenessSOA(NumericTable & table, bool allowNaN, services::Status & 
 const size_t BLOCK_SIZE       = 8192;
 const size_t THREADING_BORDER = 262144;
 
+////////// AVX512 Implementation ///////////
+
 template <typename DataType>
 DataType sumWithAVX512(size_t n, const DataType * dataPtr)
 {
@@ -186,6 +188,30 @@ DataType sumWithAVX512(size_t n, const DataType * dataPtr)
         __m512d * ptr512 = (__m512d *)dataPtr;
         for (size_t i = 0; i < n / nPerInstr; i++) sums = _mm512_add_pd(sums, ptr512[i]);
         sum = _mm512_reduce_add_pd(sums);
+    }
+    for (size_t i = (n / nPerInstr) * nPerInstr; i < n; ++i) sum += dataPtr[i];
+
+    return sum;
+}
+
+template <typename DataType>
+DataType sumWithAVX2(size_t n, const DataType * dataPtr)
+{
+    const size_t nPerInstr = 32 / sizeof(DataType);
+    DataType sum;
+    if (sizeof(DataType) == 4)
+    {
+        __m256 sums     = _mm256_set1_ps(0);
+        __m256 * ptr256 = (__m256 *)dataPtr;
+        for (size_t i = 0; i < n / nPerInstr; i++) sums = _mm256_add_ps(sums, ptr256[i]);
+        sum = _mm256_reduce_add_ps(sums);
+    }
+    else
+    {
+        __m256d sums     = _mm256_set1_pd(0);
+        __m256d * ptr256 = (__m256d *)dataPtr;
+        for (size_t i = 0; i < n / nPerInstr; i++) sums = _mm256_add_pd(sums, ptr256[i]);
+        sum = _mm256_reduce_add_pd(sums);
     }
     for (size_t i = (n / nPerInstr) * nPerInstr; i < n; ++i) sum += dataPtr[i];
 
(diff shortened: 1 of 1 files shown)


cpp/daal/src/data_management/finiteness_checker.cpp
```diff
@@ -192,6 +192,30 @@ DataType sumWithAVX512(size_t n, const DataType * dataPtr)
     return sum;
 }
 
+template <typename DataType>
+DataType sumWithAVX2(size_t n, const DataType * dataPtr)
+{
+    const size_t nPerInstr = 32 / sizeof(DataType);
+    DataType sum;
+    if (sizeof(DataType) == 4)
```
icfaust, line 200: Am I missing something about how the compiler works, or why is this branching necessary??

icfaust: Now I understand

icfaust:
test fail related to rbf kernel, which doesn't use this code

Alexandr-Solovev (commented):
In general it looks good to me, but, it's better to change the order of the new functions. Like firstly avx2 then avx512. Also, please use onedal-benchmarks with the kernel profiler as well as intelex benchmarks to check what kernels have been improved. I would like to say that you will get similar performance benefits in both.

cpp/daal/src/data_management/finiteness_checker.cpp
```diff
@@ -76,6 +57,7 @@ bool valuesAreNotFinite(const double * dataPtr, size_t n, bool allowNaN)
 template <typename DataType, daal::CpuType cpu>
 DataType computeSum(size_t nDataPtrs, size_t nElementsPerPtr, const DataType ** dataPtrs)
 {
+    std::cout << "standard sum \n";
```
samir-nasibli, line 60: remove some debug prints

cpp/daal/src/data_management/finiteness_checker_cpu.cpp
```diff
@@ -0,0 +1,917 @@
...
+
+            __mmask16 expAreOnes   = _mm512_cmpeq_epi32_mask(exp512Mask, expBits);
+            __mmask16 fracAreZeros = _mm512_cmpeq_epi32_mask(zero512, fracBits);
+
+            // "values aren't finite" = "exponent bits are ones" AND ( "fraction bits are zeros" OR NOT "NaN is allowed" )
+            __mmask16 orMask    = _kor_mask16(fracAreZeros, notAllowNaNMask);
+            __mmask16 finalMask = _kand_mask16(expAreOnes, orMask);
+            endMask             = _kor_mask16(endMask, finalMask);
+        }
+        if (_cvtmask16_u32(endMask) != 0) notFinitePtr[iBlock] = true;
```
icfaust, line 686: Note to reviewers: I have changed these lines to remove branching in the for loop. This was otherwise hidden due to the movement of the file, but otherwise the AVX512 portion is unchanged.

icfaust:
Private CI just shows an unrelated issue with LibLinear convergence issues, which shouldn't be touched by this code / is likely sporadic.

icfaust:
private CI run with intel/scikit-learn-intelex#1759 http://intel-ci.intel.com/eedfc7b0-6419-f133-b20a-a4bf010d0e2e

icfaust:
wrote a special sklearnex version which will print warnings when an inf or nan is observed, which should show up in pytest for sklearn. https://github.com/intel/scikit-learn-intelex/compare/main...icfaust:scikit-learn-intelex:test/warning_finite?expand=1

I have run a special private CI run with this branch to see if it shows up at all in sklearn conformance tests, to see how much sklearn tests _assert_all_finite, specifically to see if it is actually activated: http://intel-ci.intel.com/eee83e31-ce3d-f14a-985d-a4bf010d0e2e

icfaust:
> wrote a special sklearnex version which will print warnings when an inf or nan is observed, which should show up in pytest for sklearn. https://github.com/intel/scikit-learn-intelex/compare/main...icfaust:scikit-learn-intelex:test/warning_finite?expand=1
> 
> I have run a special private CI run with this branch to see if it shows up at all in sklearn conformance tests, to see how much sklearn tests _assert_all_finite, specifically to see if it is actually activated: http://intel-ci.intel.com/eeea9220-3038-f1df-9c06-a4bf010d0e2e (running against onedal-src/oneDAL/main)

So after activating a warning for when an inf or nan is spotted and by running all sklearn conformance testing, the only time an inf or nan occurs in the sklearn testing is here: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/tests/test_validation.py#L981 Therefore intel/scikit-learn-intelex#1759 is sadly necessary

cpp/daal/src/data_management/finiteness_checker.h
```diff
@@ -0,0 +1,52 @@
+/** file finiteness_checker.h */
+/*******************************************************************************
+* Copyright 2024 Intel Corporation
```
Vika-F, line 3: Please align copyrights with the new UXL requirements. 
```suggestion
* Copyright contributors to the oneDAL project
```

cpp/daal/src/data_management/finiteness_checker.h
```diff
@@ -0,0 +1,52 @@
...
+{
+using namespace daal::internal;
+
+typedef daal::data_management::NumericTable::StorageLayout NTLayout;
+
+const uint32_t floatExpMask  = 0x7f800000u;
+const uint32_t floatFracMask = 0x007fffffu;
+const uint32_t floatZeroBits = 0x00000000u;
+
+const uint64_t doubleExpMask  = 0x7ff0000000000000uLL;
```
Vika-F, line 37: `uLL` at the end looks strange. Please choose either 'ull' or 'ULL'.

cpp/daal/src/data_management/finiteness_checker_cpu.cpp
```diff
@@ -0,0 +1,917 @@
...
+        *((uint64_t *)(&inf)) = doubleExpMask;
+    return inf;
+}
+
+    #if defined(__AVX2__) and !defined(__AVX512F__)
+
+template <typename DataType>
+DataType sumWithAVX2(size_t n, const DataType * dataPtr)
+{
+    const size_t nPerInstr = 32 / sizeof(DataType);
```
Vika-F, line 147: Those calculations are not obvious.
It's better to have:
```suggestion
    constexpr size_t avx2RegisterLength = 256;
    constexpr size_t numberOfBitsInByte = 8;
    constexpr size_t nPerInstr = avx2RegisterLength / (numberOfBitsInByte * sizeof(DataType));
```

cpp/daal/src/data_management/finiteness_checker_cpu.cpp
```diff
@@ -0,0 +1,917 @@
...
+        size_t ptrIdx        = iBlock / nBlocksPerPtr;
+        size_t blockIdxInPtr = iBlock - nBlocksPerPtr * ptrIdx;
+        size_t start         = blockIdxInPtr * nPerBlock;
+        size_t end           = blockIdxInPtr == nBlocksPerPtr - 1 ? start + nPerBlock + nSurplus : start + nPerBlock;
+
+        pSums[iBlock] = sumWithAVX2<DataType>(end - start, dataPtrs[ptrIdx] + start);
+    });
+
+    return sumWithAVX2<DataType>(nTotalBlocks, pSums);
+}
```
Vika-F, line 210: This code is almost 100% copy-paste from AVX512 code.
I think it can be unified if sumWithXXX and computeSumXXXImpl functions would have a `cpu` template parameter.

cpp/daal/src/data_management/finiteness_checker_cpu.cpp
```diff
@@ -0,0 +1,917 @@
...
+    return computeSumAVX2Impl<float>(nDataPtrs, nElementsPerPtr, dataPtrs);
+}
+
+template <>
+double computeSum<double, avx2>(size_t nDataPtrs, size_t nElementsPerPtr, const double ** dataPtrs)
+{
+    return computeSumAVX2Impl<double>(nDataPtrs, nElementsPerPtr, dataPtrs);
+}
+
+double computeSumSOAAVX2Impl(NumericTable & table, bool & sumIsFinite, services::Status & st)
```
Vika-F, line 224: This is also a big copy-paste. I think there should be ways to avoid it.

icfaust:
This will likely pass CI, but performance benchmarks are necessary due to the underlying changes in the CPU function dispatching.

icfaust:
Things required before re-review: a privateCI run for checking avx512, and oneDAL performance benchmarks of changes to function dispatching.

icfaust:
Run with an avx512 build: http://intel-ci.intel.com/ef007c41-cb1f-f115-9514-a4bf010d0e2e failures due to un-related GPU issues.

icfaust:
private CI failures due to unrelated GPU/dpc timeouts

icfaust:
private CI run with last sklearnex master (includes _assert_all_finite tests coming from intel/scikit-learn-intelex#1759) http://intel-ci.intel.com/ef012d7e-a408-f166-adc8-a4bf010d0e2e

cpp/daal/src/data_management/finiteness_checker_avx2_impl.i
```diff
@@ -0,0 +1,234 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#ifndef __FINITENESS_CHECKER_AVX2_IMPL_I__
+#define __FINITENESS_CHECKER_AVX2_IMPL_I__
+
+template <>
+float sumWithAVX<float, avx2>(size_t n, const float * dataPtr)
```
Vika-F, line 22: I think further there might be non-AVX implementation of those functions. We already have 3 SIMD variants supported: SSE, AVX and Arm's SVE. There might be more in the future.

```suggestion
float sumWithSIMD<float, avx2>(size_t n, const float * dataPtr)
```

Vika-F, line 22: Please add the description. Something like this.
```suggestion
/*
// Computes sum of the elements of input array of type `float` with AVX2 instructions.
*/
float sumWithAVX<float, avx2>(size_t n, const float * dataPtr)
```

cpp/daal/src/data_management/finiteness_checker_avx2_impl.i
```diff
@@ -0,0 +1,234 @@
...
+template <>
+services::Status checkFinitenessInBlocks<avx2>(const float ** dataPtrs, bool inParallel, size_t nTotalBlocks, size_t nBlocksPerPtr, size_t nPerBlock,
+                                               size_t nSurplus, bool allowNaN, bool & finiteness)
+{
+    services::Status s;
+    constexpr size_t avx2RegisterLength = 256;
+    constexpr size_t numberOfBitsInByte = 8;
+    constexpr size_t nPerInstr          = avx2RegisterLength / (numberOfBitsInByte * sizeof(float));
+    services::internal::TArray<bool, avx2> notFiniteArr(nTotalBlocks);
+    bool * notFinitePtr = notFiniteArr.get();
```
Vika-F, line 103: This is just a FYI. No action is needed.

It might be better (in terms of performance) to have those kind of arrays as a 'thread local storage' (see daal::tls).

Having them as a regular arrays might lead to cache coherency issues when i-th thread updates i-th element of notFinitePtr array, and (i+1)-th thread updates (i+1)-th element of the same array. In this case both threads might need to synchronize the information in their caches as both i-th and (i+1)-th elements of the array might reside in both caches of i-th and (i+1)-th threads and can become unsynchronized after the update.

cpp/daal/src/data_management/finiteness_checker_cpu.cpp
```diff
@@ -0,0 +1,320 @@
...
+    const size_t nCols                          = table.getNumberOfColumns();
+    NumericTableDictionaryPtr tableFeaturesDict = table.getDictionarySharedPtr();
+
+    daal::TlsMem<bool, cpu, services::internal::ScalableCalloc<bool, cpu> > tlsNotFinite(1);
+
+    daal::threader_for_break(nCols, nCols, [&](size_t i, bool & needBreak) {
+        bool * localNotFinite = tlsNotFinite.local();
+        DAAL_CHECK_MALLOC_THR(localNotFinite);
+
+        switch ((*tableFeaturesDict)[i].getIndexType())
```
Vika-F, line 199: Don't we need something that checks for integer overflow for integer features here? How is it done in array API?

cpp/daal/src/data_management/finiteness_checker_cpu.cpp
```diff
@@ -0,0 +1,320 @@
...
+    {
+        sum = computeSumSOA<cpu>(table, sumIsFinite, s);
+    }
+    else
+    {
+        ReadRows<DataType, cpu> dataBlock(table, 0, nRows);
+        DAAL_CHECK_BLOCK_STATUS(dataBlock);
+        const DataType * dataPtr = dataBlock.get();
+
+        sum = computeSum<DataType, cpu>(1, nElements, &dataPtr);
```
Vika-F, line 283: Another FYI. No action is required.

It is better in terms of both performance and memory consumption to move ReadRows inside the parallel for that happens in computeSum.
Because ReadRows in some cases might copy the whole table and convert it -> +1 pass through the memory.
When ReadRows is made by blocks and is combined with the computational part it is usually more efficient.

cpp/daal/src/data_management/finiteness_checker_cpu.cpp
```diff
@@ -0,0 +1,320 @@
...
+    #if (__CPUID__(DAAL_CPU) == __avx2__)
+
+        #include "finiteness_checker_avx2_impl.i"
+
+    #endif // __CPUID__(DAAL_CPU) == __avx2__
+
+#endif
+
+template <typename DataType, daal::CpuType cpu>
+services::Status allValuesAreFiniteImpl(NumericTable & table, bool allowNaN, bool * finiteness)
```
Vika-F, line 260: Please add the description.

Vika-F (commented):
Thank you for the restructuring effort.
The code became much more understandable.

My final comment is that 'AVX' needs to be replaced with something else like 'SIMD'. Otherwise the code is good to go.

icfaust:
Rerun due to CI timeouts: http://intel-ci.intel.com/ef01f546-5586-f1d1-863c-a4bf010d0e2e

Vika-F (approved):
LGTM

PR #2672, opened 2024-02-21
Makefile refactoring to factor out common build code

keeranroth:
# Description
Following on from pull request https://github.com/oneapi-src/oneDAL/pull/2614 to add Arm as a supported CPU platform, we update the Makefile structure to make it easier to add other platforms in the future.

The motivation behind this change is to have a minimal set of variables to change when adding a new platform. We also look to re-use makefile code as much as possible in the existing build system, so that it's less likely for builds on different platforms to start to diverge.

The changes to the makefiles require no changes to the build commands.

Most of the changes are copy and paste, with some changes being made to allow the evaluation of make commands to be deferred until all variables required are in scope. No new functionality is added, and no functionality is removed.

- Add new makefiles to keep platform, arch and uarch specific make variables
- Add new makefiles to keep compiler and backend specific make variables


dev/make/compiler_definitions/clang.ref.32e.mk
```diff
@@ -0,0 +1,25 @@
+# file: clang.ref.32e.mk
+#===============================================================================
+# Copyright 2024 UXL Foundation
```
napetrov, line 3:
```suggestion
# Copyright contributors to the oneDAL project
```

dev/make/compiler_definitions/gnu.32e.mk
```diff
@@ -1,5 +1,5 @@
 #===============================================================================
-# Copyright 2023 Intel Corporation
+# Copyright 2024 UXL Foundation
```
napetrov, line 2:
```suggestion
# Copyright contributors to the oneDAL project
```

dev/make/compiler_definitions/gnu.mk
```diff
@@ -1,5 +1,5 @@
 #===============================================================================
-# Copyright 2023 Intel Corporation
+# Copyright 2024 UXL Foundation
```
napetrov, line 2: # Copyright contributors to the oneDAL project

napetrov, line 2:
```suggestion
# Copyright contributors to the oneDAL project
```

dev/make/function_definitions/32e.mk
```diff
@@ -0,0 +1,108 @@
+#===============================================================================
```
napetrov, line 1: Should we rename this to be aligned with ARM? 

This is quite a legacy name as one of variants between 32bit, em64t and itanium64 variants.  

Any ideas on what this would be?  basically this is generic x86 architecture

dev/make/function_definitions/mac32e.mk
```diff
@@ -0,0 +1,34 @@
...
+
+# Used as $(eval $(call set_daal_rt_deps))
+define set_daal_rt_deps
+  $$(eval daaldep.mac32e.rt.thr := -L$$(RELEASEDIR.tbb.soia) -ltbb -ltbbmalloc \
+          $$(daaldep.mac32e.rt.$$(COMPILER)))
+  $$(eval daaldep.mac32e.rt.seq := $$(daaldep.mac32e.rt.$$(COMPILER)))
+  $$(eval daaldep.mac32e.threxport := export_mac.def)
+
+  $$(eval daaldep.mac.threxport.create = grep -v -E '^(EXPORTS|;|$$$$$$$$)' $$$$< $$$$(USECPUS.out.grep.filter) | sed -e 's/^/-u /')
+endef
```
napetrov, line 34:
```suggestion
endef

```

dev/make/function_definitions/32e.mk
rakshithgb-fujitsu: how about x86_64.mk?

keeranroth: Agreed that this could be a better name, and x86_64 sounds decent. However, changing this requires updates outside of the make setup (there are ci scripts that depend on this value), and I think it would be better done as another merge request to change the name. What do you think? I'm happy to put something up for that after this.

napetrov: yes, x86_64 looks decent. - the only thing to check is _ handling and this would depend on how wide we would like to have this implemented, but in longerterm worth having better name.

yes, it's fine to keep this separate as renaming would require updates for internal CI. @homksei @ethanglaser  - this is something that we can look on and rename legacy 32e naming to have better match/separation in future.

keeranroth: https://github.com/oneapi-src/oneDAL/pull/2674 as a draft merge request, for what this would look like on top of this commit.

Is there anything else that needs changing on my end for this merge request? I can see that all the public CI passes, but the Intel CI is failing

dev/make/compiler_definitions/gnu.32e.mk
```diff
@@ -1,5 +1,5 @@
 #===============================================================================
-# Copyright 2023 Intel Corporation
+# Copyright 2024 contributors to the oneDAL project
```
napetrov, line 2:
```suggestion
# Copyright contributors to the oneDAL project
```

Current copyright check expects no year defined and i think this is what was suggested by UXL. 
same true for other files

keeranroth:
Looks like another error. Are there some files that I missed the license text on, or is there something else?

ethanglaser:
> Looks like another error. Are there some files that I missed the license text on, or is there something else?

All (6) new files in function_definitions/ directory are failing copyright check. Will look into why this may be happening. Other copyright checks are passing (including files in compiler_definitions/)

keeranroth:
Hi @ethanglaser thanks for having a look at those. Is there anything that I can do on my side? 

Also, is it worth putting an updated copyright check into upstream as well? It would make it easier for me next time round to check this locally (I usually defer to CI checks before pushing a merge request)

ethanglaser:
Hi @keeranroth, it appear that the `Copyright 2024 UXL Foundation` is triggering the internal CI fail

ethanglaser:
Good idea about adding copyright check to public CI - we'll discuss as a team and keep you updated.

Apologies on if I missed discussion on this but did we agree on adding `Copyright 2024 UXL Foundation` to top of new files? If so we should be able to accommodate this internally, otherwise can remove it and checks should pass.

keeranroth:
Ah, my bad on the copyright checks. @napetrov did say that the agreed style doesn't have a year in the statement, so I've updated those

makefile
```diff
@@ -610,25 +511,19 @@ $(CORE.tmpdir_y)/inc_y_folders.txt: makefile.lst | $(CORE.tmpdir_y)/. $(CORE.inc
...
+$(eval $(call subst_arch_cpu_in_var,template_source_cpp))
 
-$1: $(template_source_cpp) ; $(value C.COMPILE)
+$2: $(template_source_cpp) ; $(value C.COMPILE)
 endef
 
-$(foreach a,$(CORE.objs_a),$(eval $(call .compile.template.ay,$a,$(CORE.tmpdir_a))))
-$(foreach a,$(CORE.objs_y),$(eval $(call .compile.template.ay,$a,$(CORE.tmpdir_y))))
+$(foreach a,$(CORE.objs_a),$(eval $(call .compile.template.ay,$(ARCH),$a)))
+$(foreach a,$(CORE.objs_y),$(eval $(call .compile.template.ay,$(ARCH),$a)))
```
ethanglaser, line 526: is ARCH needed to be passed? I don't see use of $1 in .compile.template.ay

dev/make/compiler_definitions/clang.32e.mk
```diff
@@ -16,17 +16,15 @@
 #===============================================================================
 
 #++
-#  Clang definitions for makefile
+#  Clang definitions for makefile.
+#  This file contains definitions common to clang on a 32e (intel64) platform.
+#  It should only be included from files which have more specializations (e.g.
+#  clang.mkl.32e.mk
```
ethanglaser, line 22:
```suggestion
#  clang.mkl.32e.mk)
```

dev/make/compiler_definitions/clang.mk
```diff
@@ -17,34 +17,19 @@
 
 #++
 #  Clang definitions for makefile
+#  This file contains definitions common to clang on all platforms.
+#  It should only be included from files which have more specializations (e.g.
+#  clang.32e.mk
```
ethanglaser, line 22:
```suggestion
#  clang.32e.mk)
```

ethanglaser (approved):
Really nice work here, overall just minor questions/comments. Thanks for improving makefile organization! And CI is in good shape

keeranroth (commented):
Thanks for the reviews @ethanglaser , and sorry for not having seen this before. Must have missed the emails with your comments in.

I've left one point unaddressed regarding single vs double brackets in a bash script. Whilst I also have a preference for the double brackets, I changed this to match the existing style in the script. Anything else, let me know

deploy/local/vars_lnx.sh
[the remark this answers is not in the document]
ethanglaser: looks good

dev/make/function_definitions/lnx32e.mk
```diff
@@ -0,0 +1,37 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+BACKEND_CONFIG ?= mkl
+ARCH = 32e
+ARCH_DIR_ONEDAL = intel64
```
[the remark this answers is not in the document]
ethanglaser: yeah good point, this works for me

PR #2675, opened 2024-02-27
Fix logloss test failure

avolkov-intel:
# Description
Their was a problem with computation of gth logloss because the probability value was too close to 0 or 1 so log(prob) produced a nan value. I added probability clipping to avoid problems with logloss computation (the same approach is used in original algorithm)

### DIFF
--- a/cpp/oneapi/dal/backend/primitives/objective_function/test/fixture.hpp
+++ b/cpp/oneapi/dal/backend/primitives/objective_function/test/fixture.hpp
@@ -291,6 +291,18 @@ public:
         }
     }
 
+    float_t clip_prob(float_t prob) {
+        constexpr float_t bottom = sizeof(float_t) > 4 ? 1e-15 : 1e-7;
+        constexpr float_t top = float_t(1.0) - bottom;
+        if (prob < bottom) {
+            prob = bottom;
+        }
+        if (prob > top) {
+            prob = top;
+        }
+        return prob;
+    }
+
     float_t test_predictions_and_logloss(const ndview<float_t, 2>& data_host,
                                          const ndview<float_t, 1>& params_host,
                                          const ndview<std::int32_t, 1>& labels_host,
@@ -313,7 +325,7 @@ public:
             if (fit_intercept) {
                 pred += params_host.at(0);
             }
-            float_t prob = 1 / (1 + std::exp(-pred));
+            float_t prob = clip_prob(float_t(1.0) / (1 + std::exp(-pred)));
             logloss -=
                 labels_host.at(i) * std::log(prob) + (1 - labels_host.at(i)) * std::log(1 - prob);
             float_t out_val = probabilities.at(i);


cpp/oneapi/dal/backend/primitives/objective_function/test/fixture.hpp
```diff
@@ -291,6 +291,18 @@ class logloss_test : public te::float_algo_fixture<Param> {
         }
     }
 
+    float_t clip_prob(float_t prob) {
```
ahuber21, line 294: Could you add a small comment to explain why you pick `1e-15` and `1e-7` and why we can't use `std::numeric_limits` ?

ahuber21 (changes requested):

ahuber21:
Two more general comments on this file

* `using oneapi::dal::detail::operator<<;` doesn't seem to be used / have an effect. Can it be removed?
* I don't like the `check_val()` helper function. It shadows where the original error happened. I think it should be a macro instead. (Or am I running the test incorrectly?)

```
$ /localdisk2/mkl/huberand/oneDAL/bazel-bin/cpp/oneapi/dal/backend/primitives/objective_function/test_logloss_dpc

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test_logloss_dpc is a Catch v2.13.10 host application.
Run with -? for options

-------------------------------------------------------------------------------
test random input - double without L1 - no fit intercept - double
-------------------------------------------------------------------------------
cpp/oneapi/dal/backend/primitives/objective_function/test/logloss_dpc.cpp:53
...............................................................................

cpp/oneapi/dal/backend/primitives/objective_function/test/fixture.hpp:50: FAILED:
  REQUIRE( abs(real - expected) < atol )
with expansion:
  nan < 0.000001
```

This only points into the `check_val()` function, and does not have a backtrace from which line it came. So I cannot tell which of the many tests using `check_val()` actually failed. I really don't like that.

cpp/oneapi/dal/backend/primitives/objective_function/test/fixture.hpp
avolkov-intel: In fact I took these numbers from scikit-learn implementation

avolkov-intel:
> Two more general comments on this file
> 
> * `using oneapi::dal::detail::operator<<;` doesn't seem to be used / have an effect. Can it be removed?
> * I don't like the `check_val()` helper function. It shadows where the original error happened. I think it should be a macro instead. (Or am I reading the log incorrectly?)
> 
> ```
> $ ./bazel-bin/cpp/oneapi/dal/backend/primitives/objective_function/test_logloss_dpc
> 
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> test_logloss_dpc is a Catch v2.13.10 host application.
> Run with -? for options
> 
> -------------------------------------------------------------------------------
> test random input - double without L1 - no fit intercept - double
> -------------------------------------------------------------------------------
> cpp/oneapi/dal/backend/primitives/objective_function/test/logloss_dpc.cpp:53
> ...............................................................................
> 
> cpp/oneapi/dal/backend/primitives/objective_function/test/fixture.hpp:50: FAILED:
>   REQUIRE( abs(real - expected) < atol )
> with expansion:
>   nan < 0.000001
> ```
> 
> This only points into the `check_val()` function, and does not have a backtrace from which line it came. So I cannot tell which of the many tests using `check_val()` actually failed. I really don't like that.



> Two more general comments on this file
> 
> * `using oneapi::dal::detail::operator<<;` doesn't seem to be used / have an effect. Can it be removed?
> * I don't like the `check_val()` helper function. It shadows where the original error happened. I think it should be a macro instead. (Or am I reading the log incorrectly?)
> 
> ```
> $ ./bazel-bin/cpp/oneapi/dal/backend/primitives/objective_function/test_logloss_dpc
> 
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> test_logloss_dpc is a Catch v2.13.10 host application.
> Run with -? for options
> 
> -------------------------------------------------------------------------------
> test random input - double without L1 - no fit intercept - double
> -------------------------------------------------------------------------------
> cpp/oneapi/dal/backend/primitives/objective_function/test/logloss_dpc.cpp:53
> ...............................................................................
> 
> cpp/oneapi/dal/backend/primitives/objective_function/test/fixture.hpp:50: FAILED:
>   REQUIRE( abs(real - expected) < atol )
> with expansion:
>   nan < 0.000001
> ```
> 
> This only points into the `check_val()` function, and does not have a backtrace from which line it came. So I cannot tell which of the many tests using `check_val()` actually failed. I really don't like that.

I agree with this point. Actually I'm adding some new functionality and updating the tests in a separate branch, I think I will change it to MACROS there.

ahuber21 (approved):
LGTM

cpp/oneapi/dal/backend/primitives/objective_function/test/fixture.hpp
Vika-F, line 294: Please add the comment what this function does and for what.
Something like "Fit the input probability value into the interval [epsilon, 1.0 - epsilon]. This is needed because too small or too close to 1.0 probability values can lead to infinity results in logloss computations leading to test failures".
And I agree to Andreas's comment: need a comment about the choice of 'epsilon' values. Even if they were taken from sklearn, ok. Please write it in the comment.
Also it would be good to give them names, i.e. assign to some constexpr variables. Having magic numbers in the code is not recommended by our coding guidelines.

ahuber21: Sorry, should have doubled down on this. We need to explanation in code, not just in GH :) Agree with @Vika-F that @avolkov-intel should add this in a follow-up PR. Thanks!

PR #2680, opened 2024-03-01
chore(deps): update GPU FPK

homksei:

PR #2667, opened 2024-02-16
[CI] Update dpcpp compiler for public CI

icfaust:
# Description
Currently set to 2023.2.1, upgrade compiler to 2024.2.0 

Changes proposed in this pull request:
- upgrade GPG keys
- upgrade dpcpp compiler

PR #2681, opened 2024-03-04
Update DPC++ compiler and MKL GPU FPK microlibs versions

Vika-F:
DPC++ compiler versions are updated to 2024.0
MKL FPK GPU microlibs are updated to 2024-02-22

### DIFF
--- a/.ci/env/apt.sh
+++ b/.ci/env/apt.sh
@@ -22,18 +22,17 @@ function update {
 }
 
 function add_repo {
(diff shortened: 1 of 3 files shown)


icfaust:
#2680 #2667 Which of these three should we merge?

icfaust:
I think this may be the better PR, it looks like the problem is that that apt doesn't contain the necessary version to support the MKL GPU FPK. What to do next?

PR #2700, opened 2024-03-27
Minor refactor of the top-level CI build script

keeranroth:
In preparation for adding some more targets, making a few cosmetic adjustments to the top level build script

- Naming of flags consistently uses '-' instead of '_'
- Added help text
- The OS and ARCH variables are sub-fields of the PLAT, so just pass the PLAT
- Set pipefail and exit on failure, which means that failures are not silently ignored. This is only a problem (currently) locally, but it helps a lot in refactoring build scripts.
- Paths are made relative to the oneDAL root directory, rather than the current working directory.

### DIFF
--- a/.ci/scripts/build.sh
+++ b/.ci/scripts/build.sh
@@ -153,12 +153,12 @@ fi
 echo "Calling make"
 echo $CXX
 echo $CC
-echo make ${target:-daal} ${make_op} \
+echo make ${target:-onedal_c} ${make_op} \
     COMPILER=${compiler} \
-    REQCPU=\"${optimizations}\" \
-    BACKEND_CONFIG=\"${backend_config}\" \
+    REQCPU="${optimizations}" \
+    BACKEND_CONFIG="${backend_config}" \
     PLAT=${PLAT}
-make ${target:-daal} ${make_op} \
+make ${target:-onedal_c} ${make_op} \
     COMPILER=${compiler} \
     REQCPU="${optimizations}" \
     BACKEND_CONFIG="${backend_config}" \
--- a/.ci/scripts/build.sh
+++ b/.ci/scripts/build.sh
@@ -16,39 +16,22 @@
 # limitations under the License.
 #===============================================================================
 
-# Obtain platform, OS and arch details automatically
-PLATFORM=$(bash dev/make/identify_os.sh)
-OS=${PLATFORM::3}
-ARCH=${PLATFORM:3:3}
+set -eo pipefail
 
-# set default values for optimisation based on arch, these values can be overidden by passed arguments to the script.
-if [[ "${ARCH}" == "32e" ]]
-then
-optimizations=${optimizations:-avx2}
-elif [[ "${ARCH}" == "arm" ]]
-then
-optimizations=${optimizations:-sve}
-else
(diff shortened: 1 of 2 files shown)


.ci/scripts/build.sh
```diff
@@ -160,21 +153,13 @@ fi
...
-    PLAT=$PLAT
-make ${target:-daal_c} ${make_op} \
+    REQCPU=\"${optimizations}\" \
+    BACKEND_CONFIG=${backend_config} \
+    PLAT=${PLAT}
+make ${target:-daal} ${make_op} \
     COMPILER=${compiler} \
     REQCPU="${optimizations}" \
-    BACKEND_CONFIG="${backend_config}" \
+    BACKEND_CONFIG=${backend_config} \
```
ethanglaser, line 164: any reason for reversing order here?

.ci/scripts/build.sh
```diff
@@ -160,21 +153,13 @@ fi
...
     PLAT=${PLAT}
-err=$?
-
-if [ ${err} -ne 0 ]; then
-    status_ex="$(date +'%H:%M:%S') BUILD FAILED with errno ${err}"
-    GLOBAL_RETURN=${err}
-fi
-
-exit ${GLOBAL_RETURN}
```
ethanglaser, line 180: this is no longer necessary due to `set -eo pipefail`?

ethanglaser (approved):

.ci/scripts/build.sh
```diff
@@ -127,14 +120,14 @@ fi
 echo "Call env scripts"
 if [ "${backend_config}" == "mkl" ]; then
     echo "Sourcing MKL env"
-    $(pwd)/dev/download_micromkl.sh with_gpu=${with_gpu}
+    $ONEDAL_DIR/dev/download_micromkl.sh with_gpu=${with_gpu}
```
napetrov, line 123:
```suggestion
    ${ONEDAL_DIR}/dev/download_micromkl.sh with_gpu=${with_gpu}
```

.ci/scripts/build.sh
keeranroth: Yeah, that's right. The script will exit at any error (including in pipes) now. That requires a bit of care with things like grep, which return non-zero values on success, but is fine for the scripts as they stand.

.ci/scripts/build.sh
keeranroth: I can't see that the order was reversed, but I did lose some quotes. I put those back in, and set the default target to daal instead of daal_c

napetrov: Should we put onedal_c ?

keeranroth: Happy to change it. Also, no idea what the escaping was that I added, so reverted back

ethanglaser: Looks good to me

ethanglaser: Functionally daal_c vs. onedal_c should be fine either way since its just default value but in cases where script is called we pass target arg so that is used instead

keeranroth:
Is there any more to do on this merge request? I think that I've addressed all the comments

napetrov (approved):

PR #2708, opened 2024-04-04
FIX: kmeans_init bazel test enabling

ethanglaser:
- Update kmeans_init BUILD file so tests are ran from CI bazel test scope (in current main branch tests are NOT ran)
- Fix infinite loop kmeans_init++ that led to timeout in testing


Alexandr-Solovev (approved):
In general LGTM, but please clarify the dbi metric via onedal-benchmarks

ethanglaser:
> In general LGTM, but please clarify the dbi metric via onedal-benchmarks

This should have no impact on performance

PR #2711, opened 2024-04-08
Refactor OpenBLAS build script

keeranroth:
Getting ready to add a clang cross-compile path, we tidy up the OpenBLAS build script a little. LLVM is the preferred compiler toolchain for RISC-V at the moment, but the changes to come start with adding an aarch64 cross-compilation build.

As a bonus, because I recently learned about shellcheck, run this on the openblas.sh and build.sh scripts and tidy up warnings. The shellcheck changes are mostly to do with escaping.

Changes proposed in this pull request for openblas.sh CI build script:
- Add help text. Use the 'column' command to indent this properly
- Make the command-line arguments use a '-' instead of '_' to match build.sh
- Add an option to set the source directory of OpenBLAS. This makes it easier to run this script locally, and build out-of-source
- Make paths relative to the oneDAL root directory, rather than the current working directory. This enables out-of-source builds
- Change the install directory to at least contain the architecture being built for. Also update the build script to point to the correct custom install location. This just makes things a little neater when building for different platforms.

### DIFF
--- a/.ci/env/openblas.sh
+++ b/.ci/env/openblas.sh
@@ -16,6 +16,26 @@
 # limitations under the License.
 #===============================================================================
 
+set -eo pipefail
+
+SCRIPT_PATH=$(readlink -f "${BASH_SOURCE[0]}")
+SCRIPT_DIR=$(dirname "${SCRIPT_PATH}")
+ONEDAL_DIR=$(readlink -f "${SCRIPT_DIR}/../..")
+OPENBLAS_DEFAULT_SOURCE_DIR="${ONEDAL_DIR}/__work/openblas"
+
+show_help() {
+  echo "Usage: $0 [--help]"
+  column -t -s":" <<< '--help:Display this information
+--target <target>:The OpenBLAS target to build for
+--target-arch <arch>:The target architecture to build for. This is used when generating install directory names, and should match the target architecture for building oneDAL
+--compiler <path>:Path to the C compiler executable to use
+--host-compiler <path>:Path to the host compiler (suggests that that the target compiler given with '--compiler' is for another platform)
+--cflags <flags>:Any extra flags to be added to the C compile line
+--cross-compile:Indicates that we are doing cross compilation
+--blas-src:The path to an existing OpenBLAS source dircetory. The source is cloned if this parameter is omitted
+'
+}
+
 while [[ $# -gt 0 ]]; do
     key="$1"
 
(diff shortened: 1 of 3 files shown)


keeranroth:
@rakshithgb-fujitsu do you want to have a look at this as well? I'm doing some small updates to the build scripts to make them LLVM compatible in a few commits' time

.ci/env/openblas.sh
```diff
@@ -46,23 +76,49 @@ done
...
 
 sudo apt-get update
 sudo apt-get -y install build-essential gcc gfortran
-git clone https://github.com/xianyi/OpenBLAS.git
-CoreCount=$(lscpu -p | grep -Ev '^#' | wc -l)
-pushd OpenBLAS
-  make clean
+blas_src_dir=${blas_src_dir:-$OPENBLAS_DEFAULT_SOURCE_DIR}
+if [[ ! -d "${blas_src_dir}" ]] ; then
+  git clone https://github.com/xianyi/OpenBLAS.git "${blas_src_dir}"
```
rakshithgb-fujitsu, line 87: Can we change this to the latest updated link? `https://github.com/OpenMathLib/OpenBLAS`

keeranroth: Yeah, absolutely

napetrov (approved):
Looks good to me @rakshithgb-fujitsu  @keeranroth  if no other changes we can merge this after CI. 

Reason for even running internal CI - we do have checks on IP/Personal Information/etc and in some cases it might caught something from scripts that is not used internally.

.ci/env/openblas.sh
```diff
@@ -46,23 +76,49 @@ done
...
 
 sudo apt-get update
 sudo apt-get -y install build-essential gcc gfortran
-git clone https://github.com/xianyi/OpenBLAS.git
-CoreCount=$(lscpu -p | grep -Ev '^#' | wc -l)
-pushd OpenBLAS
-  make clean
+blas_src_dir=${blas_src_dir:-$OPENBLAS_DEFAULT_SOURCE_DIR}
+if [[ ! -d "${blas_src_dir}" ]] ; then
+  git clone https://github.com/OpenMathLib/OpenBLAS "${blas_src_dir}"
```
rakshithgb-fujitsu, line 87: Can we also do a shallow clone as well?

keeranroth: Yeah, good idea. As part of this, I've added options to pass in the version to check out, as well as a prefix variable. In future commits, I want to split the CI out into building TBB and OpenBLAS themselves, so that we can cache them. We don't need to build OpenBLAS and TBB every time we build

napetrov: With bazel builds in future this also can be cashed within bazel itself

keeranroth:
Was there anything that the internal CI showed up @napetrov?

ethanglaser:
No problems with internal CI

PR #2713, opened 2024-04-08
Dev/eglaser catch followup

ethanglaser:
# Description
Skip select_indexed_rows_dpc bazel test for now - it is causing problems with validation


md-shafiul-alam (approved):
assuming green CI

PR #2724, opened 2024-04-15
Add system related parameters to performance related algorithms' parameters

Vika-F:
Following changed were added:
- `system_parameters` class holding system-related performance parameters implemented.
- all the performance-related hyperparameters classes like `covariance::detail::compute_parameters` are now derived from `system_parameters` class because in the future those system-related parameters might be chosen on per-algorithm basis.


Vika-F:
@keeranroth and @rakshithgb-fujitsu,
Can you please take a look on this PR?
It adds new API that deals with the enabled CPU instruction set and other system-related parameters in oneDAL.

cpp/oneapi/dal/detail/cpu_info_impl.cpp
```diff
@@ -80,6 +81,9 @@ void cpu_info_impl::print_any(const std::any& value, std::stringstream& ss) cons
     else if (ti == typeid(cpu_vendor)) {
         print<cpu_vendor>(value, ss);
     }
+    else {
+        throw unimplemented{ dal::detail::error_messages::unsupported_data_type() };
```
keeranroth, line 85: Can you pass the typeid of `ti` so that this message is a little more specific? It would be nice to know what the unknown type is

cpp/oneapi/dal/detail/system_parameters.cpp
```diff
@@ -0,0 +1,39 @@
...
+*******************************************************************************/
+
+#include "oneapi/dal/detail/system_parameters.hpp"
+
+namespace oneapi::dal {
+
+namespace detail {
+
+system_parameters::system_parameters()
+        : impl_(detail::pimpl<system_parameters_impl>(new system_parameters_impl())) {}
```
keeranroth, line 24: Avoid the use of bare new. Use std::make_unique here. `pimpl` is a shared pointer, but going from unique to shared pointer is easy. It should probably be looked at in the future, whether `impl_` should be a unique pointer in here. It is a private field and the constructor creates the object, so it seems like ownership is clear here.

This might be opening a can of worms, but the ownership of objects is a little lax in the rest of the code. I'd like to see more unique pointers, especially in new code. Maybe that's just me, though.

```suggestion
system_parameters::system_parameters()
        : impl_(std::make_unique<system_parameters_impl>()) {}
```

cpp/oneapi/dal/detail/system_parameters.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include "oneapi/dal/detail/system_parameters_impl.hpp"
+
+namespace oneapi::dal {
+
+namespace detail {
+
+/// Stores system-related parameters that affect the performance of the algorithms.
+/// Those parameters can differ from the `get_global_context().get_cpu_info()`.
+///
+/// `cpu_info` reports the parameters available in hardware, when `system_parameters`
```
keeranroth, line 31: nit:

```suggestion
/// `cpu_info` reports the parameters available in hardware, where `system_parameters`
```

cpp/oneapi/dal/detail/system_parameters.hpp
```diff
@@ -0,0 +1,70 @@
...
+/// are the software-enabled parameters that can differ from `cpu_info`.
+class system_parameters : public base {
+public:
+    /// Creates a new default ``system_parameters`` instance.
```
keeranroth, line 35: Use of double backticks here, where single backticks are used in other places in this file. Keep the usage consistent

```suggestion
    /// Creates a new default `system_parameters` instance.
```

cpp/oneapi/dal/detail/system_parameters.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+#ifdef ONEDAL_DATA_PARALLEL
+    /// Logs host and device parameters in the format: name_1: value_1; ... ; name_N: value_N.
+    ///
+    /// @param queue                  The SYCL* queue object
+    std::string dump(sycl::queue& queue) const;
+#endif
+
+private:
+    detail::pimpl<system_parameters_impl> impl_;
```
keeranroth, line 66: Can this be a unique pointer?
```suggestion
    std::unique_ptr<system_parameters_impl> impl_;
```

cpp/oneapi/dal/detail/system_parameters_impl.cpp
```diff
@@ -0,0 +1,68 @@
...
+void system_parameters_impl::print_any(const std::any& value, std::ostringstream& ss) const {
+    const std::type_info& ti = value.type();
+    if (ti == typeid(cpu_extension)) {
+        ss << to_string(std::any_cast<cpu_extension>(value));
+    }
+    else if (ti == typeid(std::uint32_t)) {
+        ss << std::any_cast<std::uint32_t>(value);
+    }
+    else {
+        throw unimplemented{ dal::detail::error_messages::unsupported_data_type() };
```
keeranroth, line 53: Can information about `ti` be printed in the exception as well?

Update: Looking at the code, this is not a simple ask, so ignore me. I've raised issue https://github.com/oneapi-src/oneDAL/issues/2742

cpp/oneapi/dal/detail/system_parameters_impl.cpp
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+std::string system_parameters_impl::dump() const {
+    std::ostringstream ss;
+    for (auto it = sys_info_.begin(); it != sys_info_.end(); ++it) {
```
keeranroth, line 59:
```suggestion
    for (auto const& it : sys_info_) {
```

or if we can assume C++17:

```
  for (auto const &[name, val] : sys_info_) {
```

cpp/oneapi/dal/detail/system_parameters_impl.hpp
```diff
@@ -0,0 +1,56 @@
...
+    std::uint32_t get_max_workgroup_size(sycl::queue& queue) const;
+#endif
+
+    std::string dump() const;
+
+#ifdef ONEDAL_DATA_PARALLEL
+    std::string dump(sycl::queue& queue) const;
+#endif
+
+protected:
```
keeranroth, line 48: nit: I don't think that this is ever used as a base class, so this should be `private`

keeranroth (commented):
Overall looks good. Just some cosmetic comments

Alexandr-Solovev (commented):
Overall looks good to me with previous comments

cpp/oneapi/dal/detail/system_parameters.cpp
Vika-F: Thank you for this valid observation. I have replaced `new` with `make_unique`.

Regarding making `impl_` a unique pointer.
I'd prefer not to do it in the scope of this PR because:

1. it would lead to the inconsistency with the rest of oneDAL's code
2. modifying `detail::pimpl`'s definition will require extensive testing, including performance testing as `pimpl` is used widely in the code. For me it looks like a separate task with unknown outcome.

I've forwarded your observations to my team. We'll think on how to prioritize this.

cpp/oneapi/dal/detail/system_parameters_impl.cpp
Vika-F: C++17 variant was implemented here and in `cpu_info`

keeranroth (approved):
All looks good to me. Thanks for the updates

Alexandr-Solovev (approved):
Please fix conflicts and wait private CI checks

Vika-F:
Re-started CI: http://intel-ci.intel.com/ef12a9dd-06d4-f1a9-94c3-a4bf010d0e2e

Vika-F:
Re-started CI: http://intel-ci.intel.com/ef137714-0989-f1e2-b2cf-a4bf010d0e2e

PR #2725, opened 2024-04-15
TEST: bazel test of gpu primitives in CI scope

ethanglaser:
# Description
Add primitives to BUILD file so tests run in CI/nightly scope, fix failures (by adding condition for hardware)


Alexandr-Solovev (approved):

ethanglaser:
Update: search tests have fails, need to address before merge

PR #2727, opened 2024-04-16
fix: Distributed Pca singular values computation fix

Alexandr-Solovev:
# Description
The pr contains fix for correct singular values computation. We discovered that singular values don't match between batch and spmd. The issue was in computational logic, where singular values compute function takes the local number of rows, instead of global. The fix has been added.


ethanglaser:
Please clang and rebase

samir-nasibli (commented):
Generally looks good to me, but it would be better to have meaningful description (What and when was the issue? and how it fixes it?) for the PR.

PR #2719, opened 2024-04-11
Run Arm GNU cross-compiled tests in CI

keeranroth:
Add the running of cross-compiled GNU Arm builds to CI, running through QEMU. Changes are mad with a view of making things easier for future additions of cross-compilers and architectures:

* Install qemu-user-static in CI so that we can run aarch64 executables without needing to prefix qemu-aarch64 to the start of command lines
*  Rename toolchain file, so that it fits the pattern of `<arch>-<compiler>-crosscompile-toolchain.cmake`, so that we can easily include other compiler and arches without needing to update build scripts
* Pass the platform to run on into the test.sh script, and indicate when we are cross-compiling
* Export the CMAKE_PREFIX_PATH with the oneTBB install location prefixed, so that the test build can pick up the right TBB
* Print a message for failing tests. Currently the error output was not printed
* Examples CMake identifies the processor to run on using CMAKE_SYSTEM_PROCESSOR, instead of CMAKE_HOST_SYSTEM_PROCESSOR, which was leading to Intel being in directory names containing Arm ELFs
* Added files to list tests not to be run under emulation

PR #2728, opened 2024-04-17
Cache BLAS and TBB builds in CI

keeranroth:
# Description
In the case that a CI pipeline builds OpenBLAS or oneTBB from source, those should be cached and reused. The version of the libraries will be the same from run to run, so that we can leverage the Azure caching capability.

Changes made:
- build.sh will not try to build OpenBLAS or oneTBB in the case that explicit install paths are passed in
- OpenBLAS and oneTBB install scripts explicitly take a version string, so that we can use this as part of CI, and pass it through to the build scripts
- Use a cache task in the Arm and x86 CI builds when OpenBLAS or oneTBB is used as the backend

### DIFF
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -23,10 +23,15 @@ pr:
     - docs
     - .ci/pipeline/docs.yml
 
+variables:
+  OPENBLAS_VERSION : 'v0.3.27'
+  TBB_VERSION : 'v2021.10.0'
+  VM_IMAGE : 'ubuntu-22.04'
+
 jobs:
 - job: 'ClangFormat'
   pool:
-    vmImage: 'ubuntu-22.04'
+    vmImage: '$(VM_IMAGE)'
   steps:
     - script: |
         .ci/env/apt.sh clang-format
@@ -42,7 +47,7 @@ jobs:
     release.dir: '__release_lnx_gnu'
     platform.type : 'lnx32e'
   pool:
-    vmImage: 'ubuntu-22.04'
+    vmImage: '$(VM_IMAGE)'
   steps:
   - script: |
       .ci/env/apt.sh dev-base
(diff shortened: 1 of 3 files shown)


keeranroth:
This is a commit on top of pull request https://github.com/oneapi-src/oneDAL/pull/2719

keeranroth:
Are there any comments? In the latest run there were cache hits in the Azure pipeline, if you want to check that it is working

.ci/pipeline/ci.yml
```diff
@@ -143,6 +165,8 @@ jobs:
   variables:
     release.dir: '__release_lnx_gnu'
     platform.type : 'lnx32e'
+    OPENBLAS_VERSION: 'v0.3.27'
```
napetrov, line 168: do we need per pipeline definitions vs centralized global versioning for TBB and OPENBLAS?

napetrov (approved):
Have no objections here - some improvements suggested but they can be part of separate PRs

.ci/pipeline/ci.yml
keeranroth: I had thought about that, and decided to go for per pipeline for now. The use case being errata in the build for compiler / OS / architecture. But that is not the case at the moment (including for the RISC-V build I have locally), so I could put these as global for now, and we can make them per-pipeline if the need arises. What is your preference here?

.ci/pipeline/ci.yml
```diff
@@ -104,11 +108,29 @@ jobs:
   - script: |
       .ci/scripts/describe_system.sh
     displayName: 'System info'
+  - task: Cache@2
+    inputs:
+      key: '"gcc" | "aarch64" | "openblas" | "$(OPENBLAS_VERSION)"'
```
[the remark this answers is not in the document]
keeranroth: Yeah, that is a good point. Whilst I was here, it seemed a good opportunity to update the global definition of the TBB and OpenBLAS variables, as you suggested in your other comment.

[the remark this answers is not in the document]
keeranroth: I've left the GCC key out, as it takes the GCC from the package manager, rather than defining a particular version, so the OS fully defines the GCC version

.ci/pipeline/ci.yml
napetrov: i think ether would work. We can keep as is

PR #2731, opened 2024-04-17
Feature: Data type support in read function

DDJHB:
- Ability to set data type in read function for tables

### DIFF
--- a/cpp/oneapi/dal/io/csv/common.hpp
+++ b/cpp/oneapi/dal/io/csv/common.hpp
@@ -68,14 +68,16 @@ protected:
 using v1::data_source_tag;
 using v1::data_source_impl;
 using v1::data_source_base;
+using v1::is_valid_float_v;
 
 } // namespace detail
 
 namespace v1 {
 
-template <typename Float = double>
+template <typename Float = float>
 class data_source : public detail::data_source_base {
-    // static_assert(detail::is_valid_float_v<Float>);
+    static_assert(detail::is_valid_float_v<Float>);
+
 public:
     using float_t = Float;
 
--- a/cpp/oneapi/dal/io/csv/common.hpp
+++ b/cpp/oneapi/dal/io/csv/common.hpp
@@ -78,6 +78,7 @@ namespace v1 {
 ///
 /// @tparam Float The type of the floating-point that the data source will operate with.
 ///               Must be a floating-point type.
+template <typename Float>
 class data_source : public detail::data_source_base {
     static_assert(detail::is_valid_float_v<Float>);
 
--- a/cpp/oneapi/dal/io/csv/common.hpp
+++ b/cpp/oneapi/dal/io/csv/common.hpp
@@ -78,7 +78,7 @@ namespace v1 {
 ///
 /// @tparam Float The type of the floating-point that the data source will operate with.
 ///               Must be a floating-point type.
-template <typename Float>
+template <typename Float = float>
 class data_source : public detail::data_source_base {
     static_assert(detail::is_valid_float_v<Float>);
 
--- a/cpp/oneapi/dal/io/csv/common.hpp
+++ b/cpp/oneapi/dal/io/csv/common.hpp
@@ -74,32 +74,41 @@ using v1::is_valid_float_v;
 
 namespace v1 {
 
-template <typename Float = float>
+/// Used for the specification of data source configuration.
+///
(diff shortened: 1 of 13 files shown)


cpp/oneapi/dal/io/csv/common.hpp
```diff
@@ -65,13 +68,19 @@ class ONEDAL_EXPORT data_source_base : public base {
...
 using v1::data_source_base;
+using v1::is_valid_float_v;
 
 } // namespace detail
 
 namespace v1 {
 
+template <typename Float = float>
 class data_source : public detail::data_source_base {
+    static_assert(detail::is_valid_float_v<Float>);
```
Vika-F, line 79: I assume it would be better to move this line inside the constructors.

examples/oneapi/cpp/source/misc/read_batch.cpp
```diff
@@ -0,0 +1,44 @@
...
+    const auto data_file_name = get_data_path("covcormoments_dense.csv");
+
+    const auto table =
+        dal::read<dal::table>(dal::csv::data_source<TableReadType>{ data_file_name });
+
+    const auto type = table.get_metadata().get_data_type(0);
+    switch (type) {
+        case dal::data_type::float64: std::cout << "table data type: double " << std::endl; break;
+        case dal::data_type::float32: std::cout << "table data type: float " << std::endl; break;
+        default: std::cout << "table data type: null " << std::endl; break;
```
Vika-F, line 33: I think there should be an exception that might show that something goes wrong with the example. The same applies to GPU example.

cpp/oneapi/dal/io/csv/common.hpp
```diff
@@ -65,13 +68,19 @@ class ONEDAL_EXPORT data_source_base : public base {
...
 using v1::data_source_impl;
 using v1::data_source_base;
+using v1::is_valid_float_v;
 
 } // namespace detail
 
 namespace v1 {
 
+template <typename Float = float>
 class data_source : public detail::data_source_base {
```
Vika-F, line 78: Please add the description to this class and its methods.
We use doxygen notations. You can look [here](https://github.com/oneapi-src/oneDAL/blob/main/cpp/oneapi/dal/table/row_accessor.hpp) for an example of how the classes/methods descriptions look like in oneDAL.

Vika-F (commented):
The code looks good to me in general.
But it needs to be polished a bit to go to mainline.

Please see my comments. Also adding the description to a PR is a good practice.

cpp/oneapi/dal/io/csv/common.hpp
DDJHB: Have put it here following the convention in algo's descriptor definitions. Besides, since static_asserts are checked during compile time, why is it better to move it inside the constructors?

examples/oneapi/cpp/source/misc/read_batch.cpp
DDJHB: Not sure what did you mean here. If you are referring to invalid data type usage, then it is gonna fail during compile due to the way static_asserts work. If it is anything else, could you please provide more details?

cpp/oneapi/dal/io/csv/common.hpp
Vika-F: My assumption was that this code might be optimized out by the compiler if it's not on the main execution path.
Now I see that the code is in alignment with other algo's descriptors that are covered by various tests.
So it is better to leave this code as is.

examples/oneapi/cpp/source/misc/read_batch.cpp
Vika-F: I think it would be better just to remove printing in the `default` case then.

Vika-F:
@aepanchi 
Alexandra, can you or Emmanuella (cannot tag her here) please review the comments added into this public header
[cpp/oneapi/dal/io/csv/common.hpp](https://github.com/oneapi-src/oneDAL/pull/2731/files#diff-8d2a7707b3939ef9ff66ca510fe7e4ee740c8948f7b2c8e1f51832828545c0b6) ?

emmwalsh (approved):
Hi @Vika-F the comments added to this file look good to me! Thanks

Alexandr-Solovev (approved):
LGTM! Nice work!

PR #2736, opened 2024-04-22
Add Arm LLVM cross-compilation CI pipeline

keeranroth:
# Description
For Clang cross-compilation we need to do a few extra things when compared to GNU. First of all, we need to have a sysroot with all the libraries installed that we want. As such, we debootstrap a sysroot, which is to be cached in the Azure CI pipelines. We also install a more recent version of the LLVM compiler. Whilst not necessary for the Arm build, this is necessary to get some more recent support for RISC-V, in particular the vector extensions, which will be coming in a future commit.

Changes proposed:
- Install clang from llvm.org, rather than default apt
- Build a sysroot with necessary packages, and cache this in the CI
- Add a CMake toolchain file for Clang on aarch64
- Update OpenBLAS build to pass appropriate options for clang cross-compilation
- Add a variable to the oneDAL makefile to detect Clang cross-compilation, and set appropriate flags for building
- Update list of failing tests under emulation


keeranroth:
@rakshithgb-fujitsu You might want to have a look at this as well

@napetrov @Alexandr-Solovev I've added an LLVM cross-compiled Arm pipeline. This is because the LLVM toolchain for RISC-V is more mature, and I am going to put up another pull request for the RISC-V build on top of this one. I want to show that the RISC-V cross-build is very similar to the aarch64 one

.ci/env/apt.sh
```diff
@@ -62,6 +62,34 @@ function install_qemu_emulation {
...
+    # where the architecture and OS name need to be recognised by debootstrap,
+    # e.g. arch=arm64, os-name=jammy. The output directory path is relative to
+    # the working directory
+    mkdir -p "$1"
+    pushd "$1" || exit
+    sudo apt-get install -y debootstrap build-essential
+    sudo debootstrap --arch="$2" --verbose --include=fakeroot,symlinks,libatomic1 --resolve-deps --variant=minbase --components=main,universe "$3" "$4"
+    sudo chroot "$4" symlinks -cr .
+    sudo chown "${USER}" -R "$4"
+    rm -rf "${4:?}"/{dev,proc,run,sys,var}
```
Alexandr-Solovev, line 86: Can you please explain how and for what these rm functions will work? Can it affect user?

.ci/env/apt.sh
```diff
@@ -81,8 +109,14 @@ elif [ "${component}" == "dev-base" ]; then
...
+elif [ "${component}" == "llvm-version" ] ; then
+    update
+    install_llvm_version "$2"
+elif [ "${component}" == "build-sysroot" ] ; then
+    update
+    build_sysroot "$2" "$3" "$4" "$5"
 else
     echo "Usage:"
-    echo "   $0 [dpcpp|mkl|arm-compiler|clang-format|dev-base]"
+    echo "   $0 [dpcpp|mkl|arm-compiler|clang-format|dev-base|qemu-emulation|llvm-version|build-sysroot]"
```
Alexandr-Solovev, line 120: Thanks for adding qemu-emulation!

.ci/env/arm-clang-crosscompile-toolchain.cmake
```diff
@@ -0,0 +1,14 @@
+SET(MAKE_CROSS_COMPILING TRUE)
```
Alexandr-Solovev, line 1: Please add copyright here

examples/daal/cpp/target_excludes.cmake
```diff
@@ -47,4 +67,4 @@ if((CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64") AND
         "pca_metrics_dense_batch"
         "pivoted_qr_dense_batch"
     )
-endif()
+endif()
```
Alexandr-Solovev, line 70: add a newline here please

.ci/pipeline/ci.yml
```diff
@@ -163,6 +164,96 @@ jobs:
     condition: failed()
     continueOnError: true
 
+- job: 'LinuxMakeLLVM_OpenBLAS_ARM'
```
Alexandr-Solovev, line 167: Thanks for adding this CI check

Alexandr-Solovev (commented):
In general LGTM, just have one question and request minor fixes. Lets wait private CI

.ci/env/apt.sh
keeranroth: This is just to save on a few bytes when caching the sysroot. We only need the sysroot for includes and libraries, and not for running a chroot, so we rm some of the executables. The use of "${4:?}" will ensure that we don't remove /dev /proc, etc. The list might not be exhaustive here, but it is at least a start for what can be trimmed away.

Alexandr-Solovev: I see, thanks! I relaunched a private ci job, previous one was pretty green, but it's just for doublechecking. When it will be finished, I'll approve this pr.

Alexandr-Solovev (approved):

PR #2737, opened 2024-04-22
Add RISC-V clang build

keeranroth:
# Description
We add a RISC-V target. We are interested in higher-performance cores, so make sure that the vector extension is in place for this. The functionality is on par with aarch64 cross builds. CI is added as well. We assume that the RISC-V target has vector extensions enabled, and that this is at least 128-bits long. This is in line with RISC-V International guidelines that application class processors have a vector length >= 128-bits

There are some modifications of the toolchain that are required to get vector support up and running. In particular, a newer qemu is required, and this needs to be set up with binfmt support, as the deb package does not automatically setup binfmt.

Changes proposed:
- Add ability to install LLVM from an upstream repo
- Pass the architecture to install GNU cross-compilers to apt.sh
- Add ability to install qemu from a Debian package
- Add a file for setting up binfmt to run RISC-V binaries without specifying QEMU on the command line
- Add clang CMake toolchain file for RISC-V
- Add a CI pipeline for running and testing (under emulation) RISC-V builds using LLVM
- Allow the architecture string in the PLATFORM variable to be more than 3 characters long, so that we can use riscv64 as the architecture name in the oneDAL build
- Update openBLAS builds to use RISC-V optimized builds with 128-bit vector length
- Add a precompile macro TARGET_RISCV64 to identify when the build is targeting RISC-V, in order to target RISC-V specific parts of the code base
- Add structs, macros and enums, as per the SVE build, so that future commits can target specific micro-architectures more easily, and so that the current build completes for RISC-V
- Add a cpu_vendor to all builds for riscv64
- Add lnxriscv64 as a platform to use in the makefile


.ci/env/apt.sh
```diff
@@ -54,35 +54,80 @@ function install_dev-base-conda {
...
+function install_qemu_emulation_apt {
     sudo apt-get install -y qemu-user-static
 }
 
+function install_qemu_emulation_deb {
+    sudo apt-get install -y binfmt-support
+    qemu_deb=qemu-user-static_8.2.2+ds-2+b1_amd64.deb
+    wget http://ftp.de.debian.org/debian/pool/main/q/qemu/${qemu_deb}
+    sudo dpkg -i ${qemu_deb}
+    sudo update-binfmts --importdir "$1" --import "qemu-$2"
```
luhenry, line 70: That's is already done as part of the dpkg -i above

.ci/env/apt.sh
```diff
@@ -54,35 +54,80 @@ function install_dev-base-conda {
...
 
-function install_qemu_emulation {
+function install_qemu_emulation_apt {
     sudo apt-get install -y qemu-user-static
 }
 
+function install_qemu_emulation_deb {
+    sudo apt-get install -y binfmt-support
+    qemu_deb=qemu-user-static_8.2.2+ds-2+b1_amd64.deb
+    wget http://ftp.de.debian.org/debian/pool/main/q/qemu/${qemu_deb}
```
luhenry, line 68: Is there an https link instead? I'm not sure how the signature of the package is checked otherwise.

dev/make/compiler_definitions/clang.ref.riscv64.mk
```diff
@@ -0,0 +1,44 @@
...
+
+# Linker flags
+link.dynamic.lnx.clang = clang++ \
+                         $(COMPILER.lnx.clang.target) \
+                         $(COMPILER.sysroot)
+
+pedantic.opts.lnx.clang = $(pedantic.opts.clang)
+
+# For the base ISA
+rv64_OPT.clang = $(-Q)march=rv64gc_v1p0
```
luhenry, line 43: You can add the Zvl128b extension to ensure at the compiler level that the Vector length is at least 128b. It also serves as documentation in the code

examples/cmake/setup_examples.cmake
```diff
@@ -108,6 +108,8 @@ function (add_examples examples_paths)
             set(CPU_ARCHITECTURE "intel_intel64")
         elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
             set(CPU_ARCHITECTURE "arm_aarch64")
+        elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "riscv64")
+            set(CPU_ARCHITECTURE "riscv64_riscv64")
```
luhenry, line 112: Nit: Should it be "riscv_rv64" following the scheme of "arm_aarche64"?

.ci/env/apt.sh
keeranroth: Good question. I have checked just now with wget, and see the following error message
```
Connecting to ftp.de.debian.org (ftp.de.debian.org)|141.76.2.4|:443... connected.
ERROR: no certificate subject alternative name matches
        requested host name ‘ftp.de.debian.org’.
To connect to ftp.de.debian.org insecurely, use `--no-check-certificate'.
```

So it looks like there is no signature verification for the package

.ci/env/apt.sh
keeranroth: Local testing didn't update the binfmt. I tried installing the .deb package using `apt-get install` to the same effect

examples/cmake/setup_examples.cmake
keeranroth: Yeah, I had thought about that. It's just naming, and I'm not sure why the repeated values are here, but I'm copying the style. Ideally, I would like to see just a single architecture string (and then maybe have x86 rather than Intel). I think that would be another pull request, as the changes then are not associated with just RISC-V

.ci/env/apt.sh
keeranroth: Aha. Thanks Mark. That does work for me

dev/make/compiler_definitions/clang.ref.riscv64.mk
keeranroth: Yes, good point

luhenry (approved):

keeranroth:
@Alexandr-Solovev are there some things that need fixing? The public CI is green, and there aren't any outstanding comments

PR #2753, opened 2024-04-29
chore: add OpenSSF Scorecard GitHub Action

homksei:
## Summary

This Pull Request adds a new GitHub Actions workflow that runs the OpenSSF Scorecard tool on our repository.

## Changes

- Added a new workflow file `openssf-scorecard.yml` in the `.github/workflows` directory.

## Motivation

The OpenSSF Scorecard is an automated security tool that performs a series of checks on our codebase and provides a score based on the results. By adding this workflow, we can automate the process of running the Scorecard tool and publishing the results.

The workflow is triggered on the following events:
- On a schedule, every Friday at 21:00 UTC.
- When a push is made to the `main` branch.

PR #2773, opened 2024-05-09
[CI] add repository check to openssf-scorecard github action

icfaust:
# Description
A new github action with a cron job was added to oneDAL (via #2753). As it is currently setup, this will run for all forks every friday at 9PM UTC.  To prevent this, a repository check will be added, and the step will be skipped for everything but the onedal-src/oneDAL repository. This matches changes made to the [docker validation nightly](https://github.com/oneapi-src/oneDAL/blob/main/.github/workflows/docker-validation-nightly.yml#L29)

- Add repository check to openssf-scorecard github action

### DIFF
--- a/.github/workflows/openssf-scorecard.yml
+++ b/.github/workflows/openssf-scorecard.yml
@@ -16,6 +16,7 @@ permissions: read-all
 jobs:
   analysis:
     name: Scorecard analysis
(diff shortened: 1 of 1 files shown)


ethanglaser (approved):

icfaust:
I'll let public CI run green and then will merge.

PR #2777, opened 2024-05-13
fix: CPATH variable fix

Alexandr-Solovev:
# Description
The pr fixes the value of CPATH environment variable. In both oneDAL layout it must be /include, not /include/dal


napetrov:
@maria-Petrova  - if we would like to integrate it - probably worth validation via self-service prior to drop

ethanglaser:
What's reason for change? Please add description

ethanglaser (approved):

PR #2800, opened 2024-05-24
fix: adding tbb_schedule_handler 

Alexandr-Solovev:
# Description
Adding tbb::task_scheduler_handle static object to avoid undefined order of static tbb objects deinitialization. oneTBB will not be unloaded until finalize of task_scheduler_handle.

- [x] Adding tbb::scheduler_handle_task

- [x] Initialize TBB in all potential scenarios

- [x] Update destructor and free functions

- [x] Remove void setDynamicLibraryThreadingTypeOnWindows(LibraryThreadingType type);

### DIFF
--- a/cpp/daal/src/threading/service_thread_pinner.cpp
+++ b/cpp/daal/src/threading/service_thread_pinner.cpp
@@ -202,6 +202,7 @@ class thread_pinner_impl_t : public tbb::task_scheduler_observer
     bool do_pinning;
     AtomicInt is_pinning;
     tbb::enumerable_thread_specific<cpu_mask_t *> thread_mask;
+    tbb::task_scheduler_handle scheduler_handle;
     tbb::task_arena pinner_arena;
     void (*topo_deleter)(void *);
 
@@ -235,6 +236,7 @@ thread_pinner_impl_t::thread_pinner_impl_t(void (*read_topo)(int &, int &, int &
     : pinner_arena(nthreads = daal::threader_get_threads_number()), tbb::task_scheduler_observer(pinner_arena), topo_deleter(deleter)
 {
     pinner_arena.initialize();
+    scheduler_handle = tbb::task_scheduler_handle(tbb::attach {});
     do_pinning = (nthreads > 0) ? true : false;
     is_pinning.set(0);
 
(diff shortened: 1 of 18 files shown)


cpp/daal/src/threading/threading.cpp
```diff
@@ -61,6 +64,29 @@ DAAL_EXPORT void _daal_tbb_task_scheduler_free(void *& globalControl)
     }
 }
 
+DAAL_EXPORT void _daal_tbb_task_scheduler_handle_free()
+{
+    std::mutex global_mutex;
+    std::lock_guard<std::mutex> guard(global_mutex);
+    if (isInitialized)
+    {
+        tbb::finalize(*globalSchedulerHandle, std::nothrow);
```
avolkov-intel, line 73: Add isInitialized = false

cpp/daal/src/threading/service_thread_pinner.cpp
avolkov-intel, line 1: Add tbb::finalize to destructor

cpp/daal/src/threading/threading.cpp
```diff
@@ -42,6 +42,9 @@
 
 using namespace daal::services;
 
+static std::shared_ptr<tbb::task_scheduler_handle> globalSchedulerHandle = nullptr;
```
avolkov-intel, line 45: Why do we need static shared_ptr if we make sure that the object stored in it is deleted before the deletion of smart pointer? Could we use a normal pointer instead?

cpp/daal/src/threading/threading.cpp
```diff
@@ -54,21 +56,48 @@ DAAL_EXPORT void _threaded_scalable_free(void * ptr)
...
+        tbb::task_scheduler_handle * handle = static_cast<tbb::task_scheduler_handle *>(schedulerHandle);
+        tbb::finalize(*handle, std::nothrow);
+        delete handle;
+        schedulerHandle = nullptr;
+    }
+}
+
+DAAL_EXPORT void _daal_tbb_task_scheduler_handle_free(void *& schedulerHandle)
+{
+    std::mutex global_mutex;
```
olegkkruglov, line 81: is it ok that global_mutex is redefined here?

ethanglaser:
please fix clang and try to rebase for open blas public CI fails

PR #2801, opened 2024-05-27
Changed openblas build to ILP64

DhanusML:
# Issue
OpenBLAS uses 32-bit integers by default . But DAAL_INT is 64-bit integer type. This causes errors during LAPACK calls, where the argument is expected to be a pointer to 32-bit integer, but a pointer to 64-bit integer is given. This causes many examples that use LAPACK to fail.

# Fix
Build openBLAS with ILP64 interface by adding `BINARY=64` and `INTERFACE64=1` while building. An option to switch ILP64 interface has been added to the openblas build script (by default ILP64 interface is built). 

This change fixes the following examples which have been removed from the exclude list.

1. out_detect_mult_dense_batch
2. pca_cor_dense_batch
3. pca_cor_dense_distr
4. pca_cor_dense_online
5. qr_dense_batch
6. qr_dense_distr
7. qr_dense_online
8. svd_dense_batch

Changes were tested on Graviton3 (openblas+gcc build).

### DIFF
--- a/.ci/env/apt.sh
+++ b/.ci/env/apt.sh
@@ -63,7 +63,7 @@ function install_qemu_emulation_apt {
 }
 
 function install_qemu_emulation_deb {
-    qemu_deb=qemu-user-static_8.2.3+ds-2_amd64.deb
+    qemu_deb=qemu-user-static_8.2.1+ds-1~bpo12+1_amd64.deb
     set -eo pipefail
     wget http://ftp.debian.org/debian/pool/main/q/qemu/${qemu_deb}
     sudo dpkg -i ${qemu_deb}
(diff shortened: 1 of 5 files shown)


DhanusML:
This PR requires rebuilding openblas. Examples will fail if the cached version of openblas is used.

napetrov:
> This PR requires rebuilding openblas. Examples will fail if the cached version of openblas is used.

You can create new input to the cache key and include LP mode state - https://github.com/oneapi-src/oneDAL/blob/main/.ci/pipeline/ci.yml#L117

napetrov:
@keeranroth  - FYI

keeranroth:
@napetrov thanks for the heads up. This is a great catch, and good to see that so many tests are able to be cleared as well (from looking at the exclude lists being cleaned). CI will also need to have the key updated in the cached builds of OpenBLAS, as these will still pull down the 32-bit builds. That's likely the cause of the errors in CI pipelines

keeranroth:
Looks like the QEMU package has moved again. Can you set
`qemu_deb=qemu-user-static_8.2.1+ds-1~bpo12+1_amd64.deb`

in the script `.ci/env/apt.sh`? That has been up there since February, and should be good enough for us to use. Not sure why 8.2.3 should have vanished ¯\_(ツ)_/¯

keeranroth (approved):
Looks good to me

napetrov (approved):

PR #2803, opened 2024-05-28
fix: reimplement custom daal calls for VC

Alexandr-Solovev:
# Description
Reimplement custom daal function calls from DAAL without auto usage. Its necessary for avoiding issues with VC compiler. 

- [x] Update LinReg online

- [x] Update Pca online 

- [x] Update BS online


icfaust (approved):
Good to merge after green CI, just looks like localized internal interfacing so no benching etc. are necessary.

PR #2810, opened 2024-06-03
[bug] fix dpc compilation for OS=win and COMPILER=vc

icfaust:
# Description
This PR is initially to show the issue, with the solution commented out (to replicate the bug).  This shows that with VC the \$(-Q) setting in the makefile clashes with the definition of the dependency file generation here: https://github.com/oneapi-src/oneDAL/blob/main/dev/make/deps.mk#L115 needed for the icpx/icx compiler in the dpc step.

- remove unneeded definition for dpcpp in the deps file
- add an if statement for the icx compiler dependency


icfaust:
Note the CI oneDAL build example/ build Windows which shows the error. Next I will apply the changes.

Alexandr-Solovev (approved):

PR #2811, opened 2024-06-04
Fix memory leak in hashmap

md-shafiul-alam:
# Description
Fixes memory leak in hashmap caught in test with valgrind for DBSCAN. 
"delete[] entries_" is not enough to deallocate the individual hash_map_entry objects.

### DIFF
--- a/cpp/oneapi/dal/detail/hash_map.hpp
+++ b/cpp/oneapi/dal/detail/hash_map.hpp
@@ -71,6 +71,14 @@ public:
     }
 
     ~hash_map() {
+        for (std::int64_t i = 0; i < capacity_; i++) {
+            entry_ptr current = entries_[i];
+            while (current) {
+                entry_ptr next = current->get_next();
+                delete current;
+                current = next;
+            }
+        }
         delete[] entries_;
         entries_ = nullptr;
         capacity_ = 0;


cpp/oneapi/dal/detail/hash_map.hpp
```diff
@@ -71,6 +71,14 @@ class hash_map {
     }
 
     ~hash_map() {
+        for (std::int64_t i = 0; i < capacity_; i++) {
```
icfaust, line 74: dal::detail::threader_for or is it too much overhead ? @Alexandr-Solovev

icfaust (commented):
Good catch, should it be multithreaded because of the independence of the pointers?

cpp/oneapi/dal/detail/hash_map.hpp
Alexandr-Solovev: I'd like to say that its less safety here, and for me the current solution looks good

Alexandr-Solovev (approved):

samir-nasibli:
@md-shafiul-alam thank you for the fix.
Looks good. What should be done for some CI fails?

md-shafiul-alam:
> @md-shafiul-alam thank you for the fix. Looks good. What should be done for some CI fails?

All CI checks passed now

PR #2825, opened 2024-06-21
fix: replace CL/sycl.hpp includes

ethanglaser:
# Description

-
-
-


ethanglaser:
does not resolve the problem, need fix on MKL side

PR #2848, opened 2024-07-16
Add a chapter about threading layer into the docs

Vika-F:
- Added a chapter about internal threading primitives into oneDAL documentation
- Improved commenting in the header specifying those primitives APIs (`threading.h`)


Vika-F:
@emmwalsh Please review this PR. As it mostly deals with oneDAL docs.

Vika-F:
Produces docs can be found here: https://dev.azure.com/daal/DAAL/_build/results?buildId=36713&view=artifacts&pathAsName=false&type=publishedArtifacts

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Here is a variant of parallel dot product computation with static scheduling:
+
+.. include:: ../includes/threading/dot-static-parallel.rst
+
+Nested parallelism
+******************
+
+It is allowed to have nested parallel loops within oneDAL.
+What is important to know is that
```
bdmoore1, line 139:
```suggestion
It is important to know that:
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+    "when a parallel construct calls another parallel construct, a thread can obtain a task
+     from the outer-level construct while waiting for completion of the inner-level one."
+
+    -- `oneTBB documentation <https://www.intel.com/content/www/us/en/docs/onetbb/developer-guide-api-reference/2021-13/work-isolation.html>`_
+
+In practice this means that, for example, a thread-local variable might unexpectedly
```
bdmoore1, line 146:
```suggestion
In practice, this means that a thread-local variable might unexpectedly
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+oneTBB provides ways to isolate execution of a parallel construct, for its tasks
+to not interfere with other simultaneously running tasks.
+
+Those options are preferred when the parallel loops are initially written as nested.
+But in oneDAL there are cases when one parallel algorithm, the outer one,
+calls another parallel algorithm, the inner one, within a parallel region.
+
+The inner algorithm in this case can also be called solely, without additional nesting.
+And we do not want to always make it isolated.
```
bdmoore1, line 161:
```suggestion
And we do not always want to make it isolated.
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+For the cases like that oneDAL provides ``daal::ls``. Its ``local()`` method always
```
bdmoore1, line 163:
```suggestion
For the cases like that, oneDAL provides ``daal::ls``. Its ``local()`` method always
```

bdmoore1 (approved):

emmwalsh (approved):
LGTM!!

cpp/daal/src/threading/threading.h
```diff
@@ -255,6 +287,35 @@ inline void threader_for_int32ptr(const int * begin, const int * end, const F &
...
+/// This means that the work is always scheduled in the same way across the threads:
+/// each thread processes the same set of iterations on each invocation of this loop.
+///
+/// It is recommended to use this parallel loop if each iteration of the loop
+/// performs equal amount of work.
+///
+/// Let ``t`` be the number of threads available to oneDAL.
+///
+/// Then the number of iterations processed by each threads (except maybe the last one)
+/// is copmputed as:
```
ethanglaser, line 303:
```suggestion
/// is computed as:
```

cpp/daal/src/threading/threading.h
```diff
@@ -255,6 +287,35 @@ inline void threader_for_int32ptr(const int * begin, const int * end, const F &
...
+/// ``nI = (n + t - 1) / t``
+///
+/// Here is how the work in split across the threads:
```
ethanglaser, line 306:
```suggestion
/// Here is how the work is split across the threads:
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+.. highlight:: cpp
+
+Threading Layer
+^^^^^^^^^^^^^^^
+
+oneDAL uses Intel\ |reg|\  oneAPI Threading Building Blocks (Intel\ |reg|\  oneTBB) to do parallel
+computations on CPU.
+
+But oneTBB is not used in the code of oneDAL algorithms directly. The algorithms rather
+use custom primitives that either wrap oneTBB functionality or are inhome developed.
```
ethanglaser, line 26:
```suggestion
use custom primitives that either wrap oneTBB functionality or are developed in-house.
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+Those primitives form oneDAL's threading layer.
+
+This is done in order not to be dependent on possible oneTBB API changes and even
+on the particular threading technology.
+
+The API of the layer is defined in
+`threading.h <https://github.com/oneapi-src/oneDAL/blob/main/cpp/daal/src/threading/threading.h>`_.
+Please be aware that those APIs are not publicly defined. So they can be changed at any time
```
ethanglaser, line 34:
```suggestion
Please be aware that those APIs are not publicly defined, so they can be changed at any time
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+mappings of the loop's iteration space to the available threads.
+This strategy is benefitial when it is hard to estimate the amount of work performed
+by each iteration.
+
+In the cases when it is known that the iterations perform equal amount of work it
+might be benefitial to use pre-defined mapping of the loop's iterations to threads.
+This is what static work scheduling does.
+
+``daal::static_threader_for`` and ``daal::static_tls`` allow to implement static
+work scheduling within oneDAL.
```
ethanglaser, line 129:
```suggestion
``daal::static_threader_for`` and ``daal::static_tls`` allow implementation of
static work scheduling within oneDAL.
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Here is a variant of parallel dot product computation with static scheduling:
+
+.. include:: ../includes/threading/dot-static-parallel.rst
+
+Nested parallelism
+******************
+
+It is allowed to have nested parallel loops within oneDAL.
```
ethanglaser, line 138:
```suggestion
OneDAL supports nested parallel loops.
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+threader_for
+************
+
+Lets consider you need to compute an elementwise sum of two arrays and store the results
+into another array.
+Here is a variant of sequential implementation:
+
+.. include:: ../includes/threading/sum-sequential.rst
+
+There are several options available in oneDAL's threading layer to let the iterations of this code
```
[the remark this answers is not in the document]
Vika-F: That is a new information for me. Thank you!
I had fixed all the mentions in this PR.

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Lets consider you need to compute a dot product of two arrays.
+Here is a variant of sequential implementation:
+
+.. include:: ../includes/threading/dot-sequential.rst
+
+Parallel computations can be performed in two steps:
+
+    1. Compute partial dot product at each threaded.
+    2. Perform a reduction: Sum the partial results from all threads to compute the final dot product.
```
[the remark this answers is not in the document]
Vika-F: Probably yes. Not sure as I am not a native speaker. Modified to "add the partial results".

Vika-F:
Thank you @bdmoore1 , @ethanglaser and @emmwalsh for thorough reviews.
I've made all the proposed modifications. Hope to merge soon.

napetrov:
@keeranroth @rakshithgb-fujitsu - please check on this one - how clear topic is described and if you would like to get any additional details.


And for the fire - topics that might worth adding beyond threading topic

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+By default, oneTBB uses
+`dynamic work scheduling <https://oneapi-src.github.io/oneTBB/main/tbb_userguide/How_Task_Scheduler_Works.html>`_
+and work stealing.
+It means that two different runs of the same parallel loop can produce different
+mappings of the loop's iteration space to the available threads.
+This strategy is benefitial when it is difficult to estimate the amount of work performed
+by each iteration.
+
+In the cases when it is known that the iterations perform an equal amount of work, it
+might be benefitial to use predefined mapping of the loop's iterations to threads.
```
ethanglaser, line 125:
```suggestion
might be beneficial to use predefined mapping of the loop's iterations to threads.
```

ethanglaser (approved):
Just one more minor spelling adjustment mentioned above. Otherwise this looks great!

cpp/daal/src/threading/threading.h
```diff
@@ -480,10 +617,23 @@ class static_tls
...
+/// @tparam F  Type of the data located in the storage
 template <typename F>
 class ls : public tlsBase
 {
 public:
+    /// Initialize a local storage.
+    ///
+    /// @tparam lambdaType  Lambda function of type [/* captures */]() -> F
+    ///
+    /// @param lambda   Lambda function that initializes a local storage
```
keeranroth, line 634: For some of the comments, you specify that the `@param` is `[in]`, `[out]`, or `[in,out]`. I really like those annotations. Can we have them for all the comments?

```suggestion
    /// @param lambda   Lambda function that initializes local storage
```

cpp/daal/src/threading/threading.h
```diff
@@ -480,10 +617,23 @@ class static_tls
...
+    /// @param isTls    if true, then local storage is a thread-local storage (daal::tls)
```
keeranroth, line 635:
```suggestion
    /// @param isTls    If ``true``, then local storage is a thread-local storage (``daal::tls``)
```

cpp/daal/src/threading/threading.h
```diff
@@ -480,10 +617,23 @@ class static_tls
...
+    ///                 and might have problems in case of nested parallel regions.
```
keeranroth, line 636: I think it's worth putting a note somewhere in the documentation about nested parallelism, and what problems that can create for threads. Then from the comments in this file, you can link to the note. Just saying that there might be problems opens up questions, rather than resolving questions.

cpp/daal/src/threading/threading.h
```diff
@@ -499,13 +649,23 @@ class ls : public tlsBase
         lsPtr = _isTls ? _daal_get_tls_ptr(a, tls_func<lambdaType>) : _daal_get_ls_ptr(a, tls_func<lambdaType>);
     }
 
+    /// Destroys the memory associated with a local storage.
```
keeranroth, line 652:
```suggestion
    /// Destroys the memory associated with local storage.
```

cpp/daal/src/threading/threading.h
```diff
@@ -499,13 +649,23 @@ class ls : public tlsBase
...
+    ///
+    /// @note LS does not release the memory allocated by a lambda-function
```
keeranroth, line 654:
```suggestion
    /// @note ``ls`` does not release the memory allocated by a lambda-function
```

cpp/daal/src/threading/threading.h
```diff
@@ -499,13 +649,23 @@ class ls : public tlsBase
...
+    ///       provided to the constructor.
+    ///       Developers are responsible for deletion of that memory.
     virtual ~ls()
     {
         d->del(voidLambda);
         delete d;
         _isTls ? _daal_del_tls_ptr(lsPtr) : _daal_del_ls_ptr(lsPtr);
     }
 
+    /// Access a local data of a thread by value.
```
keeranroth, line 664:
```suggestion
    /// Access the local data of a thread by value.
```

cpp/daal/src/threading/threading.h
```diff
@@ -517,6 +677,12 @@ class ls : public tlsBase
         if (!_isTls) _daal_release_ls_local(lsPtr, p);
     }
 
+    /// Sequential reduction.
+    ///
+    /// @tparam lambdaType  Lambda function of type [/* captures */](F) -> void
```
keeranroth, line 682:
```suggestion
    /// @tparam lambdaType  Lambda function of type ``[/* captures */](F) -> void``
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+oneDAL uses Intel\ |reg|\  oneAPI Threading Building Blocks (Intel\ |reg|\  oneTBB) to do parallel
+computations on CPU.
+
+But oneTBB is not used in the code of oneDAL algorithms directly. The algorithms rather
+use custom primitives that either wrap oneTBB functionality or are in-house developed.
+Those primitives form oneDAL's threading layer.
+
+This is done in order not to be dependent on possible oneTBB API changes and even
+on the particular threading technology.
```
keeranroth, line 30: Do you mean POSIX vs non-POSIX threads when you mention the threading technology? It's worth giving a name of the type of threads that you mean.

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+The API of the layer is defined in
+`threading.h <https://github.com/oneapi-src/oneDAL/blob/main/cpp/daal/src/threading/threading.h>`_.
+Please be aware that those APIs are not publicly defined, so they can be changed at any time
```
keeranroth, line 34: The API is public, as I can take the file without a licence and look at it. Maybe something like:

```
Please be aware that the threading API is not constrained by a specification. Work is ongoing to improve the threading layer, and in that time, the functions in the threading API may be changed without notice.
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+.. include:: ../includes/threading/sum-sequential.rst
+
+There are several options available in the threading layer of oneDAL to let the iterations of this code
+run in parallel.
+One of the options is to use ``daal::threader_for`` as shown here:
+
+.. include:: ../includes/threading/sum-parallel.rst
+
+The iteration space here goes from ``0`` to ``n-1``.
+The last argument is the lambda function that defines a function object that proceeds ``i``-th
```
keeranroth, line 55:
```suggestion
The last argument is a function object that performs a single iteration of the loop, given loop index ``i``
```

I've not seen discussion around the independence of loop iterations. I think we are talking about data independent loops here, right? It would be good to state the assumed interaction between different loop iterations.

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+iteration of the loop.
+
+Blocking
+--------
+
+To have more control over the parallel execution and to increase
+`cache locality <https://en.wikipedia.org/wiki/Locality_of_reference>`_ oneDAL usually splits
```
keeranroth, line 62:
```suggestion
`cache locality <https://en.wikipedia.org/wiki/Locality_of_reference>`, oneDAL usually splits
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+**************************
+
+Lets consider you need to compute a dot product of two arrays.
+Here is a variant of sequential implementation:
+
+.. include:: ../includes/threading/dot-sequential.rst
+
+Parallel computations can be performed in two steps:
+
+    1. Compute partial dot product at each threaded.
```
keeranroth, line 79:
```suggestion
    1. Compute partial dot product in each thread.
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+    2. Perform a reduction: Add the partial results from all threads to compute the final dot product.
+
+``daal::tls`` provides a local storage where each thread can accumulate its local results.
+Following code allocates memory that would store partial dot products for each thread:
```
keeranroth, line 83:
```suggestion
The following code allocates memory that would store partial dot products for each thread:
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+.. include:: ../includes/threading/dot-parallel-init-tls.rst
+
+``SafeStatus`` in this code denotes a thread-safe counterpart of the ``Status`` class.
+``SafeStatus`` allows to collect errors from all threads and report them to user using
+``detach()`` method as it will be shown later in the code.
```
keeranroth, line 89:
```suggestion
``SafeStatus`` allows to collect errors from all threads and report them to the user using the
``detach()`` method. An example will be shown later in the documentation.
```

Would be worth linking to the example of how to use the detach method

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Checking the status right after the initialization code won't show the allocation errors,
+because oneTBB uses lazy evaluation and the lambda function passed to the constructor of the TLS
+is evaluated in the moment of the TLS's first use.
```
keeranroth, line 93:
```suggestion
is evaluated on first use of the thread local storage (TLS).
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Again, there are several options available in the threading layer of oneDAL to compute the partial
```
keeranroth, line 95:
```suggestion
There are several options available in the threading layer of oneDAL to compute the partial
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+dot product results at each thread.
+One of the options is to use the already mentioned ``daal::threader_for`` and blocking approach
+as shown here:
+
+.. include:: ../includes/threading/dot-parallel-partial-compute.rst
+
+To compute the final result it is requred to reduce TLS's partial results over all threads
```
keeranroth, line 102:
```suggestion
To compute the final result it is required to reduce each thread's partial results
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+as shown here:
+
+.. include:: ../includes/threading/dot-parallel-reduction.rst
+
+Local memory of the threads should also be released when it is no longer needed.
```
keeranroth, line 107:
```suggestion
Local memory of the threads should be released when it is no longer needed.
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+The complete parallel verision of dot product computations would look like:
```
keeranroth, line 109:
```suggestion
The complete parallel version of dot product computations would look like:
```

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+By default, oneTBB uses
+`dynamic work scheduling <https://oneapi-src.github.io/oneTBB/main/tbb_userguide/How_Task_Scheduler_Works.html>`_
+and work stealing.
+It means that two different runs of the same parallel loop can produce different
+mappings of the loop's iteration space to the available threads.
+This strategy is beneficial when it is difficult to estimate the amount of work performed
+by each iteration.
+
+In the cases when it is known that the iterations perform an equal amount of work, it
```
keeranroth, line 124: In all cases, where you know how much work there is to do in the parallel regions, it will be faster to map the work onto threads using a dynamic schedule. Drop the use of `it might be beneficial to use ...` in favour of `it is more performant to use ...`

docs/source/contribution/threading.rst
```diff
@@ -0,0 +1,166 @@
...
+
+oneDAL supports nested parallel loops.
+It is important to know that:
+
+    "when a parallel construct calls another parallel construct, a thread can obtain a task
+     from the outer-level construct while waiting for completion of the inner-level one."
+
+    -- `oneTBB documentation <https://www.intel.com/content/www/us/en/docs/onetbb/developer-guide-api-reference/2021-13/work-isolation.html>`_
+
+In practice, this means that a thread-local variable might unexpectedly
```
keeranroth, line 146: Can you define what you mean by thread-local? I usually think of thread local data to mean that only a single thread can read or write the value to thread-local storage. The thread-local storage class `tls` is intended to be used for thread local storage, but it cannot do so in the case of nested parallelism. The only thread local storage which is guaranteed to be thread local is in the leaf nodes of a graph of threads.

So something along the lines of:
```
If we define thread-local data to be data that can only be read and written by a single thread, then any thread which spawns other threads in general cannot have thread-local variables. This is true for variables which are defined as thread-local. This means that great care is required when using thread-local storage constructs in nested parallel regions to avoid issues caused by concurrent data access and ownership (e.g. data corruption due to races; segmentation faults; and deadlocks)
```

keeranroth (changes requested):

cpp/daal/src/threading/threading.h
```diff
@@ -223,14 +223,33 @@ inline void threader_func_break(int i, bool & needBreak, const void * a)
     lambda(i, needBreak);
 }
 
+/// Execute the for loop defined by the input parameters in parallel.
+/// The maximal number of iterations in the loop is 2^31 - 1.
+/// The work is scheduled dynamically across threads.
+///
+/// @tparam F   Lambda function of type ``[/* captures */](int i) -> void``,
+///             where ``i`` is the loop's iteration index, ``0 <= i < n``.
```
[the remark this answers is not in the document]
Vika-F: Thanks for noticing this. We are using not only doxygen, but also restructured text + some python pre-processing.
Double backtick is valid but it results in italic text.
I'm going to replace it with single backtick to have monospace.

cpp/daal/src/threading/threading.h
```diff
@@ -223,14 +223,33 @@ inline void threader_func_break(int i, bool & needBreak, const void * a)
     lambda(i, needBreak);
 }
 
+/// Execute the for loop defined by the input parameters in parallel.
+/// The maximal number of iterations in the loop is 2^31 - 1.
+/// The work is scheduled dynamically across threads.
```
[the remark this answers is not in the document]
Vika-F: The idea of this API is that it uses the default scheduling provided by the underlying implementation.
Of course, the execution might not be parallel for various reasons. For example, if only one thread is available.

It is expected that the iterations of the loop are logically independent. i.e. there are no recurrence among the iterations, but they might access the same data on read or write.
In the latter case, additional synchronization like the use of atomics, mutual execution or thread-local storage is required.

I will reword the description to make it more clear.

cpp/daal/src/threading/threading.h
```diff
@@ -239,12 +258,25 @@ inline void threader_for_int64(int64_t n, const F & lambda)
     _daal_threader_for_int64(n, a, threader_func<F>);
 }
 
+/// Execute the for loop defined by the input parameters in parallel.
+/// The maximal number of iterations in the loop is 2^31 - 1.
+/// The work is scheduled dynamically across threads.
+/// The iteration space is chunked using oneTBB ``simple_partitioner``
```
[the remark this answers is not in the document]
Vika-F: I think it is impossible to say whether TBB will always be a threading layer or not. For now we do not have plans to migrate to other threading technologies like OpenMP, C++ STD threads and so on, but who knows what happens in the future?

I think I need to reword this and remove TBB mentioning from the description.

The idea of this API is that the iterations are not grouped together; each iteration is considered as a separate task.
Because other APIs can group iterations in the way that a consequent block of iterations is executed by one thread. This API tries to avoid this if possible.

cpp/daal/src/threading/threading.h
```diff
@@ -321,10 +397,20 @@ class tls_deleter_ : public tls_deleter
     virtual void del(void * a) { delete static_cast<lambdaType *>(a); }
 };
 
+/// Thread-local storage (TLS).
+/// Can change its local variable after a nested parallel constructs.
+/// @note Use carefully in case of nested parallel regions.
```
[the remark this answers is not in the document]
Vika-F: I think the confusion here is because those APIs are not intended to be used by oneDAL's users.
The APIs are just not available as part of oneDAL's release header files.
Those are only available to the developers.

Inside oneDAL we use nested parallelism in many cases. So it is important to warn the developers when using a certain constructs might lead to issues.

Will reformulate the note as well.

cpp/daal/src/threading/threading.h
```diff
@@ -339,19 +425,35 @@ class tls : public tlsBase
...
         d->del(voidLambda);
         delete d;
         _daal_del_tls_ptr(tlsPtr);
     }
 
+    /// Access a local data of a thread by value
+    ///
+    /// @return When first invoked by a thread, a lambda provided to the constructor is
+    ///         called to initialize the local data of the thread and return it.
+    ///         All the following invocations just return the same thread-local data.
```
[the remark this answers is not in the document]
Vika-F: I don't think so.
It is just an intermediate variable needed to cast from C-style API to C++.

cpp/daal/src/threading/threading.h
```diff
@@ -470,6 +604,9 @@ class static_tls
         }
     }
 
+    /// Full number of threads.
+    ///
+    /// @return Number of threads available.
```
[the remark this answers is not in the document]
Vika-F: This function returns the total number of threads available to oneDAL regardless of nesting. Will try to clarify this.

docs/source/contribution/threading.rst
Vika-F: By thread-local data I mean the data that is managed by a particular thread.
The issue here is that there is a difference between the threads and logical tasks (loop's iterations here).
One thread can be assigned to execute several iterations of the loop.

In case of nested loops the specifics of oneTBB is that when the thread executes an outer iteration `i` of a nested loop it becomes a so called waiting thread while it waits for a completion of the inner parallel loop. And then this waiting thread can be assigned to perform another iteration `j` of an outer loop.
So, it might first complete the iteration `j` of the outer loop which will change the state of thread-local values, and then return back to iteration `i` .

I tried to reflect this behavior in the documentation's text.

In general I do not agree that the thread that spawns other threads cannot have thread-local data. But I agree that thread-local data should be used with great care in those cases.

Vika-F:
@keeranroth, I think I have addressed most of your feedbacks.
Can you please take a look on this PR one more time?

keeranroth (commented):
Looking good. Mostly picked up on some typos, but wanted to double check on the documentation mentioning passing lambda functions to the threading layer. I don't think that is a restriction, and using a functor would have a similar effect, and there is no restriction on using a function that updates global state either (although that seems like a bad idea, it's not explicitly forbidden)?

cpp/daal/src/threading/threading.h
```diff
@@ -223,14 +223,41 @@ inline void threader_func_break(int i, bool & needBreak, const void * a)
...
 
+/// Pass a function to be executed in a for loop to the threading layer.
+/// The maximal number of iterations in the loop is 2^31 - 1.
+/// The default scheduling of the threading layer is used to assign
+/// the iterations of the loop to threads.
+/// The iterations of the loop should be logically independent.
+/// Data dependencies between the iterations are allowed, but may requre the use
+/// of synchronization primitives.
+///
+/// @tparam F   Lambda function of type `[/* captures */](int i) -> void`,
```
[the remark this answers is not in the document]
Vika-F: Changed the mentions of 'lambda function' to more general 'callable object'. Is it more or less Ok?

To your main comment about the possibility of updating the global state.
Yes, there are currently no programmatical restrictions on updating the global state. But this PR is focused on documenting the present state of the things. Not on the fixing/improving the code.
Also, this API is not exposed to oneDAL users. Only oneDAL developers can use this functionality. And I hope it is used and will be used with care.

keeranroth (approved):
Looks good to me. Thanks for the update

PR #2853, opened 2024-07-24
Correct PCA nComponents description

Alexsandruss:
# Description
PCA's nComponents parameter description didn't mention it works for batch mode only. This doc string is reused in [daal4py docs](https://intelpython.github.io/daal4py/algorithms.html#daal4py.pca) resulting in confusion that parameter is supported for all modes.

Addresses https://github.com/intel/scikit-learn-intelex/issues/759.

### DIFF
--- a/cpp/daal/include/algorithms/pca/pca_types.h
+++ b/cpp/daal/include/algorithms/pca/pca_types.h
@@ -670,7 +670,7 @@ public:
     BaseBatchParameter();
 
     DAAL_UINT64 resultsToCompute; /*!< 64 bit integer flag that indicates the results to compute */
-    size_t nComponents;           /*!< number of components for reduced implementation */
+    size_t nComponents;           /*!< number of components for reduced implementation (applicable for batch mode only) */
     bool isDeterministic;         /*!< sign flip if required */
     bool doScale;                 /*!< scaling if required */
     bool isCorrelation;           /*!< correlation is provided */


cpp/daal/include/algorithms/pca/pca_types.h
```diff
@@ -670,7 +670,7 @@ class DAAL_EXPORT BaseBatchParameter : public daal::algorithms::Parameter
     BaseBatchParameter();
 
     DAAL_UINT64 resultsToCompute; /*!< 64 bit integer flag that indicates the results to compute */
-    size_t nComponents;           /*!< number of components for reduced implementation */
+    size_t nComponents;           /*!< number of components for reduced implementation (applicable for batch mode only) */
```
Alexandr-Solovev, line 673: It should be applicable for distributed and online mode as well

olegkkruglov: It is tested on sklearnex side here https://github.com/intel/scikit-learn-intelex/blob/b38d0da09eff78264834ad95155c7270356c4c70/sklearnex/preview/decomposition/tests/test_incremental_pca.py#L242 that nComponents affects computations for online mode. @Alexsandruss could you please clarify how you faced the issue? Anyway we need that functionality for all modes that's why it is better to investigate and open PR with fix than with docstring change.

Alexsandruss: @Alexandr-Solovev @olegkkruglov @md-shafiul-alam

You are confusing DAAL and oneDAL interfaces. DAAL has nComponents parameter for batch mode only and daal4py autogenerated docs follow batch parameters while `daal4py.pca` is unified instance for batch, online and distributed modes.

```python
import daal4py as d4p
import numpy as np


data = np.random.rand(1000, 10)
for distributed_mode in [False, True]:
    alg = d4p.pca(nComponents=3, distributed=distributed_mode)
    result = alg.compute(data)
    print(f"distributed: {distributed_mode}, eigenvectors shape: {result.eigenvectors.shape}")

```

```
$ python test_d4p_pca.py
distributed: False, eigenvectors shape: (3, 10)
distributed: True, eigenvectors shape: (10, 10)
```

Alexandr-Solovev: I see, you are right, but may be it makes more sense to fix it in DAAL?

Alexsandruss: Knowing oneDAL interface is to replace DAAL, no.

Alexandr-Solovev (approved):

PR #2862, opened 2024-07-30
qemu version change

md-shafiul-alam:
# Description
version 9.0.1 was removed, using a slightly older version to avoid frequent update.

-
-
-

### DIFF
--- a/.ci/env/apt.sh
+++ b/.ci/env/apt.sh
@@ -63,12 +63,30 @@ function install_qemu_emulation_apt {
 }
 
 function install_qemu_emulation_deb {
-    qemu_deb=qemu-user-static_8.2.4+ds-1_amd64.deb
+    set +e
+
+    versions=(9.0.2 9.0.1 8.2.4)
+    found_version=""
+    for version in ${versions[@]}; do
+        qemu_deb="qemu-user-static_${version}+ds-1_amd64.deb"
+        echo "Checking for http://ftp.debian.org/debian/pool/main/q/qemu/${qemu_deb}"
+        if wget -q --method=HEAD http://ftp.debian.org/debian/pool/main/q/qemu/${qemu_deb} &> /dev/null;
+        then
+            echo "Found qemu version ${version}"
+            found_version=${qemu_deb}
+            break
+        fi
+    done
+
     set -eo pipefail
-    wget http://ftp.debian.org/debian/pool/main/q/qemu/${qemu_deb}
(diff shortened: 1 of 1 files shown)


napetrov:
@keeranroth  @rakshithgb-fujitsu  - please review/comment

rakshithgb-fujitsu:
Looks fine to me. Some CI cases are failing though.

.ci/env/apt.sh
```diff
@@ -63,7 +63,7 @@ function install_qemu_emulation_apt {
 }
 
 function install_qemu_emulation_deb {
-    qemu_deb=qemu-user-static_9.0.1+ds-1~bpo12+1_amd64.deb
+    qemu_deb=qemu-user-static_8.2.4+ds-1_amd64.deb
```
keeranroth, line 66: This has been flaky in the past as well. Let's pass a few version numbers through, and get the first one that exists. Something like the following seems to work (feel free to tidy up / change as necessary

```
set +e

versions=(9.0.2 9.0.1 8.2.4)
found_version=""
for version in ${versions[@]}; do
    qemu_deb="qemu-user-static_${version}+ds-1_amd64.deb"
    echo "Checking for http://ftp.debian.org/debian/pool/main/q/qemu/${qemu_deb}"
    if wget -q --method=HEAD http://ftp.debian.org/debian/pool/main/q/qemu/${qemu_deb} &> /dev/null;
    then
        echo "Found qemu version ${version}"
        found_version=${qemu_deb}
        break
    fi
done

set -eo pipefail
if [[ -z "${found_version}" ]] ; then
    # If nothing is found, error out and fail
    echo "No qemu found with version numbers ${versions[@]}"
    false
fi

wget http://ftp.debian.org/debian/pool/main/q/qemu/${found_version}
sudo dpkg -i ${found_version}
```

keeranroth (changes requested):
Maybe try adding a list of packages, so that if one goes away for a few days, we have a backup

keeranroth (approved):
Looks good to me. I don't think that the failures in the other pipelines are related, from a quick look

md-shafiul-alam:
> Looks good to me. I don't think that the failures in the other pipelines are related, from a quick look

agree, thanks for the great suggestion

ethanglaser (approved):

ethanglaser:
LinuxSklearnex fail will be addressed by https://github.com/intel/scikit-learn-intelex/pull/1982

PR #2863, opened 2024-07-31
[CI, Enhancement] generate a nightly oneDAL build for sklearnex Github Actions

icfaust:
### Description 
This PR introduces a nightly Github action CI job which will create a DPCPP oneDAL build using the main oneDAL branch.  This will be reused for CI Github action (in intel/scikit-learn-intelex#1845) which will build sklearnex using this oneDAL build for testing.  This enables public CI testing of sklearnex with a day-old oneDAL version. This contrasts from current public CI which uses the last released version of oneDAL, which can be missing features.

- Linux will store the complete __release_lnx from an icx build.
- Windows will store the dpcpp compiler, opencl cpu runtime, and __release_win from an vc/icx combination build. The additional items are necessary due to lack of dpcpp/opencl-rt availability on windows package managers and the time necessary to build and install them.


icfaust:
Temporarily running on pull requests just to verify operation. Will be removed upon green.

.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,132 @@
...
+    if: github.repository == 'oneapi-src/oneDAL'
+    runs-on: windows-2022
+    timeout-minutes: 120
+
+    steps:
+      - name: Checkout oneDAL
+        uses: actions/checkout@v4
+        with:
+          repository: oneapi-src/oneDAL
+      - name: Install DPC++
```
napetrov, line 70: can we unify things with bat scripts used in azure? https://github.com/oneapi-src/oneDAL/blob/main/.ci/pipeline/ci.yml#L695

so we will be using same process regardless of pipeline?

icfaust: @napetrov thank you for the suggestion. i have converted aspects into powershell files and made adaptations to minimize hard drive usage. Unfortunately it looks like even then the DevOps runner memory is insufficient. A second worry is that it pushes VC build times close to 3 hours. I could solve this by doing a similar DPCPP compiler preparation in a special DevOps nightly runner generating a compiler artifact. That would remove the 45 min compiler install and may reduce the memory footprint, but i would need to learn a bit more about DevOps and would need to test. even then it may not work (not sure how much memory would be saved, whether the compiler would work on a separate DevOps runner). let me know your thoughts. I would lean towards a follow up task for this.

.ci/scripts/install_dpc.ps1
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+echo "Download intel DPC++ compiler"
+(new-object System.Net.WebClient).DownloadFile("https://registrationcenter-download.intel.com/akdlm/IRC_NAS/7991e201-ca0f-4689-bdb6-1ed73a8246fd/w_dpcpp-cpp-compiler_p_2024.2.0.491_offline.exe", "dpcpp_installer.exe")
+# wget -q -O dpcpp_installer.exe
```
Alexsandruss, line 18: What does this comment mean? I guess it should be above previous line.

.ci/scripts/install_dpc.ps1
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+#===============================================================================
+echo "Download intel DPC++ compiler"
+(new-object System.Net.WebClient).DownloadFile("https://registrationcenter-download.intel.com/akdlm/IRC_NAS/7991e201-ca0f-4689-bdb6-1ed73a8246fd/w_dpcpp-cpp-compiler_p_2024.2.0.491_offline.exe", "dpcpp_installer.exe")
+# wget -q -O dpcpp_installer.exe 
+echo "Unpacking DPC++ installer"
+Start-Process ".\dpcpp_installer.exe" -ArgumentList "--s --x --f oneAPI" -Wait
+echo "Installing DPC++ compiler"
+Start-Process ".\oneAPI\bootstrapper.exe" -ArgumentList "-s --eula=accept --install-dir=dpcpp" -Wait
+# This can take 20 minutes...
```
Alexsandruss, line 23: Explicit or clear comments are better and faster understandable
```suggestion
# Removal of files can take up to 20 minutes
```
```suggestion
# Removal of files can take nearly 20 minutes
```

.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,116 @@
...
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  build_lnx:
+    name: build Linux
+    if: github.repository == 'oneapi-src/oneDAL'
+    runs-on: ubuntu-latest
```
Alexsandruss, line 31: Accidental failures due to non-pinned image version might cause interruption of development process.
```suggestion
    runs-on: ubuntu-22.04
```
```suggestion
    runs-on: ubuntu-24.04
```

Alexsandruss (changes requested):

icfaust:
Re-enabled for running on PR to verify operation with ```ubuntu-22.04``` instead of ```ubuntu-latest```

Alexsandruss (approved):

PR #2876, opened 2024-08-16
add sve capability check at runtime

rakshithgb-fujitsu:
# Description
oneDAL is currently forced to use only SVE on arm platforms, with this new additional check it will validate the presence of SVE.

It checks for the SVE using hardware capabilities mentioned here - https://www.kernel.org/doc/html/v5.15/arm64/sve.html

### DIFF
--- a/.ci/scripts/clang-format.sh
+++ b/.ci/scripts/clang-format.sh
@@ -20,9 +20,11 @@ echo "Starting format check..."
 
 RETURN_CODE=0
 
+CLANG_FORMAT_EXE=${CLANG_FORMAT_EXE:-clang-format-14}
+
 for sources_path in cpp/daal cpp/oneapi examples/oneapi examples/daal samples/oneapi samples/daal; do
     pushd ${sources_path} || exit 1
-    for filename in $(find . -type f | grep -P ".*\.(c|cpp|h|hpp|cl|i)$"); do clang-format-14 -style=file -i "${filename}"; done
+    for filename in $(find . -type f | grep -P ".*\.(c|cpp|h|hpp|cl|i)$"); do ${CLANG_FORMAT_EXE} -style=file -i "${filename}"; done
 
     git status | grep "nothing to commit" > /dev/null
 
--- a/.ci/scripts/clang-format.sh
+++ b/.ci/scripts/clang-format.sh
@@ -22,7 +22,7 @@ RETURN_CODE=0
 
 for sources_path in cpp/daal cpp/oneapi examples/oneapi examples/daal samples/oneapi samples/daal; do
     pushd ${sources_path} || exit 1
-    for filename in $(find . -type f | grep -P ".*\.(c|cpp|h|hpp|cl|i)$"); do clang-format -style=file -i "${filename}"; done
+    for filename in $(find . -type f | grep -P ".*\.(c|cpp|h|hpp|cl|i)$"); do clang-format-14 -style=file -i "${filename}"; done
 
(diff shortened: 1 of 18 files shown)


rakshithgb-fujitsu:
@napetrov @keeranroth have a look at this PR, this lays the base for use to push SVE tuned kernels.

keeranroth (changes requested):
Looks good. Just thinking, if SVE is not supposed to be Arm specific, is it worth renaming the macro from `TARGET_ARM` to `TARGET_AARCH64` to show that it's the ISA that is being checked?

rakshithgb-fujitsu:
> Looks good. Just thinking, if SVE is not supposed to be Arm specific, is it worth renaming the macro from `TARGET_ARM` to `TARGET_AARCH64` to show that it's the ISA that is being checked?

I think its better to keep arm, eventually when neon support is added it can be the default path.

.ci/scripts/clang-format.sh
```diff
@@ -22,7 +22,7 @@ RETURN_CODE=0
 
 for sources_path in cpp/daal cpp/oneapi examples/oneapi examples/daal samples/oneapi samples/daal; do
     pushd ${sources_path} || exit 1
-    for filename in $(find . -type f | grep -P ".*\.(c|cpp|h|hpp|cl|i)$"); do clang-format -style=file -i "${filename}"; done
+    for filename in $(find . -type f | grep -P ".*\.(c|cpp|h|hpp|cl|i)$"); do clang-format-14 -style=file -i "${filename}"; done
```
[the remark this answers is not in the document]
rakshithgb-fujitsu: The current ci depends on this version, but yes this is a neater solution. Thanks!

keeranroth (approved):

Alexandr-Solovev (approved):

.ci/scripts/clang-format.sh
```diff
@@ -20,9 +20,11 @@ echo "Starting format check..."
 
 RETURN_CODE=0
 
+CLANG_FORMAT_EXE=${CLANG_FORMAT_EXE:-clang-format-14}
```
ethanglaser, line 23: Is it necessary to set specific version here as default? Because we generally do not update this file, the default clang-format option would be used which provides some flexibility

Alexsandruss, line 23: Since specific version is going to be used, clang-format alias setup should be removed from `.ci/env/apt.sh`:
https://github.com/oneapi-src/oneDAL/blob/2b18f4d4e1f3c3851c6af7b42d59d4b388728e40/.ci/env/apt.sh#L43-L47

Alexsandruss (changes requested):

.ci/scripts/clang-format.sh
rakshithgb-fujitsu: I observed this issue since the current CI uses `clang-format-14`, our systems had newer versions. The newer formatting was not compatible with `clang-format-14` and hence failing the CI check. This allows to use the default version based on CI version.

rakshithgb-fujitsu: This is used by the CI right? It is installing a specific version which allows us to set the default version to this.

Alexsandruss: Last two lines set generic `clang-format` alias which should be removed for consistency since specific version is going to be used.

rakshithgb-fujitsu: done.

Alexsandruss (approved):

PR #2878, opened 2024-08-22
Update deprecated functions in RF and GBT Examples

md-shafiul-alam:
# Description
There are a few functions in Random Forest and Gradient Boosted tree in DAAL that have been deprecated and replaced by newer definitions which expects one or two parameters more. Because of the deprecation flags CI throws warnings on windows. The deprecated functions are being used from examples and some tests on ci. This PR modifies the examples to use newer definitions. Public CI failure will be addressed on https://github.com/oneapi-src/oneDAL/pull/2891

### DIFF
--- a/cpp/daal/include/algorithms/decision_forest/decision_forest_classification_model_builder.h
+++ b/cpp/daal/include/algorithms/decision_forest/decision_forest_classification_model_builder.h
@@ -119,14 +119,6 @@ public:
     }
 
     /**
-    *  \DAAL_DEPRECATED
-    */
-    DAAL_DEPRECATED NodeId addLeafNode(const TreeId treeId, const NodeId parentId, const size_t position, const size_t classLabel)
-    {
-        return addLeafNode(treeId, parentId, position, classLabel, 0);
-    }
-
-    /**
     *  Create Leaf node and add it to certain tree
     *  \param[in] treeId          Tree to which new node is added
     *  \param[in] parentId        Parent node to which new node is added (use noParent for root node)
@@ -141,14 +133,6 @@ public:
         _status |= addLeafNodeByProbaInternal(treeId, parentId, position, proba, cover, resId);
         services::throwIfPossible(_status);
         return resId;
-    }
-
-    /**
-    *  \DAAL_DEPRECATED
-    */
-    DAAL_DEPRECATED NodeId addLeafNodeByProba(const TreeId treeId, const NodeId parentId, const size_t position, const double * const proba)
-    {
-        return addLeafNodeByProba(treeId, parentId, position, proba, 0);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


ethanglaser:
I am not sure what the plan for removal of the deprecated interfaces is but I would leave these in the repo for now. Would it be possible just to update the examples so that they are using the existing non-deprecated replacement interfaces?

md-shafiul-alam:
> I am not sure what the plan for removal of the deprecated interfaces is but I would leave these in the repo for now. Would it be possible just to update the examples so that they are using the existing non-deprecated replacement interfaces?

It was an interesting decision to deprecate functions which would have been just fine to add the new parameters with default values. The deprecated functions do exactly that. It doesn't seem to me that the deprecation was necessary. For the matter of removing the deprecated functions, there is actually no functional difference. If anyone calls with previous definitions, they will still work. However, I am open to go either way.

ethanglaser:
@Alexsandruss thoughts?

md-shafiul-alam:
> @Alexsandruss thoughts?

+ @Alexandr-Solovev

cpp/daal/include/algorithms/decision_forest/decision_forest_classification_model_builder.h
```diff
@@ -110,22 +110,14 @@ class DAAL_EXPORT ModelBuilder
...
         return resId;
     }
 
-    /**
-    *  \DAAL_DEPRECATED
-    */
-    DAAL_DEPRECATED NodeId addLeafNode(const TreeId treeId, const NodeId parentId, const size_t position, const size_t classLabel)
-    {
-        return addLeafNode(treeId, parentId, position, classLabel, 0);
-    }
```
Alexsandruss, line 127: Major binary version of library should be increased since ABI would change and I'm unsure if it's feasible for next major release as some additional work related to versioned shared libraries is needed.
I'm OK with other non-deletion changes, but you should verify binary backward compatibility on both of Linux and Windows.

Alexsandruss (changes requested):

cpp/daal/include/algorithms/decision_forest/decision_forest_classification_model_builder.h
md-shafiul-alam: Do you suggest not deleting for now?

Alexsandruss: Yes, postpone PR merge until next major release and then bump major binary version.

md-shafiul-alam:
@Alexsandruss Ignored any API changes for this PR. Would you please review again?

Alexandr-Solovev:
Should we remove deprecated functions in the same time as update examples?

Alexandr-Solovev:
and we increased major binary version, so I guess now it makes sense to return function deletions

cpp/daal/include/algorithms/decision_forest/decision_forest_classification_model_builder.h
```diff
@@ -110,22 +110,14 @@ class DAAL_EXPORT ModelBuilder
     *  \param[in] cover           Cover (Hessian sum) of the node
     *  \return Node identifier
     */
-    NodeId addLeafNode(const TreeId treeId, const NodeId parentId, const size_t position, const size_t classLabel, const double cover)
+    NodeId addLeafNode(const TreeId treeId, const NodeId parentId, const size_t position, const size_t classLabel, const double cover = 0.0)
```
ethanglaser, line 113: For example, I would not modify this line

md-shafiul-alam:
Private CI is supposed to fail because new_ts tests build step will fail because of API break. Here is the job with updated infra branch: https://intel-ci.intel.com/ef6b44cb-fdf0-f1fb-b313-a4bf010d0e2e

ethanglaser (approved):

Alexandr-Solovev (approved):

md-shafiul-alam:
The CI timeout and warning are unrelated to the PR.

PR #2880, opened 2024-08-22
Performance fix for ensemble classifier inference

Alexsandruss:
# Description

Disables ensemble level parallelism in classifier inference to avoid excessive memory and CPU usage on machines with higher number of cores.


Alexsandruss:
Private CI is bugged in this PR for some reason, replacement PR: https://github.com/oneapi-src/oneDAL/pull/2881

PR #2881, opened 2024-08-26
Perf. fix for decision forest classifier inference

Alexsandruss:
# Description

Disables ensemble level parallelism in classifier inference to avoid excessive memory and CPU usage on machines with higher number of cores.


Alexsandruss:
@mergifyio rebase

Alexandr-Solovev (approved):

PR #2882, opened 2024-08-28
[fix] Fix online SPMD algorithms finalize call

olegkkruglov:
- Fix finalize method for all four online SPMD algorithms, now the call does not affect partial results since all necessary tables are copied. It allows to call finalize between calls of partial_fit to obtain intermediate results
- Add ONEDAL_DATA_PARALLEL to finalize implementations where they were forgotten

### DIFF
--- a/cpp/oneapi/dal/algo/basic_statistics/backend/gpu/finalize_compute_kernel_dense_impl_dpc.cpp
+++ b/cpp/oneapi/dal/algo/basic_statistics/backend/gpu/finalize_compute_kernel_dense_impl_dpc.cpp
@@ -153,16 +153,21 @@ result_t finalize_compute_kernel_dense_impl<Float>::operator()(const descriptor_
     const auto nobs_nd = pr::table2ndarray_1d<Float>(q, input.get_partial_n_rows());
 
     auto rows_count_global = nobs_nd.get_data()[0];
+    auto is_distributed = (comm_.get_rank_count() > 1);
     {
         ONEDAL_PROFILER_TASK(allreduce_rows_count_global);
-        comm_.allreduce(rows_count_global, spmd::reduce_op::sum).wait();
+        if (is_distributed) {
+            comm_.allreduce(rows_count_global, spmd::reduce_op::sum).wait();
+        }
     }
     if (res_op.test(result_options::min)) {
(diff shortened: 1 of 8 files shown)


cpp/oneapi/dal/algo/basic_statistics/backend/gpu/finalize_compute_kernel_dense_impl_dpc.cpp
```diff
@@ -174,23 +176,44 @@ result_t finalize_compute_kernel_dense_impl<Float>::operator()(const descriptor_
     }
 
     if (res_op_partial.test(result_options::sum)) {
-        const auto sums_nd =
+        auto sums_nd =
             pr::table2ndarray_1d<Float>(q, input.get_partial_sum(), sycl::usm::alloc::device);
+        if (comm_.get_rank_count() > 1) {
```
Alexandr-Solovev, line 181: Its better to call it only once, and copy all partial results in one if condition. Also you can potentially get minor perf improvements on 1 gpu, if you call allreduce inside the if.

cpp/oneapi/dal/algo/basic_statistics/test/online_spmd.cpp
```diff
@@ -103,8 +104,9 @@ class basic_statistics_online_spmd_test
                 }
                 partial_results.push_back(partial_result);
             }
-            const auto compute_result = this->finalize_compute_override(bs_desc, partial_results);
-
+            auto compute_result = this->finalize_compute_override(bs_desc, partial_results);
+            base_t::check_compute_result(compute_mode, data, table{}, compute_result);
+            compute_result = this->finalize_compute_override(bs_desc, partial_results);
```
Alexandr-Solovev, line 109: I guess its better to add a separate test with partial_fit finalize_fit partial_fit and finalize_fit, instead of checking result of twice called finalize_fit

olegkkruglov: it would require significant code size and test runtime increase, current test change allows to check if finalize works correctly.  i'll add such scenario in intelex tests.

olegkkruglov:
http://intel-ci.intel.com/ef65eab3-b487-f188-a43a-a4bf010d0e2e

Alexandr-Solovev (approved):

PR #2890, opened 2024-09-04
[bug] Coverity fixes 2025.0

icfaust:
# Description
Make coverity fixes for 2025.0

-  Remove bad bounds check from KNN classification using kdtrees

### DIFF
--- a/cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
@@ -492,7 +492,7 @@ algorithmFpType KNNClassificationTrainBatchKernel<algorithmFpType, training::def
     algorithmFpType * subSamples, size_t subSampleCapacity, Status & status)
 {
     algorithmFpType samples[__KDTREE_MEDIAN_RANDOM_SAMPLE_COUNT + 1];
-    const size_t sampleCount = sizeof(samples) / sizeof(samples[0]);
+    const size_t sampleCount = __KDTREE_MEDIAN_RANDOM_SAMPLE_COUNT + 1;
 
     if (end - start <= sampleCount)
     {
--- a/cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
@@ -589,6 +589,10 @@ algorithmFpType KNNClassificationTrainBatchKernel<algorithmFpType, training::def
 
     size_t sumMid = 0;
     size_t i      = 0;
+    // this iterates through the masterHist histogram and finds the median
+    // in almost all circumstances the break in the if statement will trigger
+    // unless masterHist does not contain sufficient data to exceed
+    // (end - start)/2 by the last bin.
     for (; i < sampleCount - 1; ++i)
     {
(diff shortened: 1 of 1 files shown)


cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
```diff
@@ -589,7 +589,7 @@ algorithmFpType KNNClassificationTrainBatchKernel<algorithmFpType, training::def
 
     size_t sumMid = 0;
     size_t i      = 0;
-    for (; i < sampleCount; ++i)
+    for (; i < __KDTREE_MEDIAN_RANDOM_SAMPLE_COUNT; ++i)
```
Alexandr-Solovev, line 592: algorithmFpType samples[__KDTREE_MEDIAN_RANDOM_SAMPLE_COUNT + 1];
    const size_t sampleCount = sizeof(samples) / sizeof(samples[0]);
    should it be __KDTREE_MEDIAN_RANDOM_SAMPLE_COUNT + 1?

icfaust: good question, see this line https://github.com/icfaust/oneDAL/blob/dev/coverity-2025_0/cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i#L601  when this for loop terminates ```i == sampleCount``` (```__KDTREE_MEDIAN_RANDOM_SAMPLE_COUNT + 1```) so when it is used on that line, it will exceed the array bound.

Alexandr-Solovev: good point, agree with you,  but probably its a bug, check the line
https://github.com/oneapi-src/oneDAL/pull/2890/files#diff-e5b67a6ac7d03badd31a155779cec86abb98f260008cd311acf5166a0102b123R519
does it make sense to do the same fix here? My point that if we replace sampleCount var with __KDTREE_MEDIAN_RANDOM_SAMPLE_COUNT it makes sense to remove it everywhere, but probably its more correct to fix it as its done at the line 519

avolkov-intel: What do you think about initialize sampleCount this way in this PR:
const size_t sampleCount =  __KDTREE_MEDIAN_RANDOM_SAMPLE_COUNT + 1

avolkov-intel:
What I'm concerned about is that masterHist array is filled from 0 to sampleCount - 1 inclusive https://github.com/oneapi-src/oneDAL/pull/2890/files#diff-e5b67a6ac7d03badd31a155779cec86abb98f260008cd311acf5166a0102b123R580

If we apply the change from the PR the last value from masterHist array will not be used at all and seems wrong to me, so there is some mistake here. Maybe it is in the line https://github.com/oneapi-src/oneDAL/pull/2890/files#diff-e5b67a6ac7d03badd31a155779cec86abb98f260008cd311acf5166a0102b123L601 ?

cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
icfaust: done. Still don't understand why they had it as sizeof, I think @Alexandr-Solovev had a good theory.

avolkov-intel (approved):
LGTM, assume CI is green

Alexandr-Solovev (approved):

PR #2894, opened 2024-09-09
fix: resolve omp simd error in mdd

ethanglaser:
Avoids combination of vectorization and debug mode by only adding omp simd flag when debug flag is not added. This bypasses an error on windows build with -MDd: `using /MTd or /MDd with '#pragma omp simd' may lead to unexpected fails due to the debugging version of iterators that cannot be vectorized correctly`

Job without these changes: http://intel-ci.intel.com/ef6eac3e-3f98-f19a-9f5d-a4bf010d0e2e
Job with these changes: http://intel-ci.intel.com/ef6efb69-4c4f-f1a5-b523-a4bf010d0e2e
(only consider windows OneDAL build step, other errors exist for other reasons)

### DIFF
--- a/cpp/daal/src/externals/service_math_ref.h
+++ b/cpp/daal/src/externals/service_math_ref.h
@@ -72,17 +72,13 @@ struct RefMath<double, cpu>
 
     static void vPowx(SizeType n, const double * in, double in1, double * out)
     {
-#ifdef NDEBUG
 #pragma omp simd
-#endif
         for (SizeType i = 0; i < n; ++i) out[i] = pow(in[i], in1);
     }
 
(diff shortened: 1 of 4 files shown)


icfaust:
Looks simple enough, @ethanglaser could you give a description as to why it needs to only be implemented when MSVC_RT_is_release is true?

ethanglaser:
> Looks simple enough, @ethanglaser could you give a description as to why it needs to only be implemented when MSVC_RT_is_release is true?

Yes good point, description added

icfaust (approved):

homksei (approved):

syakov-intel (approved):
approved to merge

PR #2896, opened 2024-09-12
Add DEPENDENTLOAD flag to vc linker on Windows

olegkkruglov:
# Description
- DEPENDENTLOADFLAG is added to vc linker to prevent loading DLLs from current directory. https://learn.microsoft.com/en-us/cpp/build/reference/dependentloadflag?view=msvc-170


homksei (approved):
Looks good, but please run CI once more.

PR #2901, opened 2024-09-17
chore(ci): add `read` permissions for `label-enforcement` workflow

homksei:


Alexsandruss:
Why is `read` permission required if check seems to work fine?

homksei:
> Why is `read` permission required if check seems to work fine?

It does not affect functionality. It is a security reason: https://github.com/ossf/scorecard/blob/ea7e27ed41b76ab879c862fa0ca4cc9c61764ee4/docs/checks.md#token-permissions

Alexsandruss (approved):

PR #2905, opened 2024-09-19
Add sycl headers to bazel toolchain config for DPCPP through icx

david-cortes-intel:
# Description
There is a step in the bazel test pipeline for `//cpp/oneapi/dal:tests` in which it probes the compiler for include paths of headers to add to the toolchain config:
https://github.com/oneapi-src/oneDAL/blob/b6b6a65a59b46d6770198abfc11cfa476c1338a2/dev/bazel/toolchains/cc_toolchain_lnx.bzl#L103

It does so by requesting the default paths from the compiler with this command:
```shell
<compiler> -E -xc++ - -v
```

In the case of DPCPP, this resolves to:
```shell
icx -E -xc++ - -v
```

By default, `icx` doesn't include sycl linkage, thus the command above does not return the paths of SYCL headers. Under the default `icx` factory settings, that leads to observing an error:
```
the source file 'cpp/oneapi/dal/algo/pca/test/serialization.cpp' includes the following non-builtin files with absolute paths (if these are builtin files, make sure these paths are in your toolchain):
  '/opt/intel/oneapi/compiler/2024.2/include/sycl/sycl.hpp'
```

This PR should address it by adding the necessary `-fsycl` flag when probing the compiler for its header paths.

### DIFF
--- a/dev/bazel/toolchains/cc_toolchain_lnx.bzl
+++ b/dev/bazel/toolchains/cc_toolchain_lnx.bzl
@@ -129,7 +129,7 @@ def _preapre_builtin_include_directory_paths(repo_ctx, tools):
             "-xc++",
             get_no_canonical_prefixes_opt(repo_ctx, tools.dpcc) +
             _add_gcc_toolchain_if_needed(repo_ctx, tools.dpcc) +
-            _add_sycl_linkage(repo_ctx, tools.dpcc),
+            _add_sycl_linkage(repo_ctx, tools.dpcc) if tools.is_dpc_found else [],
         ) +
         required_tmp_includes,
     )
--- a/dev/bazel/toolchains/cc_toolchain_lnx.bzl
+++ b/dev/bazel/toolchains/cc_toolchain_lnx.bzl
@@ -128,7 +128,8 @@ def _preapre_builtin_include_directory_paths(repo_ctx, tools):
             tools.dpcc,
             "-xc++",
             get_no_canonical_prefixes_opt(repo_ctx, tools.dpcc) +
-            _add_gcc_toolchain_if_needed(repo_ctx, tools.dpcc),
+            _add_gcc_toolchain_if_needed(repo_ctx, tools.dpcc) +
+            _add_sycl_linkage(repo_ctx, tools.dpcc),
         ) +
         required_tmp_includes,
     )
@@ -151,6 +152,12 @@ def _get_gcc_toolchain_path(repo_ctx):
 def _add_gcc_toolchain_if_needed(repo_ctx, cc):
     if ("clang" in cc) or ("icpx" in cc):
         return ["--gcc-toolchain=" + _get_gcc_toolchain_path(repo_ctx)]
+    else:
+        return []
+
+def _add_sycl_linkage(repo_ctx, cc):
+    if ("icx" in cc) or ("icpx" in cc):
+        return ["-fsycl"]
     else:
         return []
 


dev/bazel/toolchains/cc_toolchain_lnx.bzl
```diff
@@ -154,6 +155,12 @@ def _add_gcc_toolchain_if_needed(repo_ctx, cc):
     else:
         return []
 
+def _add_sycl_linkage(repo_ctx, cc):
+    if ("icx" in cc) or ("icpx" in cc):
```
napetrov, line 159: i think we should check also for mode. i.e. we can use icx for CPU only mode. with this change we always would be adding fsycl

david-cortes-intel: Thanks for pointing this out. I pushed a commit which I think should fix it, by checking that the tooling has 'dpc' as done elsewhere in the script.

Although I'm not sure if that's the correct solution - it seems it will just assume everything needs DPCPP linkage whenever the compiler is 'icpx':
https://github.com/oneapi-src/oneDAL/blob/9b315cf8124c700f79bdc3007aac103df3b831bf/dev/bazel/toolchains/cc_toolchain.bzl#L23

which is also used for deciding flags in other parts of the build script:
https://github.com/oneapi-src/oneDAL/blob/9b315cf8124c700f79bdc3007aac103df3b831bf/dev/bazel/toolchains/cc_toolchain_lnx.bzl#L334

david-cortes-intel: I _suspect_ that the failing CI job after this last change:
```
.ci/scripts/test.sh: line 248: 46771 Aborted                 (core dumped) ${run_command} > "${e}".res 2>&1
terminate called after throwing an instance of 'daal::services::interface1::Exception'
  what():  NoErrorMessageFound
07:06:15 FAILED		pca_svd_dense_batch with errno 134
```

might have uncovered a different issue: it looks like the job `LinuxMakeGNU_OpenBLAS_x86` is executing SYCL examples, as those are auto-generated under `__release_lnx_gnu`.

The GCC compiler doesn't support SYCL so the examples shouldn't include them in that build, but apparently the job was somehow running them.

napetrov: examples/daal/cpp doesn't contain any SYCL/GPU related stuff.  I can't tell tis ether from log message. 

O would try rerunning public job again to start with. But it would be interesting to see results from internal runs as well

dev/bazel/toolchains/cc_toolchain_lnx.bzl
```diff
@@ -154,6 +155,12 @@ def _add_gcc_toolchain_if_needed(repo_ctx, cc):
...
+        return ["-fsycl"]
```
Alexandr-Solovev, line 160: Do we want to extend these flags with -fsycl-device-code-split=per_kernel, we use it in dpcpp.mk?

Alexandr-Solovev (commented):
Overall looks good to me, I started private ci job with extended number of test cases via /intelci:run , lets wait the private job before merging

dev/bazel/toolchains/cc_toolchain_lnx.bzl
david-cortes-intel: This is currently only being used in order to get the include paths of the compiler when it uses SYCL. From some experiments, it looks like adding `-fsycl-device-code-split=per_kernel` on top of `-fsycl` doesn't add any new include paths.

The flag `-fsycl-device-code-split=per_kernel` is however added as a compiler argument later on:
https://github.com/oneapi-src/oneDAL/blob/488c4006a5a2f29c272c1c48e71b7a3a2ece5a61/dev/bazel/toolchains/cc_toolchain_lnx.bzl#L248

Alexandr-Solovev (approved):

napetrov (approved):

PR #2910, opened 2024-09-24
fix: adding daal exports for vc compiler

Alexandr-Solovev:
# Description

-
-
-

### DIFF
--- a/cpp/daal/src/algorithms/elastic_net/elastic_net_training_result_fpt.cpp
+++ b/cpp/daal/src/algorithms/elastic_net/elastic_net_training_result_fpt.cpp
@@ -40,7 +40,7 @@ using namespace daal::services;
  * \param[in] method Computation method for the algorithm
  */
 template <typename algorithmFPType>
-Status Result::allocate(const daal::algorithms::Input * input, const Parameter * parameter, const int method)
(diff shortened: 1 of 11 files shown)


cpp/daal/src/algorithms/elastic_net/elastic_net_training_result_fpt.cpp
```diff
@@ -40,7 +40,7 @@ using namespace daal::services;
  * \param[in] method Computation method for the algorithm
  */
 template <typename algorithmFPType>
-Status Result::allocate(const daal::algorithms::Input * input, const Parameter * parameter, const int method)
+DAAL_EXPORT services::Status Result::allocate(const daal::algorithms::Input * input, const Parameter * parameter, const int method)
```
david-cortes-intel, line 43: Question: shouldn't this `DAAL_EXPORT` go into the header that offers the user-accessible function?

david-cortes-intel: Ok, so I see that the external-facing headers (path `cpp/daal/include`) are not included in these `.cpp` files and that's why the `declspec` doesn't propagate to them.

I guess then this PR should solve the issue, but perhaps could be safer to include the headers in the .cpp files just in case to make sure that the declarations match the implementations up to details like these declspecs.

david-cortes-intel (approved):

PR #2923, opened 2024-09-30
Cover more topics in installation instructions

david-cortes-intel:
# Description
This PR updates the installation instructions to more accurately reflect the requirements when building from source, and to point at ways of executing the provided examples and unit tests.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,7 +20,7 @@
 Required Software:
 * C/C++ Compiler
 * [DPC++ Compiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/dpc-compiler.html) if building with SYCL support
-* Python version 3.7 or higher
+* Python version 3.9 or higher
 * TBB library (repository contains script to download it)
 * Microsoft Visual Studio\* (Windows\* only)
 * [MSYS2](http://msys2.github.io) (Windows\* only)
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -19,7 +19,9 @@
 
 Required Software:
 * C/C++ Compiler
-* [DPC++ Compiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/dpc-compiler.html)
+* [DPC++ Compiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/dpc-compiler.html) if building with SYCL support
+* Python version 3.7 or higher
```
napetrov, line 23: i think 3.7 as well as 3.8 are not officially supported

david-cortes-intel: Thanks. Updated it to mention version 3.9 or higher.

INSTALL.md
```diff
@@ -19,7 +19,9 @@
 
 Required Software:
 * C/C++ Compiler
-* [DPC++ Compiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/dpc-compiler.html)
+* [DPC++ Compiler](https://www.intel.com/content/www/us/en/developer/tools/oneapi/dpc-compiler.html) if building with SYCL support
+* Python version 3.9 or higher
+* TBB library (repository contains script to download it)
```
Alexandr-Solovev, line 24: there is a tricky thing, that depends on tbb version we must use MKL_FPK_GPU_VERSION_LINE=2024.0. This var controls what layout of tbb will be used. If tbb < 2024.7 its not necessary to use this flag, if higher, its necessary.

Alexandr-Solovev (approved):
LGTM!

INSTALL.md
david-cortes-intel: Questions in this regard:
* Isn't the latest version of TBB set at `2021.13.0`?
* If the build script sources the TBB script that sets the environment variables, would it be possible to retrieve the version from the `.pc` file? Under the default oneapi-basekit install, it should be under `/opt/intel/oneapi/tbb/latest/lib/pkgconfig/tbb.pc`.

napetrov (approved):

PR #333, opened 2020-02-02
Decision forest model builder extension for probabilities has been added

masdevas:
Changes proposed in this pull request:
- New interface for building models

PR #2925, opened 2024-09-30
fix: kdd tree knn example segfault

Alexandr-Solovev:
# Description

-
-
-


cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
```diff
@@ -334,39 +334,37 @@ Status KNNClassificationTrainBatchKernel<algorithmFpType, training::defaultDense
...
-                        if (b.lower > dx[indexes[i]])
-                        {
-                            b.lower = dx[indexes[i]];
-                        }
-                        if (b.upper < dx[indexes[i]])
-                        {
-                            b.upper = dx[indexes[i]];
-                        }
-                    }
+            const size_t first = iBlock * rowsPerBlock;
```
Vika-F, line 338: Please add error handling. The check on the line #333 is probably useless as the lambda at the line #319 is called only when the .local() method is invoked inside daal::threader_for.
```suggestion
            DAAL_CHECK_MALLOC_THR(bboxLocal);
            const size_t first = iBlock * rowsPerBlock;
```

cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
```diff
@@ -334,39 +334,37 @@ Status KNNClassificationTrainBatchKernel<algorithmFpType, training::defaultDense
...
+                if (bboxLocal->upper < b.upper)
+                {
+                    bboxLocal->upper = b.upper;
+                }
+                if (bboxLocal->lower > b.lower)
+                {
+                    bboxLocal->lower = b.lower;
+                }
             }
         });
```
Vika-F, line 370: Error handling is needed here:
DAAL_CHECK_SAFE_STATUS()

cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
Alexandr-Solovev: resolved

cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_predict_dense_default_batch_impl.i
```diff
@@ -173,115 +172,107 @@ Status KNNClassificationPredictKernel<algorithmFpType, defaultDense, cpu>::compu
...
         return ptr;
     });
 
     DAAL_CHECK_STATUS_OK((status.ok()), status);
-
     const auto maxThreads     = threader_get_threads_number();
+    auto nThreads             = (maxThreads < 1) ? 1 : maxThreads;
     const size_t xColumnCount = x->getNumberOfColumns();
-    const auto rowsPerBlock   = (xRowCount + maxThreads - 1) / maxThreads;
+    const auto rowsPerBlock   = 128;
```
samir-nasibli, line 206: @Alexandr-Solovev could you please update the description?
I didn't understand why this number hardcoded here?

cpp/daal/src/externals/service_math.h
```diff
@@ -50,18 +50,32 @@ struct Math
 
     static fpType sPowx(fpType in, fpType in1) { return _impl<fpType, cpu>::sPowx(in, in1); }
 
+    static fpType xsPowx(fpType in, fpType in1) { return _impl<fpType, cpu>::xsPowx(in, in1); }
```
david-cortes-intel, line 53: Could you please leave some comments in this header about the differences between the functions starting with 's' and the functions starting with 'xs' and when should each be called.

cpp/daal/src/externals/service_math_ref.h
```diff
@@ -58,48 +58,88 @@ struct RefMath<double, cpu>
...
+
     static void vExp(SizeType n, const double * in, double * out)
     {
 #pragma omp simd
         for (SizeType i = 0; i < n; ++i) out[i] = exp(in[i]);
     }
 
+    static void xvExp(SizeType n, const double * in, double * out)
+    {
+        for (SizeType i = 0; i < n; ++i) out[i] = exp(in[i]);
```
david-cortes-intel, line 140: Functions like this would run faster when using multiple threads. MKL runs then multi-threaded as far as I can tell.

cpp/daal/src/externals/service_math_ref.h
```diff
@@ -58,48 +58,88 @@ struct RefMath<double, cpu>
...
+
     static double sCeil(double in) { return ceil(in); }
 
+    static double xsCeil(double in) { return ceil(in); }
+
     // Not implemented
     static double sErfInv(double in) { return std::numeric_limits<double>::quiet_NaN(); }
 
+    // Not implemented
+    static double xsErfInv(double in) { return std::numeric_limits<double>::quiet_NaN(); }
```
david-cortes-intel, line 71: I haven't looked at what exact compilation arguments go into the `icx` invocations, but isn't the default for it to have the equivalent of GCC's `-ffast-math` which implicitly assumes that NaNs and Infs will not be encountered?

If so, would this NaN be checked somewhere? Could it throw an exception instead?

And how about taking these functions from the `cephes` library from Netlib which is public domain?

cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_predict_dense_default_batch_impl.i
Alexandr-Solovev: it was for testing purposes only, already returned

cpp/daal/src/externals/service_math_ref.h
Alexandr-Solovev: yes, you are right, but I wanted to add the section about these functions usage in threader_for(multithread for), because in such case to avoid calling multithread function inside the multithread loop we must use xFunc(single thread)

cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_predict_dense_default_batch_impl.i
```diff
@@ -39,7 +39,7 @@
 #include "src/algorithms/k_nearest_neighbors/kdtree_knn_classification_model_impl.h"
 #include "src/algorithms/k_nearest_neighbors/kdtree_knn_impl.i"
 #include "src/algorithms/k_nearest_neighbors/knn_heap.h"
-
+#include <iostream>
```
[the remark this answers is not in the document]
Alexandr-Solovev: yes, fixed

Alexandr-Solovev:
> Was anything else merged that would impact kd tree knn seg faults? Because these are no longer showing up in CI

https://github.com/oneapi-src/oneDAL/pull/2951

PR #2926, opened 2024-09-30
feature: adding -fsycl-max-parallel-link-jobs in make

Alexandr-Solovev:
# Description
Adding linking flag to improve linking of onedal_sycl lib in oneDAL


icfaust (approved):
Just for reference on whoever looks at this PR in the future: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2024-2/fsycl-max-parallel-link-jobs.html was as of 2024.2 considered an experimental feature that could possibly lead to compilation issues. I think its still fine to include it.

napetrov:
@Alexandr-Solovev  can we confirm improvements in link time from CI with this PR?

Alexandr-Solovev:
> @Alexandr-Solovev can we confirm improvements in link time from CI with this PR?

its a little bit tricky, because the different machines and active jobs in it, but in local make process the result was:
1)main make onedal_dpc -j 224 5:11
2)this pr 4:05
I guess we can try to submit more jobs, but currently its not clear how to verify on CI

icfaust:
> > @Alexandr-Solovev can we confirm improvements in link time from CI with this PR?
> 
> its a little bit tricky, because the different machines and active jobs in it, but in local make process the result was: 1)main make onedal_dpc -j 224 5:11 2)this pr 4:05 I guess we can try to submit more jobs, but currently its not clear how to verify on CI

Yeah, the problem with public CI is the limitations on threads.  When the number of threads are low, this change won't show itself well, because the linking is a smaller fraction of the total time.  I would say use the times from private CI.

Alexandr-Solovev:
> > > @Alexandr-Solovev can we confirm improvements in link time from CI with this PR?
> > 
> > 
> > its a little bit tricky, because the different machines and active jobs in it, but in local make process the result was: 1)main make onedal_dpc -j 224 5:11 2)this pr 4:05 I guess we can try to submit more jobs, but currently its not clear how to verify on CI
> 
> Yeah, the problem with public CI is the limitations on cores. When the number of cores are low, this change won't show itself well, because the linking is a smaller fraction of the total time. I would say use the times from private CI.

we have the same issue in private, like machine specific, some of the machines are faster than others, at least in real scenario(local build) it works and make performance more fine

PR #2928, opened 2024-10-01
feature: dynamic setting deps layouts

Alexandr-Solovev:
## Description

Add dynamic checker for TBB layout

**Testing**


makefile
```diff
@@ -214,15 +214,17 @@ TBBDIR.include := $(if $(TBBDIR.2),$(TBBDIR.2)/include/tbb $(TBBDIR.2)/include)
 
 TBBDIR.libia.prefix := $(TBBDIR.2)/lib
 
-ifeq ($(MKL_FPK_GPU_VERSION_LINE),2024.0.0)
+OLD_TBB_LAYOUT := $(if $(wildcard $(call frompf1,$(TBBDIR.libia.prefix))/$(_IA)),yes,no)
```
napetrov, line 217: do we need to support old tbb layout?

Alexandr-Solovev: Is it possible scenario that user will try to use tbb with old layout in build? I guess that just since 2021.12 or 13 new layout exists, it makes sense to support old layout, but not sure

homksei (approved):

PR #2941, opened 2024-10-08
update version to 2025.1.0

mvtiurin:


icfaust (approved):
Looks like it closely matches #2819

PR #2943, opened 2024-10-09
feature: adding -fsycl-max-parallel-link-jobs in bazel

Alexandr-Solovev:
## Description

Adding linking flag to improve linking of onedal_sycl lib in oneDAL(bazel)

**Testing**


Alexandr-Solovev:
Time of linking has been reduced from 120 seconds to 56 seconds with these changes. In case if machine has > 1 proc, it will provide speedup.

PR #2968, opened 2024-11-05
feature: rng primitive refactoring

Alexandr-Solovev:
### Description:
**Feature: RNG primitive refactoring**

#### Summary:
This PR updates the oneDAL rng primitive. It includes various fixes and modifications for RNG primitive.

#### Key Changes:
1. **New generators have been added:**
   - mrg32k3a and philox engines have been added in DAAL/oneDAL.

2. **Host and DPC engines have been refactored and added:**
   - Opportunity to use RNG on all devices.

SKLEARNEX related pr: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2228
---

**Testing**

**Performance**


cpp/oneapi/dal/backend/primitives/rng/rng.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+#ifdef ONEDAL_DATA_PARALLEL
+
+#include "oneapi/dal/backend/primitives/rng/engine_gpu.hpp"
+
+#endif
+
+namespace oneapi::dal::backend::primitives {
+template <typename Type, typename Size = std::int64_t>
+class rng {
```
Vika-F, line 29: What is the purpose of this class? Why all the public methods cannot be just free non-member functions?
Previously it was holding daal engine as a state, but now there is no state and no purpose in grouping the methods together.

cpp/oneapi/dal/backend/primitives/rng/rng_types.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <daal/include/algorithms/engines/mt2203/mt2203.h>
+#include <daal/include/algorithms/engines/mcg59/mcg59.h>
+#include <daal/include/algorithms/engines/mrg32k3a/mrg32k3a.h>
+#include <daal/include/algorithms/engines/philox4x32x10/philox4x32x10.h>
+#include <daal/include/algorithms/engines/mt19937/mt19937.h>
+
+namespace oneapi::dal::backend::primitives {
+
+enum class engine_list { mt2203, mcg59, mt19937, mrg32k3a, philox4x32x10 };
```
Vika-F, line 27: `_list` here reminds of the linked list data type. Maybe rename to `engine_method` or simply `method`?

docs/source/daal/algorithms/engines/mrg32k3a.rst
```diff
@@ -0,0 +1,63 @@
+.. ******************************************************************************
+.. * Copyright 2020 Intel Corporation
```
Vika-F, line 2: Need to fix comments in many places.

docs/source/daal/algorithms/engines/mrg32k3a.rst
```diff
@@ -0,0 +1,63 @@
...
+.. * distributed under the License is distributed on an "AS IS" BASIS,
+.. * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+.. * See the License for the specific language governing permissions and
+.. * limitations under the License.
+.. *******************************************************************************/
+
+mrg32k3a
+========
+
+The engine is based on the 59-bit multiplicative congruential generator.
```
Vika-F, line 20:
```suggestion
The engine based on a 32-bit combined multiple recursive generator
with two components of order 3, optimized for batch processing.
```

Alexandr-Solovev:
sklearnex ref pr: https://github.com/uxlfoundation/scikit-learn-intelex/pull/2228

david-cortes-intel:
Couple questions from my side:
* Why would the default be left as MT2203 instead of something with better properties? Why not philox for example? It appears to be substantially faster.
    * All of the generators introduced here appear to produce 32-bit numbers, but some of the usages are for e.g. generating 64-bit integers, which would require two draws for one output. Wouldn't it be better to make the default dependent on the templated parameters that an algorithm will use? Might be rather complicated though so not sure how feasible it is.
* Is the idea to simply include all generators from MKL, or to provide reasonably good choices for oneDAL? We already know for example that MT-family generators have issues when used like oneDAL does. How about removing those?
    * Do we know if these generators have state quality issues when initializing them from a single seed instead of a seed sequence? For example, the current default generator passed from sklearnex (I think MT19937 but not sure) had the issue (pointed out by @icfaust) of producing very biased first samples when initialized like that with a low-number seed.

Alexandr-Solovev:
@david-cortes-intel 

> Why would the default be left as MT2203 instead of something with better properties? Why not philox for example?

For RF algorithm the feature to choose engine_method will be added in the follow-up pr(likely here https://github.com/uxlfoundation/oneDAL/pull/3029). I hold MT2203 to save result compatibility with previous releases. Not sure that its the best strategy in terms of performance, but I expect potential accuracy/mse with philox might be worse. Moreover it can make the review process more complex.
For other algorithms the default engine method can be replaced with the new one, but I didn't want to do it in this pr, due to potential metrics degradations. Can be easily done in the next prs imho.

I probably didnt get the idea of 

> All of the generators introduced here appear to produce 32-bit numbers, but some of the usages are for e.g. generating 64-bit integers, which would require two draws for one output. Wouldn't it be better to make the default dependent on the templated parameters that an algorithm will use? Might be rather complicated though so not sure how feasible it is.

Do you mean create a dispatcher before, for example unifrom, or inside the unifrom, which will choose the best potential engine method?

david-cortes-intel:
> I hold MT2203 to save result compatibility with previous releases.

Note that NumPy itself, which is used by scikit-learn, does not have such compatibility guarantees when using their `random.Generator` module:
https://numpy.org/doc/stable/reference/random/compatibility.html

Thus, I don't think it should be a big deal to make breaking changes in produced random numbers in sklearnex.

> I expect potential accuracy/mse with philox might be worse

I'd expect it should actually be the opposite, since (a) we are seeding MT with a single integer instead of a sequence, which leaves it with issues for the first draws; (b) `philox` appears to do better at statistical tests according to the paper that introduced it - see for example that it passed BigCrush:
https://www.thesalmons.org/john/random123/papers/random123sc11.pdf
.. whereas MT fail some of it: https://arxiv.org/pdf/1910.06437

> Do you mean create a dispatcher before, for example unifrom, or inside the unifrom, which will choose the best potential engine method?

On a deeper look, it seems all generators from MKL are 32-bit only, so please ignore earlier comment.

> I guess it depends on implementation on oneMKL side. Not sure, but as I know, by default oneMKL uses N threads sub-engines in mt2203 and mt19937. May be its the reason of such behavior. As I know mrg32 philox and mcg59 are implemented without sub-engines inside.

Philox is a counter-based generator, so parallelizing it and jumping states should be pretty straightforward, without needing to keep sub-engines.

david-cortes-intel:
> I am not against this removing, but not sure that we can easily change the default behavior.

Let's leave the change in defaults for a different PR then.

david-cortes-intel:
@Alexandr-Solovev Are the issues from the CI meant to be solved with the PR from the sklearnex side?
```
In file included from /home/vsts/work/1/s/build/daal4py_cpp.cpp:2:
/home/vsts/work/1/s/build/daal4py_cpp.h: In instantiation of ‘engines_mrg32k3a_manager<fptype, method>::engines_mrg32k3a_manager() [with fptype = double; daal::algorithms::engines::mrg32k3a::Method method = daal::algorithms::engines::mrg32k3a::defaultDense]’:
/home/vsts/work/1/s/build/daal4py_cpp.cpp:423:94:   required from here
/home/vsts/work/1/s/build/daal4py_cpp.h:2358:22: error: ‘daal::algorithms::engines::mrg32k3a::interface1::Batch<algorithmFPType, method>::Batch(size_t) [with algorithmFPType = double; daal::algorithms::engines::mrg32k3a::Method method = daal::algorithms::engines::mrg32k3a::defaultDense; size_t = long unsigned int]’ is protected within this context
 2358 |         _algob.reset(new algob_type());
      |                      ^~~~~~~~~~~~~~~~
In file included from /home/vsts/work/1/daal/latest/include/daal.h:307,
                 from /home/vsts/work/1/s/src/daal4py.h:24,
                 from /home/vsts/work/1/s/src/daal4py_dist.h:23,
                 from /home/vsts/work/1/s/build/daal4py_cpp.h:3,
                 from /home/vsts/work/1/s/build/daal4py_cpp.cpp:2:
/home/vsts/work/1/daal/latest/include/algorithms/engines/mrg32k3a/mrg32k3a.h:152:5: note: declared protected here
  152 |     Batch(size_t seed = 777) { initialize(); }
      |     ^~~~~
```

Alexandr-Solovev:
@david-cortes-intel Thanks for the comments!


> I'd expect it should actually be the opposite, since (a) we are seeding MT with a single integer instead of a sequence, which leaves it with issues for the first draws; (b) philox appears to do better at statistical tests according to the paper that introduced it - see for example that it passed BigCrush:
https://www.thesalmons.org/john/random123/papers/random123sc11.pdf
.. whereas MT fail some of it: https://arxiv.org/pdf/1910.06437

I will be glad to change it, but based on the testing I temporary disabled one test, so, need to investigate, but overall lets do it.

> Let's leave the change in defaults for a different PR then.

For sure!

> Are the issues from the CI meant to be solved with the PR from the sklearnex side?

Yes, it should be fixed here https://github.com/uxlfoundation/scikit-learn-intelex/pull/2228. I am waiting for combined CI results

david-cortes-intel:
Looks like something went wrong in the examples:
```
 cpp/oneapi/dal/backend/primitives/optimizers/test/newton_cg_dpc.cpp:146:27: error: expected ';' after expression
   146 |         primitives::engine eng(4014 + n_);
```

Alexandr-Solovev:
> Looks like something went wrong in the examples:
> 
> ```
>  cpp/oneapi/dal/backend/primitives/optimizers/test/newton_cg_dpc.cpp:146:27: error: expected ';' after expression
>    146 |         primitives::engine eng(4014 + n_);
> ```
> 
> ```
>  cpp/oneapi/dal/backend/primitives/rng/rng.hpp:108:6: note: candidate function template not viable: requires at least 6 arguments, but 5 were provided
>    108 | void uniform(sycl::queue& queue,
>        |      ^       ~~~~~~~~~~~~~~~~~~~
>    109 |              Size count,
>        |              ~~~~~~~~~~~
>    110 |              Type* dst,
>        |              ~~~~~~~~~~
>    111 |              dpc_engine<EngineType>& engine_,
>        |              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>    112 |              Type a,
>        |              ~~~~~~~
>    113 |              Type b,
>        |              ~~~~~~~
>    114 |              const event_vector& deps = {});
> ```

Thanks for highlight it, will be fixed soon

Alexandr-Solovev:
duplication of https://github.com/uxlfoundation/oneDAL/pull/3040

PR #2976, opened 2024-11-14
ENH: Implement GPU version of spectral decomposition as linear regression fallback

david-cortes-intel:
## Description

As a follow up from https://github.com/oneapi-src/oneDAL/pull/2930

This PR adds a GPU version of the fallback for linear regression based on spectral decomposition, which mirrors the CPU version. It reuses the same tests, but this time executing them with fp32 too.

A few things I wasn't sure about:
* Here I'm taking the minimum of the diagonal of the Cholesky factorization using a sycl reduction class, but I'm not sure if that's the usual way in oneDAL.
* I am also not 100% sure if it is guaranteed that this operation (minimum of the diagonal) with an 'at_device' at the end would force execution to wait until the operation is completed before returning the value.
* There's an operation here for which oneMKL doesn't provide the equivalent LAPACK functionality, which is: "column-major matrix divided by a broadcasted row vector" (i.e. multiply each row of the matrix by a scalar from the vector). I implemented it as a 2D for-loop with a multiplication by the inverse (which is faster but less precise), but I don't know if this is better than a 1D loop that would take a whole row, as rows are contiguous in memory.

---

**Testing**

**Performance**

Not applicable (addition is a fallback for cases that would have failed previously).


cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
```diff
@@ -59,6 +61,167 @@ inline sycl::event beta_copy_transform(sycl::queue& queue,
...
+    ndview<Float, 2>& b, // t(X)*y, RHS, solution is outputted here
+    sycl::event& event_b,
+    const std::int64_t dim_A,
+    const std::int64_t nrhs) {
+    constexpr auto alloc = sycl::usm::alloc::device;
+
+    /* Decompose: A = Q * diag(l) * t(Q) */
+    /* Note: for NRHS>1, this will overallocate in order to reuse the memory as buffer later on */
+    auto eigenvalues = array<Float>::empty(queue, dim_A * nrhs, alloc);
+    auto eigenvalues_view = ndview<Float, 1>::wrap(eigenvalues);
```
Alexandr-Solovev, line 95: Does it make sense to replace ndview and array with ndarray?

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
```diff
@@ -59,6 +61,167 @@ inline sycl::event beta_copy_transform(sycl::queue& queue,
...
+                                             { gemm_right_event });
+    }
+}
+
+/* Returns the minimum value among entries in the diagonal of a square matrix */
+template <typename Float>
+Float diagonal_minimum(sycl::queue& queue,
+                       const Float* Matrix,
+                       const std::int64_t dim_matrix,
+                       sycl::event& event_Matrix) {
```
Alexandr-Solovev, line 212: Can it be potentially reused in scenario with more than 1 event? Should it be extended with const bk::event_vector& deps

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
```diff
@@ -59,6 +61,167 @@ inline sycl::event beta_copy_transform(sycl::queue& queue,
...
+    auto eigenvalues = array<Float>::empty(queue, dim_A * nrhs, alloc);
+    auto eigenvalues_view = ndview<Float, 1>::wrap(eigenvalues);
+    sycl::event syevd_event =
+        syevd<mkl::job::vec, uplo, Float>(queue, dim_A, A, dim_A, eigenvalues_view, { event_A });
+    const Float eps = std::numeric_limits<Float>::epsilon();
+
+    /* Discard too small singular values */
+    std::int64_t num_discarded;
+    {
+        /* This is placed inside a block because the array created here is not used any further */
```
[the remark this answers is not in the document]
david-cortes-intel: It requires finding the index of the first value in a sorted array that is above a certain threshold. I couldn't think of any fast way of doing it on GPU which wouldn't involve just transferring the data to CPU, either at once or one-by-one.

Do you have a suggestion for some better way of doing it?

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
```diff
@@ -70,12 +233,34 @@ sycl::event solve_system(sycl::queue& queue,
...
     auto [nxty, xty_event] = copy<ndorder::c, Float, ylayout, alloc>(queue, xty, dependencies);
     auto [nxtx, xtx_event] = copy<ndorder::c, Float, xlayout, alloc>(queue, xtx, dependencies);
+    const std::int64_t dim_xtx = xtx.get_dimension(0);
+
+    sycl::event solution_event;
 
     opt_array<Float> dummy{};
-    auto potrf_event = potrf_factorization<uplo>(queue, nxtx, dummy, { xtx_event });
-    auto potrs_event = potrs_solution<uplo>(queue, nxtx, nxty, dummy, { potrf_event, xty_event });
+    try {
```
Alexandr-Solovev, line 241: As I remember we use try catch only in tests/examples. Does it make sense to avoid it here?

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
david-cortes-intel: It should be theoretically possible, but the function right now is only used in this file.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
david-cortes-intel: It's meant to catch an exception thrown by oneMKL in the call to `potrf`, over which we have no control AFAIK. An alternative would be to modify the wrapper, but it's also used elsewhere in places where the exception is not catched (so a user error is thrown).

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
Alexandr-Solovev: There is magic conversion that works:
https://github.com/oneapi-src/oneDAL/blob/main/cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp#L77
here ndarray has been provided in syevd

david-cortes-intel: Thanks. And what would be the difference between `array` and `ndarray`? Looks like they use the same allocator.

david-cortes-intel:
Another thing that I'm now not very sure about is: should calls to oneMKL lapack manually call `queue.wait_and_throw()` in order to catch runtime exceptions from them?

david-cortes-intel:
After some investigation regarding the `diagonal_minimum` function, turns out the issue with the function for diagonal minimum was that it wasn't initializing the reducer object, despite what the docs mention:
https://github.khronos.org/SYCL_Reference/iface/reduction-variables.html#known-identities

Adding a manual initialization did the trick.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
Alexandr-Solovev: In general would be no difference, but you can avoid allocation of ndview

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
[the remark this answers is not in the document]
Alexandr-Solovev: no, sounds good, thanks

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
david-cortes-intel: Changed to ndarray.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
```diff
@@ -59,6 +61,171 @@ inline sycl::event beta_copy_transform(sycl::queue& queue,
...
+        /* This is placed inside a block because the array created here is not used any further */
+        /* Note: the array for 'eigenvalues' is allocated with size [dimA, nrhs],
+        but by this point, only 'dimA' elements will have been written to it. */
+        auto eigenvalues_cpu =
+            ndview<Float, 1>::wrap(eigenvalues.get_data(), dim_A).to_host(queue, { syevd_event });
+        const Float* eigenvalues_cpu_ptr = eigenvalues_cpu.get_data();
+        const Float largest_ev = eigenvalues_cpu_ptr[dim_A - 1];
+        if (largest_ev <= eps) {
+            throw std::runtime_error(
+                "Could not solve linear system. Problem has too small singular values.");
```
Vika-F, line 111: Please add new error message into `error_handling.hpp` and use it like it is done in other algorithms:
https://github.com/oneapi-src/oneDAL/blob/main/cpp/oneapi/dal/algo/knn/backend/gpu/train_kernel_brute_force_dpc.cpp#L38

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
```diff
@@ -59,6 +61,171 @@ inline sycl::event beta_copy_transform(sycl::queue& queue,
...
+            const std::size_t row = id[1];
+            Q_mutable[row + col * dim_A] *= ev_mutable[col];
+        });
+    });
+
+    /* Now calculate the actual solution: Qis * Qis' * B */
+    const std::int64_t eigenvectors_offset = num_discarded * dim_A;
+    if (nrhs == 1) {
+        sycl::event gemv_right_event =
+            mkl::blas::column_major::gemv(queue,
```
Vika-F, line 146: Please use existing wrappers to oneMKL:
https://github.com/oneapi-src/oneDAL/blob/main/cpp/oneapi/dal/backend/primitives/blas/gemv.hpp

Or add a new wrapper if it is needed.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
Vika-F: I am Ok with having a `try` in the code; but let's make it more clear for the readers - please add the description of what is happening here - why the system is tries to be solved twice in some cases.

Vika-F (commented):
The overall changes look good to me, they just need to be aligned a bit with the codebase. Please see the comments.

Vika-F:
> Another thing that I'm now not very sure about is: should calls to oneMKL lapack manually call `queue.wait_and_throw()` in order to catch runtime exceptions from them?

@david-cortes-intel Good point. I think yes, `.wait_and_throw()` is needed after the last oneMKL function call in `try` to catch the exception.
Otherwise it can be missed as oneMKL calls run asynchronously on device.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
```diff
@@ -59,6 +61,171 @@ inline sycl::event beta_copy_transform(sycl::queue& queue,
     });
 }
 
+/*
```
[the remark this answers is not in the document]
david-cortes-intel: Changed comments format to three backslashes and added a comment about the expected dimensions of the input arrays.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
```diff
@@ -59,6 +61,171 @@ inline sycl::event beta_copy_transform(sycl::queue& queue,
...
+}
+
+/* Returns the minimum value among entries in the diagonal of a square matrix */
+template <typename Float>
+Float diagonal_minimum(sycl::queue& queue,
+                       const Float* Matrix,
+                       const std::int64_t dim_matrix,
+                       sycl::event& event_Matrix) {
+    constexpr auto alloc = sycl::usm::alloc::device;
+    auto diag_min_holder = array<Float>::empty(queue, 1, alloc);
```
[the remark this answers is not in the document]
david-cortes-intel: I think it should get overwritten by passing `sycl::property::reduction::initialize_to_identity()`: https://github.khronos.org/SYCL_Reference/iface/reduction-variables.html#id2

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
david-cortes-intel: Added a comment about the logic for the try-catch block.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
david-cortes-intel: Moved the error message to the file with internal error messages.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
david-cortes-intel: Changed them to the internal wrappers for `gemm` and `gemv`, but I'd say their logic is quite hard to follow.

david-cortes-intel:
> > Another thing that I'm now not very sure about is: should calls to oneMKL lapack manually call `queue.wait_and_throw()` in order to catch runtime exceptions from them?
> 
> @david-cortes-intel Good point. I think yes, `.wait_and_throw()` is needed after the last oneMKL function call in `try` to catch the exception. Otherwise it can be missed as oneMKL calls run asynchronously on device.

Thanks for confirming. Added a manual call to `wait_and_throw`.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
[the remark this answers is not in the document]
Vika-F: It looks like it shouldn't. It gets overwritten only for integral data types. Please take a look at the table here:
https://github.khronos.org/SYCL_Reference/iface/reduction-variables.html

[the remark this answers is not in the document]
david-cortes-intel: For `sycl::minimum`, it mentions under "Available Only When":
```
std::is_floating_point_v<AccumulatorT> || std::is_same_v<std::remove_cv_t<AccumulatorT>, sycl::half>
```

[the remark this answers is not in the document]
david-cortes-intel: It has two rows in the table for `sycl::minimum` BTW.

cpp/oneapi/dal/backend/primitives/lapack/solve_dpc.cpp
```diff
@@ -70,12 +220,39 @@ sycl::event solve_system(sycl::queue& queue,
...
+    /// Note: this is wrapped in a try-catch block in order to catch a potential exception
+    /// thrown by oneMKL when the Cholesky factorization ('potrs') fails due to the matrix
+    /// not being positive-definite. In such case, it will then fall back to spectral
+    /// decomposition-based inversion, which is able to handle such problems.
+    try {
+        sycl::event potrf_event = potrf_factorization<uplo>(queue, nxtx, dummy, { xtx_event });
+        queue.wait_and_throw();
+        const Float diag_min = diagonal_minimum(queue, nxtx.get_data(), dim_xtx, potrf_event);
+        if (diag_min <= 1e-6)
+            throw mkl::lapack::computation_error("", "", 0);
```
Vika-F, line 237: That should be the last portion of comments.

1. Please provide some descriptive name for 1e-6 threshold. We are trying to reduce the number of magic constants in our code.

2. Throwing oneMKL exception here looks like a hack to call the code in catch.
It would be better just to call spectral decomposition here and in catch. Maybe re-write it a bit in order not to duplicate the code with array copying.

david-cortes-intel: Added a comment about the threshold and changed the logic of the fallback to a named section.

Vika-F: Maybe I do not understand something, but why this block cannot be just a function? For the sake of simplicity.

`goto` in DPC++ code looks pretty weird. And besides the performance aspect, the use of `goto` make the code harder to understand. As the execution order and the call stack can be very hard to track.

david-cortes-intel: Changed it into a function.

Vika-F (approved):
Let's wait for a [semi-]green CI and LGTM.

PR #2991, opened 2024-11-27
Fix rule of three violations in AlgorithmContainer classes

Vika-F:
Default constructors are set to 'default' implementations;
Copy constructors and copy assignment operators are deleted in base algorithm container classes.

---

**Testing**

### DIFF
--- a/cpp/daal/include/algorithms/algorithm_container_base.h
+++ b/cpp/daal/include/algorithms/algorithm_container_base.h
@@ -158,7 +158,6 @@ template <ComputeMode mode>
 class AlgorithmContainerImpl : public AlgorithmContainer<mode>
 {
 public:
-    AlgorithmContainerImpl()                               = default;
     AlgorithmContainerImpl(const AlgorithmContainerImpl &) = delete;
 
     /**
--- a/cpp/daal/include/algorithms/algorithm_container_base.h
+++ b/cpp/daal/include/algorithms/algorithm_container_base.h
@@ -65,8 +65,9 @@ public:
 class AlgorithmContainerIfaceImpl : public AlgorithmContainerIface
 {
 public:
-    AlgorithmContainerIfaceImpl()                                    = default;
-    AlgorithmContainerIfaceImpl(const AlgorithmContainerIfaceImpl &) = delete;
+    AlgorithmContainerIfaceImpl()                                                      = default;
+    AlgorithmContainerIfaceImpl(const AlgorithmContainerIfaceImpl &)                   = delete;
+    AlgorithmContainerIfaceImpl & operator=(const AlgorithmContainerIfaceImpl & other) = delete;
 
     /**
      * Default constructor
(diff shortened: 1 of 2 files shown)


david-cortes-intel:
Thanks for looking into this. What about the "Input" classes?

This one for example:
https://github.com/oneapi-src/oneDAL/blob/f32ae79a7bd09ffd03eda33bb3f53034c5e00d21/cpp/daal/include/algorithms/pivoted_qr/pivoted_qr_types.h#L101

which has a "copy" without "assignment" (operator=). Do they also inherit from a common class where `operator=` could be deleted?

Vika-F:
@david-cortes-intel
Thanks for reminding about Input classes.
I haven't added the changes for them into this PR because firstly I wanted to check that the solution works in general. Now we do not know whether it solves Coverity hits or not.
Maybe we'll need to do something else.

cpp/daal/include/algorithms/algorithm_container_base.h
```diff
@@ -65,6 +65,9 @@ class AlgorithmContainerIface
 class AlgorithmContainerIfaceImpl : public AlgorithmContainerIface
 {
 public:
+    AlgorithmContainerIfaceImpl()                                    = default;
```
david-cortes-intel, line 68: Is this one actually needed? Does it fail to compile without it?

Vika-F: Yes, it fails to compile with this error:

```
In file included from ./__release_lnx/daal/latest/include/algorithms/algorithm_base.h:29:
./__release_lnx/daal/latest/include/algorithms/algorithm_container_base_batch.h:56:5: error: explicitly defaulted default constructor is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
   56 |     AlgorithmContainer()                           = default;
      |     ^
./__release_lnx/daal/latest/include/algorithms/algorithm_container_base_batch.h:53:35: note: default constructor of 'AlgorithmContainer<daal::batch>' is implicitly deleted because base class 'AlgorithmContainerIfaceImpl' has no default constructor
   53 | class AlgorithmContainer<batch> : public AlgorithmContainerIfaceImpl
      |                                   ^
./__release_lnx/daal/latest/include/algorithms/algorithm_container_base_batch.h:56:54: note: replace 'default' with 'delete'
```

david-cortes-intel: Got it. But then, does deleting the other constructor have some effect if it's not called explicitly?

Vika-F: It could have the effect of eliminating the "rule of three violation" hits in Coverity. I hope.
And it seems I have to mark copy assignment operator deleted as well.

david-cortes-intel: > And it seems I have to mark copy assignment operator deleted as well.

Yes, that was also highlighted by coverity. Hopefully that should do.

icfaust (approved):

david-cortes-intel:
Error that the CI shows is a time out, better to run it again.

david-cortes-intel:
Update: there's also rule-of-three findings for this class:
https://github.com/oneapi-src/oneDAL/blob/f32ae79a7bd09ffd03eda33bb3f53034c5e00d21/cpp/daal/include/algorithms/algorithm_quality_metric_set_types.h#L49

PR #3008, opened 2024-12-04
Documentation for oneAPI data management backend primitives

Vika-F:
`ndview`, `ndarray` and related primitives are documented.

---


david-cortes-intel:
Would be helpful if we could merge this PR first:
https://github.com/uxlfoundation/oneDAL/pull/3004

So that we could then check that this one isn't introducing new warnings that might be masked by previous ones in the CI logs.

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -285,11 +376,22 @@ class ndview : public ndarray_base<axis_count, order> {
     }
 
 #ifdef ONEDAL_DATA_PARALLEL
+    /// Returns a copy of the element of 1d immutable view located in USM at specified location.
```
david-cortes-intel, line 379: Does it execute operations in a queue if the data depends on it?

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -539,14 +712,26 @@ inline sycl::event fill(sycl::queue& q,
 }
 #endif
 
+/// Multidimensional array
```
david-cortes-intel, line 715: How does it differ from sycl's own class?

docs/source/api/data-management/backend-primitives.rst
```diff
@@ -0,0 +1,39 @@
...
+Programming interface
+---------------------
+
+All types and functions in this section are declared in the
+``oneapi::dal::backend::primitives`` namespace and be available via inclusion of the
+``oneapi/dal/backend/primitives/ndarray.hpp`` header file.
+
+.. toctree::
+
+   backend/ndarray.rst
```
david-cortes-intel, line 39: What's the relationship between "tables" and "arrays"?

cpp/oneapi/dal/backend/primitives/ndarray.hpp
Vika-F: I do not think sycl has own multidimensional data representations.

cpp/oneapi/dal/backend/primitives/ndarray.hpp
Vika-F: I think I do not fully understand the question. This operation is about the same as other sycl operations, it will be included into the flow of dependencies along with the other operations.

cpp/oneapi/dal/backend/primitives/ndarray.hpp
david-cortes-intel: But it does have a 1D `marray`. Lots of the usage for views is as 1D containers, so I was curious about their usage over sycl's.

docs/source/api/data-management/backend-primitives.rst
Vika-F: Good point. Will an example on how tables and ndarrays/ndviews are used together and the description of the differences.

Short answer:
- tables are user-facing API, always 2-dimensional.
- ndarray is internal convenience API, not available in release headers. Multidimensional.

cpp/oneapi/dal/backend/primitives/ndarray.hpp
david-cortes-intel: Suppose that the array is the result of a computation like $t(X)*X$.

The operation that calculates the numbers in the array returns a queue, and might not have been completed by the time that `.at` is called.

Will the call to `.at` force calculations of the operations in the queues before returning a scalar?

cpp/oneapi/dal/backend/primitives/ndarray.hpp
Vika-F: I think besides of multidimensionality, our data structures are interoperable in the sense one can easily view the same data block as dal::table, dal::array, dal::backend::primitives::ndview, etc.

Also here we implement convenience functions like slice, transpose, to_host, etc.

With sycl own structures the amount of code inside oneDAL algorithmic kernels would be much bigger. Anyway we would need to wrap some common operations into functions. This class groups those common operations on the data together.

cpp/oneapi/dal/backend/primitives/ndarray.hpp
Vika-F: Sure, in case the events triggered by $t(X) \cdot X$ are passed into the `deps` vector correctly.

Vika-F:
@rakshithgb-fujitsu, @keeranroth can you please provide the feedback?

cpp/oneapi/dal/backend/primitives/ndarray.hpp
david-cortes-intel: Would be helpful to add a hint to prefer these custom classes instead of sycl's.

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -184,69 +194,145 @@ class ndview : public ndarray_base<axis_count, order> {
...
 
+    /// Returns whether the view contains mutable data or not.
     bool has_mutable_data() const {
         return this->has_data() && this->data_is_mutable_;
     }
 
+    /// Returns a reference to the element of 1d immutable view at specified location.
+    ///
+    /// @note This method does not perform boundary checks in release build.
+    ///       In debug build, the method asserts that the index is out of the bounds.
```
keeranroth, line 313:
```suggestion
    ///       In debug build, the method asserts that the index is not out of bounds.
```

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -184,69 +194,145 @@ class ndview : public ndarray_base<axis_count, order> {
...
+    ///
+    /// @note Should be used carefully in performance-critical parts of the code
```
keeranroth, line 315: nit: The fact that the method is in the header means that it is implicitly inline. It is likely to be inlined, but not guaranteed to be. The comment should reflect this

```
@note Should be used carefully in performance-critical parts of the code, as inlining is not guaranteed.
```

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -184,69 +194,145 @@ class ndview : public ndarray_base<axis_count, order> {
...
     const T& at(std::int64_t id) const {
         ONEDAL_ASSERT(has_data());
         ONEDAL_ASSERT((this->get_dimension(0) >= id) && (id >= 0));
         return *(get_data() + id);
     }
 
+    /// Returns a reference to the element of 2d immutable view at specified location.
+    ///
+    /// @note This method does not perform boundary checks in release build.
+    ///       In debug build, the method asserts that the index is out of the bounds.
```
keeranroth, line 329:
```suggestion
    ///       In debug build, the method asserts that the index is not of bounds.
```

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -184,69 +194,145 @@ class ndview : public ndarray_base<axis_count, order> {
...
+    ///
+    /// @note Should be used carefully in performance-critical parts of the code
```
keeranroth, line 331: nit: The fact that the method is in the header means that it is implicitly inline. It is likely to be inlined, but not guaranteed to be. The comment should reflect this

```
@note Should be used carefully in performance-critical parts of the code, as inlining is not guaranteed.
```

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -340,39 +480,74 @@ class ndview : public ndarray_base<axis_count, order> {
...
+    /// @param queue The SYCL* queue object.
+    ///
+    /// @return The SYCL* event object indicating the availability of the data in the view for reading
+    ///         and writing.
     sycl::event prefetch(sycl::queue& queue) const {
         return queue.prefetch(data_, this->get_count());
     }
 #endif
 
+    /// Returns the iterator to the first element of the 1-dimensional view.
```
keeranroth, line 530: nit: This is not returning a generic iterator, it is specifically pointers that are dealt with. Similar comments for the methods below, where `iterator` is mentioned

```suggestion
    /// Returns the pointer to the first element of the 1-dimensional view.
```

cpp/oneapi/dal/backend/primitives/ndarray.hpp
```diff
@@ -539,14 +714,26 @@ inline sycl::event fill(sycl::queue& q,
 }
 #endif
 
+/// Multidimensional array
+///
+/// @tparam T           The type of the memory block elements within the multidimensional array.
+///                     :literal:`T` can represent :expr:`float`, :expr:`double` or :expr:`std::int32_t`.
```
keeranroth, line 720: Is there a restriction on the data type? I can't see anything in this header that suggests the type of data does need to be restricted. Can this comment be removed?

keeranroth (changes requested):
The comments look good for the code that is present, so I've refrained from commenting on the code. But for my own information, is there a plan to make the views constexpr, or to have the layout separate from the data, so that the offsets into arrays / pointers / multi-dimensional arrays can be calculated at compile time? And does the implementation support arbitrary shapes and strides, e.g. having a layout of the shape (a, (b, c), d), which is a 3 dimensional object, where the second dimension is a 2-dimensional object. This is useful in terms of doing things like cache-blocking. Not for this code review, but it would be great to see functionality like this, if it doesn't already exist

Vika-F:
@keeranroth 
> is there a plan to make the views constexpr, or to have the layout separate from the data, so that the offsets into arrays / pointers / multi-dimensional arrays can be calculated at compile time?

No, there are no such plans for now.
Currently in our code base the sizes of ndarrays are usually related to the sizes of input/output data tables which is run-time information and cannot be constexpr.

But I understand that it might be beneficial to have ndarrays of constexpr sizes to represent some data blocks used in intermediate computations.

> And does the implementation support arbitrary shapes and strides ...?

I think it would be more correct to answer "no" to this question. Because:

- `ndarray` and `ndview` do not support strides at all. The layout of the data is defined only with the data's dimensions. For now we do not have the plans to support strided data layouts.
- the arbitrary shapes like (a, (b, c), d) are not supported also. For the purposes of caching we usually use regular (or parallel) loops  creating ndviews that cache the respective data blocks on each iteration. It is allowed to create multiple ndviews pointing to different parts of the data table, for example.
So, this (a, (b, c), d) structure might be emulated with loops + ndviews of lower dimension.

keeranroth (approved):
Looks good to me. Couldn't find anything obviously out of place

david-cortes-intel (approved):

PR #3012, opened 2024-12-09
Fix illegal instructions failures in examples

Vika-F:
The implementations of the methods that could potentially be the cause of illegal instructions issues in the examples are moved away from header files.

---

**Testing**

### DIFF
--- a/cpp/daal/include/algorithms/algorithm_container_base.h
+++ b/cpp/daal/include/algorithms/algorithm_container_base.h
@@ -65,7 +65,7 @@ public:
 class AlgorithmContainerIfaceImpl : public AlgorithmContainerIface
 {
 public:
-    AlgorithmContainerIfaceImpl()                                                      = default;
+    AlgorithmContainerIfaceImpl();
     AlgorithmContainerIfaceImpl(const AlgorithmContainerIfaceImpl &)                   = delete;
     AlgorithmContainerIfaceImpl & operator=(const AlgorithmContainerIfaceImpl & other) = delete;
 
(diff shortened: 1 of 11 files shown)


david-cortes-intel:
Looks like it wasn't enough to fix the problem unfortunately:
```
16:52:48 FAILED		svm_two_class_thunder_csr_batch with errno 132
.ci/scripts/test.sh: line 248: 46928 Illegal instruction     (core dumped) ${run_command} > "${e}".res 2>&1
```

There's a few more PRs I merged with the same pattern of using `= delete` in header files:
https://github.com/uxlfoundation/oneDAL/pull/2998/files
https://github.com/uxlfoundation/oneDAL/pull/2997/files
https://github.com/uxlfoundation/oneDAL/pull/2996/files

Plus this one which defines a constructor in a header - not sure if relevant as it was there before already:
https://github.com/uxlfoundation/oneDAL/pull/2994/files

And this one which added `= default` in headers:
https://github.com/uxlfoundation/oneDAL/pull/3006/files

I think a valgrind job would be really helpful here to validate that these are actually the culprits.

Vika-F:
@david-cortes-intel yes, but #2996, #2997, #2998 are in `cpp/oneapi/dal`, which should not have those problems. Only `cpp/daal` build sub-system is fragile in terms of illegal instructions.
The fix for the cases from #3006 is also included into this PR.

I will dig further - planning to reproduce locally.

david-cortes-intel:
In case it helps, the breakage seems to have started with this commit:
https://github.com/uxlfoundation/oneDAL/pull/2991

This PR has CI jobs with reproducers - succeds before that PR, fails after adding it:
https://github.com/uxlfoundation/oneDAL/pull/3014

Vika-F:
> In case it helps, the breakage seems to have started with this commit: #2991
> 
> This PR has CI jobs with reproducers - succeds before that PR, fails after adding it: #3014

@david-cortes-intel Yes, that helps very much. Thank you! Didn't understood that during the call.

david-cortes-intel (approved):
Perhaps these patterns could be added as coding guidelines for implementation of algorithms in the docs.

PR #3024, opened 2024-12-13
Disable LinuxMakeGNU_OpenBLAS_x86 step in CI

Vika-F:
Disabling the step to get rid of illegal instructions failures and save the CI time.

---

**Testing**


.ci/env/openblas.sh
```diff
@@ -163,7 +163,10 @@ pushd "${blas_src_dir}"
         NO_FORTRAN=1
         USE_OPENMP=0
         USE_THREAD=0
-        USE_LOCKING=1)
+        USE_LOCKING=1
+        DYNAMIC_ARCH=1
+        DYNAMIC_LIST="NEHALEM HASWELL"
```
david-cortes-intel, line 168: Does it still throw the same 'illegal instruction' error if you let it use the default "sapphire rapids" instead of manually listing these kernels here?

Vika-F: It does not fail with the default build settings, yes. But I thought maybe to reduce the build time by listing the specific kernels.

david-cortes-intel (approved):

.ci/env/openblas.sh
```diff
@@ -158,22 +158,28 @@ pushd "${blas_src_dir}"
...
+    make install "${make_options[@]}" PREFIX="${blas_prefix}"
+  else
+    # Download and extract the build
+    BLAS_VERSION_SHORT=$(echo -n ${BLAS_VERSION} | sed 's/^.//')
+    OPENBLAS_ARCHIVE="OpenBLAS-${BLAS_VERSION_SHORT}-x64-64.zip"
+    wget "https://github.com/OpenMathLib/OpenBLAS/releases/download/${BLAS_VERSION}/${OPENBLAS_ARCHIVE}"
+    if [[ ! -d "${blas_prefix}" ]] ; then
+      mkdir -p "${blas_prefix}"
+    fi
+    unzip -o -q ${OPENBLAS_ARCHIVE} -d "${blas_prefix}"
```
david-cortes-intel, line 182: Shouldn't this one copy only the file `lib/libopenblas.a`?

david-cortes-intel:
The error shown in the logs is really curious:
```
collect2: fatal error: ld terminated with signal 11 [Segmentation fault], core dumped
```

... that is, the _linker_ from GNU is segfaulting when it tries to do its jobs, which is not supposed to happen.

Looks like we could just revert to building from source now.

rakshithgb-fujitsu:
> @rakshithgb-fujitsu FYI
> 
> On change itself - i think we don't need x86 openblas variant at all and just keep openblas ARM variant?

Yeah I agree on this. Reduces the load on CI as well.

.ci/pipeline/ci.yml
```diff
@@ -378,6 +378,7 @@ jobs:
 
 - job: 'LinuxMakeGNU_OpenBLAS_x86'
```
david-cortes-intel, line 379: Why not remove this entry altogether instead?

Vika-F: Done.

david-cortes-intel (approved):
LGTM, but let's wait for CI results for the other openblas jobs.

napetrov (approved):

PR #3037, opened 2025-01-08
Fix GCC build errors in PCA example

Vika-F:
Add missing copy constructor or assignment operators to fix GCC "implicitly-declared `operator=` is deprecated" and similar errors in `pca_cor_dense_batch` example.

---

PR should start as a draft, then move to ready for review state after CI is passed and all applicable checkboxes are closed.
This approach ensures that reviewers don't spend extra time asking for regular requirements.

You can remove a checkbox as not applicable only if it doesn't relate to this PR in any way.
For example, PR with docs update doesn't require checkboxes for performance while PR with any change in actual code should have checkboxes and justify how this code change is expected to affect performance (or justification should be self-evident).

Checklist to comply with **before moving PR from draft**:

**PR completeness and readability**

- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have added a respective label(s) to PR if I have a permission for that.
- [x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [x] I have run it locally and tested the changes extensively.
- [x] All CI jobs are green or I have provided justification why they aren't.

PR #3039, opened 2025-01-08
Reproducer for PCA example failures (#3037)

Vika-F:
Add `-Werror=deprecated-copy` option to GCC examples build to reproduce the validation tests failures.

---

**Testing**

**Performance**

### DIFF
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -221,6 +221,14 @@ for link_mode in "${link_modes[@]}"; do
             cmake_options+=(-DCMAKE_TOOLCHAIN_FILE="${ONEDAL_DIR}"/.ci/env/"${ARCH}"-"${compiler}"-crosscompile-toolchain.cmake)
         fi
 
+        if [ "$compiler" == "gnu" ] ; then
+            if [ "$TEST_KIND" = "examples" ] ; then
+                # Build examples with -Werror=deprecated-copy enabled to catch deprecated copy constructors
+                # and assignment operators in the code
+                cmake_options+=(-DCMAKE_CXX_FLAGS="-Werror=deprecated-copy")
+            fi
+        fi
+
         echo cmake "${cmake_options[@]}"
         cmake "${cmake_options[@]}"
         err=$?
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -221,12 +221,10 @@ for link_mode in "${link_modes[@]}"; do
             cmake_options+=(-DCMAKE_TOOLCHAIN_FILE="${ONEDAL_DIR}"/.ci/env/"${ARCH}"-"${compiler}"-crosscompile-toolchain.cmake)
         fi
 
-        if [ "$compiler" == "gnu" ] ; then
-            if [ "$TEST_KIND" = "examples" ] ; then
-                # Build examples with -Werror=deprecated-copy enabled to catch deprecated copy constructors
-                # and assignment operators in the code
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
Looks like it didn't work.

How about adding it to env. var `$CXXFLAGS` to see if it works there?

.ci/scripts/test.sh
```diff
@@ -221,6 +221,14 @@ for link_mode in "${link_modes[@]}"; do
             cmake_options+=(-DCMAKE_TOOLCHAIN_FILE="${ONEDAL_DIR}"/.ci/env/"${ARCH}"-"${compiler}"-crosscompile-toolchain.cmake)
         fi
 
+        if [ "$compiler" == "gnu" ] ; then
```
david-cortes-intel, line 224: Looks like the flag is also supported by clang:
https://clang.llvm.org/docs/DiagnosticsReference.html#wdeprecated-copy

and by extension, should also be supported by icx. Perhaps it could be added unconditionally.

Vika-F:
@david-cortes-intel 
> Looks like it didn't work.
> 
> How about adding it to env. var `$CXXFLAGS` to see if it works there?

I do not like the idea of modifying environment variables. They could have impact on other processes besides cmake.

david-cortes-intel:
> @david-cortes-intel
> 
> > Looks like it didn't work.
> > How about adding it to env. var `$CXXFLAGS` to see if it works there?
> 
> I do not like the idea of modifying environment variables. They could have impact on other processes besides cmake.

Have you been able to reproduce this locally by adding the same flag?

Vika-F:
@david-cortes-intel 
No. I've added the flag locally, but couldn't get the error.
Anyway I think that the changed in #3037 are valid.

david-cortes-intel:
> @david-cortes-intel No. I've added the flag locally, but couldn't get the error. Anyway I think that the changed in #3037 are valid.

I wasn't able to reproduce it either, neither with gcc nor clang nor icx.

I notice the issue was raised with the "dynamic" variant of the examples. Just to double check: are the examples here compiled against the static or the dynamic-link library?

david-cortes-intel:
@Vika-F On a deeper look, I was able to reproduce the error when using the exact same compiler version where the issue was reported (gcc==11.3).

Since this happens only with GCC, and only with certain older versions, it leads me to believe that there might actually not be any issue in the code and we might instead be seeing a compiler bug like this:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92145

Vika-F:
> > @david-cortes-intel No. I've added the flag locally, but couldn't get the error. Anyway I think that the changed in #3037 are valid.
> 
> I wasn't able to reproduce it either, neither with gcc nor clang nor icx.
> 
> I notice the issue was raised with the "dynamic" variant of the examples. Just to double check: are the examples here compiled against the static or the dynamic-link library?

Both modes are tested here: static and dynamic. See this log, for example: https://dev.azure.com/daal/c48b0cee-5374-446e-aea4-3de8ba2ac9ca/_apis/build/builds/41717/logs/116

david-cortes-intel (approved):
While this didn't end up reproducing the original issue, it's still helpful to have this flag enabled anyway, so no harm in merging.

Vika-F:
@david-cortes-intel 
It looks that this bug is not exactly the same issue that we have in our code. In our code we have clear rule of 3 violation in `Argument` class.
I've created a stand-alone reproducer that replicates what happens in our code, and it fails even with the newer versions of GCC:
https://godbolt.org/z/1ae6Wqesd

PR #3048, opened 2025-01-21
FIX: Fix incorrect mapping of thread IDs

david-cortes-intel:
## Description

We have an issue when calling functions that use the static thread for function being passed `tid` (=thread ID) numbers that can be beyond the number of threads being used, even if the actual computations happen on a lower number of threads.

I _think_ this PR fixes it but I'm not 100% sure yet that it won't break anything else.

---

**Testing**

**Performance**

Not applicable.

### DIFF
--- a/cpp/daal/src/threading/threading.cpp
+++ b/cpp/daal/src/threading/threading.cpp
@@ -427,7 +427,8 @@ DAAL_EXPORT void _daal_parallel_reduce_tls(void * tlsPtr, void * a, daal::tls_re
     size_t n                                    = 0;
     tbb::enumerable_thread_specific<void *> * p = static_cast<tbb::enumerable_thread_specific<void *> *>(tlsPtr);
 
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
Wonder if it'd be correct to also shorten the `reduce` calls that iterate over all threads:
https://github.com/uxlfoundation/oneDAL/blob/5d3d62cbdc0033ca9a643d9535f1b7f52b1ca49c/cpp/daal/src/threading/threading.h#L615

That variable `_nThreads` will point to the **maximum** number of threads regardless of how many were used. But 'm not sure how its usage would play along other forms of parallel calls, apart from the static parallel_for being modified here.

david-cortes-intel:
Looks like this might potentially be causing a segfault in the TSNE tests for daal4py. Will investigate.

david-cortes-intel:
CI failure is unrelated to the changes here.

Vika-F (approved):

PR #3058, opened 2025-02-03
ENH: Non-batched linear regression for high-dimensional problems

david-cortes-intel:
## Description

This PR adds a preliminary version of a non-batched version of the linear regression normal regression algorithm for high-dimensional problems, which calculates the aggregates $\mathbf{X}^T \mathbf{X}$ and $\mathbf{X}^T \mathbf{y}$ separately through corresponding calls to BLAS functions.

The code route here is only meant for high-dimensional problems where the current approach does not lead to better performance. The idea is to make these thresholds configurable through the same parameters system used for e.g. the batch size in the regular route, but adding new configurable parameters appears to be a much trickier job so I'm starting with this PR with hard-coded thresholds in the meantime.

A few notes:
* ~The function that computes the aggregates takes a flag `initializeResult`, but if this flag is `false` and the results are not zeroed-out beforehand, it will lead to failing some bazel tests when the code route here is used.~ This was due to a bug in the PR code, has been fixed by now.
* There are methods to calculate statistics such as column sums from the `NumericTable` class in which the data is passed, but the inputs have `const` qualifiers and those methods that add the statistics to them modify the inputs, so cannot be used with `const` data.
    * Hence, I am calculating those by vector-matrix products with an array of all-ones. ~I wasn't sure if there's dedicated procedures to allocate aligned arrays, so I'm just using a regular unique pointer for them.~ Changed to use the dedicated `TArray` class.
* ~While the new code route being introduced here is meant for high-dimensional data only, for testing purposes, I'm enabling it for all input sizes. It works only for row-major data as it requires contiguous arrays though.~ Modified the tests to test both the regular route and this route for all linear regression cases on CPU.

**Testing**

Tests will be left for a future PR once the thresholds for triggering this mode are made configurable.

**Performance**

Performance comparisons were shared internally, without sklearn_bench as the situations to trigger it are rather specific.

### DIFF
--- a/cpp/daal/src/algorithms/linear_model/linear_model_hyperparameter_impl.h
+++ b/cpp/daal/src/algorithms/linear_model/linear_model_hyperparameter_impl.h
@@ -41,8 +41,11 @@ namespace internal
  */
 enum HyperparameterId
 {
-    denseUpdateStepBlockSize = 0,
-    hyperparameterIdCount    = denseUpdateStepBlockSize + 1
+    denseUpdateStepBlockSize     = 0,
+    denseUpdateMaxColsBatched    = 1,
+    denseSmallRowsThreshold      = 2,
+    denseSmallRowsMaxColsBatched = 3,
+    hyperparameterIdCount        = 4
 };
 
 enum DoubleHyperparameterId
--- a/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
+++ b/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
@@ -178,10 +178,11 @@ template <typename algorithmFPType, CpuType cpu>
 Status computeNonBatchedAggregates(const DAAL_INT nRows, const DAAL_INT nCols, const DAAL_INT nResponses, bool initializeResult, bool interceptFlag,
                                    const algorithmFPType * xPtr, const algorithmFPType * yPtr, algorithmFPType * xtx, algorithmFPType * xty)
 {
-    DAAL_INT nBetasIntercept = nCols + static_cast<int>(interceptFlag);
-    DAAL_INT one_int         = 1;
-    algorithmFPType one      = 1;
-    algorithmFPType zero     = 0;
+    DAAL_INT nBetasIntercept            = nCols + static_cast<int>(interceptFlag);
+    DAAL_INT one_int                    = 1;
+    algorithmFPType one                 = 1;
+    algorithmFPType zero                = 0;
+    algorithmFPType * mult_current_data = initializeResult ? &zero : &one;
     std::unique_ptr<algorithmFPType[]> ones;
     if (interceptFlag)
     {
@@ -189,18 +190,17 @@ Status computeNonBatchedAggregates(const DAAL_INT nRows, const DAAL_INT nCols, c
         std::fill(ones.get(), ones.get() + nRows, algorithmFPType(1));
     }
 
-    BlasInst<algorithmFPType, cpu>::xsyrk("U", "N", &nCols, &nRows, &one, xPtr, &nCols, &zero, xtx, &nBetasIntercept);
+    BlasInst<algorithmFPType, cpu>::xsyrk("U", "N", &nCols, &nRows, &one, xPtr, &nCols, mult_current_data, xtx, &nBetasIntercept);
     if (interceptFlag)
     {
-        BlasInst<algorithmFPType, cpu>::xgemv("N", &nCols, &nRows, &one, xPtr, &nCols, ones.get(), &one_int, initializeResult ? &zero : &one,
+        BlasInst<algorithmFPType, cpu>::xgemv("N", &nCols, &nRows, &one, xPtr, &nCols, ones.get(), &one_int, mult_current_data,
                                               xtx + static_cast<size_t>(nBetasIntercept) * static_cast<size_t>(nCols), &one_int);
         xtx[static_cast<size_t>(nBetasIntercept) * static_cast<size_t>(nBetasIntercept) - 1] = nRows;
     }
 
     if (nResponses == 1)
     {
-        BlasInst<algorithmFPType, cpu>::xgemv("N", &nCols, &nRows, &one, xPtr, &nCols, yPtr, &one_int, initializeResult ? &zero : &one, xty,
-                                              &one_int);
+        BlasInst<algorithmFPType, cpu>::xgemv("N", &nCols, &nRows, &one, xPtr, &nCols, yPtr, &one_int, mult_current_data, xty, &one_int);
         if (interceptFlag)
         {
             const algorithmFPType last_val = BlasInst<algorithmFPType, cpu>::xxdot(&nRows, yPtr, &one_int, ones.get(), &one_int);
@@ -212,12 +212,12 @@ Status computeNonBatchedAggregates(const DAAL_INT nRows, const DAAL_INT nCols, c
     }
     else
     {
-        BlasInst<algorithmFPType, cpu>::xgemm("N", "T", &nCols, &nResponses, &nRows, &one, xPtr, &nCols, yPtr, &nResponses,
-                                              initializeResult ? &zero : &one, xty, &nBetasIntercept);
+        BlasInst<algorithmFPType, cpu>::xgemm("N", "T", &nCols, &nResponses, &nRows, &one, xPtr, &nCols, yPtr, &nResponses, mult_current_data, xty,
+                                              &nBetasIntercept);
         if (interceptFlag)
         {
-            BlasInst<algorithmFPType, cpu>::xgemv("N", &nResponses, &nRows, &one, yPtr, &nResponses, ones.get(), &one_int,
-                                                  initializeResult ? &zero : &one, xty + nCols, &nBetasIntercept);
+            BlasInst<algorithmFPType, cpu>::xgemv("N", &nResponses, &nRows, &one, yPtr, &nResponses, ones.get(), &one_int, mult_current_data,
+                                                  xty + nCols, &nBetasIntercept);
         }
     }
     return Status();
--- a/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
+++ b/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
@@ -277,7 +277,7 @@ Status UpdateKernel<algorithmFPType, cpu>::compute(const NumericTable & xTable, 
         DAAL_CHECK_BLOCK_STATUS(yBlock);
         const algorithmFPType * xPtr = xBlock.get();
         const algorithmFPType * yPtr = yBlock.get();
-        return computeNonBatchedAggregates<algorithmFPType, cpu>(nRows, nCols, nResponses, true, interceptFlag, xPtr, yPtr, xtx, xty);
+        return computeNonBatchedAggregates<algorithmFPType, cpu>(nRows, nCols, nResponses, initializeResult, interceptFlag, xPtr, yPtr, xtx, xty);
     }
 
     /* Initialize output arrays by zero in case of batch mode */
--- a/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
+++ b/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
@@ -259,7 +259,7 @@ Status UpdateKernel<algorithmFPType, cpu>::compute(const NumericTable & xTable, 
     //                              && (nResponses == 1 || yTable.getDataLayout() == NumericTable::StorageLayout::aos);
     /// For testing purposes, will enable it regardless of input sizes, but this should be changed later.
     bool use_non_batched_route =
-        getDataLayout() == NumericTable::StorageLayout::aos && (nResponses == 1 || yTable.getDataLayout() == NumericTable::StorageLayout::aos);
+        xTable.getDataLayout() == NumericTable::StorageLayout::aos && (nResponses == 1 || yTable.getDataLayout() == NumericTable::StorageLayout::aos);
     if (use_non_batched_route)
     {
         /// Note: this is only implemented for row-major arrays, because there's
--- a/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
+++ b/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
@@ -194,7 +194,11 @@ Status computeNonBatchedAggregates(const DAAL_INT nRows, const DAAL_INT nCols, c
     {
         BlasInst<algorithmFPType, cpu>::xgemv("N", &nCols, &nRows, &one, xPtr, &nCols, ones.get(), &one_int, mult_current_data,
                                               xtx + static_cast<size_t>(nBetasIntercept) * static_cast<size_t>(nCols), &one_int);
-        xtx[static_cast<size_t>(nBetasIntercept) * static_cast<size_t>(nBetasIntercept) - 1] = nRows;
+        const size_t idx_last = static_cast<size_t>(nBetasIntercept) * static_cast<size_t>(nBetasIntercept) - 1;
+        if (initializeResult)
+            xtx[idx_last] = nRows;
+        else
+            xtx[idx_last] += nRows;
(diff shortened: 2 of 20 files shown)


cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
```diff
@@ -193,6 +243,40 @@ Status UpdateKernel<algorithmFPType, cpu>::compute(const NumericTable & xTable,
...
+    /// data grows, the potential speed gains from calculating both aggregates simultaneously
+    /// decreases, and the memory requirements increase, which can become a problem when there are
+    /// many threads in the system. Hence, if the number of columns is too large, it will compute
+    /// both aggregates independently, in separate calls to BLAS functions, while if the number of
+    /// columns is reasonably small, will prefer the batched procedure which typically ends up
+    /// being faster.
+
+    /// These are the thresholds where the non-batched route should be used.
+    // bool use_non_batched_route = (nBetas >= 4096 || (nRows <= 10000 && nBetas >= 1024)) && getDataLayout() == NumericTable::StorageLayout::aos
+    //                              && (nResponses == 1 || yTable.getDataLayout() == NumericTable::StorageLayout::aos);
```
Vika-F, line 259: Checking the data layout in algorithmic kernels contradicts the idea of oneDAL's data management primitives.
`aos` cannot guarantee, for example, that all columns in the data have the same type.
So, some conversions can happen anyway.

The idea of the whole implementations is to pass high-level `NumericTable` objects to the kernels and have one path for all types of data layouts: row-major, column-major, heterogeneous, chunked, whatever. And not having various types ok kernels for various data layouts.

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
```diff
@@ -193,6 +243,40 @@ Status UpdateKernel<algorithmFPType, cpu>::compute(const NumericTable & xTable,
...
+    /// For testing purposes, will enable it regardless of input sizes, but this should be changed later.
+    bool use_non_batched_route =
+        getDataLayout() == NumericTable::StorageLayout::aos && (nResponses == 1 || yTable.getDataLayout() == NumericTable::StorageLayout::aos);
+    if (use_non_batched_route)
```
Vika-F, line 263: This needs to be moved after `if (initializeResult)` to let the algorithm work correctly.
Because by default `xtx` and `xty` might contain the partial results obtained on the previous iteration of online computation.
If not, they must be initialized with zeros explicitly before the computations.

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
```diff
@@ -173,6 +174,55 @@ Status UpdateKernel<algorithmFPType, cpu>::compute(const NumericTable & xTable,
...
+    if (interceptFlag)
+    {
+        ones = std::unique_ptr<algorithmFPType[]>(new algorithmFPType[nRows]);
+        std::fill(ones.get(), ones.get() + nRows, algorithmFPType(1));
+    }
+
+    BlasInst<algorithmFPType, cpu>::xsyrk("U", "N", &nCols, &nRows, &one, xPtr, &nCols, &zero, xtx, &nBetasIntercept);
+    if (interceptFlag)
+    {
+        BlasInst<algorithmFPType, cpu>::xgemv("N", &nCols, &nRows, &one, xPtr, &nCols, ones.get(), &one_int, initializeResult ? &zero : &one,
```
Vika-F, line 195: Please move this `initializeResult ? &zero : &one` into some pre-computed value.
This code duplication makes it harder to read.

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
```diff
@@ -173,6 +174,55 @@ Status UpdateKernel<algorithmFPType, cpu>::compute(const NumericTable & xTable,
...
 
+template <typename algorithmFPType, CpuType cpu>
+Status computeNonBatchedAggregates(const DAAL_INT nRows, const DAAL_INT nCols, const DAAL_INT nResponses, bool initializeResult, bool interceptFlag,
+                                   const algorithmFPType * xPtr, const algorithmFPType * yPtr, algorithmFPType * xtx, algorithmFPType * xty)
+{
+    DAAL_INT nBetasIntercept = nCols + static_cast<int>(interceptFlag);
+    DAAL_INT one_int         = 1;
+    algorithmFPType one      = 1;
+    algorithmFPType zero     = 0;
+    std::unique_ptr<algorithmFPType[]> ones;
```
Vika-F, line 185: Please use `TArray<algorithmFPType, cpu>` instead.

cpp/daal/src/externals/service_blas_ref.h
```diff
@@ -46,14 +46,14 @@ struct OpenBlas<double, cpu>
 {
     typedef DAAL_INT SizeType;
 
-    static void xsyrk(char * uplo, char * trans, DAAL_INT * p, DAAL_INT * n, double * alpha, double * a, DAAL_INT * lda, double * beta, double * ata,
-                      DAAL_INT * ldata)
+    static void xsyrk(const char * uplo, const char * trans, const DAAL_INT * p, const DAAL_INT * n, const double * alpha, const double * a,
```
Vika-F, line 49: I am Ok with the change. But strange that it hadn't triggered any compilation warnings.

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
david-cortes-intel: @Vika-F Do you suggest copying the data to a continuous array then?

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
david-cortes-intel: @Vika-F The current values in `xtx` and `xty` in such case would either be set to zero or added to the new aggregates inside of the BLAS function calls.

But the issue is: if I make them respect the parameter `initializeResult` (by updating them as e.g. xtx += t(X)*X instead of overwriting the contents as xtx := t(X)*X), then it leads to some test failures in bazel:
* `//cpp/oneapi/dal/algo/linear_regression:test_online_dpc`
* `//cpp/oneapi/dal/algo/linear_regression:test_online_host`

Whereas forcibly overwriting the contents regardless of `initializeResult` , results in all tests passing.

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
david-cortes-intel: Switched to `TArray`.

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
david-cortes-intel: Actually turns out I had a bug in my code around the initialization, after fixing it things work correctly when honoring the `initializeResult` flag that gets passed from outside.

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
Vika-F: ReadRows does it for you. No need in additional copying.

david-cortes-intel: @Vika-F Do you have an example? Is it different from the current instantiation?
```cpp
ReadRowsType xBlock(const_cast<NumericTable &>(xTable), 0, nRows);
```

Vika-F: It's the same. ReadRows always return the data in row-major format regardless of the input table data format. So, you can just remove the layout checks and everything should work fine.
[Only the performance would probably differ, but it's another story]

david-cortes-intel: Thing is: if the data is column-major, the procedure could be done more efficiently without transposing anything.

Is there some way of generating a similar transposed ReadRows when "X" is soa?

Vika-F: I understand that. But having multiple branches of code for various input/output data layouts is not the solution we wanted to have/support.

> Is there some way of generating a similar transposed ReadRows when "X" is soa?
No.

With DAAL we were targeting mostly the cases of tall and skinny inputs, so blocking by rows was the main strategy.
We just do not have APIs to return blocks of data (or the whole dataset) in column-major format.

david-cortes-intel: Ok, removed the condition about layouts. Still working on making it controllable through the parameters system, will take a while for that.

david-cortes-intel:
@Vika-F I've now made the thresholds for choosing the non-batched route configurable through the parameters system, by adding 3 new parameters, but I'm not sure that I've correctly added them in all the places where they are necessary - could you take a look?

Also, is there some way that the tests could be expanded to execute the exact same linear regression tests but with custom parameters? (so that it'd trigger this route).

cpp/daal/src/algorithms/linear_model/linear_model_hyperparameter_impl.h
```diff
@@ -41,8 +41,11 @@ namespace internal
  */
 enum HyperparameterId
 {
-    denseUpdateStepBlockSize = 0,
-    hyperparameterIdCount    = denseUpdateStepBlockSize + 1
+    denseUpdateStepBlockSize     = 0,
+    denseUpdateMaxColsBatched    = 1,
+    denseSmallRowsThreshold      = 2,
+    denseSmallRowsMaxColsBatched = 3,
+    hyperparameterIdCount        = 4
```
Vika-F, line 48: It would be less changes when some new ID is added:
```suggestion
    denseUpdateStepBlockSize = 0,
    denseUpdateMaxColsBatched,
    denseSmallRowsThreshold,
    denseSmallRowsMaxColsBatched,
    hyperparameterIdCount = denseSmallRowsMaxColsBatched + 1
```

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
```diff
@@ -193,6 +246,47 @@ Status UpdateKernel<algorithmFPType, cpu>::compute(const NumericTable & xTable,
...
+        status = hyperparameter->find(denseSmallRowsThreshold, smallRowsThreshold);
+        DAAL_CHECK(smallRowsThreshold > 0, services::ErrorIncorrectDataRange);
+        DAAL_CHECK_STATUS_VAR(status);
+
+        status = hyperparameter->find(denseSmallRowsMaxColsBatched, smallRowsMaxColsBatched);
+        DAAL_CHECK(smallRowsMaxColsBatched > 0, services::ErrorIncorrectDataRange);
+        DAAL_CHECK_STATUS_VAR(status);
+    }
+
+    const bool use_non_batched_route = nBetas >= maxColsBatched || (nRows >= smallRowsThreshold && nBetas >= smallRowsMaxColsBatched);
```
Vika-F, line 277: I'd better change the logic a bit. I think there might be some benefit from blocking when both the number of columns __and__ the number of rows are large.
```suggestion
    const bool use_non_batched_route = (nBetas >= maxColsBatched && nColumns / nRows < threshold) || (nRows >= smallRowsThreshold && nBetas >= smallRowsMaxColsBatched);
```

cpp/oneapi/dal/algo/linear_regression/backend/cpu/finalize_train_kernel_norm_eq.cpp
```diff
@@ -55,10 +55,21 @@ static daal_lr_hyperparameters_t convert_parameters(const detail::train_paramete
     using daal_lr::internal::HyperparameterId;
 
     const std::int64_t block = params.get_cpu_macro_block();
+    const std::int64_t max_cols_batched = params.get_cpu_max_cols_batched();
```
Vika-F, line 58: I know it wasn't your doing, but don't you think it worth moving this copy-paste from "train", "partial_train" and "finalize_train" to some common place?

Vika-F: Also I am not sure that those hyperparameters are needed at all in "finalize". But that's another story.

cpp/oneapi/dal/algo/linear_regression/test/fixture.hpp
```diff
@@ -57,6 +57,35 @@ class lr_test : public te::crtp_algo_fixture<TestType, Derived> {
...
 
+    bool use_non_batched_route = false;
+
+    template <typename... Args>
+    auto train(Args&&... args) {
+        if (!this->use_non_batched_route || !this->get_policy().is_cpu())
+            return te::crtp_algo_fixture<TestType, Derived>::train(std::forward<Args>(args)...);
+        else {
+            detail::train_parameters parameters{};
+            parameters.set_cpu_max_cols_batched(1);
```
Vika-F, line 68: The test tests 1 parameter out of 3.
I think it would be better to extend the respective parameters test to vary all 3 parameters.

Vika-F:
@david-cortes-intel The hyperparameters were added correctly. I have only some minor comments.
Regarding the testing, I'd prefer to extend the "parameters" test, not the "batch".

cpp/daal/src/algorithms/linear_model/linear_model_hyperparameter_impl.h
david-cortes-intel: @Vika-F But there's no guarantee that the numbers will be mapped in a sequence. In theory, the compiler would be free to map `denseSmallRowsMaxColsBatched` to '100' and then we'd end up with an incorrect count of parameters.

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
david-cortes-intel: Not really - the non-blocked version is still faster for input sizes like 250,000 x 4,096.

Then there's also the issue that the blocked version requires allocating one array of dimension $n^2$ per thread, which can be problematic for high-dimensional problems.

cpp/oneapi/dal/algo/linear_regression/backend/cpu/finalize_train_kernel_norm_eq.cpp
Vika-F: I'd put it into some new file like "train_kernel_common.hpp". Because it looks like currently there is no common place that can share implementation with train, partial_train and finalize_train.

cpp/daal/src/algorithms/linear_model/linear_model_hyperparameter_impl.h
Vika-F: I understand your concern. Yes, you are right in theory. Let's leave it like this.

cpp/oneapi/dal/algo/linear_regression/backend/cpu/finalize_train_kernel_norm_eq.cpp
david-cortes-intel: Moved to a shared file "train_kernel_common.hpp".

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
Vika-F: Thanks for the explanation. Let's leave it as is then.

cpp/oneapi/dal/algo/linear_regression/test/fixture.hpp
david-cortes-intel: Expanded the parameter tests to cover all possible combinations of outcomes.

But I purposefully shortened the amount of combinations that go into it, because the test execution was taking a lot of time. It takes ~1.5 minutes to run the parameters tests after these changes.

Vika-F (approved):
Let's wait for a green CI and merge.

PR #3069, opened 2025-02-11
Resolve Symbol Conflicts Between oneMKL and oneDAL

Alexandr-Solovev:
#### Overview  
This PR addresses symbol conflicts between oneMKL and oneDAL that emerged after the oneMKL migration. These conflicts occur when both libraries define identical symbols, leading to linker errors or runtime issues.  

#### Solution  
To resolve this, we introduced the `--exclude-libs` linker flag to explicitly exclude specific oneMKL libraries from linking, ensuring no overlap with oneDAL symbols.  

### 🛠 Symbol Conflict Resolution  
- Added `--exclude-libs` to exclude the following oneMKL libraries:  
  - `libmkl_intel_ilp64.a`  
  - `libmkl_core.a`  
  - `libmkl_sycl.a`  
  - `libmkl_tbb_thread.a`  

  This ensures that oneMKL symbols do not conflict with oneDAL symbols during linking.  
 
- Added `--exclude-libs` list for OpenBLAS libraries

### 🚀 Dynamic and Maintainable Build System  
- Used `foreach` to dynamically generate the `EXCLUDE_LIBS` variable, making it easy to extend the exclusion list in the future.  
- Added comments and notifications to document the purpose of the exclusion list and remind developers to update it if needed.  

**Testing**

### DIFF
--- a/dev/make/deps.ref.mk
+++ b/dev/make/deps.ref.mk
@@ -31,6 +31,9 @@ daaldep.math_backend.seq := $(OPENBLASDIR.libia)/libopenblas.$a
 daaldep.math_backend.incdir := $(OPENBLASDIR.include)
 daaldep.math_backend_oneapi.incdir := $(OPENBLASDIR.include)
 
+# List of OpenBLAS libraries to exclude from linking.
+# This list is used to generate the `--exclude-libs` linker options.
+# If you need to exclude additional libraries, extend this list by appending the library names.
 MATH_LIBS_TO_EXCLUDE := libopenblas.$a 
 
 ifeq ($(RNG_OPENRNG), yes)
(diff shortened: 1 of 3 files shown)


dev/make/deps.ref.mk
```diff
@@ -44,6 +44,8 @@ ifeq ($(RNG_OPENRNG), yes)
 	daaldep.math_backend.incdir += $(daaldep.rng_backend.incdir)
 endif
 
+MATH_LIBS_TO_EXCLUDE :=
```
david-cortes-intel, line 47: It also uses static linkage for OpenBLAS. Those should also be removed.

I think it'd be anyway safer to just add `-Wl,--exclude-libs=ALL`. I don't think there's any reason why we'd want to have exported symbols from any statically-linked library.

Alexandr-Solovev: I am not sure why, but  with `-Wl,--exclude-libs=ALL` it  fails examples

david-cortes-intel: Do you have error logs for it?

Alexandr-Solovev: my bad, I doublechecked, it works, so should we exclude all symbols or just oneMKL? @Vika-F any thoughts?

Vika-F: I do not have much experience with symbols hiding, so I'd better exclude only the symbols that we definitely need to exclude: MKL and OpenBLAS. I'd leave the rest as is.

PR #3080, opened 2025-02-18
[WIP, CI, enhancement] enforce ABI checking of linux DPCPP build

icfaust:
## Description

Address unwritten rule to oneDAL codebase with testing.

---

**Testing**

**Performance**

### DIFF
--- a/.ci/scripts/build.sh
+++ b/.ci/scripts/build.sh
@@ -36,6 +36,7 @@ show_help() {
 --tbb-dir:The TBB installation directory to use to build oneDAL with in the case that the backend is given as `ref`. If the installation directory does not exist, attempts to build this from source
 --use-openrng:Set this to yes if openrng is to be used as RNG backend. Use this only with the `ref` backend.
 --sysroot:The sysroot to use, in the case that clang is used as the cross-compiler
+--debug:Enable debug symbols in build
 '
 }
 
@@ -75,6 +76,9 @@ while [[ $# -gt 0 ]]; do
         shift;;
         --use-openrng)
         use_openrng="$2"
+        shift;;
+        --debug)
+        use_debug="yes"
         shift;;
         --help)
         show_help
@@ -245,6 +249,10 @@ fi
 
 if [ "${use_openrng}" == "yes" ]; then
     make_options+=(RNG_BACKEND=openrng)
+fi
+
+if [ "${use_debug}" == "yes" ]; then
+    make_options+=(REQDBG=yes)
 fi
 
 echo "Calling make"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ concurrency:
 
 jobs:
   LinuxMakeDPCPP:
-    name: LinuxMakeDPCPP
+    name: Linux Make DPCPP with Debug
     if: github.repository == 'uxlfoundation/oneDAL'
     runs-on: uxl-gpu-4xlarge
     timeout-minutes: 120
@@ -59,17 +59,19 @@ jobs:
       - name: Make daal
         run: |
           source /opt/intel/oneapi/setvars.sh
-          .ci/scripts/build.sh --compiler icx  --optimizations avx2 --target daal
+          .ci/scripts/build.sh --compiler icx  --optimizations avx2 --target daal --debug
       - name: Make onedal
         run: |
           source /opt/intel/oneapi/setvars.sh
-          .ci/scripts/build.sh --compiler icx  --optimizations avx2 --target onedal
+          .ci/scripts/build.sh --compiler icx  --optimizations avx2 --target onedal --debug
       - name: Archive build
+        if: github.event_name == "push"
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: __release_lnx
           path: ./__release_lnx
       - name: Archive oneDAL environment
+        if: github.event_name == "push"
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
           name: oneDAL_env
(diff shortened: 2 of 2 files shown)


icfaust:
more advanced CI run showing no problems (ignore infrastructure issues, tests still ran through properly) http://intel-ci.intel.com/effdbfb8-1661-f1b1-9595-a4bf010d0e2d

icfaust:
Performance benchmarks are available upon request.

makefile
```diff
@@ -129,6 +129,7 @@ y           := $(notdir $(filter $(_OS)/%,lnx/so win/dll mac/dylib))
 -Q          := $(if $(OS_is_win),$(if $(COMPILER_is_vc),-,-Q),-)
 -cxx17      := $(if $(COMPILER_is_vc),/std:c++17,$(-Q)std=c++17)
 -fPIC       := $(if $(OS_is_win),,-fPIC)
+-visibility := $(if $(OS_is_win),,-fvisibility=hidden)
```
david-cortes-intel, line 132: What happens if you build it with clang on windows?

cpp/daal/include/services/daal_defines.h
```diff
@@ -50,11 +50,11 @@
...
 #endif
 
-#if defined(_WIN32) || defined(_WIN64)
-    #ifdef __DAAL_IMPLEMENTATION
+#ifdef __DAAL_IMPLEMENTATION
+    #if defined(_WIN32) || defined(_WIN64)
         #define DAAL_EXPORT __declspec(dllexport)
     #else
-        #define DAAL_EXPORT
+        #define DAAL_EXPORT __attribute__((visibility("default")))
```
david-cortes-intel, line 57: Since we're using C++17, how about `[[gnu::visibility("default")]]` instead?

makefile
icfaust: Good question, the default windows macro is used adding ```__declspec(dllexport)``` (daal_defines.h and dal/common.hpp).

david-cortes-intel: But would that still have an effect if not passing the right flags to clang on windows?

icfaust:
Another good question, all daal algorithms are linked to daal4py using the parser scripts (https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/generator/gen_daal4py.py), and we aren't sure who is using daal algorithms directly without scikit-learn-intelex.  In essence 2 reasons: 1) to make sure that scikit-learn-intelex ABI is maintained and 2) for possible users of the DAAL ABI which aren't reflected in our common uses cases and possibly testing.  Luckily in these circumstances, tests will fail if I didn't do it (including obscure algorithms like BACON, stump, etc.) And adaboost, for example is exposed still in daal4py, but not in sklearnex: https://github.com/uxlfoundation/scikit-learn-intelex/blob/a4105452b7e0dae7f5ca279f347aeddd740b013a/daal4py/sklearn/ensemble/AdaBoostClassifier.py#L40 . I have to say, it made this PR more difficult...

makefile
david-cortes-intel: @icfaust I mean: the makefile here is not setting the flag `-fvisibility=hidden` on windows.

icfaust: [dllexport](https://learn.microsoft.com/en-us/cpp/cpp/dllexport-dllimport?view=msvc-170) marks a function as available via the windows version of the [dynamic table](https://learn.microsoft.com/en-us/cpp/build/reference/module-definition-dot-def-files?view=msvc-170) (the module definition file *.def). Therefore by definition, Windows has default visibility of hidden in dll files unless ```__declspec(dllexport)``` is added.  This is to say, our Windows builds are already have explicitly defined symbols in the dll, which are exposed when ```_WIN32``` or ```_WIN64``` are defined by the compiler. This made my job significantly easier, as the only problems I needed to solve is how the ```__declspec(dllexport)``` is treated differently than ```__attribute__((visibility("default"))``` by the various compilers (which was a pain as GNU has different rules from Clang in visiblity on Linux, which are more strict than ```__declspec(dllexport)``` usage).

icfaust: Also rest assured, we test with both Visual Studio C++ and Intel DPC++ Windows builds, the latter being a clang-derived compiler. If there were issues, there would be failures in intelci associated with it.

david-cortes-intel: @icfaust I'm checking the windows .dll files produced with ICX by the CI as it is right now, without the flag `-fvisibility=hidden`. I'm not 100% sure but it looks like it is exporting all symbols.

For example, if I check it with winetools like this:
```shell
winedump -C -j export onedal.3.dll
```

Then I see symbols like these:
```
_onedal_parallel_reduce_int32ptr_int64_simple

?allocate@homogen_table_builder@v1@detail@dal@oneapi@@QEAA?A?<auto>@@_J0@Z

void __cdecl oneapi::dal::detail::threader_for_blocked_size<class `void __cdecl oneapi::dal::backend::copy_convert<struct oneapi::dal::backend::cpu_dispatch_avx512,double,unsigned int>(class oneapi::dal::detail::v1::host_policy const & __ptr64,double const * __ptr64,__int64,unsigned int * __ptr64,__int64,__int64)'::`1'::<lambda_1> >(unsigned __int64,unsigned __int64,class `void __cdecl oneapi::dal::backend::copy_convert<struct oneapi::dal::backend::cpu_dispatch_avx512,double,unsigned int>(class oneapi::dal::detail::v1::host_policy const & __ptr64,double const * __ptr64,__int64,unsigned int * __ptr64,unsigned int * __ptr64,unsigned int * __ptr64)'::`1'::<lambda_1> const & __ptr64)
```

If I understand it correctly, the logic for windows builds with ICX would not change with this PR, which means they'd still be exported.

Vika-F:
> adaboost_predict_dense_default_batch_fpt_dispatcher.cpp

@david-cortes-intel 

Those are the part of DAAL API. Here is an example that uses that API:
https://github.com/uxlfoundation/oneDAL/blob/main/examples/daal/cpp/source/boosting/adaboost_dense_batch.cpp#L106

Some algorithms like AdaBoost are implemented only on DAAL side and do not have oneDAL API at all. They cannot be just removed from the library.

Though I do not fully understand why Windows build was not failing without those exports.

cpp/oneapi/dal/algo/decision_tree/common.hpp
```diff
@@ -98,6 +98,9 @@ class node_info : public base {
     impl_t* impl_;
 };
 
+// template class ONEDAL_EXPORT node_info<task::classification>;
+// template class ONEDAL_EXPORT node_info<task::regression>;
+
```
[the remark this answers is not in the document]
icfaust: Oh good catch!

cpp/daal/src/algorithms/assocrules/assoc_rules_apriori_batch_fpt_dispatcher.cpp
```diff
@@ -29,5 +29,24 @@ namespace daal
...
+}
+
+using BatchType = Batch<DAAL_FPTYPE, association_rules::apriori>;
+template <>
+DAAL_EXPORT BatchType::Batch(const BatchType & other) : input(other.input), parameter(other.parameter)
+{
+    initialize();
+}
+} // namespace interface1
+} // namespace association_rules
```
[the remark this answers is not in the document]
icfaust: I totally agree with you, and thought about this, we could do the standard constructor that way. Unfortunately the copy constructor is different for the various algorithms because of differences on what is in the member initializer list (which was/is very annoying). This is because the contents of ```initialize()``` is not standardized and sometimes handled in the initializer. We could go back and redefine all of the initialize() functions to be standard, but I thought it was a bit out of scope of the PR. Let me know what you think @Vika-F .

makefile
icfaust: That's right, the topic of the PR is focused on getting visibility correct for Linux only. So far my research online has only matched that the default visibility for windows is hidden, maybe a follow up would be to check this circumstance and if necessary create a follow-up ticket/PR for windows.

david-cortes-intel: > That's right, the topic of the PR is focused on getting visibility correct for Linux make only, in fact Linux bazel builds are still full visibility. So far my research online has only matched that the default visibility for windows is hidden, maybe a follow up would be to check this circumstance and if necessary create a follow-up ticket/PR for windows.

As far as I'm aware, that depends on the compiler, linker, arguments, and presence of other files like .def. MSVC for example defaults to hiding everything, but it appears ICX doesn't, or at least is not doing so in the CI builds.

But since the Makefile here is not meant to work with every possible compiler, I guess it should be enough to pass `-fvisibility=hidden` on windows for the compilers that support such flag (mingw, clang, icx).

david-cortes-intel: From some further investigations:
* The flag `-fvisibility` is not supported by ICX on windows.
* It is not exporting all of the symbols in windows either, or at least not when it comes to ICX builds from the Makefile.
* But it is exporting more than needed - for example, it has `ONEDAL_EXPORT` applied to the internal functions used for threading:
https://github.com/uxlfoundation/oneDAL/blob/8d49822b4fdebdbfadb05dc05597f456f89b76f1/cpp/oneapi/dal/detail/threading.hpp#L72

david-cortes-intel (approved):

cpp/daal/src/algorithms/assocrules/assoc_rules_apriori_batch_fpt_dispatcher.cpp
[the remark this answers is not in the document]
icfaust: I just saw your general comment, it looks like you came to the same conclusion. Should I template the standard constructor? Let me know. I would like to create a follow up task to see if we can make all initialize calls similar. @Vika-F

Vika-F (approved):

PR #3101, opened 2025-03-05
feature: add noise_variance computation on oneDAL side

Alexandr-Solovev:
## Changes Summary

1. **Added `noise_variance` computation**  
   Implemented calculation of noise variance as part of the PCA result options. This allows better estimation of the unexplained variance in the dataset.

2. **Added `syevr`-based eigen decomposition function**  
   Introduced a new function utilizing LAPACK's `syevr` routine to improve the performance of eigenvector and eigenvalue computations for symmetric matrices.

---

**Testing**

**Performance**


cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -223,6 +231,54 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
     return services::Status();
 }
```
Alexandr-Solovev, line 251: doublecheck results

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -223,6 +231,54 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+    char jobz  = 'V';
+    char range = 'I';
+    char uplo  = 'U';
+
+    DAAL_INT il = static_cast<DAAL_INT>(nFeatures - nComponents + 1);
+    DAAL_INT iu = static_cast<DAAL_INT>(nFeatures);
+    DAAL_INT m;
+    DAAL_INT info;
+    // Could be modified to be a function parameter
+    algorithmFPType abstol = 1e-8;
```
Alexandr-Solovev: less than zero

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
icfaust: Note to self?

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -223,6 +233,55 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+
+    DAAL_INT il = static_cast<DAAL_INT>(nFeatures - nComponents + 1);
+    DAAL_INT iu = static_cast<DAAL_INT>(nFeatures);
+    DAAL_INT m;
+    DAAL_INT info;
+    // Could be modified to be a function parameter
+    algorithmFPType abstol = -1;
+
+    DAAL_INT lwork  = 26 * nFeatures;
+    DAAL_INT liwork = 10 * nFeatures;
```
icfaust: Could you provide a code comment for the hardcoded values of ```26``` and ```10``` to provide guidance to future development?

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -223,6 +233,55 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+{
+    char jobz  = 'V';
+    char range = 'I';
+    char uplo  = 'U';
+
+    DAAL_INT il = static_cast<DAAL_INT>(nFeatures - nComponents + 1);
+    DAAL_INT iu = static_cast<DAAL_INT>(nFeatures);
+    DAAL_INT m;
+    DAAL_INT info;
+    // Could be modified to be a function parameter
```
icfaust: I'd add a little snippet referring to the application notes for values < 0 (https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-0/syevx.html)

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -223,6 +233,55 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+    LapackInst<algorithmFPType, cpu>::xsyevr(&jobz, &range, &uplo, (DAAL_INT *)(&nFeatures), eigenvectors, (DAAL_INT *)(&nFeatures), nullptr, nullptr,
+                                             &il, &iu, &abstol, &m, temp_eigenvalues.get(), temp_eigenvectors.get(), (DAAL_INT *)(&nFeatures),
+                                             isuppz.get(), work.get(), &lwork, iwork.get(), &liwork, &info);
+
+    if (info != 0 || m != static_cast<DAAL_INT>(nComponents))
+    {
+        return services::Status(services::ErrorPCAFailedToComputeCorrelationEigenvalues);
+    }
+
+    for (size_t i = 0; i < nComponents; ++i)
```
icfaust: Any way to add vectorization pragmas here? (thoughts @Vika-F ?)

cpp/oneapi/dal/algo/pca/backend/cpu/finalize_train_kernel_cov.cpp
```diff
@@ -163,6 +163,21 @@ static train_result<Task> call_daal_kernel_finalize_train(const context_cpu& ctx
         result.set_variances(homogen_table::wrap(arr_vars, 1, column_count));
     }
 
+    if (desc.get_result_options().test(result_options::noise_variance)) {
+        auto eigvals = arr_eigval.get_data();
+        auto total_variance = arr_vars.get_data();
+
+        double noiseVariance = 0.0;
```
icfaust, line 195: pragmas here too?

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_cov.cpp
```diff
@@ -126,6 +126,21 @@ static result_t call_daal_kernel(const context_cpu& ctx,
         result.set_singular_values(homogen_table::wrap(arr_singular_values, 1, component_count));
     }
 
+    if (desc.get_result_options().test(result_options::noise_variance)) {
```
icfaust, line 135: Is this done twice once for batch and once for incremental?

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_svd.cpp
```diff
@@ -142,6 +142,21 @@ static result_t call_daal_kernel(const context_cpu& ctx,
         result.set_variances(homogen_table::wrap(arr_vars, 1, column_count));
     }
 
+    if (desc.get_result_options().test(result_options::noise_variance)) {
```
icfaust, line 145: I guess now that I am seeing this 4x, is there a way to put this in a common header as an inline func?

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
Alexandr-Solovev: yes

cpp/oneapi/dal/algo/pca/backend/gpu/finalize_train_kernel_cov_impl_dpc.cpp
```diff
@@ -57,7 +56,7 @@ result_t finalize_train_kernel_cov_impl<Float>::operator()(const descriptor_t& d
 
     auto result = train_result<task_t>{}.set_result_options(desc.get_result_options());
 
-    const auto nobs_host = pr::table2ndarray<Float>(q, input.get_partial_n_rows());
+    const auto nobs_host = pr::table2ndarray<Float>(input.get_partial_n_rows());
```
icfaust, line 59: underlying bug in the impl? Let me know.

cpp/oneapi/dal/algo/pca/backend/gpu/finalize_train_kernel_cov_impl_dpc.cpp
```diff
@@ -116,46 +116,49 @@ result_t finalize_train_kernel_cov_impl<Float>::operator()(const descriptor_t& d
     auto [eigvals, syevd_event] =
         syevd_computation(q, data_to_compute, { cov_event, corr_event, vars_event });
 
-    auto flipped_eigvals_host = flip_eigenvalues(q, eigvals, component_count, { syevd_event });
+    if (desc.get_result_options().test(result_options::noise_variance)) {
+        auto eigvals_host_tmp = eigvals.to_host(q, { syevd_event });
+        std::int64_t rows_count_ = rows_count_global;
+        auto range = std::min(rows_count_, column_count) - component_count;
+        auto noise_variance =
+            compute_noise_variance_on_host(q, eigvals_host_tmp, range, { syevd_event });
```
icfaust: If I am reading this right, this would mean that the ```compute_noise_variance``` is only done on host? If so please note this in the code and in the description (possibly as a follow up, or why it is done this way).

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -191,6 +286,36 @@ auto compute_means(sycl::queue& queue,
...
+                                      pr::ndarray<Float, 1> eigenvalues,
+                                      std::int64_t range,
+                                      const dal::backend::event_vector& deps = {}) {
+    ONEDAL_PROFILER_TASK(compute_noise_variance_on_host);
+    ONEDAL_ASSERT(eigenvalues.has_mutable_data());
+
+    auto eigvals_ptr = eigenvalues.get_data();
+
+    Float sum = 0;
+    for (std::int64_t i = 0; i < range; ++i) {
```
icfaust: reason not to have an equivalent add ```reduce``` call on GPU?

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -231,6 +356,43 @@ auto compute_explained_variances_on_host(sycl::queue& queue,
...
+    auto vars_ptr = vars_host.get_data();
+    auto explained_variances_ratio_ptr = explained_variances_ratio.get_mutable_data();
+
+    Float sum = 0;
+    for (std::int64_t i = 0; i < column_count; ++i) {
+        sum += vars_ptr[i];
+    }
+
+    ONEDAL_ASSERT(sum > 0);
+    const Float inverse_sum = 1.0 / sum;
```
icfaust: Is there a reason why this was done on CPU?

icfaust (commented):
Hopefully this review is useful, I see some basic for loops which could use pragmas, and then some locations where its done on CPU where maybe a GPU solution could also be done without too much work.

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -196,10 +185,31 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeCorrelationEig
     DAAL_CHECK_BLOCK_STATUS(eigenvaluesBlock);
     algorithmFPType * eigenvaluesArray = eigenvaluesBlock.get();
 
-    copyArray(nFeatures * nComponents, fullEigenvectorsArray, eigenvectorsArray);
-    copyArray(nComponents, fullEigenvaluesArray, eigenvaluesArray);
+    // SYEVR branch
+    // In this case, we compute only nComponents eigenvectors and then sort them in descending order
```
Vika-F, line 193: Sorting is mentioned in the comment, but there is no sorting in the code. Is it Ok?

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
Vika-F: I haven't merged the PR with pragmas change yet :-\ need to collect the perf data.
But I think the inner loop should be easily vectorizable, maybe even without the pragmas. But pragmas can help to guide the compiler.

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
Alexandr-Solovev: Have been replaced via fake call of xsyevr to get the exact size

cpp/oneapi/dal/algo/pca/backend/gpu/finalize_train_kernel_cov_impl_dpc.cpp
Alexandr-Solovev: not a bug, but an inefficient usage of memory. The array is used only for get single value for host, but allocates in shared memory

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
Alexandr-Solovev: My opinion that vectorization pragmas could be added later

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_cov.cpp
Alexandr-Solovev: no, in general it should be done once for each of method: 
batch: cov, svd, precomputed
incrementa: svd, cov

cpp/oneapi/dal/algo/pca/backend/cpu/train_kernel_svd.cpp
Alexandr-Solovev: yes, good point, will be aligned in the same style as for all other optional results

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
Alexandr-Solovev: I guess its a performance question, usually the range is kinda small and reduce(parallel_for kerenl initialization) takes more time than move data to cpu and compute via cpu

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
Alexandr-Solovev: The sorting for syevr is already included in computeEigenVectorsInplaceSyevr, but I can split the functions

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
icfaust: Make/share a ticket for that. I don't think it would be too bad if @Vika-F 's PR is merged soon.

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
icfaust: I'd leave a comment behind that design choice then, as that should help guide developers who may see this in the future.

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
Vika-F: To my opinion, it would be better to align the behavior of `computeEigenvectorsInplace` and `computeEigenvectorsInplaceSyevr`: either to include sorting into both, or to exclude sorting from both.

Otherwise, it is harder to get what's happening in the code.

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -91,6 +91,60 @@ auto syevd_computation(sycl::queue& queue,
     return std::make_tuple(eigenvalues, syevd_event);
 }
 
+template <typename Float>
+auto prepare_eigenvectors_svd(sycl::queue& queue,
```
Vika-F, line 139: Please add a description.

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
david-cortes-intel: All lapack functions for symmetric eigenvalues produce the results in sorted order by design.

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -196,10 +185,31 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeCorrelationEig
...
+    // inside the 'computeEigenvectorsInplaceSyevr' function
+    if (nComponents < nFeatures)
```
david-cortes-intel: Could it also limit the components by the number of rows in the data by this point? Is that info available here?

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -223,6 +233,69 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+    char jobz  = 'V';
+    char range = 'I';
+    char uplo  = 'U';
+
+    DAAL_INT il = static_cast<DAAL_INT>(nFeatures - nComponents + 1);
+    DAAL_INT iu = static_cast<DAAL_INT>(nFeatures);
+    DAAL_INT m;
+    DAAL_INT info;
+
+    algorithmFPType abstol = 2.0 * std::numeric_limits<algorithmFPType>::min();
```
david-cortes-intel: This would leave a tolerance in the order of 10^-300, which is not numerically achievable for lapack.

Better to pass a negative number, so that it would use the default value.

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -223,6 +233,69 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+        return services::Status(services::ErrorPCAFailedToComputeCorrelationEigenvalues);
+    }
+
+    for (size_t i = 0; i < nComponents; ++i)
+    {
+        size_t idx     = nComponents - 1 - i;
+        eigenvalues[i] = temp_eigenvalues[idx];
+        for (size_t j = 0; j < nFeatures; ++j)
+        {
+            eigenvectors[j + i * nFeatures] = temp_eigenvectors[j + idx * nFeatures];
```
david-cortes-intel: There's a dedicated lapack function to reorder columns:
https://www.netlib.org/lapack/explore-3.2.1-html/dlapmt.f.html

Plus there's C++ 'reverse' for vectors:
https://en.cppreference.com/w/cpp/algorithm/reverse.html

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -223,6 +233,69 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+
+    if (info != 0 || m != static_cast<DAAL_INT>(nComponents))
+    {
+        return services::Status(services::ErrorPCAFailedToComputeCorrelationEigenvalues);
+    }
+
+    for (size_t i = 0; i < nComponents; ++i)
+    {
+        size_t idx     = nComponents - 1 - i;
+        eigenvalues[i] = temp_eigenvalues[idx];
```
david-cortes-intel: Shouldn't this be discarding components with too small eigenvalues?

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -91,6 +91,60 @@ auto syevd_computation(sycl::queue& queue,
...
+            const Float abs_val = std::abs(val);
+            if (abs_val > abs_max) {
+                abs_max = abs_val;
+                max_val = val;
+            }
+        }
+
+        if (max_val < 0) {
+            for (std::int64_t j = 0; j < column_count; j++) {
+                row[j] = -row[j];
```
david-cortes-intel: Why would this be necessary? If I'm getting the logic correctly, they would afterwards get summed, so it should be possible to flip the sign of only the sum.

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -91,6 +91,60 @@ auto syevd_computation(sycl::queue& queue,
...
+
+    auto eigenvectors_host = eigenvectors.to_host(queue, deps);
+    Float* data = eigenvectors_host.get_mutable_data();
+
+    for (std::int64_t i = 0; i < row_count; i++) {
+        Float* row = data + i * column_count;
+
+        Float max_val = row[0];
+        Float abs_max = std::abs(row[0]);
+        for (std::int64_t j = 1; j < column_count; j++) {
```
david-cortes-intel: Would it perhaps be faster to use `idamax` from blas?

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -231,6 +356,43 @@ auto compute_explained_variances_on_host(sycl::queue& queue,
...
+    auto explained_variances_ratio =
+        pr::ndarray<Float, 1>::empty(queue, { component_count }, alloc::device);
+
+    auto eigvals_ptr = eigenvalues.get_data();
+    auto vars_host = vars.to_host(queue);
+    auto vars_ptr = vars_host.get_data();
+    auto explained_variances_ratio_ptr = explained_variances_ratio.get_mutable_data();
+
+    Float sum = 0;
+    for (std::int64_t i = 0; i < column_count; ++i) {
```
david-cortes-intel: Could oneDPL be used for this kind of things?

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -460,6 +622,32 @@ auto compute_singular_values_on_host(sycl::queue& queue,
...
+    ONEDAL_PROFILER_TASK(compute_singular_values_on_gpu);
+    ONEDAL_ASSERT(eigenvalues.has_mutable_data());
+
+    const std::int64_t component_count = eigenvalues.get_dimension(0);
+    auto singular_values = pr::ndarray<Float, 1>::empty(queue, { component_count }, alloc::device);
+
+    auto eigvals_ptr = eigenvalues.get_data();
+    auto singular_values_ptr = singular_values.get_mutable_data();
+
+    const Float factor = row_count - 1;
```
david-cortes-intel, line 638: Why '-1' here? Wouldn't this make the result not meet the necessary property that $(\mathbf{X} \mathbf{Z})^T (\mathbf{X} \mathbf{Z}) = \mathbf{I}$

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -460,6 +622,32 @@ auto compute_singular_values_on_host(sycl::queue& queue,
...
+
+    auto compute_event = queue.submit([&](sycl::handler& h) {
+        h.depends_on(deps);
+        h.parallel_for(sycl::range<1>(component_count), [=](sycl::id<1> i) {
+            singular_values_ptr[i] = sycl::sqrt(factor * eigvals_ptr[i]);
```
david-cortes-intel, line 643: syevr only generates results that are valid up to numerical tolerance. So a very small eigenvalue that should in theory be positive or zero could still end up as a very small negative number.

cpp/daal/src/algorithms/pca/pca_dense_base_impl.i
```diff
@@ -60,6 +60,37 @@ services::Status PCADenseBase<algorithmFPType, cpu>::computeExplainedVariancesRa
...
+    const algorithmFPType * const variancesBlockArray = variancesBlock.get();
+    double noiseVariance                              = 0.0;
+    for (size_t i = 0; i < nColumns; i++)
+    {
+        noiseVariance += variancesBlockArray[i];
+    }
+
+    for (size_t i = 0; i < nComponents; i++)
+    {
+        noiseVariance -= eigenValuesArray[i];
```
david-cortes-intel: Would be more accurate if precomputing the sum and then subtracting it from the earlier variable.

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
david-cortes-intel: Sorry, took a deeper look at what it does.

My impression is that this is not needed at all, unless for some reason you need to exactly match the numbers from other solvers. The solution with different signs is still valid and meets the necessary properties.

Alexandr-Solovev: We have deterministic value in descriptor, and user can disable this sign_flip

cpp/daal/src/algorithms/pca/pca_dense_base.h
```diff
@@ -45,6 +45,8 @@ class PCADenseBase : public Kernel
     services::Status computeExplainedVariancesRatio(const data_management::NumericTable & eigenvalues,
                                                     const data_management::NumericTable & variances,
                                                     data_management::NumericTable & explained_variances_ratio);
+    services::Status computeNoiseVariances(const data_management::NumericTable & eigenvalues, const data_management::NumericTable & variances,
+                                           double & noise_variance);
```
Alexandr-Solovev: @Vika-F does it make sense to name the variable noiseVariance for daal like style?

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
Alexandr-Solovev: I mean descending and ascending ordering sort

cpp/daal/src/algorithms/pca/pca_dense_base_impl.i
```diff
@@ -60,6 +60,40 @@ services::Status PCADenseBase<algorithmFPType, cpu>::computeExplainedVariancesRa
...
+    double explainedVariance = 0.0;
+    for (size_t i = 0; i < nComponents; i++)
+    {
+        explainedVariance += eigenValuesArray[i];
+    }
+
+    double noiseVariance = totalVariance - explainedVariance;
+    if(nColumns <= nComponents)
+        {
+            services::throwIfPossible(services::Status(services::ErrorIncorrectParameter));
```
david-cortes-intel: Looks like this might be triggering the sklearnex CI failures.

But either way: in such cases, the remaining components (beyond nColumns) should be zeros.

cpp/oneapi/dal/algo/pca/backend/cpu/finalize_train_kernel_cov.cpp
Alexandr-Solovev: I think no

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -224,24 +243,95 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+                                                                                      algorithmFPType * eigenvectors, algorithmFPType * eigenvalues)
+{
+    const algorithmFPType eps = std::numeric_limits<algorithmFPType>::epsilon();
+
+    for (size_t i = 0; i < nComponents / 2; i++)
+    {
+        algorithmFPType tmp     = eigenvalues[i];
+        algorithmFPType tmp_rev = eigenvalues[nComponents - 1 - i];
+
+        if (std::fabs(tmp) < eps) tmp = algorithmFPType(0);
```
david-cortes-intel: A covariance by definition cannot have negative eigenvalues (oneDAL's is imprecise so it can happen when it gets too inexact). Should look at just the signed value, not the absolute value.

Alexandr-Solovev: thanks! Fixed

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -224,24 +243,95 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+                                                                                      algorithmFPType * eigenvectors, algorithmFPType * eigenvalues)
+{
+    const algorithmFPType eps = std::numeric_limits<algorithmFPType>::epsilon();
+
+    for (size_t i = 0; i < nComponents / 2; i++)
+    {
+        algorithmFPType tmp     = eigenvalues[i];
+        algorithmFPType tmp_rev = eigenvalues[nComponents - 1 - i];
+
+        if (tmp < 0) tmp = algorithmFPType(0);
```
david-cortes-intel: When a singular value should in theory be zero when calculated to infinite precision, lapack will very likely output a number (either positive or negative) in the neighborhood of zero. The approach of discarding values below machine epsilon is more sensible given the way lapack works.

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
Alexandr-Solovev: I will add a todo for investigation. For now looks like makes the pr bigger

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
Alexandr-Solovev: I will add a todo mark, but may be done in next pr

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
Alexandr-Solovev: Can you clarify please in terms of code whats your suggestion?

david-cortes-intel: Suggestion is to compare the (signed) values against a (positive) threshold - e.g.:
```
if (tmp < thr) {
    // discard
} else {
    // take
```

syakov-intel:
@Alexandr-Solovev please rebase

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -171,35 +171,54 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeCorrelationEig
...
     WriteOnlyRows<algorithmFPType, cpu> eigenvaluesBlock(eigenvalues, 0, 1);
     DAAL_CHECK_BLOCK_STATUS(eigenvaluesBlock);
     algorithmFPType * eigenvaluesArray = eigenvaluesBlock.get();
 
-    copyArray(nFeatures * nComponents, fullEigenvectorsArray, eigenvectorsArray);
-    copyArray(nComponents, fullEigenvaluesArray, eigenvaluesArray);
+    // SYEVR branch
+    // In this case, we compute only nComponents eigenvectors and then sort them in descending order
+    // inside the 'computeEigenvectorsInplaceSyevr' function
+    if (nComponents < 0.25 * nFeatures)
```
david-cortes-intel, line 202: Please add a TODO comment that this constant should be a tunable parameter controlled by the regular parameters system used elsewhere, and create a ticket about it (CC @Vika-F @avolkov-intel ).

cpp/daal/src/algorithms/pca/pca_dense_correlation_base_impl.i
```diff
@@ -223,6 +242,59 @@ services::Status PCACorrelationBase<algorithmFPType, cpu>::computeEigenvectorsIn
...
+
+    algorithmFPType abstol = -1;
+
+    // Workspace query
+    algorithmFPType work_query;
+    DAAL_INT iwork_query;
+    DAAL_INT lwork  = -1;
+    DAAL_INT liwork = -1;
+
+    TArray<DAAL_INT, cpu> dummy_isuppz(2);
```
david-cortes-intel: Perhaps you could use stack-allocated variable instead (`DAAL_INT[2]`).

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -91,104 +118,214 @@ auto syevd_computation(sycl::queue& queue,
     return std::make_tuple(eigenvalues, syevd_event);
 }
 
-///  A wrapper that flips 2d array of eigenvectors from the syevd result in necessary order
+///  Prepares eigenvectors from the gesvd result in the required order.
```
david-cortes-intel, line 121: Is there some implementation of the whole algorithm using gesvd? Is this function used exclusively for that code path?

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -91,104 +118,214 @@ auto syevd_computation(sycl::queue& queue,
...
-    flip_event.wait_and_throw();
-    auto flipped_eigenvectors_host = eigenvectors.to_host(queue);
+    copy_event.wait_and_throw();
 
-    return flipped_eigenvectors_host;
+    return trimmed;
 }
 
-///  A wrapper that flips 1d array of eigenvalues from syevd result in descending order
+///  A wrapper that flips 2d array of eigenvectors from the syevd result in necessary order
```
david-cortes-intel: 'flip' here is quite ambiguous. Perhaps you could provide a better description. Also 'swap' might be a better match for something that reorders things.

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
david-cortes-intel: Please remember about the TODO mark.

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -204,30 +341,39 @@ auto compute_means(sycl::queue& queue,
...
         sum += vars_ptr[i];
     }
+
     ONEDAL_ASSERT(sum > 0);
     const Float inverse_sum = 1.0 / sum;
-    for (std::int64_t i = 0; i < component_count; ++i) {
-        explained_variances_ratio_ptr[i] = eigvals_ptr[i] * inverse_sum;
-    }
+
+    auto compute_event = queue.submit([&](sycl::handler& h) {
```
david-cortes-intel, line 391: If you mentioned in an earlier comment that taking the sum of the array was faster on CPU because the number of elements cannot be large, shouldn't this also be faster on CPU?

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
david-cortes-intel: Bumping up comment.

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
david-cortes-intel: Bumping up comment. Is the array clipped by the point it reaches this function?

cpp/oneapi/dal/algo/pca/backend/gpu/train_kernel_svd_impl_dpc.cpp
```diff
@@ -128,50 +127,42 @@ result_t train_kernel_svd_impl<Float>::operator()(const descriptor_t& desc, cons
     auto [U, S, V_T, gesvd_event] =
         svd_decomposition(q_, data_nd, component_count, { scaled_event, mean_centered_event });
 
-    auto S_host = S.to_host(q_);
-    auto eigenvalues = compute_eigenvalues_on_host(q_, S_host, row_count, { gesvd_event });
+    auto eigenvalues = compute_eigenvalues(q_, S, row_count, { gesvd_event });
```
david-cortes-intel, line 133: Does this call gesvd at any point?

Alexandr-Solovev: Didn't undestand the question. It works with S matrix, whis is a result of gesvd.
`auto [U, S, V_T, gesvd_event] =
        svd_decomposition(q_, data_nd, component_count, { scaled_event, mean_centered_event });`

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
Alexandr-Solovev: It is used with GESVD to finalize the result in the way scikit-learn requires.

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
```diff
@@ -612,14 +624,17 @@ auto compute_singular_values(sycl::queue& queue,
...
+    ONEDAL_ASSERT(eigenvalues.has_data());
 
     const std::int64_t component_count = eigenvalues.get_dimension(0);
     auto singular_values = pr::ndarray<Float, 1>::empty(queue, { component_count }, alloc::device);
 
     auto eigvals_ptr = eigenvalues.get_data();
     auto singular_values_ptr = singular_values.get_mutable_data();
 
+    // Recover singular values from eigenvalues of the unbiased sample covariance:
+    // singular_value = sqrt((row_count - 1) * eigenvalue).
```
david-cortes-intel, line 636: @Alexandr-Solovev As far as I can tell, this uses the biased covariance, which divides by the number of rows. Could you point me to the part where it divides by rows-1 on the covariance side?

Alexandr-Solovev: For GPU https://github.com/uxlfoundation/oneDAL/blob/main/cpp/oneapi/dal/backend/primitives/stat/cov_dpc.cpp#L85 
Actually pr with addition https://github.com/uxlfoundation/oneDAL/pull/2581 and the CPU is https://github.com/uxlfoundation/oneDAL/pull/2581/changes#diff-db34ebfebd30eccffea1cc69b10ca3eede93b98e2139988b726030b84edb50a5R446

cpp/oneapi/dal/algo/pca/backend/gpu/misc.hpp
david-cortes-intel: On a deeper look, it appears scikit-learn suffers from the same problem. Might need fixing in the future though.

david-cortes-intel (approved):

david-cortes-intel:
CI error:
```
cpp/oneapi/dal/algo/pca/backend/cpu/sign_flip_cpu.cpp:17:10: fatal error: oneapi/dal/algo/pca/backend/sign_flip.hpp: No such file or directory
   17 | #include "oneapi/dal/algo/pca/backend/sign_flip.hpp"
```

PR #3102, opened 2025-03-05
[CI, Enhancement] Refine github actions dpcpp linux make build using AVX512 and debug symbols

icfaust:
## Description

This makes an AVX512 debug build to be stored for analysis, and then runs AVX512 testing along with GPU testing.  On push to main, it will store the build in the cache (rather than as an artifact) in order to guarantee longer longevity, as the cache will clear the main build over time.  Workflow_dispatch is added to allow for regenerating the cache as necessary, if it is pushed out. The number of threads is limited as the used github actions runner lists a much higher cpu count vs the number of threads available. Therefore it is hard coded to a reasonable value.

Additional features are added:
1) limiting the number of jobs for CI testing and building shell scripts via the ```--jobs``` flag, which will be useful on thread-limited hardware
2) Add a special attribute to REQDBG such that when ```REQDBG=symbols``` it will generate only the symbols, and not include the ```ONEDAL_ASSERT``` infrastructure
3) Add ```--debug``` flag option to build.sh to allow for building in CI with debug symbols

---

**Testing**

**Performance**

- [x] I have extended benchmarking suite and provided corresponding scikit-learn_bench PR if new measurable functionality was inroduced in this PR.


makefile
```diff
@@ -1116,7 +1117,9 @@ Flags:
       possible values: $(CORE.ALGORITHMS.CUSTOM.AVAILABLE)
   REQCPU - list of CPU optimizations to be included into library
       possible values: $(CPUs)
-  REQDBG - flag that enables build in debug mode
+  REQDBG - flag that enables build in debug mode. Debug mode turns on oneDAL
```
david-cortes-intel, line 1120: Here it could mention the values that it allows.

.github/workflows/ci.yml
```diff
@@ -56,24 +58,37 @@ jobs:
         run: |
           source /opt/intel/oneapi/setvars.sh
           .ci/scripts/describe_system.sh
-      - name: Make daal
+      - name: Make daal debug
         run: |
           source /opt/intel/oneapi/setvars.sh
-          .ci/scripts/build.sh --compiler icx  --optimizations avx2 --target daal
-      - name: Make onedal
+          .ci/scripts/build.sh --compiler icx  --optimizations avx512 --target daal --debug symbols --jobs 20
```
david-cortes-intel, line 64: Shouldn't this one also build the DPC part?

makefile
icfaust: Not sure what to write, as it accepts all values in order to activate. The PR just adds a special check for the value "symbols". Let me know what you think.

.github/workflows/ci.yml
icfaust: Good question with a subtle answer: It does allow for that, but I need to re-use the daal part of the build in order to save time in building the complete oneDAL package. This also follows convention here: https://github.com/uxlfoundation/oneDAL/blob/main/.ci/pipeline/ci.yml.

makefile
david-cortes-intel: Then how about adding an explanation that other values turn on both options.

.github/workflows/ci.yml
david-cortes-intel: But then, if this one doesn't build the DPC target, does this PR end up producing any DPC build with debug symbols on it?

icfaust: Gotcha, I understand now. The next step in the CI job: https://github.com/icfaust/oneDAL/blob/dev/DPCPP_upgrade/.github/workflows/ci.yml#L70 will build the dpcpp part first with debug symbols, and then without: https://github.com/icfaust/oneDAL/blob/dev/DPCPP_upgrade/.github/workflows/ci.yml#L91

david-cortes-intel (approved):

makefile
icfaust: Added!

.github/workflows/ci.yml
```diff
@@ -56,24 +58,37 @@ jobs:
...
-          path: ./__release_lnx
-      - name: Archive oneDAL environment
-        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
-        with:
-          name: oneDAL_env
-          path: .ci/env
+          path: ./__release_lnx_main
+      - name: Make onedal
+        run: |
+          # generate new onedal portion for use in examples testing (due to issues with dpc debug build)
```
Alexandr-Solovev, line 87: do we really have issues with dpc debug build? 
As far as I remember, for dynamic libs(so) we disble debug symbols, but for static it works

icfaust: Good question, the .ci/scripts/test.sh file it will test both static and dynamic examples. If I use the dpc debug build, the dynamic part of test.sh will crash the github_actions runner due to memory usage in linking.  I didn't want to reduce dpc testing, so my compromise was to rebuild the dpc portion without debug symbols to make sure the dynamic dpc portion was tested

Alexandr-Solovev (approved):

PR #3110, opened 2025-03-07
Fix sparse K-means failure on PVC when the number of rows in large

Vika-F:
## Description

SYCL dependencies flow was fixed in sparse K-means;
Additional .wait() statements were added in `assign_clusters` kernel to workaround the issue.

Testcase that reproduces the failure on PVC was added.
The output of the test before the fix in source files:

```
Abort was called at 80 line in file:
./shared/source/command_stream/linear_stream.h
-------------------------------------------------------------------------------
KMmeans sparse cases on large number of rows - kmeans_types - 1
-------------------------------------------------------------------------------
cpp/oneapi/dal/algo/kmeans/test/batch.cpp:325
...............................................................................

cpp/oneapi/dal/algo/kmeans/test/batch.cpp:325: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
```

---

**Testing**

**Performance**

The change might have some performance impact, but without the change the algorithm aborts on PVC.

### DIFF
--- a/cpp/oneapi/dal/algo/kmeans/backend/gpu/kernels_csr_impl.hpp
+++ b/cpp/oneapi/dal/algo/kmeans/backend/gpu/kernels_csr_impl.hpp
@@ -151,7 +151,7 @@ sycl::event assign_clusters(sycl::queue& q,
 
     sycl::event transpose_event = transpose(q, centroids, centroids_transposed, deps);
 
-    // Workaround to sparse K-means crash on PVC.
+    // Workaround to sparse K-means abort on PVC.
     transpose_event.wait();
 
     // Compute dot products of each data point and each cluster centroid:
@@ -165,7 +165,7 @@ sycl::event assign_clusters(sycl::queue& q,
                                Float(0),
                                { transpose_event });
 
-    // Workaround to sparse K-means crash on PVC.
+    // Workaround to sparse K-means abort on PVC.
     dist_event.wait();
 
     // Select min_j (D_ij) and idxmin_j (D_ij),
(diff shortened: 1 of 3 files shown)


cpp/oneapi/dal/algo/kmeans/backend/gpu/kernels_csr_impl.hpp
```diff
@@ -162,6 +165,9 @@ sycl::event assign_clusters(sycl::queue& q,
                                Float(0),
                                { transpose_event });
 
+    // Workaround to sparse K-means abort on PVC.
+    dist_event.wait();
```
Alexandr-Solovev, line 169: Are you sure that these waits are necessary? Because looks like at least select function contains inside cgh.depends_on checker

Alexandr-Solovev (approved):

cpp/oneapi/dal/algo/kmeans/backend/gpu/kernels_csr_impl.hpp
ethanglaser: agreed - seems like cpp/oneapi/dal/algo/kmeans/backend/gpu/train_kernel_lloyd_csr_dpc.cpp had a true sycl event problem whereas these might not be necessary

Vika-F: @Alexandr-Solovev , @ethanglaser 
I understand that those two waits look super-weird, but without them the problem remains. I can remove and we'll see what happen in CI.

PR #3074, opened 2025-02-16
[Enhancement] Add float datatype support to rocAucScore

icfaust:
## Description

The algorithm rocAucScore is currently forced to do all computation in doubles. This does not follow the coding standards of the codebase where it should template both the cpu ISA as well as the datatype (float or double).  In order to properly interface to this code outside of DAAL (i.e. oneDAL), it must first maintain a float version as well.  While introducing a datatype template, a default value of double is set in order to guarantee compatibility to daal4py. These changes do not have any impact on the performance or operation of daal4py or any code which currently interfaces rocAucScore, therefore performance benchmarking is unnecessary.

This PR is a precursor to enabling a oneDAL algorithm of rocAucScore. This completes issue #2740

---

PR should start as a draft, then move to ready for review state after CI is passed and all applicable checkboxes are closed.
This approach ensures that reviewers don't spend extra time asking for regular requirements.

You can remove a checkbox as not applicable only if it doesn't relate to this PR in any way.
For example, PR with docs update doesn't require checkboxes for performance while PR with any change in actual code should have checkboxes and justify how this code change is expected to affect performance (or justification should be self-evident).

Checklist to comply with **before moving PR from draft**:

**PR completeness and readability**

- [x] I have reviewed my changes thoroughly before submitting this pull request.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes or created a separate PR with update and provided its number in the description, if necessary.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have added a respective label(s) to PR if I have a permission for that.
- [x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [x] I have run it locally and tested the changes extensively.
- [x] All CI jobs are green or I have provided justification why they aren't.
- [x] I have extended testing suite if new functionality was introduced in this PR.

**Performance**

- [x] I have measured performance for affected algorithms using [scikit-learn_bench](https://github.com/IntelPython/scikit-learn_bench) and provided at least summary table with measured data, if performance change is expected.
- [x] I have provided justification why performance has changed or why changes are not expected.
- [x] I have provided justification why quality metrics have changed or why changes are not expected.
- [x] I have extended benchmarking suite and provided corresponding scikit-learn_bench PR if new measurable functionality was introduced in this PR.

PR #3112, opened 2025-03-10
[CI, enhancement] enforce ABI checking of linux DPCPP build

icfaust:
## Description

Address unwritten rule to oneDAL codebase with testing.

This uses the [abidiff tool](https://sourceware.org/libabigail/manual/abidiff.html) to compare to the last debug main build for changes in the Linux makefile AVX512 debug build.  This will make oneDAL have a well-defined, explicit, and maintained ABI going forward.

It creates a follow-on GitHub actions job using the uxl-xlarge runner to prevent overutilization of public resources.  An ABI check will generally take 5 minutes or less and is dependent on the number of changes to the ABI in the PR.

This builds off of #3102 which introduces important aspects to the Linux make DPCPP build.

An example ABI change can be found in #3074 which causes a failure in the ABI verification job.

The ABI checker can be skipped by adding the "API/ABI Breaking Change" label to the PR, which will signal to reviewers that it is a API/ABI change and should be treated differently.

No performance benchmarks necessary.

---

**Testing**

**Performance**


.github/.abignore
```diff
@@ -0,0 +1,51 @@
...
+  name_regexp = daal::internal::.*
+
+[suppress_variable]
+  name_regexp = daal::internal::.*
+
+[suppress_function]
+  name_regexp = daal::internal::.*
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;; oneDAL DESELECTIONS ;;;;
```
napetrov, line 31: dal::preview - is also namespace that not required to follow ABI

.github/workflows/ci.yml
```diff
@@ -61,13 +61,13 @@ jobs:
       - name: Make daal debug
         run: |
           source /opt/intel/oneapi/setvars.sh
-          .ci/scripts/build.sh --compiler icx  --optimizations avx512 --target daal --debug symbols --jobs 20
+          .ci/scripts/build.sh --compiler icx  --optimizations avx512 --target daal --debug symbols --jobs 16
```
napetrov, line 64: for ABI - shouldn't it be all code paths?  i.e. for avx512 enforcement we would get only avx512 code compatibility.

icfaust: @napetrov very good point, you are right that this will only check SSE and AVX512. We could have the LinuxMakeDPCPP build for the additional code paths at the cost of increased CI runtime. Let me know what you think.

.github/.abignore
icfaust: Added.

.github/workflows/ci.yml
napetrov: It might be ok to keep current setup just know that there will be no symbols associated with other paths.  
Might be just note this in documentation that this validation have certain limitations. But in current setup it would have good coverage already.

icfaust: Sounds good! Will do.

icfaust: Done.

napetrov (approved):
Looks good to me, but i would wait for another +1 approval here

.github/workflows/ci.yml
```diff
@@ -105,3 +105,49 @@ jobs:
         run: |
             source /opt/intel/oneapi/setvars.sh
             .ci/scripts/test.sh --test-kind samples --build-dir __release_lnx --compiler gnu --interface daal/cpp/mpi --conda-env ci-env --build-system cmake
+
+  LinuxABICheck:
+    name: ABI Conformance(AVX512)
+    needs: LinuxMakeDPCPP
+    if: |
```
ahuber21, line 112: Should we limit to only when we merge into `main`?

.github/workflows/ci.yml
```diff
@@ -105,3 +105,49 @@ jobs:
...
+        run: .ci/env/apt.sh abigail
+      - name: Check Cache
+        id: get-cache-key
+        run: |
+          CACHE_KEY=`gh cache list --sort created_at --key __release_lnx --json key --jq .[0].key`
+          echo "key=${CACHE_KEY}" >> "$GITHUB_OUTPUT"
+          echo $CACHE_KEY
+        env:
+          GH_TOKEN: ${{ github.token }}
+      - name: Download oneDAL main build
```
ahuber21, line 134: Two thoughts

* It wouldn't hurt to verify we get a recent build from cache. I'm thinking about the failure mode where we stop caching nightly builds for some reason and a few months later we're wondering why this test lights up.
  * Maybe check that it's no older than 7 days, plus check that it has the same major version?
* I would prefer if the step that archives the build makes it clear it's a debug build, e.g. by appending `_dpg` to the build directory name, or something along that lines.

ahuber21 (changes requested):

.ci/scripts/abi_check.sh
```diff
@@ -0,0 +1,41 @@
+#! /bin/bash
```
Alexsandruss, line 1:
```suggestion
#!/bin/bash
```

.github/workflows/ci.yml
icfaust: Sorry I wasn't more explicit, the nightly build is different than the build used here. While the nightly generates an artifact, this does something different. When a branch is pulled to main the build will be stored in the persistent cache (a different storage mechanism with a much longer lifetime). If enough other things fill the cache, the build could be pushed out leaving a similar scenario as you describe. In that circumstance, a specific failure should be observed on pull requests, and there is a text written at the failure that is left to the developer for the steps to re-insert the value into the cache (by running workflow_dispatch of the CI on main). Due to permissions issues (requiring write access), the workflow should not do it itself, but should be done by someone with proper privileges.

For the second point, appending the ```_dbg``` to the main build will require a different PR. The writing to cache must be done by the main branch (again write privledges), so modifying it here will break the PR in a way that cannot be corrected.  If you would like that done, let me know and I can make a quick PR for it. It would then need to be merged, and CI would need to complete, then I could make changes here.

.github/workflows/ci.yml
icfaust: Good call. Will change

icfaust: I went back and looked, and forgot that that I set the whole CI job to only currently triggered on push and pull requests to main: https://github.com/uxlfoundation/oneDAL/blob/main/.github/workflows/ci.yml#L22 That means its already limited.

.github/workflows/ci.yml
ahuber21: Cool. If you think the error message is clear then I'm fine with not adding anything particular in the get cache step.

For the second point: That's clear. I think it's more consistent. If I see it correctly, we also append the compiler name in some output directories, so why not the build type? A separate PR would be nice! :)

.github/workflows/ci.yml
ahuber21: Of course, thanks for checking!

ahuber21 (approved):

Alexsandruss (approved):

PR #3117, opened 2025-03-13
Reduce memory requirements of sparse K-means test

Vika-F:
The test that tests SYCL implementation of sparse K-means on the large number of rows was modified:
* It was switched off on CPUs;
* On all GPUs except Intel(R) Data Center GPU Max series, the number of rows in the testing data set was reduced 100 times to prevent out of memory crashes.

---

### DIFF
--- a/cpp/oneapi/dal/algo/kmeans/test/batch.cpp
+++ b/cpp/oneapi/dal/algo/kmeans/test/batch.cpp
@@ -318,8 +318,6 @@ TEMPLATE_LIST_TEST_M(kmeans_batch_test,
     }
 }
 
-#ifdef ONEDAL_DATA_PARALLEL
-
 TEMPLATE_LIST_TEST_M(kmeans_batch_test,
                      "KMmeans sparse cases on large number of rows",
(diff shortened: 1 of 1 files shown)


cpp/oneapi/dal/algo/kmeans/test/batch.cpp
```diff
@@ -318,22 +318,33 @@ TEMPLATE_LIST_TEST_M(kmeans_batch_test,
     }
 }
 
+#ifdef ONEDAL_DATA_PARALLEL
```
Alexandr-Solovev, line 321: I guess this test can be launched only in dpcpp backend, like this ifdef defined always. Not sure that it makes sense to have this #ifdef here

Alexandr-Solovev (approved):
LGTM

cpp/oneapi/dal/algo/kmeans/test/batch.cpp
Vika-F: Testing on CPU fails without this #ifdef. I will leave it as is.

PR #3126, opened 2025-03-17
Use parallel reduce threading primitive in covariance algorithm

Vika-F:
## Description

Ads new `daal::Reducer` interface class that defines the API that have to be implemented in the algorithms to allow the use of reduction primitives based on `tbb::parallel_reduce` and `tbb::parallel_deterministic_reduce`.

Two new threading primitives were added:

* `threader_reduce` implements parallel reduction using dynamic work balancing,
* `static_threader_reduce` implements parallel reduction using static work balancing and deterministic reduction.

Dense covariance algorithm in oneDAL was modified to use new `static_threader_reduce` primitive instead of `static_threader_for` + single thread reduction as it was done previously.

`tls_data_t` structure previously used as a thread local storage for partial results in Covariance algorithm was replaced with `CovarianceReduser` class which implements the interface of new `daal::Reducer` to perform parallel reduction.

This PR depends on https://github.com/uxlfoundation/oneDAL/pull/3159 because `CovarianceReducer` uses `TArrayScalableCalloc` to store the partial results.
And the performance of `TArrayScalableCalloc` is not optimal due to unaligned memory stores (takes up to 45% of the total Covariance compute time). PR #3159 fixes the issue.

---

**Testing**

No new failures comparing to the main branch.

**Performance**


cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -177,95 +183,98 @@ services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatu
...
+        }
 
         /* Threaded loop with syrk seq calls */
-        daal::static_threader_for(numBlocks, [&](int iBlock, size_t tid) {
-            struct tls_data_t<algorithmFPType, cpu> * tls_data_local = tls_data.local(tid);
-            if (!tls_data_local)
-            {
-                return;
-            }
+        tls_data_t<algorithmFPType, cpu> result = tbb::parallel_reduce(
```
david-cortes-intel, line 193: Wouldn't it be better to make a NUMA-aware version of this function that would first reduce within nodes and then globally?

Vika-F: @david-cortes-intel Thank you for looking into this.

I would prefer to go step-by-step with NUMA.
If a non-NUMA version is enough to improve performance, I would be more likely to add it first.
If not, than I'll probably add NUMA awareness into the primitives.

I also have to say that this version of the code is just an initial quick and dirty commit. The code will change a lot further. At least it should pass the testing and all the `tbb::` stuff should be moved into threading layer.

cpp/daal/src/threading/threading.cpp
```diff
@@ -835,5 +835,64 @@ DAAL_EXPORT void _daal_wait_task_group(void * taskGroupPtr)
     ((tbb::task_group *)taskGroupPtr)->wait();
 }
 
+/// The class implements the body of the parallel reduce algorithm in compliance with
+/// oneTBB ParallelReduceBody requirements:
+/// https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/named_requirements/algorithms/par_reduce_body
+///
+/// The class uses pimpl idiom to hide the implementation details of the reducer.
+class ReductionBody
```
avolkov-intel, line 843: I think we also need to add copy assignment operator to follow the rule of three

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -83,50 +83,203 @@ services::Status prepareCrossProduct(size_t nFeatures, algorithmFPType * crossPr
...
+    CovarianceReducer(NumericTable * dataTable, DAAL_INT64 numRowsInBlock, size_t numBlocks, bool isNormalized)
+        : _dataTable(dataTable),
+          _numRowsInBlock(numRowsInBlock),
+          _numBlocks(numBlocks),
+          _nFeatures(dataTable->getNumberOfColumns()),
+          _isNormalized(isNormalized)
     {
-        auto object = new tls_data_t<algorithmFPType, cpu>(isNormalized, nFeatures);
-        if (!object)
+        _crossProductArray.reset(_nFeatures * _nFeatures);
```
avolkov-intel, line 120: Add mul overflow check here

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -83,50 +83,203 @@ services::Status prepareCrossProduct(size_t nFeatures, algorithmFPType * crossPr
...
-            delete object;
-            return nullptr;
+            computeOk = false;
+            return;
         }
 
-        return object;
+        algorithmFPType alpha           = 1.0;
+        algorithmFPType beta            = 1.0;
+        const size_t numRowsInLastBlock = _numRowsInBlock + (_dataTable->getNumberOfRows() - _numBlocks * _numRowsInBlock);
```
avolkov-intel, line 173: Here we need to check for mul overflow and also that _dataTable->getNumberOfRows() >= _numBlocks * _numRowsInBlock so that we do not store negative value in size_t datatype

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -83,50 +83,203 @@ services::Status prepareCrossProduct(size_t nFeatures, algorithmFPType * crossPr
...
+
+        /// Process blocks of the input data table
+        for (size_t iBlock = begin; iBlock < end; ++iBlock)
+        {
+            size_t nRows    = ((iBlock < (_numBlocks - 1)) ? _numRowsInBlock : numRowsInLastBlock);
```
avolkov-intel, line 178:
```suggestion
            size_t nRows    = ((iBlock + 1< _numBlocks) ? _numRowsInBlock : numRowsInLastBlock);
```

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -83,50 +83,203 @@ services::Status prepareCrossProduct(size_t nFeatures, algorithmFPType * crossPr
...
+        algorithmFPType * thisCrossProduct        = crossProduct();
+        if (!computeOk || !thisCrossProduct || !other->computeOk || !otherCrossProduct)
+        {
+            computeOk = false;
+            return;
         }
 
-        sums         = sumsArray.get();
-        crossProduct = crossProductArray.get();
+        /// It is safe to use aligned loads and stores because the data is aligned
```
avolkov-intel, line 239: Let's add comment to the places where we allocate this memory that it is critical to allocate aligned memory otherwise, the result will be incorrect

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -175,110 +329,50 @@ services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatu
...
+        if (!result.crossProduct() || !result.sums())
+        {
+            return services::Status(services::ErrorMemoryAllocationFailed);
+        }
 
-            ReadRows<algorithmFPType, cpu, NumericTable> dataTableBD(dataTable, startRow, nRows);
-            DAAL_CHECK_BLOCK_STATUS_THR(dataTableBD);
-            algorithmFPType * dataBlock_local = const_cast<algorithmFPType *>(dataTableBD.get());
+        /* Reduce input matrix X into cross product Xt X and a vector of column sums */
+        const size_t grainSize = 1; // minimal number of data blocks to be processed by a thread
```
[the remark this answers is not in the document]
Vika-F: I plan to do it in a follow up PR.
I think it is already enough changes for this one PR.

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -358,15 +452,31 @@ void mergeCrossProductAndSums(size_t nFeatures, const algorithmFPType * partialC
...
+                });
+            }
+            else
+            {
+                for (size_t i = 0; i < nFeatures; i++)
                 {
-                    crossProduct[i * nFeatures + j] += partialCrossProduct[i * nFeatures + j];
-                    crossProduct[j * nFeatures + i] = crossProduct[i * nFeatures + j];
+                    PRAGMA_IVDEP
+                    PRAGMA_VECTOR_ALWAYS
```
avolkov-intel, line 472: Can we add pragma for aligned data here? It seems that operations here are performed with the memory allocated inside CovarianceReduce

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -175,110 +329,50 @@ services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatu
...
-        });
-
-        /* If data is not normalized, perform subtractions of(sums[i]*sums[j])/n */
-        if (!isNormalized && !assumeCentered)
+        }
+        else
         {
-            DAAL_ITTNOTIFY_SCOPED_TASK(gemmSums);
-            for (size_t i = 0; i < nFeatures; i++)
+            for (size_t i = 0; i < nFeatures * nFeatures; i++)
```
avolkov-intel, line 373: Can we add pragmas for vectorization here?

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -358,15 +452,31 @@ void mergeCrossProductAndSums(size_t nFeatures, const algorithmFPType * partialC
 
         if (nObsValue == 0)
         {
-            daal::threader_for(nFeatures, nFeatures, [=](size_t i) {
-                PRAGMA_IVDEP
-                PRAGMA_VECTOR_ALWAYS
-                for (size_t j = 0; j <= i; j++)
+            if (nFeatures > 512)
```
avolkov-intel, line 455: Let's define a named hyperparameter for 512 (no need to dispatch it for now)

cpp/daal/src/threading/threading.h
```diff
@@ -774,6 +800,42 @@ void conditional_static_threader_for(const bool inParallel, const size_t n, Func
...
+/// in the threading layer.
+/// The default scheduling of the threading layer is used to assign
+/// the iterations of the reduction loop to threads.
+/// Data dependencies between the iterations are allowed, but may requre the use
+/// of synchronization primitives.
+///
+/// @tparam ReducerType     A subtype of `daal::Reducer`.
+///
+/// @param[in] n         Number of iterations in the reduction loop.
+/// @param[in] grainSize Parameter reserved for the future. Currently unused.
```
avolkov-intel, line 813: I think that it is incorrect, grainSize is already used

cpp/daal/src/threading/threading.h
```diff
@@ -774,6 +800,42 @@ void conditional_static_threader_for(const bool inParallel, const size_t n, Func
...
```
avolkov-intel, line 831: Same comment here

cpp/daal/src/threading/threading.h
Vika-F: Thanks for noticing. This is a copy-paste probably, I'll fix it.

avolkov-intel:
I think it also worth adding pragmas for finalize covariance kernel, for example in this loop: https://github.com/uxlfoundation/oneDAL/blob/e77ed77c3272227126f5ea5db695ca7fa53a0b5f/cpp/daal/src/algorithms/covariance/covariance_impl.i#L562

cpp/daal/src/threading/threading.cpp
Vika-F: I've marked them as deleted, as they are not used.

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -83,50 +83,203 @@ services::Status prepareCrossProduct(size_t nFeatures, algorithmFPType * crossPr
...
+                if (!sumsPtr)
+                {
+                    computeOk = false;
+                    return;
+                }
+                /* Sum input array elements in case of non-normalized data */
+                for (DAAL_INT i = 0; i < nRows; i++)
+                {
+                    PRAGMA_IVDEP
+                    PRAGMA_VECTOR_ALWAYS
```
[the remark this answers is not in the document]
Vika-F: `dataBlock` points to user-provided data in many cases. And even if the whole block is aligned, the `_nFeatures` value usually does not allow each row in the `dataBlock` to be aligned in many cases.
So, I prefer to leave it as it is for now.

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -83,50 +83,203 @@ services::Status prepareCrossProduct(size_t nFeatures, algorithmFPType * crossPr
...
+        }
     }
 
 private:
-    TArrayScalableCalloc<algorithmFPType, cpu> sumsArray;
-    TArrayScalableCalloc<algorithmFPType, cpu> crossProductArray;
+    /// Pointer to the input data table that stores matrix X for which the cross-product matrix and sums are computed.
+    NumericTable * _dataTable;
+    /// Number of features in the input data table.
+    DAAL_INT _nFeatures;
```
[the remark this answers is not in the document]
Vika-F: Changed _numRowsInBlock type to DAAL_INT.

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -175,110 +329,50 @@ services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatu
...
-
-        /* TLS reduction: sum all partial cross products and sums */
-        tls_data.reduce([=](tls_data_t<algorithmFPType, cpu> * tls_data_local) {
-            DAAL_ITTNOTIFY_SCOPED_TASK(computeSums.reduce);
-            /* Sum all cross products */
-            if (tls_data_local->crossProduct)
+            for (size_t i = 0; i < nFeatures; i++)
             {
                 PRAGMA_IVDEP
                 PRAGMA_VECTOR_ALWAYS
```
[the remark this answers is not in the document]
Vika-F: `crossProduct` might be allocated by user and thus not aligned.
That's why additional check is needed to have `pragma vector aligned` here.
This part is not a hotspot of this algorithms, so I prefer to leave it as is for now.

cpp/daal/src/threading/threading.cpp
```diff
@@ -835,5 +835,67 @@ DAAL_EXPORT void _daal_wait_task_group(void * taskGroupPtr)
...
+{
+public:
+    /// Constructs the body of the parallel reduce algorithm from the given reducer.
+    ///
+    /// @param reducer Pointer to the reducer object.
+    ReductionBody(daal::Reducer * reducer) : _reducer(reducer), _isSplit(false) {}
```
keeranroth, line 849: Does this need to be marked `explicit` to avoid the compiler from using this as an implicit conversion operator?

cpp/daal/src/threading/threading.h
```diff
@@ -54,13 +54,39 @@ typedef int64_t (*loop_functype_int32ptr_int64)(const int32_t * start_idx_reduce
...
+
+/// Interface for the parallel reduction operation
+class Reducer
+{
+public:
+    /// Constructs a thread-local partial result and initializes it with identity value.
+    /// Must be able to run concurrently with `update` and `join` methods.
+    ///
+    /// @return Pointer to the constructed thread-local partial result.
+    virtual Reducer * create() const = 0;
```
keeranroth, line 66: How difficult would it be to use `std::unique_ptr<Reducer>` here? That would show that the create method creates an owning object. Not sure if that would have a knock-on effect in the rest of the code, but it would be nice to have.

cpp/daal/src/externals/core_threading_win_dll.cpp
```diff
@@ -787,4 +793,24 @@ DAAL_EXPORT void * _getThreadPinner(bool create_pinner, void (*read_topo)(int &,
     }
     return _getThreadPinner_ptr(create_pinner, read_topo, deleter);
 }
+
+DAAL_EXPORT void _daal_threader_reduce(const size_t n, const size_t grainSize, daal::Reducer & reducer)
+{
+    load_daal_thr_dll();
+    if (_daal_threader_reduce_ptr == NULL)
```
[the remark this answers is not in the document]
Vika-F: Can you please explain in more detail why you think this code can work incorrectly in multi-threaded environment?
As far as I understand GetProcAddress might be executed simultaneously for the same symbol here, but in this case the loading of the symbols happens successfully I think.

cpp/daal/src/threading/threading.cpp
Vika-F: Done.

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -83,50 +83,234 @@ services::Status prepareCrossProduct(size_t nFeatures, algorithmFPType * crossPr
...
+            {
+                DAAL_ITTNOTIFY_SCOPED_TASK(reducer.update.sums);
+                algorithmFPType * sumsPtr = sums();
+                if (!sumsPtr)
+                {
+                    errorCode = ErrorCode::memAllocationFailed;
+                    return;
+                }
+                /* Sum input array elements in case of non-normalized data */
+                for (DAAL_INT i = 0; i < nRows; i++)
```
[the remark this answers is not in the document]
Vika-F: sumPtr is always initialized with zeroes as it is allocated by TArrayScalable**Calloc**.

Vika-F:
Link to the latest CI: http://intel-ci.intel.com/f02505a5-9ed8-f10d-b412-a4bf010d0e2d

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -157,130 +341,87 @@ services::Status updateDenseCrossProductAndSums(bool isNormalized, size_t nFeatu
...
-                    }
-                }
+                return services::Status(services::ErrorCovarianceInternal);
             }
+        }
 
-            delete tls_data_local;
-        });
+        const algorithmFPType * resultCrossProduct = result.crossProduct();
+        if (result.errorCode != CovarianceReducer<algorithmFPType, cpu>::ok || !resultCrossProduct)
```
avolkov-intel, line 391: Should result.errorCode here be equal to memAllocationFailed?

cpp/daal/src/threading/threading.cpp
```diff
@@ -835,5 +835,91 @@ DAAL_EXPORT void _daal_wait_task_group(void * taskGroupPtr)
     ((tbb::task_group *)taskGroupPtr)->wait();
 }
 
+/// Class to manage the lifetime of the daal::Reducer object.
+/// It uses a unique_ptr with default_deleter or empty deleter to ensure that the
+/// destructor of the daal::Reducer object is called when needed.
+class ReducerUniquePtr : public std::unique_ptr<daal::Reducer, std::function<void(daal::Reducer *)> >
```
avolkov-intel, line 841: Should we use DAAL implementation of unique ptr as it is done in other places? https://github.com/uxlfoundation/oneDAL/blob/d20d38a244127a3109f5b7587a0109cbad4cac37/cpp/daal/src/services/service_unique_ptr.h#L30

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -83,50 +83,234 @@ services::Status prepareCrossProduct(size_t nFeatures, algorithmFPType * crossPr
...
+        bool isOverflow = false;
+        DAAL_OVERFLOW_CHECK_BY_MULTIPLICATION_BOOL(size_t, _numRowsInBlock, _nFeatures, isOverflow);
+        if (isOverflow)
         {
-            delete object;
-            return nullptr;
+            errorCode = ErrorCode::intOverflow;
+            return;
         }
+        const size_t numRowsInLastBlock = _numRowsInBlock + _dataTable->getNumberOfRows() - _numBlocks * _numRowsInBlock;
```
avolkov-intel, line 193: We need to check that the value on the right is non negative, so that it can be stored in size_t datatype

Vika-F: The value on the right cannot be negative as _numBlocks was computed as:
```
        size_t numBlocks = _dataTable->getNumberOfRows() / numRowsInBlock;
        if (numBlocks * numRowsInBlock < nVectors)
        {
            numBlocks++;
        }
```
So `_numRowsInBlock + _dataTable->getNumberOfRows()` is always greater than `_numBlocks * _numRowsInBlock` by definition.

cpp/daal/src/externals/core_threading_win_dll.cpp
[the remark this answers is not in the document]
keeranroth: Static initialization of `_daal_threader_reduce_ptr` is to NULL. On first call of this function we try and set the value of the pointer, without a mutex or any other synchronization primitive. If two threads are in here, if one sees `_daal_threader_reduce_ptr == NULL`, and is part way through the update of the 64-bit pointer, then it _could_ be the case that another thread sees `_daal_threader_reduce_ptr != NULL`, but with only 32-bits updated. Trying to derefence that will at best be a segfault.

That is a lot of ifs, and it is a long-shot to be hit, but I do think it is a race.

Vika-F:
Latest CI: http://intel-ci.intel.com/f02b1814-5836-f128-935a-a4bf010d0e2d
The previous one got a timeout on Linux.

avolkov-intel (approved):

cpp/daal/src/externals/core_threading_win_dll.cpp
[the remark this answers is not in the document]
Vika-F: Thank you for the explanation.
A follow up PR was created based on this discussion thread: https://github.com/uxlfoundation/oneDAL/pull/3195

PR #3139, opened 2025-03-24
init parallel trees method

Alexandr-Solovev:
## Summary

This PR introduces a new method `local_trees` for the Random Forest implementation. The method builds local trees independently on each GPU node, enabling efficient parallelization and improved scaling.

### Key Changes

- ✅ Added `local_trees` method for distributed tree construction across GPUS.
- ✅ Introduced functionality to copy local trained models in the global model.
- ✅ Added explicit calls for model serialization and deserialization(the distributed process based on the 
serialize local models -> allgather local models -> deserialize local models -> copy local models in the global model).

---

**Testing**

**Performance**


cpp/daal/src/algorithms/dtrees/dtrees_model_impl.h
```diff
@@ -495,6 +495,47 @@ class DAAL_EXPORT ModelImpl
...
+        {
+            resize(global_count); // sets _nTree = 0
+            _nTree.set(global_count);
+        }
+        // copy data if source and target pointers are valid
+        if (isValid() && other.isValid())
+        {
+            for (size_t i = 0; i < other._nTree.get(); i++)
+            {
+                (*_serializationData)[idx + i] = (*other._serializationData)[i];
```
Vika-F: Can it be replaced with `memcpy`? For example, this one:
https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/services/daal_memory.h#L71

cpp/daal/src/algorithms/dtrees/dtrees_model_impl.h
```diff
@@ -495,6 +495,47 @@ class DAAL_EXPORT ModelImpl
...
+        {
+            resize(global_count); // sets _nTree = 0
+            _nTree.set(global_count);
+        }
+        // copy data if source and target pointers are valid
+        if (isValid() && other.isValid())
+        {
+            for (size_t i = 0; i < other._nTree.get(); i++)
+            {
+                auto probtbl                   = new data_management::HomogenNumericTable<double>(0, 0, data_management::NumericTable::doAllocate);
```
Vika-F: Please use `HomogenNumericTable<double>::create(...)` instead of `new HomogenNumericTable<double>`.
Also is it correct that the data type is fixed to `double`?

cpp/daal/src/algorithms/dtrees/dtrees_model_impl.h
```diff
@@ -495,6 +495,47 @@ class DAAL_EXPORT ModelImpl
 
     const data_management::DataCollection * serializationData() const { return _serializationData.get(); }
 
+    void copy_model_cls(const ModelImpl & other, size_t idx, size_t global_count)
```
Vika-F: 1. Camel case should be used in DAAL:
```suggestion
    void copyModelCls(const ModelImpl & other, size_t idx, size_t global_count)
```
2. I think this method have to be moved into `daal::algorithms::decision_forest::classification::internal::ModelImpl`; and copy_model_reg into regression ModelImpl respectively.

cpp/oneapi/dal/algo/decision_forest/common.hpp
```diff
@@ -514,6 +516,16 @@ class descriptor : public detail::descriptor_base<Task> {
         return *this;
     }
 
+    /// The distributed trees build.
```
Vika-F: Can you please write more detailed description? This goes into the docs, and from this description it is not clear what is the purpose of this flag.

cpp/daal/include/algorithms/regression/regression_model.h
```diff
@@ -67,13 +67,30 @@ class DAAL_EXPORT Model : public daal::algorithms::Model
...
+      *  \return Number of features in the dataset was used on the training stage
+      */
     virtual size_t getNumberOfFeatures() const = 0;
+
+    /**
+      *  Sets the number of features in the dataset was used on the training stage
+      *  \DAAL_DEPRECATED
+      *  \param[in]  nFeatures  Number of features in the dataset was used on the training stage
+      */
+    virtual void setNFeatures(size_t /*nFeatures*/) {}
```
Vika-F: It looks like the implementations of all the regression algorithms models are derived from `regression::internal::ModelInternal` here:
https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/src/algorithms/regression/regression_model_impl.h

And it already contains the info about `nFeatures`. I guess the serialization can be modified properly for that class to serialize the info about `nFeatures` so you won't have to add those new APIs here.

cpp/daal/src/algorithms/dtrees/dtrees_model_impl.h
Alexandr-Solovev: yes, by default it adds bool ModelImpl::add(const TreeType & tree, size_t nClasses, size_t iTree)
{
    DAAL_CHECK_STATUS_VAR(!(size() >= _serializationData->size()));
    _nTree.inc();
    const size_t nNode = tree.getNumberOfNodes();

    auto pTbl           = new DecisionTreeTable(nNode);
    auto impTbl         = new HomogenNumericTable<double>(1, nNode, NumericTable::doAllocate);
    auto nodeSamplesTbl = new HomogenNumericTable<int>(1, nNode, NumericTable::doAllocate);
    auto probTbl        = new HomogenNumericTable<double>(0, 0, NumericTable::doAllocate);

cpp/daal/src/algorithms/dtrees/dtrees_model_impl.h
Alexandr-Solovev: you mean idx by idx, or once for the whole table? Not sure that I can really get the count of bytes to copy

cpp/daal/src/algorithms/dtrees/dtrees_model_impl.h
Alexandr-Solovev: upd: create is not working directly, due to 
`error: no matching member function for call to 'reset'
  186 |             (*_probTbl)[idx + i].reset(probtbl);`
  
  my suggestion is to hold it as it is, especially due to for add function we call new

Vika-F: `.reset` doesn't work because `create` returns a shared pointer, not a raw pointer.
Try  `(*_probTbl)[idx + i] = probtbl` instead of `.reset`.

cpp/daal/src/algorithms/dtrees/dtrees_model_impl.h
Vika-F: I didn't realize those are objects.
It would be better to leave it as is.

cpp/daal/src/algorithms/dtrees/dtrees_model_impl.h
Alexandr-Solovev: fixed in next commit

PR #3155, opened 2025-04-02
feature: onedal verbose profiler

Alexandr-Solovev:
### Summary

This PR introduces internal support for logging, tracing, and analysis in oneDAL. The functionality is conditionally enabled using the `ONEDAL_VERBOSE` environment variable.

### Changes Introduced

- Added `logger`, `tracer`, and `analyzer` components to the oneDAL internal infrastructure.
- These tools provide structured tracing of algorithm and kernel execution.
- Logging behavior is controlled via the `ONEDAL_VERBOSE` environment variable:
  - `ONEDAL_VERBOSE=0/empty`: Logging is disabled (default).
  - `ONEDAL_VERBOSE=1`: Logging enabled.
  - `ONEDAL_VERBOSE=2`: Tracer enabled.
  - `ONEDAL_VERBOSE=3`: Analyzer enabled.
  - `ONEDAL_VERBOSE=4`: Analyzer logger and tracer are enabled.
  - `ONEDAL_VERBOSE=5`: Analyzer logger and tracer are enabled with service function.

### Purpose

These additions are intended to:
- Help developers debug and profile algorithms and kernels.
- Analyze performance bottlenecks and nested execution paths.
- Provide structured insights into internal flow without affecting public APIs.

### Additional Notes

- Logging is zero-cost when disabled (no runtime overhead).
- All functionality is internal and transparent to end users.
- Future improvements may extend the analyzer with visualization or export support.

---

**Testing**

**Performance**


david-cortes-intel:
Thanks for looking into this. A couple comments:
* Please add documentation about this variable.
* It doesn't work if the environment variable is set after the library is imported. If this is hard to change, then please document this aspect too.
* I cannot find any information about execution times in the outputs. For example, this is what I see from linear regression with verbosity level 3:
    ```
    auto daal::algorithms::linear_model::normal_equations::training::internal::ThreadingTask<double, daal::avx512>::update(long long, long long, const NumericTable &, const NumericTable &)::(anonymous class)::operator()() const [algorithmFPType = double, cpu = daal::avx512]
    Profiler task_name: computeUpdate.syrkX-----------------------------------------------------------------------------
    auto daal::algorithms::linear_model::normal_equations::training::internal::ThreadingTask<double, daal::avx512>::update(long long, long long, const NumericTable &, const NumericTable &)::(anonymous class)::operator()() const [algorithmFPType = double, cpu = daal::avx512]
    Profiler task_name: computeUpdate.gemm1X
    ```
* I'm not sure if this is coming from this PR, but it looks like linear regression is printing a lot more than it should. For example, it is generating one print for each subproblem that calls syrk/gemm, whereas last time I looked at it it had only one call to kernel profiling before the loop that makes multiple of those calls.
* Given the amount of prints, perhaps this could use pure C++ streams not synchronized with C. Info: https://en.cppreference.com/w/cpp/io/ios_base/sync_with_stdio
* Looks like there's quite a bit of commented out code that needs removal.

david-cortes-intel:
@Alexandr-Solovev I'm still seeing the same kinds of prints without timings after the last updates.

Alexandr-Solovev:
@david-cortes-intel Thanks for the comments! With the latest commit, in case of usage Jupyter times should be available with tracer(ONEDAL_VERBOSE 2 4 5), you can use 4 or 5 to enable everything.

- I fixed "Given the amount of prints, perhaps this could use pure C++ streams not synchronized with C", "It doesn't work if the environment variable is set after the library is imported. If this is hard to change, then please document this aspect too."

- About the "I'm not sure if this is coming from this PR, but it looks like linear regression is printing a lot more than it should. For example, it is generating one print for each subproblem that calls syrk/gemm, whereas last time I looked at it it had only one call to kernel profiling before the loop that makes multiple of those calls." Not sure, I had the same structure, but I can take a look if its important.

david-cortes-intel:
@Alexandr-Solovev Thanks, I can see the times now. But it again looks like it generates one log per each batch of linear regression:
Status ThreadingTask<algorithmFPType, cpu>::update(DAAL_INT startRow, DAAL_INT nRows, const NumericTable & xTable, const NumericTable & yTable)

.. whereas it should be generating one entry for the whole kernel:
https://github.com/uxlfoundation/oneDAL/blob/5dd39b869bf90afd59ba9073f22203536d5e0e5c/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i#L234

@Vika-F Any comments here?

david-cortes-intel:
> @david-cortes-intel I checked and its a threading task(ThreadingTask<algorithmFPType, cpu>::update) I assume that each thread prints own information. @Vika-F I can try to add new type of tasks for threading. @david-cortes-intel Can you limit the number of threads and check changes?

I get slightly fewer prints when limiting threads, but it still looks like it prints a lot more than it should.

Alexandr-Solovev:
> > @david-cortes-intel I checked and its a threading task(ThreadingTask<algorithmFPType, cpu>::update) I assume that each thread prints own information. @Vika-F I can try to add new type of tasks for threading. @david-cortes-intel Can you limit the number of threads and check changes?
> 
> I get slightly fewer prints when limiting threads, but it still looks like it prints a lot more than it should.

It must print just once for each PROFILER_TASK call, so, may be smth is called in a loop or idk

david-cortes-intel:
I'm not sure what exactly is happening there.

For our use case, it ideally shouldn't print anything about those per-thread tasks, just about the larger task where they are called in a loop.

But still, the amount of prints doesn't seem to correspond with the number of calls.

Currently, it divides the data in batches of 256 rows, so for example if I pass 5k rows, it would amount to 20 batches, but it makes 81 prints instead:
```python
import os
os.environ["ONEDAL_VERBOSE"] = "4"
import numpy as np
from sklearnex.linear_model import LinearRegression
rng = np.random.default_rng(seed=123)
X = rng.standard_normal(size=(5000,40))
y = rng.standard_normal(size=X.shape[0])
model = LinearRegression().fit(X, y)
```
Using a single thread reduces those to 42, but that's still more than expected.

Alexandr-Solovev:
@david-cortes-intel can you launch it as an example(not through Jupyter) and check the algorithm tree to see what are nested calls?

david-cortes-intel:
@Alexandr-Solovev Attached is the full log with the analyzer and prints whenrunning it without the examples, on the same data. I see 80 prints which is just one less than before.
[log_linreg.txt](https://github.com/user-attachments/files/19790287/log_linreg.txt)

cpp/daal/src/externals/service_profiler.cpp
```diff
@@ -16,50 +16,307 @@
...
+*
+*  @returns pointer to mode
+*                      0/empty disabled
+*                      1 enabled only with logger
+*                      2 enabled only with tracer
+*                      3 enabled only with analyzer
+*                      4 enabled with logger tracer and analyzer
+*                      5 enabled with logger tracer and analyzer with service functions
+*/
+int * daal_verbose_mode()
```
Vika-F, line 117: It's not obvious why it is returned by pointer here. Can you please explain?

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Using oneDAL Verbose Mode
+=========================
+
+When building applications that call Intel® oneAPI Data Analytics Library functions, it may be useful to determine:
+
+- which computational functions are called
+- what parameters are passed to them
+- how much time is spent to execute the functions
+- (for GPU applications) which GPU device the kernel is executed on
```
david-cortes-intel, line 23:
```suggestion
- Which computational kernels are called
- What parameters are passed to them
- How much time is spent to execute the functions
- Which device (CPU/GPU) the kernel is executed on
```

cpp/daal/src/externals/service_profiler.cpp
```diff
@@ -16,50 +16,307 @@
...
+
+    daal_verbose_val = newval;
+    // read_done        = 1;
 }
 
-ProfilerTask Profiler::startTask(const char * taskName)
+std::string format_time_for_output(std::uint64_t time_ns)
 {
-    return ProfilerTask(taskName);
+    std::ostringstream out;
```
david-cortes-intel, line 76: Would there be any issue in writing directly to `stdout` if the prints from threaded kernels are removed?

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+
+You can get an application to print this information to a standard output
+device by enabling **Intel® oneAPI Data Analytics Library Verbose**.
```
david-cortes-intel, line 26:
```suggestion
device by enabling **Intel® oneAPI Data Analytics Library Verbose Mode**.
```

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+
+When Verbose mode is active in oneDAL, every call of a verbose-enabled function finishes with 
+printing a human-readable line describing the call. Even, if your application gets terminated during
```
david-cortes-intel, line 29:
```suggestion
printing a human-readable line describing the call. Even if the application gets terminated during 
```

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+You can get an application to print this information to a standard output
+device by enabling **Intel® oneAPI Data Analytics Library Verbose**.
+
+When Verbose mode is active in oneDAL, every call of a verbose-enabled function finishes with 
+printing a human-readable line describing the call. Even, if your application gets terminated during 
+the function call, information for that function will be printed. 
+The first call to a verbose-enabled function also prints a version information line.
+
+For GPU applications, additional information (one or more GPU information lines) will also 
+be printed by the first call to a verbose-enabled function, following the version information line printed
```
david-cortes-intel, line 34:
```suggestion
be printed on the first call to a verbose-enabled function, following the version information lines printed
```

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+for the host CPU.
+
+Modes of Verbose
```
david-cortes-intel, line 37:
```suggestion
Verbosity Levels
```

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+----------------
+
+We have common implementations for verbose with CPU and GPU applications.
```
david-cortes-intel, line 40:
```suggestion
We have common implementations for verbose mode with CPU and GPU applications.
```

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+
+- **Applications**:
```
david-cortes-intel, line 42:
```suggestion
- **Levels**:
```

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
++--------------------------+-----------------------------------------------+
+| Set ``ONEDAL_VERBOSE=5`` | All tools and service functions enabled       |
++--------------------------+-----------------------------------------------+
+
+
+When enabled with timing, execution timing is taken synchronously, meaning previous kernel execution will block later kernels.
+
+General Notes on Verbose Mode
+-----------------------------
+
```
david-cortes-intel, line 76: Here could add a note that the environment variable only has an effect when set before oneDAL is loaded.

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+Modes of Verbose
+----------------
+
+We have common implementations for verbose with CPU and GPU applications.
+
+- **Applications**:
+  - Disabled (default)
+  - Logger enabled
+  - Tracer enabled
+  - Analyzer enabled
```
david-cortes-intel, line 46: Could be useful to explain what this is and show an example output snippet.

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
++--------------------------+-----------------------------------------------+
+| Set ``ONEDAL_VERBOSE=3`` | Analyzer enabled                              |
++--------------------------+-----------------------------------------------+
+| Set ``ONEDAL_VERBOSE=4`` | All tools enabled (Logger + Tracer + Analyzer)|
++--------------------------+-----------------------------------------------+
+| Set ``ONEDAL_VERBOSE=5`` | All tools and service functions enabled       |
++--------------------------+-----------------------------------------------+
+
+
+When enabled with timing, execution timing is taken synchronously, meaning previous kernel execution will block later kernels.
```
david-cortes-intel, line 72: Is "enabled with timing" the same as "tracer enabled"?

cpp/daal/src/externals/service_profiler.cpp
```diff
@@ -16,50 +16,307 @@
...
+    return service_debug_value;
 }
 
-void Profiler::endTask(const char * taskName) {}
+// Check for logger enabling(PRETTY_FUNCTION + Params)
+bool profiler::is_logger_enabled()
+{
+    static const bool logger_value = [] {
+        int value = *daal_verbose_mode();
+        return value == 1 || value == 4 || value == 5;
```
Vika-F, line 145: Please define constants that represent the values `1`, `4`, `5` here and in other similar places.
It could be done in the top of the file, near `daal_verbose_val` declaration. I'd also convert some values into bit flags in order to simplify the logic of the checks.
For example:
```cpp
constexpr unsigned int DAAL_VERBOSE_ENABLED           = (1U << 0);
constexpr unsigned int DAAL_VERBOSE_TRACER_ENABLED    = (1U << 1);
constexpr unsigned int DAAL_VERBOSE_LOGGER_ENABLED    = (1U << 2);
constexpr unsigned int DAAL_VERBOSE_ANALYSER_ENABLED  = (1U << 3);
constexpr unsigned int DAAL_VERBOSE_SERVICE_FUNC_ENABLED  = (1U << 4);
constexpr unsigned int DAAL_VERBOSE_ALL_ENABLED  = (DAAL_VERBOSE_ENABLED | DAAL_VERBOSE_TRACER_ENABLED | DAAL_VERBOSE_LOGGER_ENABLED | DAAL_VERBOSE_ANALYSER_ENABLED);
constexpr unsigned int DAAL_VERBOSE_SERVICE_ENABLED = (DAAL_VERBOSE_ALL_ENABLED | DAAL_VERBOSE_SERVICE_FUNC_ENABLED);
```

In this case the check will look like:
`return (value & DAAL_VERBOSE_LOGGER_ENABLED);`

Vika-F: But the logic of `set_verbose_from_env` would have to be modified to convert from 0, 1, 2, 3... to bit flags.

I do not insist on bit flags, but the constants 1, 2, 3,... have to be named somehow.

cpp/oneapi/dal/detail/profiler.hpp
```diff
@@ -20,51 +20,200 @@
...
+inline void profiler_log_named_args(const char* /*names*/) {
+    // base case — no args
+}
+
+template <typename T, typename... Rest>
+void profiler_log_named_args(const char* names, const T& value, Rest&&... rest) {
+    const char* comma = strchr(names, ',');
+    std::string name = comma ? std::string(names, comma) : std::string(names);
+
+    name.erase(0, name.find_first_not_of(" \t\n\r"));
```
david-cortes-intel, line 155: Maybe could use [std::iswspace](https://en.cppreference.com/w/cpp/string/wide/iswspace) with a different function instead?

cpp/oneapi/dal/detail/profiler.hpp
```diff
@@ -20,51 +20,200 @@
...
+
+    std::cerr << name << ": " << value << "; ";
```
david-cortes-intel, line 157: Why is this one using `cerr` while the others use `cout`?

cpp/daal/src/externals/service_profiler.cpp
```diff
@@ -16,50 +16,307 @@
...
-ProfilerTask::ProfilerTask(const char * taskName) : _taskName(taskName)
+static volatile int daal_verbose_val = -1;
+
+void print_header()
 {
-    _handle = __itt_string_handle_create(taskName);
+    daal::services::LibraryVersionInfo ver;
 
-    __itt_task_begin(Profiler::getDomain(), __itt_null, __itt_null, _handle);
+    std::cout << "Major version:          " << ver.majorVersion << std::endl;
```
david-cortes-intel, line 38: Would be faster with newline (`'\n'`) instead of `std::endl` at each call.

cpp/daal/src/externals/service_profiler.cpp
```diff
@@ -16,50 +16,307 @@
...
+*                      1 enabled only with logger
+*                      2 enabled only with tracer
+*                      3 enabled only with analyzer
+*                      4 enabled with logger tracer and analyzer
+*                      5 enabled with logger tracer and analyzer with service functions
+*/
+int * daal_verbose_mode()
 {
-    __itt_task_end(Profiler::getDomain());
+#ifdef _MSC_VER
```
david-cortes-intel, line 119: Would all of this work on windows if using a compiler other than MSVC?

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+=========================
+
+When building applications that call Intel® oneAPI Data Analytics Library functions, it may be useful to determine:
+
+- which computational functions are called
+- what parameters are passed to them
+- how much time is spent to execute the functions
+- (for GPU applications) which GPU device the kernel is executed on
+
+You can get an application to print this information to a standard output
```
david-cortes-intel, line 25: Worth mentioning that some logs are going to `stdout` and some to `stderr`.

cpp/daal/src/externals/service_profiler.cpp
```diff
@@ -16,50 +16,307 @@
...
+            if (i + 1 < tasks_info.kernels.size())
+            {
+                const auto & next = tasks_info.kernels[i + 1];
+                if (next.level >= entry.level)
+                {
+                    is_last = false;
+                }
+            }
+
+            prefix += is_last ? "└── " : "├── ";
```
Vika-F, line 240: Is it Ok to have non-ASCII symbols in the code? Maybe it would be safer to replace with the IDs of the respective characters?

cpp/daal/src/externals/service_profiler.h
```diff
@@ -20,103 +20,136 @@
...
+
+#ifdef _WIN32
+    #define PRETTY_FUNCTION __FUNCSIG__
+#else
+    #define PRETTY_FUNCTION __PRETTY_FUNCTION__
+#endif
 
 #ifndef __SERVICE_PROFILER_H__
-#define __SERVICE_PROFILER_H__
+    #define __SERVICE_PROFILER_H__
```
Vika-F, line 44: You don't need those in case there is already `#pragma once` in the file.

cpp/oneapi/dal/detail/profiler.cpp
```diff
@@ -14,29 +14,358 @@
...
+    std::cerr << "Product status:         " << ver.productStatus << std::endl;
+    std::cerr << "Build:                  " << ver.build << std::endl;
+    std::cerr << "Build revision:         " << ver.build_rev << std::endl;
+    std::cerr << "Name:                   " << ver.name << std::endl;
+    std::cerr << "Processor optimization: " << ver.processor << std::endl;
+    std::cerr << std::endl;
 }
 
-void profiler::end_task(const char* task_name) {}
+static void set_verbose_from_env(void) {
```
Vika-F, line 52: Why not to call those directly from DAAL and not duplicate here?

Vika-F:
> @david-cortes-intel I checked and its a threading task(ThreadingTask<algorithmFPType, cpu>::update) I assume that each thread prints own information. @Vika-F I can try to add new type of tasks for threading. @david-cortes-intel Can you limit the number of threads and check changes?

I would expect that the times are grouped together in one number for all threads. Otherwise it is too much info and less useful.

david-cortes-intel:
> > @david-cortes-intel I checked and its a threading task(ThreadingTask<algorithmFPType, cpu>::update) I assume that each thread prints own information. @Vika-F I can try to add new type of tasks for threading. @david-cortes-intel Can you limit the number of threads and check changes?
> 
> I would expect that the times are grouped together in one number for all threads. Otherwise it is too much info and less useful.

@Vika-F But do we actually need numbers from per-thread tasks? Before this PR, were those somehow being logged differently than the ones from the main algorithm workflow?

Alexandr-Solovev:
> But do we actually need numbers from per-thread tasks? Before this PR, were those somehow being logged differently than the ones from the main algorithm workflow?

I think in the older one PR I used a set for task_names and the similar tasks have been grouped. But the approach of having set in profiler is incorrect, because you can call the same kernel several times(not in thread, but lets say gemm->func1->gemm, and it will cause of the incorrect output. I can add ONEDAL_PROFILER_THREADING_TASK which will be grouped. Thoughts @david-cortes-intel  @Vika-F ?

david-cortes-intel:
> > But do we actually need numbers from per-thread tasks? Before this PR, were those somehow being logged differently than the ones from the main algorithm workflow?
> 
> I think in the older one PR I used a set for task_names and the similar tasks have been grouped. But the approach of having set in profiler is incorrect, because you can call the same kernel several times(not in thread, but lets say gemm->func1->gemm, and it will cause of the incorrect output. I can add ONEDAL_PROFILER_THREADING_TASK which will be grouped. Thoughts @david-cortes-intel @Vika-F ?

Yes, would be incorrect to mix up times from different kernel calls.

But the current mechanism is not desirable either: threads shouldn't be logging their individual events separately. Especially not when they process things in batches.

Does the current logic (in main branch) have different macros or different behavior for multi-threaded sections and for the main thread?

@Vika-F are the calls for those per-thread `computeUpdate.syrkX` / `computeUpdate.gemm1X` / etc. required for something else (e.g. VTune) if there's a call for one ITTNOTIFY in the main thread that spawns those?

cpp/daal/src/externals/service_profiler.cpp
Alexandr-Solovev: Fixed, with 
`#if defined(__GNUC__) || defined(__clang__)
    if (__builtin_expect((daal_verbose_val == -1), 0))
#else
    if (daal_verbose_val == -1)
#endif
        set_verbose_from_env();`

cpp/oneapi/dal/detail/profiler.cpp
Alexandr-Solovev: Done

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,76 @@
...
+For GPU applications, additional information (one or more GPU information lines) will also 
+be printed on the first call to a verbose-enabled function, following the version information lines printed
+for the host CPU.
+
+Verbosity Levels
+----------------
+
+We have common implementations for verbose mode with CPU and GPU applications.
+
+  - **Levels**:
```
david-cortes-intel, line 42:
```suggestion
Levels:
```

cpp/daal/src/externals/service_profiler.cpp
Alexandr-Solovev: I replaced this code with default "| and -"

cpp/oneapi/dal/detail/profiler.cpp
```diff
@@ -14,29 +14,358 @@
...
 }
 
-profiler_task::profiler_task(const char* task_name, const sycl::queue& task_queue)
-        : task_name_(task_name),
-          task_queue_(task_queue) {}
+/**
+* Returns the pointer to variable that holds oneDAL verbose mode information (enabled/disabled)
+*
+*  @returns pointer to mode
+*                      0/empty disabled
```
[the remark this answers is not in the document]
Alexandr-Solovev: I think its for developers, not for customers and not a part of doc gen

docs/source/contribution/profiling.rst
Alexandr-Solovev: It has been fixed and theoretically not an issue

cpp/daal/src/externals/service_profiler.cpp
Alexandr-Solovev: then it will not be possible to call smth 
std::cerr<<"kernel name"<<formatted_time()<<std::endl;
and It will be called like
std::cerr<<"kernel name";
formatted_time();
std::cerr<<std::endl;

cpp/daal/src/externals/service_profiler.cpp
Alexandr-Solovev: it has been fixed :)

cpp/daal/src/externals/service_profiler.cpp
david-cortes-intel: I believe it should be possible to explicitly define it as a UTF8 string in C++, but looks like it would require the file itself to also be encoded in UTF8, unless going by the codes. Translating those into codes:
* `└── ` = `\u0020`, `\u2514`, `\u2500`, `\u2500`, `\u0020`
* `├── ` = `\u0020`, `\u251C`, `\u2500`, `\u2500`, `\u0020`

david-cortes-intel: Example:
image

Vika-F:
> @Vika-F are the calls for those per-thread `computeUpdate.syrkX` / `computeUpdate.gemm1X` / etc. required for something else (e.g. VTune) if there's a call for one ITTNOTIFY in the main thread that spawns those?

Until this PR those macros were also used by ITTNOTIFY, yes. But the grouping itself was not implemented by me.
When visualized by VTune, the numbers for the similar tasks are grouped together. For example, it there are call sequences: funcA -> onemkl.gemm and funcB -> onemkl.gemm, than the report contains only one number for onemkl.gemm summed from both funcA and funcB.

This makes the analysis harder sometimes, but at least the amount of data to analyze is limited in case of ITTNOTIFY.
Having too much data in the report can also complicate the things.

In my opinion it would be better to have a trade-off between the amount of the details in the report and ease of analysis. Need to see how the grouped results would look like.

david-cortes-intel:
@Alexandr-Solovev Thanks for the fixes. But I'm still struggling to understand the outputs.

There's a log from running linear regression for example:
[linreg_log.txt](https://github.com/user-attachments/files/19848113/linreg_log.txt)

From my POV, there is one important kernel which is profiled from this line:
https://github.com/uxlfoundation/oneDAL/blob/e4274a0dfb8cf7e5b3e76bae36ebbe192cf4d027/cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i#L234

That kernel is only executed once during the course of the algorithm.

So ideally, what I would expect to see here is one entry for that kernel with all the time spent inside that function. But there are lots of logs about the individual 'syrkX' and 'gemm1X' which are done in batches and in parallel. There appears to be fewer prints of those now than there were before, but there's still a large amount of those, and they increase as the number of batches increases (in the log attached here, there should be 20 such batches executed).

What are those individual prints? Is it possible to group them all together? Is there a difference between the first prints and the ones that come after "Profiler task_name: ... "?

david-cortes-intel:
@Alexandr-Solovev I'm getting a missing symbol error under the last commit for this method:
```
daal::internal::profiler_task::~profiler_task()
```

david-cortes-intel:
@Alexandr-Solovev Would it be possible to add more details to the prints using macros `__LINE__` and `__FILE__`?

Also another question: I see now there's two prints for the same task name, with one containing a time and the other not containing it:
image

What's the difference between those? Are there two tasks registering themselves with the same name?

Alexandr-Solovev:
> @Alexandr-Solovev Would it be possible to add more details to the prints using macros `__LINE__` and `__FILE__`?
> 
> Also another question: I see now there's two prints for the same task name, with one containing a time and the other not containing it: image
> 
> What's the difference between those? Are there two tasks registering themselves with the same name?

@david-cortes-intel yes, sure I will add __LINE__ and __FILE__ in the function. About the timing I would like to say that its more likely threading task, and probably it was printed in incorrect order. Like somewhere also you will probably find the time. I will try to save order

Alexandr-Solovev:
@david-cortes-intel btw for now all threading outputs are hidden, and with VERBOSE=5 are shown. So, can you check the output for now? I will try today to resolve the threading tasks with atomics

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
```diff
@@ -139,7 +139,7 @@ template <typename algorithmFPType, CpuType cpu>
 void ThreadingTask<algorithmFPType, cpu>::reduce(algorithmFPType * xtx, algorithmFPType * xty)
 {
     {
-        DAAL_ITTNOTIFY_SCOPED_TASK(computeUpdate.syrkX);
+        DAAL_PROFILER_THREADING_TASK(reduce.syrkX);
```
david-cortes-intel, line 142: These ones somehow still end up printing more than once despite having 'threading task':
image

Alexandr-Solovev: yes, I split it from default task, but didnt add atomic/other ability to print it only once. Working on it

david-cortes-intel:
> @david-cortes-intel btw for now all threading outputs are hidden, and with VERBOSE=5 are shown. So, can you check the output for now? I will try today to resolve the threading tasks with atomics

Thanks, gave it a try now and it looks like it avoids the race conditions. But it still looks like the times don't add up:
image

* The first print there 'computeUpdate' should comprise the majority of the time (that's where the threaded tasks come are executed).
* The second part 'computeFinalize' appears to have the timings from the first kernel added to it.

Alexandr-Solovev:
> Thanks, gave it a try now and it looks like it avoids the race conditions. But it still looks like the times don't add up:

The kernel profiler works as follows: when the code reaches the profiler's scope, it prints out some debug information and records the start time. As soon as it leaves the scope, the destructor is triggered and the durations are accumulated. Therefore, since finalizeCompute is called after the threading part, it indeed appears to take a long time.
You can also try to use analyzer(it shows merged threading tasks) to see how much time we spent on it

david-cortes-intel:
> > Thanks, gave it a try now and it looks like it avoids the race conditions. But it still looks like the times don't add up:
> 
> The kernel profiler works as follows: when the code reaches the profiler's scope, it prints out some debug information and records the start time. As soon as it leaves the scope, the destructor is triggered and the durations are accumulated. Therefore, since finalizeCompute is called after the threading part, it indeed appears to take a long time. You can also try to use analyzer(it shows merged threading tasks) to see how much time we spent on it

But in this case, 'finalize' is a different scoped task which gets called after 'update' has finished. The threading tasks are called within 'update', not within 'finalize', but it looks like their times are added to 'finalize' instead.

Alexandr-Solovev:
> > > Thanks, gave it a try now and it looks like it avoids the race conditions. But it still looks like the times don't add up:
> > 
> > 
> > The kernel profiler works as follows: when the code reaches the profiler's scope, it prints out some debug information and records the start time. As soon as it leaves the scope, the destructor is triggered and the durations are accumulated. Therefore, since finalizeCompute is called after the threading part, it indeed appears to take a long time. You can also try to use analyzer(it shows merged threading tasks) to see how much time we spent on it
> 
> But in this case, 'finalize' is a different scoped task which gets called after 'update' has finished. The threading tasks are called within 'update', not within 'finalize', but it looks like their times are added to 'finalize' instead.

can you share analyzer tree? it shouldnt summarize the times for finalize. You can also summarize the finalize step nested tasks to check

david-cortes-intel:
Looks like it crashes before reaching the analyzer:
```
auto daal::algorithms::linear_model::normal_equations::training::internal::ThreadingTask<float, daal::avx512>::update(long long, long long, const NumericTable &, const NumericTable &)::(anonymous class)::operator()() const [algorithmFPType = float, cpu = daal::avx512]
Profiler task_name: update.gemm1Y
update.gemm1Y 260ns
double free or corruption (!prev)
[1]    1688955 abort (core dumped)  ./_cmake_results/intel_intel64_so/lin_reg_norm_eq_dense_batch
```

Alexandr-Solovev:
> update.gemm1Y

thanks! I reproduced the issue, working on the fix

david-cortes-intel:
In case it helps, we now get nice error traces from ASAN. It requires some workarounds to make it execute the examples when the library is not compiled with GCC though, so attaching the log here in case it helps.
[asan_tracer_log.txt](https://github.com/user-attachments/files/19867026/asan_tracer_log.txt)

Alexandr-Solovev:
> In case it helps, we now get nice error traces from ASAN. It requires some workarounds to make it execute the examples when the library is not compiled with GCC though, so attaching the log here in case it helps. [asan_tracer_log.txt](https://github.com/user-attachments/files/19867026/asan_tracer_log.txt)

the issue should be fixed with the latest commit

david-cortes-intel:
Actually I'm seeing the reason for the difference in timings now: there's one missing component here which is not being profiled:
https://github.com/uxlfoundation/oneDAL/blob/d080d5e9a9b24a47f1d70a3deb34c89bd67a7a92/cpp/daal/src/algorithms/service_kernel_math.h#L841

david-cortes-intel:
@Alexandr-Solovev Could you add some extra profiler tasks for these? It still looks like these parts should be faster than the previous ones, so let's see if the times add up afterwards:
https://github.com/uxlfoundation/oneDAL/blob/d080d5e9a9b24a47f1d70a3deb34c89bd67a7a92/cpp/daal/src/algorithms/service_kernel_math.h#L841
https://github.com/uxlfoundation/oneDAL/blob/d080d5e9a9b24a47f1d70a3deb34c89bd67a7a92/cpp/daal/src/algorithms/service_kernel_math.h#L648
https://github.com/uxlfoundation/oneDAL/blob/d080d5e9a9b24a47f1d70a3deb34c89bd67a7a92/cpp/daal/src/algorithms/service_kernel_math.h#L690

david-cortes-intel:
@Alexandr-Solovev After the latest changes, now verbosity level=2 works as expected, but I'm seeing many strange behaviors:
* I think levels 1 and 2 are swapped ('1' is a lot more verbose than '2').
* Level 3 doesn't print anything.
* Levels 2 and 4 have logs from 'finalize', but not from 'update' - example:
image

Alexandr-Solovev:
> @Alexandr-Solovev After the latest changes, now verbosity level=2 works as expected, but I'm seeing many strange behaviors:
> 
> * I think levels 1 and 2 are swapped ('1' is a lot more verbose than '2').
> * Level 3 doesn't print anything.
> * Levels 2 and 4 have logs from 'finalize', but not from 'update' - example:
>   image

hi @david-cortes-intel the issue with computeUpdate logs has been fixed.(at least for 1 iteration run) About option 3 - have you used jupyter or no? I can add a separate method for getting analyzer tree in next updates, but for now its not possible to get it in notebook. Can you check pls the latest commit and let me know if some strange behavior is still exist

david-cortes-intel:
Gave it another try. I now see new prints, but I'm struggling to understand what they mean.

This is for level 2:
```
THREADING Profiler task_name: update.syrkX
THREADING Profiler task_name: update.gemm1X
THREADING Profiler task_name: update.gemmXY
THREADING Profiler task_name: update.gemm1Y
THREADING Profiler task_name: reduce.syrkX
THREADING Profiler task_name: reduce.gemmXY
update.syrkX, Main rank Duration(could be different for other ones): 1.04ms
update.gemm1X, Main rank Duration(could be different for other ones): 37.54us
update.gemmXY, Main rank Duration(could be different for other ones): 33.57ms
update.gemm1Y, Main rank Duration(could be different for other ones): 823ns
reduce.syrkX, Main rank Duration(could be different for other ones): 167.27us
reduce.gemmXY, Main rank Duration(could be different for other ones): 1.29us
computeUpdate 45.81ms
```

Does it mean that it is printing only the timings from the main thread? Is that for one batch or for all? Perhaps it could do away with those, considering that they would show up under a different logger level.

For level 1, it looks like it prints those before the general info:
```
THREADING Profiler task_name: update.syrkX
THREADING Profiler task_name: update.gemm1X
THREADING Profiler task_name: update.gemmXY
THREADING Profiler task_name: update.gemm1Y
THREADING Profiler task_name: reduce.syrkX
THREADING Profiler task_name: reduce.gemmXY
-----------------------------------------------------------------------------
File: ./cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i], Line: 234
Profiler task_name: computeUpdate
Major version:          2025
Minor version:          6
Update version:         0
Product status:         
Build:                  20250424
Build revision:         work
Name:                   oneAPI Data Analytics Library
Processor optimization: Intel(R) Xeon(R) processors based on Intel(R) Advanced Vector Extensions 512

update.syrkX, Main rank Duration(could be different for other ones): 1.02ms
update.gemm1X, Main rank Duration(could be different for other ones): 39.03us
update.gemmXY, Main rank Duration(could be different for other ones): 40.65ms
update.gemm1Y, Main rank Duration(could be different for other ones): 892ns
reduce.syrkX, Main rank Duration(could be different for other ones): 122.00us
reduce.gemmXY, Main rank Duration(could be different for other ones): 651ns
```

For level 3: yes, I am using a jupyter-based client. What exactly is supposed to trigger or not trigger it? It doesn't appear to trigger upon model object destruction either.

cpp/daal/include/services/service_profiler.h
```diff
@@ -0,0 +1,468 @@
...
+        return value == 5;
+    }();
+    return service_debug_value;
+}
+
+inline static bool is_logger_enabled()
+{
+    static const bool logger_value = [] {
+        int value = daal_verbose_mode();
+        return value == 1 || value == 4 || value == 5;
```
Vika-F, line 218: The magic constants `1`, `4`, `5`, etc. have to be named somehow.

cpp/daal/include/services/service_profiler.h
```diff
@@ -0,0 +1,468 @@
...
+class profiler_task
+{
+public:
+    inline profiler_task(const char * task_name, int idx) : task_name_(task_name), idx(idx) {}
+    inline profiler_task(const char * task_name, int idx, bool thread) : task_name_(task_name), idx(idx), is_thread(thread) {}
+    inline ~profiler_task();
+
+private:
+    const char * task_name_;
+    int idx;
```
Vika-F, line 280: Some private fields have trailing underscore like `task_name_` , other do not like `idx`. Please align the naming.

cpp/daal/include/services/service_profiler.h
```diff
@@ -0,0 +1,468 @@
...
+{
+    std::set<task_entry> kernels;
+};
+
+class profiler_task
+{
+public:
+    inline profiler_task(const char * task_name, int idx) : task_name_(task_name), idx(idx) {}
+    inline profiler_task(const char * task_name, int idx, bool thread) : task_name_(task_name), idx(idx), is_thread(thread) {}
+    inline ~profiler_task();
```
Vika-F, line 276: Need to add copy constructor and operator= or mark them deleted in order not to violate the rule of 3.

cpp/daal/include/services/service_profiler.h
```diff
@@ -0,0 +1,468 @@
...
+                std::cerr << prefix << entry.name << " time: " << format_time_for_output(entry.duration) << " " << std::fixed << std::setprecision(2)
+                          << (total_time > 0 ? (double(entry.duration) / total_time) * 100 : 0.0) << "% " << entry.count << " times"
+                          << " in a " << entry.threading_task << " region" << '\n';
+            }
+            std::cerr << "|---(end)" << '\n';
+            std::cerr << "DAAL KERNEL_PROFILER: kernels total time " << format_time_for_output(total_time) << '\n';
+        }
+    }
+
+    inline static profiler_task start_task(const char * task_name)
```
Vika-F, line 324: Please add comments helping to distinguish `start_task` and `start_threading_task`. In which situations one or another is used?

cpp/daal/include/services/service_profiler.h
```diff
@@ -0,0 +1,468 @@
...
+        return static_cast<std::uint64_t>(ns);
+    }
+
+    inline static profiler * get_instance()
+    {
+        static profiler instance;
+        return &instance;
+    }
+
+    inline void merge_tasks()
```
Vika-F, line 410: It is not obvious what this function is meant to do. Please add the comment.

cpp/daal/include/services/service_profiler.h
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    inline task & get_task() { return task_; }
+    inline std::int64_t & get_current_level() { return current_level; }
+    inline std::int64_t & get_kernel_count() { return kernel_count; }
+
+private:
+    std::int64_t current_level = 0;
+    std::int64_t kernel_count  = 0;
+    task task_;
```
Vika-F, line 453: The same comment about trailing underscore.

examples/oneapi/dpc/source/decision_forest/df_cls_hist_batch.cpp
```diff
@@ -81,7 +81,9 @@ int main(int argc, char const* argv[]) {
                   << ", " << d.get_info<sycl::info::device::name>() << "\n"
                   << std::endl;
         auto q = sycl::queue{ d };
-        run(q);
+        for (int i = 0; i < 5; i++) {
```
Vika-F, line 84: Please revert this one.

Vika-F (commented):
The added functionality looks good to me. I have only minor suggestions to the implementation.

But the biggest question I have is: Why is the implementation placed into a public header "daal/include/services/service_profiler.h"?
It looks like a functionality for internal use. And the namespace "daal::internal" also identifies that. Please provide some explanation to this.

cpp/daal/include/services/service_profiler.h
Alexandr-Solovev: Thanks! Fixed!

cpp/daal/include/services/service_profiler.h
Alexandr-Solovev: Fixed!

cpp/daal/include/services/service_profiler.h
Alexandr-Solovev: Fixed! Thanks!

Alexandr-Solovev:
> @Alexandr-Solovev I'm still seeing the same issue with redundant dashes being printed at level 1 after the last commit.

Can you please share the whole log

david-cortes-intel:
> > @Alexandr-Solovev I'm still seeing the same issue with redundant dashes being printed at level 1 after the last commit.
> 
> Can you please share the whole log

Full log:
<details>

```
Major version:          2025
Minor version:          6
Update version:         0
Product status:         
Build:                  20250425
Build revision:         work
Name:                   oneAPI Data Analytics Library
Processor optimization: Intel(R) Xeon(R) processors based on Intel(R) Advanced Vector Extensions 512

-----------------------------------------------------------------------------
File: ./cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i, Line: 234
Profiler task_name: computeUpdate
-----------------------------------------------------------------------------
-----------------------------------------------------------------------------
-----------------------------------------------------------------------------
-----------------------------------------------------------------------------
-----------------------------------------------------------------------------
-----------------------------------------------------------------------------
-----------------------------------------------------------------------------
File: ./cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_finalize_impl.i, Line: 62
Profiler task_name: computeFinalize
-----------------------------------------------------------------------------
File: ./cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_finalize_impl.i, Line: 99
Profiler task_name: computeFinalize.betaBufCopy
-----------------------------------------------------------------------------
File: ./cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_finalize_impl.i, Line: 109
Profiler task_name: computeFinalize.xtxCopy
-----------------------------------------------------------------------------
File: ./cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_finalize_impl.i, Line: 115
Profiler task_name: computeFinalize.computeBetasImpl
-----------------------------------------------------------------------------
File: ./cpp/daal/src/algorithms/service_kernel_math.h, Line: 846
Profiler task_name: solveSymmetricEquationsSystem Printed args: n: 41; nX: 1; 
-----------------------------------------------------------------------------
File: ./cpp/daal/src/algorithms/service_kernel_math.h, Line: 651
Profiler task_name: solveEquationsSystemWithCholesky Printed args: n: 41; nX: 1; 
-----------------------------------------------------------------------------
File: ./cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_finalize_impl.i, Line: 125
Profiler task_name: computeFinalize.copyBetaToResult
```

</details>

Alexandr-Solovev:
> ------------------------------------------------------

@david-cortes-intel  Thanks! The issue was that I printed task name in cout and dashes in cerr. Will be fixed in a few seconds.

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,148 @@
...
+When building applications that call Intel® oneAPI Data Analytics Library functions, it may be useful to determine:
+
+- Which computational kernels are called
+- What parameters are passed to them
+- How much time is spent to execute the functions
+- Which device (CPU/GPU) the kernel is executed on
+
+You can get an application to print this information to a standard output
+device by enabling **Intel® oneAPI Data Analytics Library Verbose Mode**.
+All logs are printed to std::cerr.
```
david-cortes-intel, line 27:
```suggestion
All logs are printed to the `stderr` stream.
```

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,148 @@
...
+
+When Verbose mode is active in oneDAL, every call of a verbose-enabled function finishes with
```
david-cortes-intel, line 29: This doesn't appear to be the case, so perhaps these docs could be updated. Otherwise, verbosity level 3 should print in python as it does indeed exit the function.

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,148 @@
...
+For GPU applications, additional information (one or more GPU information lines) will also 
+be printed on the first call to a verbose-enabled function, following the version information lines printed
+for the host CPU.
+
+Verbosity Levels
+----------------
+
+We have common implementations for verbose mode with CPU and GPU applications.
+
+  - Levels:
```
david-cortes-intel, line 43:
```suggestion
 Levels:
 
```

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,148 @@
...
+Tracer
+------
+
+**Tracer** is a subtool designed to collect and display the execution time of each ``profiler_task``.
+
+Example output (for ``cov_dense_batch``):
+
+::
+
+    reduction.reduce_by_columns        81.72ms
```
david-cortes-intel, line 104: I think this log is outdated now that there's info from threaded tasks from the main thread.

docs/source/contribution/profiling.rst
```diff
@@ -0,0 +1,148 @@
...
+
+    ONEDAL KERNEL_PROFILER: kernels total time 501.60ms
+
+The Analyzer helps visualize nested kernel calls and their contribution to the total runtime. It's especially helpful for understanding which tasks call which kernels and how much time each part consumes.
+
+
+
+General Notes on Verbose Mode
+-----------------------------
+
```
david-cortes-intel, line 148: Here could mention what are the requirements for triggering the analyzer, which doesn't show up on python.

david-cortes-intel:
> > 
> 
> @david-cortes-intel Thanks! The issue was that I printed task name in cout and dashes in cerr. Will be fixed in a few seconds.

Thanks, can confirm that it works correctly now.

I see that you are adding lots of data to the linear regression example - is it somehow needed for something in this PR?

Alexandr-Solovev:
> > > 
> > 
> > 
> > @david-cortes-intel Thanks! The issue was that I printed task name in cout and dashes in cerr. Will be fixed in a few seconds.
> 
> Thanks, can confirm that it works correctly now.
> 
> I see that you are adding lots of data to the linear regression example - is it somehow needed for something in this PR?

it was necessary just to reproduce multithread behavior. Will be reverted

david-cortes-intel (approved):
Waiting for @Vika-F 's comments about the python outputs.

cpp/daal/src/algorithms/linear_model/linear_model_train_normeq_update_impl.i
```diff
@@ -295,7 +295,7 @@ Status UpdateKernel<algorithmFPType, cpu>::compute(const NumericTable & xTable,
     }
 
     /* Split rows by blocks */
-    size_t nRowsInBlock = 128;
+    size_t nRowsInBlock = 64;
```
Alexandr-Solovev, line 298: have to revert

cpp/oneapi/dal/detail/profiler.hpp
```diff
@@ -16,55 +16,111 @@
...
 
-#define ONEDAL_PROFILER_TASK(...)                                                           \
-    oneapi::dal::detail::profiler_task ONEDAL_PROFILER_CONCAT(__profiler_task__,            \
-                                                              ONEDAL_ITTNOTIFY_UNIQUE_ID) = \
-        ONEDAL_PROFILER_GET_MACRO(__VA_ARGS__,                                              \
-                                  ONEDAL_PROFILER_MACRO_2,                                  \
-                                  ONEDAL_PROFILER_MACRO_1,                                  \
-                                  FICTIVE)(__VA_ARGS__)
+// HEADER OUTPUT
+#define ONEDAL_PROFILER_PRINT_HEADER()                                                         \
```
Alexandr-Solovev, line 33: replace with daal impl

ahuber21:
> Breaking ABI changes require a fix.

@Alexsandruss do you have further insights in the API breaks, I think it's a false positive. Let me break it down

ABI checker trips over 4 deleted symbols:

```text
  [D] 'method daal::services::interface1::Status daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<double, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)0>::_ZTSZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIdLNS2_6MethodE0ELNS_7CpuTypeE0EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_0EEERPNS3_11BoundingBoxIdEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEEUlvE0_::operator()()'    {_ZZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIdLNS2_6MethodE0ELNS_7CpuTypeE0EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_0EEERPNS3_11BoundingBoxIdEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEENKUlvE0_clEv}
  [D] 'method daal::services::interface1::Status daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<double, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)4>::_ZTSZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIdLNS2_6MethodE0ELNS_7CpuTypeE4EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_4EEERPNS3_11BoundingBoxIdEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEEUlvE0_::operator()()'    {_ZZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIdLNS2_6MethodE0ELNS_7CpuTypeE4EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_4EEERPNS3_11BoundingBoxIdEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEENKUlvE0_clEv}
  [D] 'method daal::services::interface1::Status daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<float, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)0>::_ZTSZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIfLNS2_6MethodE0ELNS_7CpuTypeE0EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_0EEERPNS3_11BoundingBoxIfEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEEUlvE0_::operator()()'    {_ZZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIfLNS2_6MethodE0ELNS_7CpuTypeE0EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_0EEERPNS3_11BoundingBoxIfEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEENKUlvE0_clEv}
  [D] 'method daal::services::interface1::Status daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<float, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)4>::_ZTSZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIfLNS2_6MethodE0ELNS_7CpuTypeE4EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_4EEERPNS3_11BoundingBoxIfEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEEUlvE0_::operator()()'    {_ZZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIfLNS2_6MethodE0ELNS_7CpuTypeE4EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_4EEERPNS3_11BoundingBoxIfEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEENKUlvE0_clEv}
```

Demangling them gives the following functions

```c++
daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<double, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)0>::buildSecondPartOfKDTree(daal::algorithms::kdtree_knn_classification::training::internal::Queue<daal::algorithms::kdtree_knn_classification::training::internal::BuildNode, (daal::CpuType)0>&, daal::algorithms::kdtree_knn_classification::training::internal::BoundingBox<double>*&, daal::data_management::interface1::NumericTable const&, daal::algorithms::kdtree_knn_classification::interface1::Model&, unsigned long*, daal::algorithms::engines::interface1::BatchBase&)::{lambda()#2}::operator()() const
daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<double, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)4>::buildSecondPartOfKDTree(daal::algorithms::kdtree_knn_classification::training::internal::Queue<daal::algorithms::kdtree_knn_classification::training::internal::BuildNode, (daal::CpuType)4>&, daal::algorithms::kdtree_knn_classification::training::internal::BoundingBox<double>*&, daal::data_management::interface1::NumericTable const&, daal::algorithms::kdtree_knn_classification::interface1::Model&, unsigned long*, daal::algorithms::engines::interface1::BatchBase&)::{lambda()#2}::operator()() const
daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<float, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)0>::buildSecondPartOfKDTree(daal::algorithms::kdtree_knn_classification::training::internal::Queue<daal::algorithms::kdtree_knn_classification::training::internal::BuildNode, (daal::CpuType)0>&, daal::algorithms::kdtree_knn_classification::training::internal::BoundingBox<float>*&, daal::data_management::interface1::NumericTable const&, daal::algorithms::kdtree_knn_classification::interface1::Model&, unsigned long*, daal::algorithms::engines::interface1::BatchBase&)::{lambda()#2}::operator()() const
daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<float, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)4>::buildSecondPartOfKDTree(daal::algorithms::kdtree_knn_classification::training::internal::Queue<daal::algorithms::kdtree_knn_classification::training::internal::BuildNode, (daal::CpuType)4>&, daal::algorithms::kdtree_knn_classification::training::internal::BoundingBox<float>*&, daal::data_management::interface1::NumericTable const&, daal::algorithms::kdtree_knn_classification::interface1::Model&, unsigned long*, daal::algorithms::engines::interface1::BatchBase&)::{lambda()#2}::operator()() const

// These are all the same function with different (template) args
KNNClassificationTrainBatchKernel<...>(...)::{lambda()#2}::operator()() const
```

So ABI checker is complaining that an anonymous, internal lambda is not exported to the lib. Something like this should not trip the check IMO. Or is it considered part of the ABI?

The real question is why is the symbol no longer there? Do you have any idea @Alexandr-Solovev? The specializations corresponding to the template parameters above seem to no longer be compiled.

Alexandr-Solovev:
> > Breaking ABI changes require a fix.
> 
> @Alexsandruss do you have further insights in the API breaks, I think it's a false positive. Let me break it down
> 
> ABI checker trips over 4 deleted symbols:
> 
> ```
>   [D] 'method daal::services::interface1::Status daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<double, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)0>::_ZTSZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIdLNS2_6MethodE0ELNS_7CpuTypeE0EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_0EEERPNS3_11BoundingBoxIdEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEEUlvE0_::operator()()'    {_ZZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIdLNS2_6MethodE0ELNS_7CpuTypeE0EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_0EEERPNS3_11BoundingBoxIdEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEENKUlvE0_clEv}
>   [D] 'method daal::services::interface1::Status daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<double, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)4>::_ZTSZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIdLNS2_6MethodE0ELNS_7CpuTypeE4EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_4EEERPNS3_11BoundingBoxIdEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEEUlvE0_::operator()()'    {_ZZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIdLNS2_6MethodE0ELNS_7CpuTypeE4EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_4EEERPNS3_11BoundingBoxIdEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEENKUlvE0_clEv}
>   [D] 'method daal::services::interface1::Status daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<float, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)0>::_ZTSZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIfLNS2_6MethodE0ELNS_7CpuTypeE0EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_0EEERPNS3_11BoundingBoxIfEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEEUlvE0_::operator()()'    {_ZZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIfLNS2_6MethodE0ELNS_7CpuTypeE0EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_0EEERPNS3_11BoundingBoxIfEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEENKUlvE0_clEv}
>   [D] 'method daal::services::interface1::Status daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<float, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)4>::_ZTSZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIfLNS2_6MethodE0ELNS_7CpuTypeE4EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_4EEERPNS3_11BoundingBoxIfEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEEUlvE0_::operator()()'    {_ZZN4daal10algorithms25kdtree_knn_classification8training8internal33KNNClassificationTrainBatchKernelIfLNS2_6MethodE0ELNS_7CpuTypeE4EE23buildSecondPartOfKDTreeERNS3_5QueueINS3_9BuildNodeELS6_4EEERPNS3_11BoundingBoxIfEERKNS_15data_management10interface112NumericTableERNS1_10interface15ModelEPmRNS0_7engines10interface19BatchBaseEENKUlvE0_clEv}
> ```
> 
> Demangling them gives the following functions
> 
> ```c++
> daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<double, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)0>::buildSecondPartOfKDTree(daal::algorithms::kdtree_knn_classification::training::internal::Queue<daal::algorithms::kdtree_knn_classification::training::internal::BuildNode, (daal::CpuType)0>&, daal::algorithms::kdtree_knn_classification::training::internal::BoundingBox<double>*&, daal::data_management::interface1::NumericTable const&, daal::algorithms::kdtree_knn_classification::interface1::Model&, unsigned long*, daal::algorithms::engines::interface1::BatchBase&)::{lambda()#2}::operator()() const
> daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<double, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)4>::buildSecondPartOfKDTree(daal::algorithms::kdtree_knn_classification::training::internal::Queue<daal::algorithms::kdtree_knn_classification::training::internal::BuildNode, (daal::CpuType)4>&, daal::algorithms::kdtree_knn_classification::training::internal::BoundingBox<double>*&, daal::data_management::interface1::NumericTable const&, daal::algorithms::kdtree_knn_classification::interface1::Model&, unsigned long*, daal::algorithms::engines::interface1::BatchBase&)::{lambda()#2}::operator()() const
> daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<float, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)0>::buildSecondPartOfKDTree(daal::algorithms::kdtree_knn_classification::training::internal::Queue<daal::algorithms::kdtree_knn_classification::training::internal::BuildNode, (daal::CpuType)0>&, daal::algorithms::kdtree_knn_classification::training::internal::BoundingBox<float>*&, daal::data_management::interface1::NumericTable const&, daal::algorithms::kdtree_knn_classification::interface1::Model&, unsigned long*, daal::algorithms::engines::interface1::BatchBase&)::{lambda()#2}::operator()() const
> daal::algorithms::kdtree_knn_classification::training::internal::KNNClassificationTrainBatchKernel<float, (daal::algorithms::kdtree_knn_classification::training::Method)0, (daal::CpuType)4>::buildSecondPartOfKDTree(daal::algorithms::kdtree_knn_classification::training::internal::Queue<daal::algorithms::kdtree_knn_classification::training::internal::BuildNode, (daal::CpuType)4>&, daal::algorithms::kdtree_knn_classification::training::internal::BoundingBox<float>*&, daal::data_management::interface1::NumericTable const&, daal::algorithms::kdtree_knn_classification::interface1::Model&, unsigned long*, daal::algorithms::engines::interface1::BatchBase&)::{lambda()#2}::operator()() const
> 
> // These are all the same function with different (template) args
> KNNClassificationTrainBatchKernel<...>(...)::{lambda()#2}::operator()() const
> ```
> 
> So ABI checker is complaining that an anonymous, internal lambda is not exported to the lib. Something like this should not trip the check IMO. Or is it considered part of the ABI?
> 
> The real question is why is the symbol no longer there? Do you have any idea @Alexandr-Solovev? The specializations corresponding to the template parameters above seem to no longer be compiled.

@ahuber21 not sure that its false positive, until the https://github.com/uxlfoundation/oneDAL/pull/3155/commits/2e39634157aedd12ca4d2382804292aa6054ee90 it was green. But IDK what I changed here

ahuber21:
If my understanding is correct, my best guess is that the compiler started inlining the lambda.

Alexsandruss (approved):

cpp/daal/src/externals/service_memory.cpp
```diff
@@ -23,9 +23,11 @@
 
 #include "src/externals/service_memory.h"
 #include "src/externals/service_service.h"
+#include "services/internal/service_profiler.h"
 
 void * daal::services::daal_malloc(size_t size, size_t alignment)
 {
+    // DAAL_PROFILER_SERVICE_TASK(daal_malloc);
```
icfaust, line 30: @Alexandr-Solovev was this meant to be as a comment?

cpp/daal/include/services/internal/service_profiler.h
```diff
@@ -0,0 +1,539 @@
...
+/*
+//++
+//  Profiler for time measurement of kernels
+//--
+*/
+#pragma once
+
+#include <chrono>
+#include <cstdint>
+#include <cstring>
```
icfaust, line 27: @Alexandr-Solovev @Vika-F @ahuber21 are these standard includes new to some of the daal algorithms? i didnt see them in the removed external profiler header. just want to check this for impact on dll dependencies. i know standard header includes in daal can be tricky. can someone comment to say this is done properly?

cpp/daal/src/externals/service_memory.cpp
Alexandr-Solovev: it was called too much times as a kernel

cpp/daal/include/services/internal/service_profiler.h
Alexandr-Solovev: Related to the ci issue in intelex I have 2 points:
1)The worst thing which can happened with adding these headers - ABI breaking, but breaking ABI doesn't break build process.
2)We have unreproducible locally issue in intelex ci, and in oneDAL ci it works -> it says, at least for me, that it makes sense to doublecheck that the process of building is the same between oneDAL private ci and intelex public nightly step.
My ideas why it fails in intelex:

a) Usage of mutex inside DllMain
This is strongly discouraged on Windows — locking, thread creation, or loading other libraries inside DllMain can cause deadlocks or loader issues, and might result in DLL load failures.

b) Heavy use of static variables and static functions
If static objects have non-trivial constructors or involve threading primitives (e.g., mutexes), they might trigger side effects during DLL initialization that cause the loader to fail silently.

PR #3158, opened 2025-04-04
FIX: Fix potential non-definitiness of BFGS approximation

david-cortes-intel:
## Description

Currently, the L-BFGS code has a logic that discards correction pairs if their curvature is exactly equal to zero.

The approximation requires strict positiveness up to numerical precision, so this condition is insufficient:
* If there's no check for Wolfe conditions, the inner product might be negative, which would still make it pass this check.
* The product can be something very small, like 1e-100, in which case it would lead to issues like numbers turning to infinite.

This PR fixes it by hard-coding a more reasonable threshold, with >= condition.

---

**Testing**

**Performance**

Not applicable.

### DIFF
--- a/cpp/daal/src/algorithms/optimization_solver/lbfgs/lbfgs_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/optimization_solver/lbfgs/lbfgs_dense_default_impl.i
@@ -41,6 +41,9 @@ namespace lbfgs
 {
 namespace internal
 {
+
+constexpr double min_curvature = 1e-8;
+
 template <CpuType cpu>
 bool getOptionalInputData(NumericTable * pCorrectionIndexInput, size_t & lastCorrectionIndex, size_t & lastIteration)
 {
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
Failure in the CI is from SVM multiclass, so not related to the changes here:
```
=================================== FAILURES ===================================
______________________ TestExNpyArray.test_svm_multiclass ______________________

self = <test_daal4py_examples.TestExNpyArray testMethod=test_svm_multiclass>

    @unittest.skipUnless(
        config.check_version(),
        f"Minimum required version {config.required_version} (have {daal_version})",
    )
    @unittest.skipUnless(missing_dep is None, f"Missing dependency: {missing_dep}")
    @unittest.skipIf(
        config.is_suspended(),
        f"Test was suspended for {config.suspended_for_n_days} days on {config.suspended_on}",
    )
    def run_test(self):
        start = time.process_time()
    
        ex = import_module_any_path(example_path / config.module_name)
    
        if not hasattr(ex, "main"):
            self.skipTest("Missing main function")
    
        result: Any = self.call_main(ex)
        if config.result_file_name and config.result_attribute:
            testdata = readcsv.np_read_csv(
                example_data_path / config.result_file_name
            )
            ra = config.result_attribute
            actual = ra(result) if callable(ra) else getattr(result, ra)
>           np.testing.assert_allclose(actual, testdata, atol=1e-05)
```

Alexsandruss (approved):

PR #3159, opened 2025-04-07
Use aligned loads and stores where possible in DAAL memory management

Vika-F:
## Description

* `service_memset_seq`, `service_calloc` and `service_scalable_calloc` functions were modified to use aligned memory loads and stores for aligned data.

* C++ standard version was increased from C++11 to C++17 in `daal_module` in bazel build. This was done to simplify the implementations of `service_calloc` and `service_scalable_calloc` by the use of C++17 features like `if constexpr`.

---

**Testing**

**Performance**

Total geomean speedup across the algorithms is **1.02** on 112-core SPR system. No significant performance changes.

The quality metrics had not changed because the computations logic is not affected.

### DIFF
--- a/cpp/daal/src/externals/service_memory.h
+++ b/cpp/daal/src/externals/service_memory.h
@@ -50,6 +50,11 @@ namespace internal
 template <typename T, CpuType cpu>
 void service_memset_seq(T * const ptr, const T value, const size_t num)
 {
+    /// TODO: 1. Add aligned branch for the case when num is greater than UINT_MAX
+    ///          and ptr is aligned.
+    ///       2. When possible, split the loop into two parts:
+    ///          - the first part uses unaligned stores until the first aligned address,
+    ///          - the second part uses aligned stores until the end of the block.
     if (num < UINT_MAX && !((DAAL_UINT64)ptr & DAAL_DEFAULT_ALIGNMENT_MASK))
     {
         /// Use aligned stores
--- a/cpp/daal/src/externals/service_memory.h
+++ b/cpp/daal/src/externals/service_memory.h
@@ -39,7 +39,7 @@ namespace services
 namespace internal
 {
 
-/// Initialize block of memory of length num with value.
+/// Initializes block of memory of length `num` with value.
 ///
 /// @tparam T   Data type of the elements in the block.
 /// @tparam cpu Variant of the CPU instruction set: SSE2, SSE4.2, AVX2, AVX512, ARM SVE, etc.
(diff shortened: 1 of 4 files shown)


Vika-F:
CI re-run: http://intel-ci.intel.com/f0153412-dab0-f1e3-a345-a4bf010d0e2d

cpp/daal/src/externals/service_memory.h
```diff
@@ -29,71 +29,197 @@
...
-    if (ptr == NULL)
+    if (num < UINT_MAX && !((DAAL_UINT64)ptr & DAAL_DEFAULT_ALIGNMENT_MASK))
     {
-        return NULL;
+        /// Use aligned stores
+        const unsigned int num32 = static_cast<unsigned int>(num);
+        PRAGMA_IVDEP
+        PRAGMA_VECTOR_ALWAYS
+        PRAGMA_VECTOR_ALIGNED
+        for (unsigned int i = 0; i < num32; i++)
```
avolkov-intel, line 60: Can we apply vectorization in case sizeof(T) > our alignment?

cpp/daal/src/externals/service_memory.h
```diff
@@ -29,71 +29,197 @@
...
+T * service_calloc(size_t size, size_t alignment = DAAL_MALLOC_DEFAULT_ALIGNMENT)
 {
-    daal::services::daal_free(ptr);
+    T * ptr = (T *)daal::services::daal_malloc(size * sizeof(T), alignment);
+    if (ptr == NULL)
+    {
+        return NULL;
+    }
+
+    if constexpr (std::is_trivially_default_constructible_v<T>)
```
avolkov-intel, line 129: Not sure, if we can use std here (for example if T is a template class that has cpu as template parameter). Likely, it's fine but maybe worth to check

cpp/daal/src/externals/service_memory.h
```diff
@@ -29,71 +29,197 @@
...
-T * service_calloc(size_t size, size_t alignment = 64)
+void service_memset_seq(T * const ptr, const T value, const size_t num)
 {
-    T * ptr = (T *)daal::services::daal_malloc(size * sizeof(T), alignment);
-    if (ptr == NULL)
+    if (num < UINT_MAX && !((DAAL_UINT64)ptr & DAAL_DEFAULT_ALIGNMENT_MASK))
     {
-        return NULL;
+        /// Use aligned stores
+        const unsigned int num32 = static_cast<unsigned int>(num);
```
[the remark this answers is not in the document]
Vika-F: > Why do we need this conversion?

Because loops with 32-bit counters are in general faster that loops with 64-bit counters.

> Is it is in general possible to vectorize loop if num >= UINT_MAX?

It is possible. And the loop will be vectorized in this case

> Maybe in this case we can split loop into few so that each of them is vectorizable

Yes, I thought about adding more branches or maybe even splitting the loop into two sub-loops: unaligned first part and aligned second part (when possible).

But the current version already solves the performance problems I have in https://github.com/uxlfoundation/oneDAL/pull/3126, and it improves the performance in most of the testcases in our benchmarks.

Further improvements are possible though.

cpp/daal/src/externals/service_memory.h
Vika-F: I think it would depend on the data type.
Also, those pragmas are just guidance to the compiler. Having them here does not force compiler to vectorize the loop, it just asks it to vectorize this loop.

cpp/daal/src/externals/service_memory.h
Vika-F: In this particular case this shouldn't be a problem as this construct is converted into `true` of `false` in compile time.
Actually, the whole `if constexpr` is evalueated in compile time resulting into a single call of `memset_default` or `memset` depending on the data type `T`.
That was the idea. Because it is not possible to call the default constructor for the type that doesn't have a default constructor.
And setting the memory to all zeros for non-POD types (as it was done previously) also looks strange to me.

avolkov-intel (approved):
Approve with minor comments: please, add TODO and create the follow up ticket to investigate observed performance degradations

napetrov:
@rakshithgb-fujitsu  as this would affect ARM as well

PR #3173, opened 2025-04-15
fix: static cast uniqueDepth to int to avoid ull underflow

ahuber21:
## Description

Remove some inconsistent assignments and comparisons between signed and unsigned (long) ints by explicitly downcasting to int and adding a boundary check. All if this just addresses theoretical over- and underflow scenarios because in reality the SHAP algos never call these internal functions with such exotic values.
Correctness is confirmed by the SHAP value checks in sklearnex.

---

**Testing**

**Performance**


ahuber21:
We don't have specific unit tests for SHAP value correctness in oneDAL (only in sklearnex) and it would be very difficult to add it unfortunately. I checked the correctness manually though.

cpp/daal/src/algorithms/dtrees/gbt/treeshap.cpp
```diff
@@ -34,44 +34,60 @@ namespace v0
...
+
+    uniquePath[depth].featureIndex  = featureIndex;
+    uniquePath[depth].zeroFraction  = zeroFraction;
+    uniquePath[depth].oneFraction   = oneFraction;
+    uniquePath[depth].partialWeight = (depth == 0 ? 1.0f : 0.0f);
 
-    const float constant = 1.0f / static_cast<float>(uniqueDepth + 1);
-    for (int i = uniqueDepth - 1; i >= 0; --i)
+    const float constant = 1.0f / static_cast<float>(depth + 1);
+    for (int i = depth - 1; i >= 0; --i)
```
Vika-F, line 51: Coverity report says there are problems when `uniqueDepth == 0` due to that `int i = uniqueDepth - 1` initialization.

It would be less invasive just to add a check for (uniqueDepth > 0) here I think.

avolkov-intel: I think we have different hits for both, so I like this solution with static_cast more

Vika-F (approved):

icfaust (approved):
Thank you for this + the comments in the code.

PR #3176, opened 2025-04-17
[CI] swap uxl self-hosted runners to github hosted

icfaust:
## Description

This PR moves the UXL foundation self-hosted runners to github hosted runners (ideally on a temporary basis).

The following changes are made:
 - ISA is set from AVX512 to AVX2 due to runner limitations
 - A matrix value ```ISA``` is defined to control the ISA of the runners, which modifies the displayed job names. Unfortunately it must be lower case because of restrictions set by ```REQCPU```.
 - The GPU testing step is disabled by checking the runner name rather than being removed (in the hope that the GPU step removal is temporary).

This does not require performance benchmarking

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,6 @@ concurrency:
 jobs:
   LinuxMakeDPCPP:
     if: github.repository == 'uxlfoundation/oneDAL'
-
     timeout-minutes: 120
     strategy:
       fail-fast: false
@@ -45,7 +44,7 @@ jobs:
           - ISA: avx2
             runner: ubuntu-24.04
 
-    runs-on: matrix.runner
+    runs-on: ${{ matrix.runner }}
     name: LinuxMakeDPCPP(${{ matrix.ISA }})
     steps:
       - name: Checkout oneDAL
@@ -55,7 +54,7 @@ jobs:
           echo ${{ matrix.runner }}
           .ci/env/apt.sh dpcpp
       - name: temp
-        if: matrix.runner != 'ubuntu-24.04'
+        if: ${{ matrix.runner }} != 'ubuntu-24.04'
         run: echo "yes"
       - name: Install MKL
         run: .ci/env/apt.sh mkl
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -157,4 +160,6 @@ jobs:
           echo "please rerun the main 'CI' workflow via GitHub's workflow dispatch to regenerate the cache."
           echo "Keep the branch up to date with oneDAL 'main' as it may otherwise not contain the latest changes "
           echo "in the ABI that can lead to erroneous failures in the ABI check."
-          .ci/scripts/abi_check.sh __release_lnx_main/daal/latest/lib/intel64/ __release_lnx/daal/latest/lib/intel64/
+          .ci/scripts/abi_check.sh __release_lnx_main/daal/latest/lib/intel64/ __release_lnx/daal/latest/lib/intel64/ || $ABI_BREAK
```
ahuber21, line 163: So the logic changes from "Don't run the check" to "Accept that it fails". Why is that necessary? Because we need to update the reference from the AVX512 build to AVX2? But wouldn't that mean that the reference gets now updated every time, even if the check fails and it's not okay?

ahuber21 (changes requested):

.github/workflows/ci.yml
icfaust: Good catch, its more that when the ABI will be known to break, that the resultant changes can be observed in the ABI check without it failing in CI, rather than not running at all. This may be useful to reviewers.

In this case its a bit tangentially related, but I wanted this such that I can observe it running through during this PR while not having it fail, otherwise to get a green CI we have to run it twice and hope that the reviewers look at the old run.

ahuber21: We just revert this PR and it will be back :)

icfaust: In order to speed this PR to being merged, I have reverted this change.

icfaust:
Note failure of the ABI checker should occur. This is because it is comparing an AVX512 build to an AVX2 build, and should light up like a christmas tree.

ethanglaser (approved):

icfaust:
The failure in acquiring the cache comes from the fact that a different OS is now being used, this is using ubuntu-24.04 whereas the uxl runner uses ubuntu-22.04. https://github.com/actions/cache?tab=readme-ov-file#cache-version  Therefore, the cache problem is likely to occur in the future with os upgrades.  For the sake of this PR, I will label the PR as an API/ABI breaking change now.

ahuber21 (approved):

PR #3180, opened 2025-04-22
fix: use throwIfPossible to fix -fno-exception build

ahuber21:
Using a bare throw breaks the `-fno-exception` build. This workaround uses `service::throwIfPossible`.


icfaust:
Where should I be looking to check for the ```-fno-exception``` build?

icfaust (approved):

ahuber21:
Will revert the throw and address the coverity hit differently

PR #3181, opened 2025-04-22
Revert "fix: static cast uniqueDepth to int to avoid ull underflow (#…

ahuber21:
…3173)"

This reverts commit c17ea8b550e407d39189cd44d5c84e33f4d75300.

Reverts uxlfoundation/oneDAL#3173

The coverity hits fixed by this PR will be addressed differently. This check is virtually impossible to trigger and therefore it shouldn't clutter or slow down the code.


Vika-F (approved):
The errors in internal CI do not look related to the revert.

PR #3186, opened 2025-04-28
fix(docs): Install dependencies fails over debian-installed packages

ahuber21:
## Description

Create a `venv` for the python packages in the install step. Also add `set -euo pipefail` to fail the correct step if a problem occurs.

## What it fixes

Building docs sometimes fails at the `Install requirements` step, preventing the installation of the requirements. The subsequent `Build documentation` will fail with `sphinx-build: not found`.

image

Example: https://dev.azure.com/daal/DAAL/_build/results?buildId=45831&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5caf77c8-9b10-50ef-b5c7-ca89c63e1c86

### DIFF
--- a/.ci/pipeline/docs.yml
+++ b/.ci/pipeline/docs.yml
@@ -33,12 +33,12 @@ steps:
 - script: |
     set -euo pipefail
     cd docs
+    sudo apt-get update
     sudo apt-get install python3-setuptools
     python3 -m venv venv
     source venv/bin/activate
     pip install --upgrade pip
     pip install -r requirements.txt
-    sudo apt-get update
     sudo apt-get install doxygen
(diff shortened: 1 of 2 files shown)


ahuber21:
The last commit before merging won't have a `docbuild` step because of the directory filters. But here's the latest one that shows that now everything is passing:
https://dev.azure.com/daal/DAAL/_build/results?buildId=45983&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5caf77c8-9b10-50ef-b5c7-ca89c63e1c86

Alexsandruss (approved):

PR #3196, opened 2025-05-08
Fix issues with Logistic Regression and Newton-CG

avolkov-intel:
## Description

This PR address multiple issues in computation of logistic loss and newton-cg optimization process:

- Issue with logloss computation in logloss_function class is fixed, additional tests were added
- Issue with incorrect sign of precomputed gradient provided to backtracking algorithm is fixed
- Additional stopping criteria is added to newton-cg algorithm
- Tolerance for solution vector is raised a little bit
- Additional test is added to compare the value at the optimal point and in the orginal point

---

**Testing**

**Performance**

### DIFF
--- a/cpp/oneapi/dal/backend/primitives/optimizers/line_search_dpc.cpp
+++ b/cpp/oneapi/dal/backend/primitives/optimizers/line_search_dpc.cpp
@@ -42,10 +42,12 @@ Float backtracking(sycl::queue queue,
     Float df0 = 0;
     dot_product(queue, grad_f0, direction, result.get_mutable_data(), &df0, deps + precompute)
         .wait_and_throw();
+    Float dir_inf_norm = 0;
+    max_abs(queue, direction, result.get_mutable_data(), &dir_inf_norm, deps).wait_and_throw();
     Float cur_val = 0;
     constexpr Float eps = std::numeric_limits<Float>::epsilon();
     bool is_first_iter = true;
-    while ((is_first_iter || cur_val - f0 > c1 * alpha * df0) && alpha > eps) {
+    while ((is_first_iter || cur_val - f0 > c1 * alpha * df0) && alpha * dir_inf_norm > eps) {
         // TODO check that conditions are the same across diferent devices
         if (!is_first_iter) {
             alpha /= 2;
@@ -62,7 +64,8 @@ Float backtracking(sycl::queue queue,
         wait_or_pass(func_event_vec).wait_and_throw();
         cur_val = f.get_value();
     }
-    if (alpha <= eps) {
+    if (alpha * dir_inf_norm <= eps) {
+        // if we add alpha * dir the vector of weights will not change
         alpha = 0;
     }
     return alpha;
--- a/cpp/oneapi/dal/backend/primitives/optimizers/line_search_dpc.cpp
+++ b/cpp/oneapi/dal/backend/primitives/optimizers/line_search_dpc.cpp
@@ -18,6 +18,8 @@
 #include "oneapi/dal/backend/primitives/optimizers/common.hpp"
 #include "oneapi/dal/backend/primitives/element_wise.hpp"
 #include "oneapi/dal/detail/profiler.hpp"
+#include <iostream>
+#include <iomanip>
 
 namespace oneapi::dal::backend::primitives {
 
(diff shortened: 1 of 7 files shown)


cpp/oneapi/dal/backend/primitives/optimizers/line_search_dpc.cpp
```diff
@@ -62,6 +62,9 @@ Float backtracking(sycl::queue queue,
         wait_or_pass(func_event_vec).wait_and_throw();
         cur_val = f.get_value();
     }
+    if (alpha <= eps) {
```
david-cortes-intel, line 65: I think this threshold might still be too liberal. Basically you'd have:
* x_val -> likely beyond +1/-1.
* alpha -> machine epsilon.
* dir_val -> likely smaller than x_val.

Thus, it's quite likely that the update will amount to zero even at values slightly above machine epsilon.

It's also a **lot** more trials being conducted than in other software - for example, for fp64, it will take 52 iterations to reach the threshold, which is perhaps too much. For example, SciPy's wolfe line search uses a maximum of 10 iterations:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.line_search.html
Other bundled line searches there also use less - for example: https://github.com/scipy/scipy/blob/3fb9ab145341b570af87cf3a4a3c5ea812ec6cd0/scipy/optimize/_lbfgsb_py.py#L97
Granted, those are both interpolating line searches, while this is a backtracking procedure so it might require more trials, but it still feels a bit too much.

Maybe you could try setting to sqrt(eps) instead. Or a hard-coded threshold like 1e-6 for both fp32 and fp64.

avolkov-intel:
image

Benchmarking results

cpp/oneapi/dal/backend/primitives/optimizers/line_search_dpc.cpp
avolkov-intel: Among the proposed ideas, I like the sqrt(eps) the most, however, it seems that if we introduce this update the validation should be done more thoroughly, on a wider set of datasets. Meanwhile, in this PR the convergence should not be affected we stop algorithm at the moment it wouldn't introduce any updates anyway. Moreover, this PR contains few fixes for critical bugs, that's why I don't want to postpone the merge till we wait for investigation.

All in all, I agree with the idea, but I think we cannot just pull a random constant out of nowhere without validation on wider set of datasets and theoretical basis. That's why let's introduce this update in separate PR.

david-cortes-intel: @avolkov-intel It's actually not a random constant - it correspond exactly to 20 line searches, which is the default in some SciPy optimizers.

avolkov-intel: Yeah, but those are different methods

david-cortes-intel: If you want another reference, IPOPT's backtracking LS uses a rather different criterion "tiny_step_tol":
https://coin-or.github.io/Ipopt/OPTIONS.html#OPT_Line_Search

which in a regular case, would also amount to less than what this PR is setting.

Granted, it's a different case (not pure Newton) and they have a lot more safeguards, but might still serve as a reference.

avolkov-intel: eps for float in my PR is 
1.1920928955078125e-07
eps for double in my PR is
2.220446049250313080847263336181640625e-16

From documentation default tiny_step_tol is 2.22045e-15 which looks similar to eps for double, so it seems they also use values around numerical limits

david-cortes-intel: That threshold on IPOPT is not on the step size, but on the values of the update.

david-cortes-intel (approved):

david-cortes-intel:
CI errors are in a CPU test plus a timeout, don't look related to this PR.

cpp/oneapi/dal/backend/primitives/optimizers/test/fixture.hpp
```diff
@@ -102,12 +106,6 @@ class quadratic_function : public base_function<Float> {
     linear_matrix_operator<Float> hessp_;
 };
 
-template <typename Float>
```
Alexandr-Solovev, line 105: what the reason of replacement this function with a macro?

avolkov-intel: Yes, when you observe failure in CI it's easier to identify where it happened if macro is used. If function is used it shows you that error happened at this function but no information about from where this function was called

PR #3210, opened 2025-05-13
enh: replace unix2dos with sed

Alexandr-Solovev:
### Summary

Replaced usage of `unix2dos` with `sed` to remove an external dependency.

### Details

The original script relied on `dos2unix`, which is not always available in all environments by default. This change replaces it with a `sed` command that achieves the same result (removing carriage return characters), thus improving portability and reducing setup requirements.

### Changes

- Replaced `unix2dos`
- Removed the dependency on the `unix2dos` utility

### Motivation

Minimize external dependencies and ensure better cross-platform compatibility.

---

**Testing**

**Performance**


makefile
```diff
@@ -1040,7 +1040,7 @@ $4: $3/$2
 $3/$2: $(DIR)/$1 | $3/. ; $(value cpy)
 	$(if $(filter %.sh %.bat dal,$2),sed -i -e 's/__DAL_MAJOR_BINARY__/$(MAJORBINARY)/' $3/$2)
 	$(if $(filter %.sh %.bat dal,$2),sed -i -e 's/__DAL_MINOR_BINARY__/$(MINORBINARY)/' $3/$2)
-	$(if $(OS_is_win),unix2dos $3/$2)
+	$(if $(OS_is_win),sed -i -n -z -e 's/\r*\n/\r\n/g;p' $3/$2)
```
yuejiaointel, line 1043: Do we need to worry about it has to work on other systems like macos(some of the sed arg is not supported I think) ?

david-cortes-intel: OneDAL currently does not support macOS so no need to worry about it.

Sed is part of most coreutils out there (like GNUs) and is required by POSIX, so even if it doesn't come out of the box on macOS (although I'm pretty sure it does), it can be installed easily afterwards.

david-cortes-intel:
@Alexandr-Solovev Could you please remove the dos2unix requirement and instructions from the INSTALL.md file?

makefile
david-cortes-intel: Sorry, I think I misread it: yes, some arguments might indeed be GNU-specific extensions. This needs to be documented.

@Alexandr-Solovev Could you please add a note that this requires the GNU variant of sed if that's the case?

Alexandr-Solovev: Its implemented only for if $(OS_is_win), so, I think its not necessary to think about other system support

INSTALL.md
```diff
@@ -26,9 +26,9 @@ Required Software:
 * oneDPL library
 * Microsoft Visual Studio\* (Windows\* only)
 * [MSYS2](http://msys2.github.io) (Windows\* only)
-* `make` and `dos2unix` tools; install these packages using MSYS2 on Windows\* as follows:
+* `make`tool; install these packages using MSYS2 on Windows\* as follows:
```
david-cortes-intel, line 29:
```suggestion
* `make`; which can be installed using MSYS2 on Windows\* as follows:
```

david-cortes-intel (approved):

Vika-F (approved):
The scripts contain Windows-style line endings after the change.

PR #3212, opened 2025-05-15
fix: use github action instead of gh-pages

yuejiaointel:
## Description

Similarly to intelex, now we are using github action instead of gh-pages. It is simpler setup as onedal does not need to track all versions

---

**Testing**

### DIFF
--- a/.github/workflows/docs-release.yml
+++ b/.github/workflows/docs-release.yml
@@ -22,6 +22,8 @@ on:
 
 permissions:
   contents: write
+  id-token: write
+  contents: read
   
 jobs:
   build-docs:
@@ -73,25 +75,19 @@ jobs:
         cd docs
         make html
 
-    - name: Deploy Documentation to gh-pages
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Setup Pages
+      uses: actions/configure-pages@v4
+
+    - name: Prepare Documentation for Deployment
       run: |
-        # Set git auth for push changes 
-        git config --global user.name "GitHub Actions"
-        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        mkdir -p _site        
+        cp -r docs/build/* _site/
+        
+    - name: Upload artifact for GitHub Pages
+      uses: actions/upload-pages-artifact@v3
+      with:
+        path: '_site'
 
-        # Copy the build folder to a temporary location
-        cp -R docs/build/. /tmp/build
-
-        # Checkout the gh-pages branch
-        git checkout gh-pages
-
-        # Replace existing files with new build files
-        git rm -rf .
-        cp -R /tmp/build/. .
-
-        # Push to gh-pages
-        git add -A
-        git commit -m "Automatic doc update for version $DOC_VERSION"
-        git push origin gh-pages
+    - name: Deploy to GitHub Pages
+      id: deployment
+      uses: actions/deploy-pages@v4
--- a/.github/workflows/docs-release.yml
+++ b/.github/workflows/docs-release.yml
@@ -22,8 +22,8 @@ on:
 
 permissions:
   contents: write
-  id-token: write
-  contents: read
+  pages: write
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-release.yml
```diff
@@ -22,6 +22,8 @@ on:
 
 permissions:
   contents: write
+  pages: write
+  id-token: write
```
homksei, line 26: Could you please move all these permissions to the job level, and leave `read-all` here (see [Token-Permissions](https://github.com/ossf/scorecard/blob/cd152cb6742c5b8f2f3d2b5193b41d9c50905198/docs/checks.md#token-permissions) criterion)? 

It would look something like this:
```yaml
permissions: read-all

jobs:
  build-docs:
    runs-on: ubuntu-24.04
    permissions:
      contents: write
      pages: write
      id-token: write

File: .github/workflows/docs-release.yml [hunk 2]
```diff
@@ -73,25 +75,19 @@ jobs:
...
-        git config --global user.name "GitHub Actions"
-        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
-
-        # Copy the build folder to a temporary location
-        cp -R docs/build/. /tmp/build
-
-        # Checkout the gh-pages branch
-        git checkout gh-pages
+    - name: Setup Pages
+      uses: actions/configure-pages@v4
```
homksei, line 79: Why was the choice made for the version `v4` instead of `v5` ? 

In any case, please pin action by its hash, it will be like this for `v4`: 
```suggestion
      uses: actions/configure-pages@1f0c5cde4bc74cd7e1254d0cb4de8d49e9068c7d # v4.0.0
```
and same thing for actions above: 

`actions/checkout@v4` ->  `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2`
`actions/setup-python@v5` -> `actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0`

.github/workflows/docs-release.yml
```diff
@@ -73,25 +75,19 @@ jobs:
...
-        # Replace existing files with new build files
-        git rm -rf .
-        cp -R /tmp/build/. .
+    - name: Prepare Documentation for Deployment
+      run: |
+        mkdir -p _site        
+        cp -r docs/build/* _site/
+        
+    - name: Upload artifact for GitHub Pages
+      uses: actions/upload-pages-artifact@v3
```
homksei, line 87:
```suggestion
      uses: actions/upload-pages-artifact@56afc609e74202658d3ffba0e8f6dda462b719fa # v3.0.1
```

.github/workflows/docs-release.yml
```diff
@@ -73,25 +75,19 @@ jobs:
...
+        path: '_site'
 
-        # Push to gh-pages
-        git add -A
-        git commit -m "Automatic doc update for version $DOC_VERSION"
-        git push origin gh-pages
\ No newline at end of file
+    - name: Deploy to GitHub Pages
+      id: deployment
+      uses: actions/deploy-pages@v4
```
homksei, line 93:
```suggestion
      uses: actions/deploy-pages@d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e # v4.0.5
```

.github/workflows/docs-release.yml
yuejiaointel: Thx, moved!

.github/workflows/docs-release.yml
yuejiaointel: Thx, replaced, using newest commit id

.github/workflows/docs-release.yml
yuejiaointel: Thx, replaced, using newest commit id

.github/workflows/docs-release.yml
yuejiaointel: Thx, replaced, using newest commit id

homksei (approved):

PR #3227, opened 2025-05-26
MAINT: Reuse common routines in covariance

david-cortes-intel:
## Description

This PR simplifies some parts of the covariance implementation on CPU to reuse common routines from either MKL or from the rest of daal. This should slightly simplify future work on compiler pragmas.

---

**Testing**


icfaust:
@david-cortes-intel can you explain why this won't need performance benchmarks? It looks like the vectorization hints are switched to an omp pragma.

icfaust:
if you want to enable this for only gcc i recommend leaving comments in the code as such. wouldn't these changes then pertain to any build not using MKL regardless of the compiler?

david-cortes-intel:
> if you want to enable this for only gcc i recommend leaving comments in the code as such. wouldn't these changes then pertain to any build not using MKL regardless of the compiler?

I'm not understanding that part. The change is from using compiler hints towards using functions from the 'service' headers. When building with MKL, the functions get taken from MKL regardless of compiler, while when building without MKL, the functions are generated by the compiler, using `omp simd` pragmas, which is an open standard supported by all the major C++ compilers (gcc, mingw, clang, icx, IBM's, oracle's, msvc with experimental features), unlike the previous pragmas which were specific to icc and some supported by icx.

david-cortes-intel:
@icfaust I've now reverted the changes on the sections that had the pragma for aligned data, although I'm pretty sure MKL would be able to make the same optimization.

cpp/daal/src/externals/service_math_ref.h
```diff
@@ -214,6 +232,12 @@ struct RefMath<float, cpu>
         for (SizeType i = 0; i < n; ++i) out[i] = sqrt(in[i]);
     }
 
+    static void vInvSqrtI(SizeType n, const float * a, const SizeType inca, float * b, const SizeType incb)
```
KateBlueSky, line 235: For this is inca and incb usually 1? If so there might be a better way writing this so the compiler optimizes the code.

Vika-F:
I agree with @icfaust and think that these changes require the benchmarks run.
Because replacing of the compiler generated loop with a call (even to some highly-optimized library) can have unpredictable performance impact.

We can have some assumptions, but they really need to be proved by actual performance data.

david-cortes-intel:
Times for an input of size 1,500x1,000, in an IceLake, mesured for 50 repetitions:

* Main branch:
    * Median: 0.1283869743347168
    * Standard deviation: 0.011898340793814566
* This branch:
    * Median: 0.12701189517974854
    * Standard deviation: 0.017115250317560733

So looks like it has a positive or neutral effect, at the expense of slightly higher variance, but the measurements are still quite noisy.

Vika-F (approved):

PR #3230, opened 2025-05-28
ENH: Non-batched route for high-dimensional covariance

david-cortes-intel:
## Description

This PR adds a non-batched computation route for dense covariance on CPU, alongside with parameters controlling the switch.

Note that there are two methods offered here: defaultDense and sumDense. MKL already implemented the equivalent of defaultDense, so in the non-batched case for that method it reuses the same current routine from it.

Since this adds a new combination of parameters to test in a combinatorial grid, I reduced the amount and size of the random datasets being generated during tests in order not to make them too slow.

~PCA will follow in a separate PR.~ It also adds the parameters for PCA in the same PR as the examples would otherwise end up with uninitialized values.

---

**Testing**

**Performance**

Performance numbers have been shared internally.


cpp/daal/src/externals/service_stat_ref.h
```diff
@@ -96,6 +98,24 @@ struct RefStatistics<double, cpu>
     typedef __int64 MethodType;
     typedef int ErrorType;
 
+    static int xmeansOnePass(const double * data, __int64 nFeatures, __int64 nVectors, double * means)
+    {
+        daal::services::internal::TArray<double, cpu> tempRowMinusMeans(nFeatures);
+        daal::services::internal::service_memset<double, cpu>(means, 0.0, nFeatures);
```
Vika-F:
```suggestion
        daal::services::internal::TArrayCalloc<double, cpu> tempRowMinusMeans(nFeatures);
        if (!tempRowMinusMeans.get()) return services::ErrorMemoryAllocationFailed;
```

cpp/oneapi/dal/algo/pca/train_types.hpp
```diff
@@ -58,9 +58,30 @@ class train_parameters : public dal::detail::system_parameters {
         return *this;
     }
 
+    std::int64_t get_cpu_max_cols_batched() const;
```
Vika-F, line 64: Please add the descriptions like it is done here:
https://github.com/uxlfoundation/oneDAL/blob/main/cpp/oneapi/dal/algo/pca/train_types.hpp#L45

I forgot to add those for covariance :(

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -335,119 +335,224 @@ inline size_t getBlockSize<DAAL_CPU_TYPE>(size_t nrows)
...
+    }
+    bool assumeCentered = parameter->assumeCentered;
+    DAAL_PROFILER_TASK_WITH_ARGS(Covariance::updateDenseCrossProductAndSums, numRowsInBlock, grainSize, maxColsBatched, smallRowsThreshold,
+                                 smallRowsMaxColsBatched, assumeCentered, isNormalized);
+
+    services::Status status;
+    const bool prefer_non_batched = (nFeatures >= maxColsBatched) || (nVectors <= smallRowsThreshold && nFeatures >= smallRowsMaxColsBatched);
+    if (prefer_non_batched)
+    {
+        if (method == sumDense || isNormalized || assumeCentered)
```
Vika-F, line 542: Why is non-batched method is enabled only on centered or normalized data, or when the sums are provided by user?
I thought the intention was to enable it for `defaultDense` method and non-normalized data (the default usage scenario).

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -335,119 +335,224 @@ inline size_t getBlockSize<DAAL_CPU_TYPE>(size_t nrows)
...
+        return services::Status(services::ErrorMemoryAllocationFailed);
+    }
 
-        const algorithmFPType * resultCrossProduct = result.crossProduct();
-        if (result.errorCode != CovarianceReducer<algorithmFPType, cpu>::ok || !resultCrossProduct)
+    if (computeSumsAndCenter)
+    {
+        algorithmFPType * means = sums;
+        StatisticsInst<algorithmFPType, cpu>::xmeansOnePass(dataPointer, nFeatures, nVectors, means);
+        for (size_t vector = 0; vector < nVectors; vector++)
```
Vika-F: Does it make sense to do this in parallel?

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -335,119 +335,224 @@ inline size_t getBlockSize<DAAL_CPU_TYPE>(size_t nrows)
...
+    {
+        return services::Status(services::ErrorMemoryAllocationFailed);
+    }
 
-        const algorithmFPType * resultCrossProduct = result.crossProduct();
-        if (result.errorCode != CovarianceReducer<algorithmFPType, cpu>::ok || !resultCrossProduct)
+    if (computeSumsAndCenter)
+    {
+        algorithmFPType * means = sums;
+        StatisticsInst<algorithmFPType, cpu>::xmeansOnePass(dataPointer, nFeatures, nVectors, means);
```
Vika-F, line 436: In `sumDense` method the `sums` array is provided by user. I do not think it is a good idea to write into user-provided data.

cpp/daal/src/algorithms/covariance/covariance_impl.i
david-cortes-intel: Because MKL already has the 'fast' version that doesn't create a centered copy of the data.

cpp/daal/src/algorithms/covariance/covariance_impl.i
david-cortes-intel: I don't think there would be any gain in parallelizing a subtraction, since it will likely go as fast as the RAM can be read with a single thread.

cpp/daal/src/algorithms/covariance/covariance_impl.i
david-cortes-intel: That's how it's currently being done in the `main` branch. Should that condition be modified?

cpp/oneapi/dal/algo/pca/train_types.hpp
david-cortes-intel: Added.

cpp/daal/src/algorithms/covariance/covariance_impl.i
Vika-F: Found that defaultDense is handled by MKL for non-batched, sorry.
But then we have different implementations in defaultDense for batched and not-batched case.
I think the batched case should also be modified to compute X-means, but in batches.

david-cortes-intel: I don't follow. Both MKL and the current code in the main branch would follow the 'fast' route that doesn't create a centered copy of the data, and both have the same properties in terms of numerical accuracy.

It sounds like you want to implement the more accurate method with centered data for the non-batched case, but that wouldn't be the same as 'defaultDense' or 'fast' from MKL - currently, that would be method 'sumDense'.

Although I would prefer to have the method with centered data as default too, since it's more accurate.

cpp/daal/src/algorithms/covariance/covariance_impl.i
Vika-F: Yes, it looks like it was my mistake =\\
We lack DAAL API testing now which leads to growing the number of errors.
oneDAL tests only test the defaultDense method :(

The idea is that sumDense method shouldn't compute or modify the sums. It just has to use the provided sums in the computations.
I think it also should be something like this here:
`CovarianceReducer<algorithmFPType, cpu> result(dataTable, numRowsInBlock, numBlocks, isNormalized || method == sumDense);`
https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/src/algorithms/covariance/covariance_impl.i#L370

cpp/daal/src/algorithms/covariance/covariance_impl.i
Vika-F: On our typical dimensions Covariance is a memory-bound algorithm. So, reading in parallel from the RAM can be really beneficial.
Anyway, it would be good to get the performance data obtained with scikit-learn_bench to be able to do data-driven decisions.

david-cortes-intel: Changed to a parallel for. Although I still don't think it'd bring any speed up unless the data is in different numa nodes or something like that.

cpp/daal/src/algorithms/covariance/covariance_impl.i
Vika-F: Ah, I see.
I was confused by vSub usage; but now I see that it's only for sumDense.

Yes, I think the final goal would be to replace our current defaultDense with the version that centers the data first.
But it now I see that it is out of the scope of this PR.

And I am still concerned with all those combinations of methods, isNormalized and assumeCentered.

david-cortes-intel: @Vika-F I modified the new route introduced here to use the custom sums when method is 'sumDense' and to not modify them according to your comment. The current code calculating the means is still there, but it's not reachable now (something to address later).

I guess there should be a new method 'sumFast' that would use the current 'defaultDense' methodology, and the 'defaultDense' method should use the more accurate centering approach (which is not reachable in the code).

Perhaps you could make a follow-up PR fixing the code in the batched route to not overwrite the 'sums' array, as it's not clear to me how all cases should be handled there.

I guess we should have:
* defaultDense -> method applying centering and then cross product.
* sumFast -> method doing cross product with SR1 correction.

But currently we have:
* defaultDense -> always does cross product + SR1, but the sums are calculated on the fly.
* sumDense -> this could either use the data centering route or the SR1 correction. After this PR, it might use one or the other depending on the amount of data.

Maybe 'sumDense' should not be a method at all, but an input parameter 'use provided sums or not', with the method controlling only whether it does centering or SR1.

david-cortes-intel:
Same CI failures observed here appear to affect other PRs too, so probably not related to the changes here.

PR #3246, opened 2025-06-03
Fix vectorization pragmas for icx compiler

Vika-F:
## Description

---

**Testing**

**Performance**

### DIFF
--- a/cpp/daal/src/algorithms/cordistance/cordistance_batch_impl.i
+++ b/cpp/daal/src/algorithms/cordistance/cordistance_batch_impl.i
@@ -29,6 +29,7 @@
 #include "src/data_management/service_numeric_table.h"
 
 static const int blockSizeDefault = 128;
+#include "src/algorithms/cordistance/cordistance_impl.i"
 #include "src/algorithms/cordistance/cordistance_full_impl.i"
 #include "src/algorithms/cordistance/cordistance_up_impl.i"
 #include "src/algorithms/cordistance/cordistance_lp_impl.i"
--- a/cpp/daal/src/algorithms/cordistance/cordistance_full_impl.i
+++ b/cpp/daal/src/algorithms/cordistance/cordistance_full_impl.i
@@ -66,54 +66,10 @@ services::Status corDistanceFull(const NumericTable * xTable, NumericTable * rTa
         algorithmFPType sum[blockSizeDefault], buf[blockSizeDefault * blockSizeDefault];
 
         /* compute sums for elements in each row of the block */
-        for (size_t i = 0; i < blockSize1; i++)
-        {
-            algorithmFPType s = (algorithmFPType)0.0;
-            PRAGMA_VECTOR_ALWAYS
-            for (size_t j = 0; j < p; j++)
-            {
-                s += x[i * p + j];
-            }
-            sum[i] = s;
-        }
+        sumByRows<algorithmFPType, cpu>(blockSize1, p, x, sum);
 
-        /* calculate sum^t * sum */
-        const algorithmFPType one(1.0);
-        const algorithmFPType zero(0.0);
-        algorithmFPType alpha = one, beta = zero;
-        char transa = 'N', transb = 'T';
-        DAAL_INT m = blockSize1, k = 1, nn = blockSize1;
-        DAAL_INT lda = m, ldb = nn, ldc = m;
-
-        BlasInst<algorithmFPType, cpu>::xxgemm(&transa, &transb, &m, &nn, &k, &alpha, sum, &lda, sum, &ldb, &beta, buf, &ldc);
-
-        /* calculate x * x^t - 1/p * sum^t * sum */
-        alpha  = one;
-        beta   = -one / (algorithmFPType)p;
-        transa = 'T';
-        transb = 'N';
-        m = blockSize1, k = p, nn = blockSize1;
-        lda = k, ldb = k, ldc = m;
-
-        BlasInst<algorithmFPType, cpu>::xxgemm(&transa, &transb, &m, &nn, &k, &alpha, x, &lda, x, &ldb, &beta, buf, &ldc);
-
-        PRAGMA_VECTOR_ALWAYS
-        for (size_t i = 0; i < blockSize1; i++)
-        {
-            if (buf[i * blockSize1 + i] > (algorithmFPType)0.0)
-            {
-                buf[i * blockSize1 + i] = (algorithmFPType)1.0 / daal::internal::MathInst<algorithmFPType, cpu>::sSqrt(buf[i * blockSize1 + i]);
-            }
-        }
-
-        for (size_t i = 0; i < blockSize1; i++)
-        {
-            PRAGMA_VECTOR_ALWAYS
-            for (size_t j = i + 1; j < blockSize1; j++)
-            {
-                buf[i * blockSize1 + j] = 1.0 - buf[i * blockSize1 + j] * buf[i * blockSize1 + i] * buf[j * blockSize1 + j];
-            }
-        }
+        /* compute upper triangle of diagonal block of the correlation distance elements */
+        computeDiagonalBlock<algorithmFPType, cpu, true>(blockSize1, p, x, sum, buf);
 
         for (size_t i = 0; i < blockSize1; i++)
         {
@@ -144,16 +100,7 @@ services::Status corDistanceFull(const NumericTable * xTable, NumericTable * rTa
         algorithmFPType sum1[blockSizeDefault];
 
         /* compute sums for elements in each row of the block x1 */
-        for (size_t i = 0; i < blockSize1; i++)
-        {
-            algorithmFPType s = (algorithmFPType)0.0;
-            PRAGMA_VECTOR_ALWAYS
-            for (size_t j = 0; j < p; j++)
-            {
-                s += x1[i * p + j];
-            }
-            sum1[i] = s;
-        }
+        sumByRows<algorithmFPType, cpu>(blockSize1, p, x1, sum1);
 
         daal::threader_for(nBlocks - k1 - 1, nBlocks - k1 - 1, [=, &safeStat, &sum1](size_t k3) {
             DAAL_INT blockSize2 = blockSizeDefault;
@@ -193,16 +140,7 @@ services::Status corDistanceFull(const NumericTable * xTable, NumericTable * rTa
             algorithmFPType sum2[blockSizeDefault];
 
             /* compute sums for elements in each row of the block x2 */
-            for (size_t i = 0; i < blockSize2; i++)
-            {
-                algorithmFPType s = (algorithmFPType)0.0;
-                PRAGMA_VECTOR_ALWAYS
-                for (size_t j = 0; j < pl; j++)
-                {
-                    s += x2[i * pl + j];
-                }
-                sum2[i] = s;
-            }
+            sumByRows<algorithmFPType, cpu>(blockSize2, p, x2, sum2);
 
             /* calculate sum1^t * sum2 */
             const algorithmFPType one(1.0);
--- a/cpp/daal/src/algorithms/cordistance/cordistance_impl.i
+++ b/cpp/daal/src/algorithms/cordistance/cordistance_impl.i
@@ -1,0 +1,120 @@
+/* file: cordistance_impl.i */
+/*******************************************************************************
+* Copyright contributors to the oneDAL project
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "src/services/service_defines.h"
+using namespace daal::internal;
+
+namespace daal
+{
+namespace algorithms
+{
+namespace correlation_distance
+{
+namespace internal
+{
+/// Compute sum of each row of the input matrix x and store the result in sum
+///
+/// \tparam algorithmFPType  Data type to use in intermediate computations for the correlation distance algorithm, double or float
+/// \tparam cpu              Type of CPU, e.g. daal::CpuType::sse2, daal::CpuType::avx2, etc.
+///
+/// \param[in] nRows            Number of rows in the input matrix x
+/// \param[in] nColumns         Number of columns in the input matrix x
+/// \param[in] x                Pointer to the input matrix x of size nRows * nColumns
+/// \param[out] sum              Pointer to the output array of size nRows, where the sum of each row of x will be stored
+template <typename algorithmFPType, CpuType cpu>
+void sumByRows(const size_t nRows, const size_t nColumns, const algorithmFPType * x, algorithmFPType * sum)
+{
+    for (size_t i = 0; i < nRows; i++)
+    {
+        algorithmFPType s = (algorithmFPType)0.0;
+        PRAGMA_OMP_SIMD(reduction(+:s))
+        for (size_t j = 0; j < nColumns; j++)
+        {
+            s += x[i * nColumns + j];
+        }
+        sum[i] = s;
+    }
+}
+
+/// Compute the diagonal block of the correlation distance matrix
+///
+/// \tparam algorithmFPType  Data type to use in intermediate computations for the correlation distance algorithm, double or float
+/// \tparam cpu              Type of CPU, e.g. daal::CpuType::sse2, daal::CpuType::avx2, etc.
+/// \tparam upper            If true, compute the upper triangle of the block, otherwise compute the lower triangle
+///
+/// \param[in] blockSize     Size of the block to compute
+/// \param[in] nColumns      Number of columns in the input matrix x
+/// \param[in] x             Pointer to the input matrix x of size blockSize * nColumns
+/// \param[in,out] sum       On input, pointer to the array of size blockSize where the sum of each row of x is be stored.
+///                          On output, it contains the diagonal elements of the block.
+/// \param[out] block        Pointer to the output block of size blockSize * blockSize, where the computed correlation distance block will be stored
+template <typename algorithmFPType, CpuType cpu, bool upper = true>
+void computeDiagonalBlock(const size_t blockSize, const size_t nColumns, const algorithmFPType * x, algorithmFPType * sum, algorithmFPType * block)
+{
+    /* calculate sum^t * sum */
+    const algorithmFPType one(1.0);
+    const algorithmFPType zero(0.0);
+    const DAAL_INT bsz = blockSize;
+    const DAAL_INT ione = 1;
+
+    BlasInst<algorithmFPType, cpu>::xxgemm("N", "T", &bsz, &bsz, &ione, &one, sum, &bsz, sum, &bsz, &zero, block, &bsz);
+
+    /* calculate x * x^t - 1/p * sum^t * sum */
+    const algorithmFPType beta = -one / (algorithmFPType)nColumns;
+    const DAAL_INT ncol = nColumns;
+
+    BlasInst<algorithmFPType, cpu>::xxgemm("T", "N", &bsz, &bsz, &ncol, &one, x, &ncol, x, &ncol, &beta, block, &bsz);
+
+    algorithmFPType * diag = sum; // reuse sum array for diagonal elements
+
+    for (size_t i = 0; i < blockSize; i++)
+    {
+        if (block[i * blockSize + i] > (algorithmFPType)0.0)
+        {
+            block[i * blockSize + i] = (algorithmFPType)1.0 / daal::internal::MathInst<algorithmFPType, cpu>::sSqrt(block[i * blockSize + i]);
+        }
+        diag[i] = block[i * blockSize + i];
+    }
+
+    if constexpr (upper)
+    {
+        for (size_t i = 0; i < blockSize; i++)
+        {
+            PRAGMA_FORCE_SIMD
+            for (size_t j = i + 1; j < blockSize; j++)
+            {
+                block[i * blockSize + j] = one - block[i * blockSize + j] * diag[i] * diag[j];
+            }
+        }
+    }
+    else
+    {
+        for (size_t i = 0; i < blockSize; i++)
+        {
+            PRAGMA_VECTOR_ALWAYS
+            for (size_t j = 0; j < i; j++)
+            {
+                block[i * blockSize + j] = one - block[i * blockSize + j] * diag[i] * diag[j];
+            }
+        }
+    }
+}
+
+}
+}
+}
+}
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
@@ -485,8 +485,7 @@ bool UnorderedRespHelperBest<algorithmFPType, cpu>::findSplitCategoricalFeature(
         if ((count < nMinSplitPart) || ((n - count) < nMinSplitPart) || (leftWeights < minWeightLeaf)
             || ((totalWeights - leftWeights) < minWeightLeaf))
             continue;
-        PRAGMA_FORCE_SIMD
-        PRAGMA_VECTOR_ALWAYS
+
         for (size_t j = 0; j < _nClasses; ++j) _impRight.hist[j] = curImpurity.hist[j] - _impLeft.hist[j];
         calcGini(leftWeights, _impLeft);
         calcGini(totalWeights - leftWeights, _impRight);
@@ -726,7 +725,6 @@ void RespHelperBase<algorithmFPType, cpu, crtp>::calcImpurity(const IndexType * 
     imp.init(this->_nClasses);
     if (noWeights)
     {
-        PRAGMA_VECTOR_ALWAYS
         for (size_t i = 0; i < n; ++i)
         {
             const ClassIndexType iClass = this->_aResponse[aIdx[i]].val;
@@ -736,7 +734,6 @@ void RespHelperBase<algorithmFPType, cpu, crtp>::calcImpurity(const IndexType * 
     }
     else
     {
-        PRAGMA_VECTOR_ALWAYS
         for (size_t i = 0; i < n; ++i)
         {
             const ClassIndexType iClass = this->_aResponse[aIdx[i]].val;
@@ -1186,8 +1183,7 @@ int UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitFewClasses(int nDi
         {
             minidx++;
 
-            PRAGMA_FORCE_SIMD
-            PRAGMA_VECTOR_ALWAYS
+            PRAGMA_OMP_SIMD(reduction(+:thisNFeatIdx))
             for (size_t iClass = 0; iClass < K; ++iClass)
             {
                 thisNFeatIdx += nSamplesPerClass[minidx * K + iClass];
@@ -1202,8 +1198,7 @@ int UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitFewClasses(int nDi
         while ((minidx < maxidx) && isZero<IndexType, cpu>(thisNFeatIdx)) thisNFeatIdx = nFeatIdx[++minidx];
         nLeft = thisNFeatIdx;
 
-        PRAGMA_FORCE_SIMD
-        PRAGMA_VECTOR_ALWAYS
+        PRAGMA_OMP_SIMD(reduction(+:minWeights))
         for (size_t iClass = 0; iClass < K; ++iClass)
         {
             minWeights += nSamplesPerClass[minidx * K + iClass];
@@ -1228,8 +1223,7 @@ int UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitFewClasses(int nDi
         while ((minidx < maxidx) && isZero<algorithmFPType, cpu>(thisNFeatIdx))
         {
             maxidx--;
-            PRAGMA_FORCE_SIMD
-            PRAGMA_VECTOR_ALWAYS
+            PRAGMA_OMP_SIMD(reduction(+:thisNFeatIdx))
             for (size_t iClass = 0; iClass < K; ++iClass)
             {
                 thisNFeatIdx += nSamplesPerClass[maxidx * K + iClass];
@@ -1260,12 +1254,12 @@ int UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitFewClasses(int nDi
         //iterate idx down to a bin with values for FinalizeBestSplit
         algorithmFPType thisNFeatIdx(0);
 
-        PRAGMA_FORCE_SIMD
-        PRAGMA_VECTOR_ALWAYS
+        PRAGMA_OMP_SIMD(reduction(+:thisNFeatIdx))
         for (size_t iC = 0; iC < K; ++iC) thisNFeatIdx += nSamplesPerClass[idx * K + iC];
         while ((minidx < idx) && isZero<algorithmFPType, cpu>(thisNFeatIdx))
         {
             idx--;
+            PRAGMA_OMP_SIMD(reduction(+:thisNFeatIdx))
             for (size_t iClass = 0; iClass < K; ++iClass)
             {
                 thisNFeatIdx += nSamplesPerClass[idx * K + iClass];
@@ -1282,18 +1276,19 @@ int UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitFewClasses(int nDi
         }
         else //sum over all to idx
         {
-            PRAGMA_FORCE_SIMD
-            PRAGMA_VECTOR_ALWAYS
             for (size_t i = minidx + 1; i <= idx; i++)
             {
+                PRAGMA_FORCE_SIMD
+                PRAGMA_VECTOR_ALWAYS
                 for (size_t iClass = 0; iClass < K; ++iClass) histLeft[iClass] += nSamplesPerClass[i * K + iClass];
             }
         }
 
-        PRAGMA_FORCE_SIMD
-        PRAGMA_VECTOR_ALWAYS
+        PRAGMA_OMP_SIMD(reduction(+:leftWeights))
         for (size_t iClass = 0; iClass < K; ++iClass)
+        {
             leftWeights += histLeft[iClass]; //histleft is forced to float, and may cause issues with algorithmFPType = double
+        }
         nLeft = leftWeights;
     }
     else
@@ -1322,8 +1317,7 @@ int UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitFewClasses(int nDi
             }
         }
 
-        PRAGMA_FORCE_SIMD
-        PRAGMA_VECTOR_ALWAYS
+        PRAGMA_OMP_SIMD(reduction(+:leftWeights))
         for (size_t iClass = 0; iClass < K; ++iClass) leftWeights += histLeft[iClass];
     }
 
@@ -1412,8 +1406,6 @@ bool UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitOrderedFeature(co
 
     if (noWeights)
     {
-        PRAGMA_FORCE_SIMD
-        PRAGMA_VECTOR_ALWAYS
         for (i = 0; i < r; ++i)
         {
             const ClassIndexType iClass = this->_aResponse[aIdx[i]].val;
@@ -1423,8 +1415,6 @@ bool UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitOrderedFeature(co
     }
     else
     {
-        PRAGMA_FORCE_SIMD
-        PRAGMA_VECTOR_ALWAYS
         for (i = 0; i < r; ++i)
         {
             const ClassIndexType iClass = this->_aResponse[aIdx[i]].val;
--- a/cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i
@@ -160,10 +160,10 @@ protected:
     {
         const algorithmFPType expThreshold = daal::internal::MathInst<algorithmFPType, cpu>::vExpThreshold();
         algorithmFPType maxArg             = arg[0];
-        PRAGMA_VECTOR_ALWAYS
+        PRAGMA_OMP_SIMD(reduction(max:maxArg))
         for (size_t i = 1; i < _nClasses; ++i)
         {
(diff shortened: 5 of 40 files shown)


cpp/daal/src/services/service_defines.h
```diff
@@ -44,59 +44,49 @@ DAAL_EXPORT bool daal_check_is_intel_cpu();
 
 #define DAAL_CHECK_CPU_ENVIRONMENT (daal_check_is_intel_cpu())
 
-#if defined(__INTEL_COMPILER)
+#if defined(__INTEL_COMPILER) || defined(__INTEL_LLVM_COMPILER)
     #define PRAGMA_FORCE_SIMD       _Pragma("ivdep")
```
david-cortes-intel: I see this has different definitions here and in another file. Plus neither of them actually 'forces' SIMD onto a loop (the compiler can still decide not to vectorize).

How about naming them differently?

cpp/daal/src/services/service_defines.h
```diff
@@ -44,59 +44,49 @@ DAAL_EXPORT bool daal_check_is_intel_cpu();
...
     #define PRAGMA_NOVECTOR         _Pragma("novector")
     #define PRAGMA_VECTOR_ALIGNED   _Pragma("vector aligned")
```
david-cortes-intel: Does this play along with `omp simd`? How about passing the alignment to the OMP pragma? It also supports `#pragma omp simd aligned(pointer_name:64)`.

cpp/daal/src/services/service_defines.h
```diff
@@ -44,59 +44,49 @@ DAAL_EXPORT bool daal_check_is_intel_cpu();
...
     #define PRAGMA_VECTOR_UNALIGNED
     #define PRAGMA_VECTOR_ALWAYS
-    #define PRAGMA_ICC_TO_STR(ARGS)
-    #define PRAGMA_ICC_OMP(ARGS)
-    #define PRAGMA_ICC_NO16(ARGS)
     #define DAAL_TYPENAME typename
 #elif defined(_MSC_VER)
-    #define PRAGMA_FORCE_SIMD
-    #define PRAGMA_NOVECTOR
+    #define PRAGMA_FORCE_SIMD _Pragma("loop(ivdep)")
```
david-cortes-intel: MSVC supports `omp simd` if enabling experimental mode:
https://learn.microsoft.com/en-us/cpp/parallel/openmp/openmp-simd?view=msvc-170

cpp/daal/src/services/service_defines.h
```diff
@@ -44,59 +44,49 @@ DAAL_EXPORT bool daal_check_is_intel_cpu();
...
+    #define PRAGMA_OMP_SIMD(ARGS)   PRAGMA_TO_STR(omp simd ARGS)
     #define DAAL_TYPENAME           typename
 #elif defined(__GNUC__)
     #if defined(TARGET_ARM)
-        #define PRAGMA_FORCE_SIMD _Pragma("omp simd")
+        #define PRAGMA_FORCE_SIMD     _Pragma("omp simd")
+        #define PRAGMA_TO_STR(ARGS)   _Pragma(#ARGS)
+        #define PRAGMA_OMP_SIMD(ARGS) PRAGMA_TO_STR(omp simd ARGS)
     #else
         #define PRAGMA_FORCE_SIMD
```
david-cortes-intel: `omp simd` is supported by GCC on all platforms as far as I am aware.

cpp/oneapi/dal/backend/common.hpp
```diff
@@ -25,12 +25,16 @@
...
-#if defined(__INTEL_COMPILER)
+#if defined(__INTEL_COMPILER) || defined(__INTEL_LLVM_COMPILER)
+#define PRAGMA_TO_STR(ARGS)     _Pragma(#ARGS)
+#define PRAGMA_OMP_SIMD(ARGS) PRAGMA_TO_STR(omp simd ARGS)
 #define PRAGMA_FORCE_SIMD    _Pragma("ivdep")
 #define PRAGMA_VECTOR_ALWAYS _Pragma("vector always")
 #else
 #if defined(TARGET_ARM)
+#define PRAGMA_TO_STR(ARGS)   _Pragma(#ARGS)
 #define PRAGMA_FORCE_SIMD _Pragma("omp simd")
```
david-cortes-intel: Wouldn't this require adding `-fopenmp` flag to the makefile?

Vika-F: No, it only requires `-qopenmp-simd` which we already have.

cpp/daal/src/algorithms/cordistance/cordistance_impl.i
```diff
@@ -0,0 +1,120 @@
...
+///
+/// \tparam algorithmFPType  Data type to use in intermediate computations for the correlation distance algorithm, double or float
+/// \tparam cpu              Type of CPU, e.g. daal::CpuType::sse2, daal::CpuType::avx2, etc.
+///
+/// \param[in] nRows            Number of rows in the input matrix x
+/// \param[in] nColumns         Number of columns in the input matrix x
+/// \param[in] x                Pointer to the input matrix x of size nRows * nColumns
+/// \param[out] sum              Pointer to the output array of size nRows, where the sum of each row of x will be stored
+template <typename algorithmFPType, CpuType cpu>
+void sumByRows(const size_t nRows, const size_t nColumns, const algorithmFPType * x, algorithmFPType * sum)
```
david-cortes-intel, line 39: Why not call MKL here? https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-1/vslsseditsums.html

cpp/daal/src/algorithms/cordistance/cordistance_impl.i
```diff
@@ -0,0 +1,120 @@
...
+
+    BlasInst<algorithmFPType, cpu>::xxgemm("T", "N", &bsz, &bsz, &ncol, &one, x, &ncol, x, &ncol, &beta, block, &bsz);
+
+    algorithmFPType * diag = sum; // reuse sum array for diagonal elements
+
+    for (size_t i = 0; i < blockSize; i++)
+    {
+        if (block[i * blockSize + i] > (algorithmFPType)0.0)
+        {
+            block[i * blockSize + i] = (algorithmFPType)1.0 / daal::internal::MathInst<algorithmFPType, cpu>::sSqrt(block[i * blockSize + i]);
```
david-cortes-intel: There is a PR adding a function for this from MKL: https://github.com/uxlfoundation/oneDAL/pull/3227

Perhaps that other one could be merged first.

cpp/daal/src/services/service_defines.h
Vika-F: This name was introduced because there is the same macro that is expanded to `#pragma ivdep` with intel compiler and to `#pragma omp simd` with GCC on ARM.
I can rename to `PRAGMA_USE_SIMD`. Or maybe you can propose something more suitable?

david-cortes-intel: How about:
* `#pragma ivdep` -> PRAGMA_IVDEP
* `#pragma omp simd` -> PRAGMA_USE_SIMD
* `#pragma omp simd aligned(...)` -> PRAGMA_USE_SIMD_ALIGNED(...)

cpp/daal/src/services/service_defines.h
Vika-F: The idea is to pass any arguments to `#pragma omp simd` with PRAGMA_OMP_SIMD() macro.
For example:
https://github.com/uxlfoundation/oneDAL/pull/3246/files#diff-61b267e32558b19a2dba159ff0128be35c0c8eafe9b140791a7073e262af17f7R1201

Alignment options can be passed similarly.

cpp/daal/src/algorithms/cordistance/cordistance_impl.i
Vika-F: I think it is out of the scope of this PR.
Here I am just trying to enable the pragmas for ICX. The refactoring was made just to reduce code duplication and not to do the same modifications in 3 files.

cpp/daal/src/services/service_defines.h
Vika-F: The problem is that now it is defined as:
```
 #if defined(__INTEL_COMPILER)
        #define PRAGMA_FORCE_SIMD     _Pragma("ivdep")
#endif
#if defined(TARGET_ARM)
        #define PRAGMA_FORCE_SIMD     _Pragma("omp simd")
#endif
```
And if the code of the algorithms we have:
```
  PRAGMA_FORCE_SIMD
  for (...)
```

With what is proposed, I'll need to go through the entire code base and choose either of PRAGMA_IVDEP or PRAGMA_USE_SIMD, or both?

david-cortes-intel: @Vika-F Since 'ivdep' was only being used on ARM, and since the original intent was to make it vectorize things, is there actually any current case where 'ivdep' would apply but 'omp simd' wouldn't? Perhaps they could be all switched to 'omp simd' and then later on 'ivdep' be used as needed on new developments.

cpp/daal/src/services/service_defines.h
david-cortes-intel: But does the compiler actually use the hint if put before `omp simd`?

cpp/daal/src/algorithms/cordistance/cordistance_impl.i
Vika-F: I am Ok to have that one merged first and reuse that functionality; but I think the performance have to be measured, as that PR clearly might have performance impact on the algorithm.

cpp/daal/src/services/service_defines.h
Vika-F: Make sense.
I'll try using this approach with the only difference: PRAGMA_OMP_SIMD[_ALIGNED] instead of PRAGMA_USE_SIMD[_ALIGNED].

cpp/daal/src/services/service_defines.h
Vika-F: Done.

dev/make/compiler_definitions/gnu.32e.mk
```diff
@@ -47,7 +47,7 @@ else
 endif
 
 COMPILER.all.gnu =  ${CXX} -m64 -fwrapv -fno-strict-overflow -fno-delete-null-pointer-checks \
-                    -Werror -Wreturn-type ${CXXFLAGS}
+                    -Werror -Wreturn-type -fopenmp-simd ${CXXFLAGS}
```
david-cortes-intel, line 50: Wouldn't this be also needed for clang?

cpp/daal/src/algorithms/cordistance/cordistance_impl.i
david-cortes-intel: PR has been merged by now and performance impact was positive. Please remember to update.

cpp/daal/src/services/service_defines.h
david-cortes-intel: @Vika-F Please remember about this one.

dev/make/compiler_definitions/gnu.32e.mk
Vika-F: Now clang goes to `else` branch which does not use `omp simd`. It's an oversight. I'll fix it.

david-cortes-intel: Be aware that clang still defines the GNU macros.

Vika-F: @david-cortes-intel Added OMP macros for Clang. The only issue: it doesn't vectorize max reduction for some reason. But in short stand-alone example the similar code gets successfully vectorized. Don't want to spend too much time on the investigation. I've added workarounds for now:
https://github.com/uxlfoundation/oneDAL/pull/3246/files#diff-1e7c103235a87e753e399150e458de51e0491e8921d96b734b03e5e7fb941fbdR160

cpp/daal/src/services/service_defines.h
```diff
@@ -46,57 +46,37 @@ DAAL_EXPORT bool daal_check_is_intel_cpu();
...
+
+#if defined(__INTEL_COMPILER) || defined(__INTEL_LLVM_COMPILER)
+    #define PRAGMA_IVDEP               _Pragma("ivdep")
+    #define PRAGMA_NOVECTOR            _Pragma("novector")
+    #define PRAGMA_VECTOR_UNALIGNED    _Pragma("vector unaligned")
+    #define PRAGMA_VECTOR_ALWAYS       _Pragma("vector always")
+    #define PRAGMA_OMP_SIMD            PRAGMA_TO_STR(omp simd)
+    #define PRAGMA_OMP_SIMD_ARGS(ARGS) PRAGMA_TO_STR_(omp simd ARGS)
+#elif defined(__GNUC__) || defined(__clang__)
+    #define PRAGMA_IVDEP
```
david-cortes-intel, line 60:
```suggestion
    #define PRAGMA_IVDEP _Pragma("ivdep")
```

cpp/daal/src/services/service_defines.h
```diff
@@ -46,57 +46,37 @@ DAAL_EXPORT bool daal_check_is_intel_cpu();
...
+    #define PRAGMA_NOVECTOR
     #define PRAGMA_VECTOR_UNALIGNED
     #define PRAGMA_VECTOR_ALWAYS
```
david-cortes-intel, line 63:
```suggestion
    #define PRAGMA_VECTOR_UNALIGNED _Pragma("omp simd")
    #define PRAGMA_VECTOR_ALWAYS _Pragma("omp simd")
```

cpp/oneapi/dal/algo/triangle_counting/backend/cpu/relabel_kernel.hpp
```diff
@@ -64,20 +64,51 @@ void parallel_prefix_sum(const std::int32_t* degrees_relabel,
...
         local_sums[block] = local_sum;
     });
 
-    std::int64_t total = 0;
-    PRAGMA_VECTOR_ALWAYS
-    for (std::int64_t block = 0; block < num_blocks; block++) {
-        part_prefix[block] = total;
-        total += local_sums[block];
+    // std::int64_t total = 0;
+    constexpr unsigned int simd_width = 4;
```
david-cortes-intel, line 74: Why 4 and not 8? Wouldn't AVX512 support 8 at a time?

cpp/daal/src/services/service_defines.h
Vika-F: Those macros were not defined for GNU and clang previously, I would leave it for a follow up PRs.

cpp/oneapi/dal/algo/triangle_counting/backend/cpu/relabel_kernel.hpp
Vika-F: I am not aware about the instruction that can shuffle 512 bit registers.

cpp/daal/src/algorithms/cordistance/cordistance_impl.i
```diff
@@ -0,0 +1,120 @@
...
+/// \param[in] nColumns      Number of columns in the input matrix x
+/// \param[in] x             Pointer to the input matrix x of size blockSize * nColumns
+/// \param[in,out] sum       On input, pointer to the array of size blockSize where the sum of each row of x is be stored.
+///                          On output, it contains the diagonal elements of the block.
+/// \param[out] block        Pointer to the output block of size blockSize * blockSize, where the computed correlation distance block will be stored
+template <typename algorithmFPType, CpuType cpu, bool upper = true>
+void computeDiagonalBlock(const size_t blockSize, const size_t nColumns, const algorithmFPType * x, algorithmFPType * sum, algorithmFPType * block)
+{
+    /* calculate sum^t * sum */
+    const algorithmFPType one(1.0);
```
Alexandr-Solovev: Probably name alpha is better here

cpp/daal/src/algorithms/cosdistance/cosdistance_impl.i
```diff
@@ -0,0 +1,87 @@
...
+/// \param[in] blockSize     Size of the block to compute
+/// \param[in] nColumns      Number of columns in the input matrix x
+/// \param[in] x             Pointer to the input matrix x of size blockSize * nColumns
+/// \param[out] block        Pointer to the output block of size blockSize * blockSize, where the computed cosine distance block will be stored
+template <typename algorithmFPType, CpuType cpu, bool upper = true>
+void computeDiagonalBlock(const size_t blockSize, const size_t nColumns, const algorithmFPType * x, algorithmFPType * block)
+{
+    /* calculate x^t * x */
+    algorithmFPType diag[blockSizeDefault];
+    constexpr algorithmFPType one(1.0);
```
Alexandr-Solovev, line 44: the same comment here

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i
```diff
@@ -139,18 +139,15 @@ public:
...
-            PRAGMA_FORCE_SIMD
+            PRAGMA_OMP_SIMD
             PRAGMA_VECTOR_ALWAYS
             for (size_t k = 0; k < _nClasses; ++k)
             {
                 const algorithmFPType pk = p[k];
                 const algorithmFPType h  = algorithmFPType(2.) * pk * (algorithmFPType(1.) - pk);
                 algorithmFPType * gh_ik  = gh + 2 * (k * nRows + iSample);
+                gh_ik[0]                 = (size_t(y[iSample]) == k) ? (pk - algorithmFPType(1.)) : pk;
                 gh_ik[1]                 = h;
```
Alexandr-Solovev: does it make sense to compute gh_ik[1] as algorithmFPType(2.) * pk * (algorithmFPType(1.) - pk); instead of additional variable h?

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i
```diff
@@ -160,12 +157,14 @@ protected:
...
         const algorithmFPType expThreshold = daal::internal::MathInst<algorithmFPType, cpu>::vExpThreshold();
         algorithmFPType maxArg             = arg[0];
-        PRAGMA_VECTOR_ALWAYS
+#ifndef __clang__ // TODO: Temporary workaround. Clang 18 fails to vectoize this simple loop
+        PRAGMA_OMP_SIMD_ARGS(reduction(max : maxArg))
+#endif
         for (size_t i = 1; i < _nClasses; ++i)
         {
-            if (maxArg < arg[i]) maxArg = arg[i];
+            maxArg = (maxArg < arg[i]) ? arg[i] : maxArg;
```
Alexandr-Solovev, line 166: offtop question: do we have plans to add more __builtin_expect ?

cpp/daal/src/algorithms/kmeans/kmeans_lloyd_impl.i
```diff
@@ -67,14 +67,13 @@ struct TaskKMeansLloyd
...
-                PRAGMA_ICC_NO16(omp simd reduction(+ : sum))
+                algorithmFPType sum(0);
+                PRAGMA_OMP_SIMD_ARGS(reduction(+ : sum))
                 for (size_t j = 0; j < dim; j++)
                 {
-                    sum += cCenters[k * dim + j] * cCenters[k * dim + j] * 0.5;
+                    sum += cCenters[k * dim + j] * cCenters[k * dim + j];
                 }
-                clSq[k] = sum;
+                clSq[k] = sum * 0.5;
```
Alexandr-Solovev, line 76: good catch

cpp/daal/src/algorithms/cordistance/cordistance_impl.i
Vika-F: Done

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i
Vika-F: I do not think __builtin_expect or [[likely]] are useful in the cases like this, as here the branch taken depends on the input data.
I think it is more useful in the scenarios like error handling. I.e. we can treat the error handling branches as unlikely situations.
For example:
```
if (__builtin_expect(error_condition == true, 0)) {
    /* unlikely branch, error handling here */
}
```

Compiler-specific macros would also be needed to do this.

cpp/daal/src/algorithms/cordistance/cordistance_impl.i
icfaust: I guess in this case this new file could possibly be dropped (by the looks of it)?

Edit: nevermind.  I was wrong. I am guess I understand what is in #3227, and wonder if there is a way to centralize this file with the copy existing in the cosdistance. If it were in the DAL side we would put it as a primitive, but I don't know if that something we do on the DAAL side.

cpp/daal/src/algorithms/cordistance/cordistance_full_impl.i
```diff
@@ -67,54 +67,10 @@ services::Status corDistanceFull(const NumericTable * xTable, NumericTable * rTa
...
-
-        for (size_t i = 0; i < blockSize1; i++)
-        {
-            PRAGMA_VECTOR_ALWAYS
-            for (size_t j = i + 1; j < blockSize1; j++)
-            {
-                buf[i * blockSize1 + j] = 1.0 - buf[i * blockSize1 + j] * buf[i * blockSize1 + i] * buf[j * blockSize1 + j];
-            }
-        }
+        /* compute upper triangle of diagonal block of the correlation distance elements */
```
icfaust, line 72: Great comment in the code

cpp/daal/src/algorithms/cordistance/cordistance_batch_impl.i
```diff
@@ -29,6 +29,7 @@
 #include "src/data_management/service_numeric_table.h"
 
 static const int blockSizeDefault = 128;
+#include "src/algorithms/cordistance/cordistance_impl.i"
```
icfaust, line 32: Just a note to look at this when changes in #3227 are integrated.

cpp/daal/src/algorithms/covariance/covariance_impl.i
```diff
@@ -270,9 +270,7 @@ public:
         }
 
         /// It is safe to use aligned loads and stores because the data in TArrayScalableCalloc data structures is aligned
-        PRAGMA_FORCE_SIMD
-        PRAGMA_VECTOR_ALWAYS
-        PRAGMA_VECTOR_ALIGNED
+        PRAGMA_OMP_SIMD_ARGS(aligned(thisCrossProduct, otherCrossProduct : DAAL_MALLOC_DEFAULT_ALIGNMENT))
```
icfaust, line 273: The more I look at this, I guess the title for the PR should possible be changed (I think its underselling a bit actually), because I bet this will improve performance on the other supported compilers.

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
```diff
@@ -1202,8 +1198,7 @@ int UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitFewClasses(int nDi
         while ((minidx < maxidx) && isZero<IndexType, cpu>(thisNFeatIdx)) thisNFeatIdx = nFeatIdx[++minidx];
         nLeft = thisNFeatIdx;
 
-        PRAGMA_FORCE_SIMD
```
icfaust, line 1211: So I must apologize here, as when I implemented it I did this a bit blindly to match the pre-existing code. Thank you for working on this.

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i
```diff
@@ -139,18 +139,14 @@ public:
...
             {
                 const algorithmFPType pk = p[k];
-                const algorithmFPType h  = algorithmFPType(2.) * pk * (algorithmFPType(1.) - pk);
                 algorithmFPType * gh_ik  = gh + 2 * (k * nRows + iSample);
-                gh_ik[1]                 = h;
-                if (size_t(y[iSample]) == k)
-                    gh_ik[0] = (pk - algorithmFPType(1.));
-                else
-                    gh_ik[0] = pk;
+                gh_ik[0]                 = (size_t(y[iSample]) == k) ? (pk - algorithmFPType(1.)) : pk; // gradient
```
icfaust: The algorithmFPType of the 1 and 2 below are compile time right? So long as that is the case I am okay with this.

cpp/daal/src/algorithms/dtrees/gbt/gbt_train_partition.i
```diff
@@ -183,11 +183,48 @@ protected:
...
+                /// The loop implementation is unrolled to improve performance by using SIMD instructions.
+                /// The original loop looks like this:
+                /// for (IndexType i = iStart; i < iEnd; ++i)
+                /// {
+                ///     if (indexedFeature[aIdx[i]] != idxFeatureValueBestSplit)
+                ///         bestSplitIdxRight[iRight++] = aIdx[i];
+                ///     else
+                ///         bestSplitIdx[iLeft++] = aIdx[i];
+                /// }
+                /// The unrolled loop:
```
icfaust: ❤️ for doing the documentation here the right way. This unrolling should improve performance right (since its more specific as what was previous done)?

cpp/daal/src/algorithms/dtrees/gbt/gbt_train_partition.i
```diff
@@ -183,11 +183,48 @@ protected:
             RowIndexType * bestSplitIdx      = buffer + 2 * iStart;
             RowIndexType * bestSplitIdxRight = bestSplitIdx + iEnd - iStart;
 
+            constexpr size_t unrollFactor = 32;                             /// Unroll factor for SIMD
```
icfaust: Is this factor is set to 32 to match AVX2 (256 bit with float)? I'm not wanting to improve on this in this PR, but if someone really want to crank the perfomance here I expect they would want to tweak this, and so an additional note may be necessary (not sure).

cpp/daal/src/algorithms/dtrees/gbt/gbt_train_updater.i
```diff
@@ -188,7 +188,7 @@ protected:
     {
         LoopHelper<cpu>::run(true, this->_data.ctx.nFeaturesPerNode(), [&](size_t i) {
             const DAAL_INT iFeature = featureSample ? featureSample[i] : i;
-            DAAL_TYPENAME super::SplitTaskType task(iFeature, this->_data, this->_node, bestSplit, this->_result->res[i]);
```
icfaust, line 191: I'd be nice to include some understanding of this change in the description of the PR (not sure how the typename was getting influenced by the vectorization pragmas).

cpp/daal/src/algorithms/dtrees/service_array.h
```diff
@@ -115,9 +115,8 @@ class TVector
 
     void resize(size_t n)
     {
-        T * ptr = (T *)Allocator::alloc((n) * sizeof(T));
-        PRAGMA_VECTOR_ALWAYS
-        for (size_t i = 0; i < (_size > n ? n : _size); ++i) ptr[i] = _data[i];
+        T * ptr = (T *)Allocator::alloc(n * sizeof(T));
+        services::internal::daal_memcpy_s(ptr, n * sizeof(T), _data, _size * sizeof(T));
```
icfaust, line 119: Great change.

cpp/daal/src/algorithms/low_order_moments/low_order_moments_impl.i
```diff
@@ -860,14 +860,21 @@ void finalize(LowOrderMomentsFinalizeTask<algorithmFPType, cpu> & task)
...
+    for (size_t i = 0; i < nFeatures; i++)
+    {
+        mean[i]     = sums[i] * invNObservations;
+        raw2Mom[i]  = sumSq[i] * invNObservations;
+        variance[i] = sumSqCen[i] * invNObservationsM1;
+    }
+
+    daal::internal::MathInst<algorithmFPType, cpu>::vSqrt(nFeatures, variance, stDev);
+
+    PRAGMA_OMP_SIMD
```
icfaust, line 874: This is okay because its the finalization step and the metrics are independent arrays? I assume that the pre-finalization is something like an AoS structure of the various metrics (sorry if this is a dumb question, I have not deeply looked into low order moments, but have some assumptions how it may operate).

cpp/daal/src/algorithms/multiclassclassifier/multiclassclassifier_predict_votebased_impl.i
```diff
@@ -112,14 +112,13 @@ public:
...
-            PRAGMA_FORCE_SIMD
+            PRAGMA_OMP_SIMD
             PRAGMA_VECTOR_ALWAYS
             for (size_t i = 0; i < nRows; ++i)
             {
-                if (y[i] >= 0)
-                    votes[i * _nClasses + iClass]++;
-                else
-                    votes[i * _nClasses + jClass]++;
+                const int isIClass(y[i] >= 0); // isIClass == 1 if y[i] >= 0; isIClass == 0 if y[i] < 0
```
icfaust, line 119: clever.

cpp/daal/src/algorithms/multiclassclassifier/multiclassclassifier_train_oneagainstone_impl.i
```diff
@@ -337,9 +337,14 @@ Status SubTaskDense<algorithmFPType, cpu>::copyDataIntoSubtable(size_t nFeatures
...
-        PRAGMA_FORCE_SIMD
+        const algorithmFPType * x = _mtX.get();
+        algorithmFPType * subsetX = this->_subsetX.get();
+        PRAGMA_OMP_SIMD
         PRAGMA_VECTOR_ALWAYS
-        for (size_t jx = 0; jx < nFeatures; jx++) this->_subsetX.get()[nRows * nFeatures + jx] = _mtX.get()[jx];
+        for (size_t jx = 0; jx < nFeatures; jx++)
+        {
+            subsetX[nRows * nFeatures + jx] = x[jx];
+        }
```
icfaust, line 347: services::internal::daal_memcpy_s ?

cpp/daal/src/algorithms/pca/pca_dense_svd_batch_impl.i
```diff
@@ -280,11 +281,17 @@ services::Status PCASVDBatchKernel<algorithmFPType, ParameterType, cpu>::normali
     if (!safeStat) return safeStat.detach();
 
     /* Convert array of variances to inverse sigma's */
-    PRAGMA_FORCE_SIMD
-    PRAGMA_VECTOR_ALWAYS
+    PRAGMA_OMP_SIMD
     for (size_t j = 0; j < nFeatures; j++)
     {
-        if (inv_sigma_total[j]) inv_sigma_total[j] = algorithmFPType(1.0) / daal::internal::MathInst<algorithmFPType, cpu>::sSqrt(inv_sigma_total[j]);
+        non_zero_sigma_total[j] = (inv_sigma_total[j] > 0.0) ? 1.0 : 0.0;
```
icfaust:
```suggestion
        non_zero_sigma_total[j] = algorithmFPType(inv_sigma_total[j] > 0.0);
```
Sorry if this a dumb suggestion, it looks like type casting was already occurring in some fashion, I figure just abuse the int bool in its current form anyway.

cpp/daal/src/services/service_defines.h
icfaust: I agree with @Vika-F on this. We need to limit scope creep on the PR due to the size and impact it has.

cpp/daal/src/services/service_defines.h
```diff
@@ -46,57 +46,37 @@ DAAL_EXPORT bool daal_check_is_intel_cpu();
...
-    #define DAAL_TYPENAME typename
-#endif
-
-#if defined __APPLE__ && defined __INTEL_COMPILER && (__INTEL_COMPILER == 1600)
-    #undef PRAGMA_ICC_TO_STR
-    #define PRAGMA_ICC_TO_STR(ARGS) _Pragma(#ARGS)
-    #undef PRAGMA_ICC_OMP
-    #define PRAGMA_ICC_OMP(ARGS) PRAGMA_ICC_TO_STR(ARGS)
-    #undef PRAGMA_ICC_NO16
-    #define PRAGMA_ICC_NO16(ARGS)
```
icfaust, line 99: A bit of a change that shouldn't matter much, but worth mentioning in the description.

cpp/oneapi/dal/algo/triangle_counting/backend/cpu/relabel_kernel.hpp
icfaust: Worth adding a comment in the code to this affect (such that with future ISAs we can look at it again if we deem it important enough (realize that this is a graph algo).

icfaust (commented):
Impressive work. I do think this needs to be prioritized and merged quickly after the code freeze is over due to the size and impact (upon performance verification).  This must improve performance in a lot of cases and is worth mentioning.

What I would like to see is a list of the various thematic changes listed in the PR description because that will help future developers.  I think out side of a few nitpicks in what is implemented, most of what I worry about is code comments and text in the description for me and others to reference in the future for understanding and utility.

Again, this is really great work @Vika-F , thank you so much for doing this.

cpp/daal/src/algorithms/cordistance/cordistance_impl.i
Vika-F: In DAAL we don't have some centralized place for primitives, but there is an ability to share the code across the algorithms.
Quick search shown that we already have such common place for distances here: https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/src/algorithms/service_kernel_math.h#L110

For this PR let me replace this `sSqrt` with `vInvSqrtI` from #3227.
I have created a follow up task to further refactor the distances code.

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i
Vika-F: I hope so, but not 100% sure. I'll define them as constants.

cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
Vika-F: I think this kind of loop can be vectorized successfully without any pragma. I've just selected a more appropriate variant, though I do not think it impacts the performance here.

cpp/daal/src/algorithms/dtrees/gbt/gbt_train_updater.i
Vika-F: Yes, the change is unrelated.
This DAAL_TYPENAME macro was in the same block of defines with the pragmas, and it was defined similarly for all the compiler variants. So, I decided to clean it up as well.

cpp/daal/src/algorithms/dtrees/gbt/gbt_train_partition.i
Vika-F: The loop before the changes is not vectorizable by the compiler.

This was the attemp to make it at least partially vectorizable. Though still the main part if the computations is not vectorized due to gather/scatter pattern. And the performance of this part was not measured by sklex_bench, so I am not sure if this loop performs well after the change.

It would probably be better to revert this part of the changes.

cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i
Vika-F: Done.

cpp/daal/src/algorithms/dtrees/gbt/gbt_train_updater.i
Vika-F: Done

cpp/daal/src/algorithms/multiclassclassifier/multiclassclassifier_train_oneagainstone_impl.i
Vika-F: For some reason, `//cpp/oneapi/dal/algo/svm:test_serialization_host` starts to fail after the change.
Probably, something is wrong here, but I won't investigate it now, sorry.

icfaust: Totally understand. Put a ticket in the backlog for maintenance and it'll get prioritized as necessary (I imagine not for a while), and I'll be cool with it.

cpp/daal/src/services/service_defines.h
Vika-F: Done.

cpp/daal/src/algorithms/pca/pca_dense_svd_batch_impl.i
Vika-F: Done.

cpp/daal/src/algorithms/multiclassclassifier/multiclassclassifier_train_oneagainstone_impl.i
Vika-F: Done.

cpp/daal/src/algorithms/outlierdetection_multivariate/outlierdetection_multivariate_dense_default_impl.i
```diff
@@ -79,8 +78,7 @@ inline void OutlierDetectionKernel<algorithmFPType, method, cpu>::mahalanobisDis
     for (size_t i = 0; i < nVectors; i++, dataCenPtr += nFeatures, dataCenInvScatterPtr += nFeatures)
     {
         distance[i] = zero;
-        PRAGMA_FORCE_SIMD
-        PRAGMA_VECTOR_ALWAYS
+        PRAGMA_IVDEP
```
david-cortes-intel: These do not look independent. They all add towards the same value.

Vika-F: The initial change was `#pragma omp simd reduction(+:distanceValue)`, but it resulted in test failures.
So, I changed that back to `#pragma ivdep` to be more aligned with what we have now in `main`.
I can create a ticket to investigate this further, but not in the scope of this PR.

david-cortes-intel: But it's quite obvious from this single-line loop that ivdep is not applicable.

Vika-F: I've removed the ivdep.

Vika-F:
> @Vika-F Would you see an improvement in forests if you were to replace those pragmas with `omp simd` where applicable instead of removing them?

@david-cortes-intel Performance rerun is a rather time consuming thing now. Last time it took 2 CI jobs to get the results from EMR ang GNR. So, I would leave the improvements for the further PRs, and now I am just trying to have no degradations to make it mergeable finally.

Alexsandruss:
@Vika-F , conda build with clang has failed: https://github.com/uxlfoundation/oneDAL/actions/runs/18882214115/job/53888157260

PR #3256, opened 2025-06-13
[bug] fix shared library AddressSanitizer integration for the gnu compiler

icfaust:
## Description

The logic was getting ugly for integrating ASan, and as compiler/linker integrations of AddressSanitizer are under development (see https://devblogs.microsoft.com/cppblog/addresssanitizer-asan-for-windows-with-msvc/).  Related linkers have different default settings and different available options, this moves them to the compiler makefile definitions in order to isolate changes and maintain the central logic.

GNU compilers don't have the -shared-libasan linker flag, while clang, icx and dpcpp do. Windows doesn't maintain any of these flags (this is still very untested, but is just trying to simplify the work when verifying its operation at a later date).

Note this also doesn't fix some issues with other sanitizers, which for Clang will default to static linking (making introspection between the oneDAL shared librarys difficul/impossible) or for GNU which will default to shared linking.  This does not touch those issues, and is necessary for fixing ASan specifically. Some compilers/linkers don't allow for these other sanitizers at all.

---

**Testing**

**Performance**


david-cortes-intel:
@icfaust This results in a compilation error for the .a files:
```
ar: creating __work_gnu/md/lnx32e/daal/lib/libonedal.a
collect2: error: ld returned 1 exit status
make: *** [makefile:520: __work_gnu/md/lnx32e/daal/lib/libonedal_core.so] Error 1
```

david-cortes-intel:
The error appears to be something along these lines:
```
/usr/bin/ld: cannot find /usr/lib64/libasan.so.6.0.0
```

The available file in this system would be as follows:
```
/usr/lib/gcc/x86_64-redhat-linux/11/libasan.so
```

Not sure what exactly is making this PR look for this different filename.

dev/make/compiler_definitions/gnu.mk
```diff
@@ -28,6 +28,9 @@ CORE.SERV.COMPILER.gnu = generic
 -Zl.gnu =
 -DEBC.gnu = -g
 
+-asanstatic.gnu = -static-libasan
```
david-cortes-intel, line 31:
```suggestion
-asanstatic.gnu =
```

It at least compiles successfully if this argument is removed. Not sure about the impact though.

icfaust: Thanks for checking, it looks like by not including it, it will default to dynamic linking (https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-static-libasan).  Seems like an issue with GCC, should we leave a note and remove it, saying that static linking with GNU compilers is temporarily unavailable?

icfaust: By remove I mean take your suggestion.

david-cortes-intel: > Thanks for checking, it looks like by not including it, it will default to dynamic linking (https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-static-libasan). Seems like an issue with GCC, should we leave a note and remove it, saying that static linking with GNU compilers is temporarily unavailable?

I cannot think of any usage for a .a version of oneDAL with ASAN, so yes maybe it could added as a note.

Although on a second look, this appears to be a redhat-specific issue with misconfigured software, and I have a feeling it might be something specific to the way our servers are set up.

david-cortes-intel: @icfaust After some experiments with my setup, turns out it's not strictly necessary to remove this line.

But I still don't think the resulting file would be of much use - for example, in the case of sklearnex, if one tries to use that .a file, one might also want to build sklearnex itself with asan, but it cannot link statically to it (I think it's a limitation of cython but not 100% sure).

icfaust:
> @icfaust PR has conflicts with the main branch in the makefile now.

@david-cortes-intel Could you give some more info on this? (i.e. a reproducer) As of now, it merged main without issue, so some guidance will save some time in fixing.

PR #3271, opened 2025-06-26
Added SVE intrinsics for postGemmPart function

shubhamsvc:
This pull request adds SVE-based implementations of postGemmPart function for both float and double types to accelerate vectorized computation on ARM.

**Average Performance (on Graviton3)**

**Float**: ~4.3× speedup over scalar
**Double**: ~1.19× speedup over scalar

### DIFF
--- a/cpp/daal/src/algorithms/kernel_function/kernel_function_rbf_helper.h
+++ b/cpp/daal/src/algorithms/kernel_function/kernel_function_rbf_helper.h
@@ -19,7 +19,11 @@
 #define __KERNEL_FUNCTION_RBF_HELPER_H__
 
 #include "src/externals/service_math.h"
-
+#if defined(TARGET_ARM)
+#if (__CPUID__(DAAL_CPU) == __sve__)
+#include <arm_sve.h>  
+#endif
+#endif
 namespace daal
 {
 namespace algorithms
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
@shubhamsvc I haven't checked all the places where this is used, but I think the 'exp' function as used in RBF kernels requires high-precision computations - otherwise quality of the results might degrade significantly in some situations (particularly Newton-type methods that rely on those computations).

The exp function from MKL that gets called on x86 is called with the 'high accuracy' mode as you can see here:
https://github.com/uxlfoundation/oneDAL/blob/13c979bb596a4fe06864627e6456fc63fbbc04f5/cpp/daal/src/externals/service_math_mkl.h#L126

Could you provide some information about the accuracy level of the 'exp' function here? Is there some reference paper analyzing the method?

@Vika-F Could you provide some info about which algorithms use this function?

Vika-F:
@david-cortes-intel  @shubhamsvc 
> @Vika-F Could you provide some info about which algorithms use this function?

Besides RBF this function is used in GBT, prediction stage; but the accuracy requirements are lower there I guess.

vExp is also used in EM GMM, AdaBoost and LogitBoost algorithms, but those are not part of sklearn-intelex, the importance of those algorithms is lower.

shubhamsvc:
@david-cortes-intel @Vika-F  Thank you for quick response. What ULP accuracy is expected for exp(x) in this case?

david-cortes-intel:
For the general vExp function which is used also in logistic regression, I don't think it'd be advantageous to use something with more than 1ULP of error even if much faster, as numerical inaccuracies there do have a noticeable effect on convergence speed and quality of results.

For SVM RBF kernel in specific I am not sure - perhaps someone more familiar with the underlying algorithm could comment. Nevertheless, would still be ideal to know what's the accuracy level of this 'exp' function, at the very least to leave it as a comment in the code.

Perhaps one potential next step could be to conduct tests with RBF SVMs using the [sklbench repository](https://github.com/IntelPython/scikit-learn_bench) (which requires [sklearnex](https://github.com/uxlfoundation/scikit-learn-intelex) built against this oneDAL branch) before and after this PR and see how the quality metrics change. @Alexsandruss  Could you provide instructions for running the SVM RBF cases from sklbench? @rakshithgb-fujitsu Could you perhaps try out the changes before/after this PR with SVM RBF kernel examples in ARM hardware?

icfaust:
To support the importance of RBF, it is already exposed in sklearnex in the onedal module, https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/onedal/primitives/kernel_functions.py#L90 and could easily become a publicly-usable sklearnex function in short order (replicating sklearn functionality: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.rbf_kernel.html)

icfaust:
I added the PR checklist to guide reviewing/development.

rakshithgb-fujitsu:
> For the general vExp function which is used also in logistic regression, I don't think it'd be advantageous to use something with more than 1ULP of error even if much faster, as numerical inaccuracies there do have a noticeable effect on convergence speed and quality of results.
> 
> For SVM RBF kernel in specific I am not sure - perhaps someone more familiar with the underlying algorithm could comment. Nevertheless, would still be ideal to know what's the accuracy level of this 'exp' function, at the very least to leave it as a comment in the code.
> 
> Perhaps one potential next step could be to conduct tests with RBF SVMs using the [sklbench repository](https://github.com/IntelPython/scikit-learn_bench) (which requires [sklearnex](https://github.com/uxlfoundation/scikit-learn-intelex) built against this oneDAL branch) before and after this PR and see how the quality metrics change. @Alexsandruss Could you provide instructions for running the SVM RBF cases from sklbench? @rakshithgb-fujitsu Could you perhaps try out the changes before/after this PR with SVM RBF kernel examples in ARM hardware?

Yes, Shubham is from Fujitsu as well. We'll share all the benchmark details and validate it on arm hardware.

david-cortes-intel:
CI failures do not look related to this PR, but @shubhamsvc please lint the files according to the [instructions](https://github.com/uxlfoundation/oneDAL/blob/main/CONTRIBUTING.md#code-style) and merge the main branch here.

PR #3275, opened 2025-06-30
Fix shared pointers initialization in system_parameters and cpu_info classes

Vika-F:
## Description

The initial idea was to make pointers to implementation `impl_` unique pointers in both `system_parameters` and `cpu_info` classes.
But it resulted in the need of similar changes in the derived classes.

So, the `impl_` pointers types were left unchanged, but their initialization was changed to `new class_name()`.

Still the use of shared pointers (instead of unique) as pointers to implementations looks as not the best design choice.

---

**Testing**

The failures in distributed tests do not look related to the changes.

**Performance**

not applicable

### DIFF
--- a/cpp/oneapi/dal/detail/cpu_info.cpp
+++ b/cpp/oneapi/dal/detail/cpu_info.cpp
@@ -31,11 +31,11 @@ namespace v1 {
 
 cpu_info::cpu_info() {
 #if defined(TARGET_X86_64)
-    impl_ = detail::pimpl<cpu_info_iface>(std::make_unique<cpu_info_x86>());
+    impl_ = detail::pimpl<cpu_info_iface>(new cpu_info_x86);
 #elif defined(TARGET_ARM)
-    impl_ = detail::pimpl<cpu_info_iface>(std::make_unique<cpu_info_arm>());
+    impl_ = detail::pimpl<cpu_info_iface>(new cpu_info_arm);
 #elif defined(TARGET_RISCV64)
-    impl_ = detail::pimpl<cpu_info_iface>(std::make_unique<cpu_info_riscv64>());
+    impl_ = detail::pimpl<cpu_info_iface>(new cpu_info_riscv64);
 #endif
 }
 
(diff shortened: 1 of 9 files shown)


Vika-F:
@david-cortes-intel 
> Thanks, this looks like an improvement, but there might still be something wrong in this same class even after all the changes here - see logs:

Thank you for the diagnostic. For now I do not understand why there is a null pointer as a result of make_shared. I will try to figure it out.

david-cortes-intel:
> @david-cortes-intel
> 
> > Thanks, this looks like an improvement, but there might still be something wrong in this same class even after all the changes here - see logs:
> 
> Thank you for the diagnostic. For now I do not understand why there is a null pointer as a result of make_shared. I will try to figure it out.

Perhaps something to do with casts done in class `base`?

cpp/oneapi/dal/detail/cpu_info.cpp
```diff
@@ -29,24 +29,28 @@
 namespace oneapi::dal::detail {
 namespace v1 {
 
-cpu_info::cpu_info() {
+cpu_info::cpu_info()
+        :
 #if defined(TARGET_X86_64)
-    impl_ = detail::pimpl<cpu_info_iface>(std::make_unique<cpu_info_x86>());
+          impl_(new cpu_info_x86)
```
david-cortes-intel: Would this type of call without parentheses leave the memory uninitialized?

Also from internal discussions it seems that it shouldn't be a problem to break ABI compatibility only around internal-only functions.

Vika-F: In C++ there is no difference between `new T` and `new T()` in case T's default constructor is defined. Anyway I've added the `()` to make the code more understandable.

> Also from internal discussions it seems that it shouldn't be a problem to break ABI compatibility only around internal-only functions.

Just made the ABI checker happier.
Also, as you pointed before, the use of `new` is more compliant with the rest of the codebase.

david-cortes-intel:
@Vika-F I managed to run UBSan in sklearnex now, got this new info which looks very relevant:
```
cpp/daal/src/services/service_topo.cpp:780:45: runtime error: left shift of negative value -1
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior cpp/daal/src/services/service_topo.cpp:780:45 
cpp/daal/src/services/service_topo.cpp:786:50: runtime error: left shift of negative value -1
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior cpp/daal/src/services/service_topo.cpp:786:50
```

PR #3293, opened 2025-07-14
FIX: Avoid logistic loss turning to infinite

david-cortes-intel:
## Description

This PR adds a workaround that prevents the logistic regression loss function from turning into infinite.

Note that this doesn't match with how scikit-learn calculates the loss for high/low values:
https://github.com/scikit-learn/scikit-learn/blob/9b7a86fb6d45905eec7b7afd01d3ae32643c8180/sklearn/_loss/_loss.pyx.tp#L728

But copying their approach would imply switching away from single calls to `vExp` towards doing row-by-row calls to the same function, which could end up making the procedure here slower than sklearn's.

As such, this doesn't make the results from sklearnex match sklearn when calling oneDAL for these calculations, but it at least avoids producing grossly suboptimal results due to optimization ending prematurely from the infinites.

---

**Testing**

### DIFF
--- a/cpp/daal/src/algorithms/objective_function/logistic_loss/logistic_loss_dense_default_batch_impl.i
+++ b/cpp/daal/src/algorithms/objective_function/logistic_loss/logistic_loss_dense_default_batch_impl.i
@@ -20,6 +20,8 @@
 //  Implementation of logloss algorithm
 //--
 */
+#include <type_traits>
+
 #include "src/services/service_data_utils.h"
 #include "src/externals/service_math.h"
 #include "src/services/service_utils.h"
@@ -323,6 +325,20 @@ services::Status LogLossKernel<algorithmFPType, method, cpu>::doCompute(const Nu
             {
                 DAAL_PROFILER_TASK(applyBeta);
                 applyBeta(xLocal, b, fPtrLocal, nRowsToProcess, p, parameter->interceptFlag);
+            }
+
+            // Note: due to the way the function is calculated here, it needs to avoid sigmoid
+            // and '1-sigmoid' values from being exactly zero or exactly one. They could in
+            // theory be calculated to a larger precision by using a different numerical route,
+            // but in order to avoid slow row-by-row exp/log calls, this just clips symmetrically
+            // to the nearest integer value that would avoid running into zeros and ones.
+            const algorithmFPType maxSigmoidInput = std::is_same<algorithmFPType, double>::value ? 35 : 16;
+            const algorithmFPType minSigmoidInput = -maxSigmoidInput;
+
+            PRAGMA_FORCE_SIMD
+            for (size_t row = 0; row < nRowsToProcess; row++)
+            {
+                fPtrLocal[row] = std::fmax(std::fmin(fPtrLocal[row], maxSigmoidInput), minSigmoidInput);
             }
 
             {


david-cortes-intel:
Failures in CI appear to be unrelated to the changes here, and happen in other PRs too.

cpp/daal/src/algorithms/objective_function/logistic_loss/logistic_loss_dense_default_batch_impl.i
```diff
@@ -325,6 +327,20 @@ services::Status LogLossKernel<algorithmFPType, method, cpu>::doCompute(const Nu
...
+            // theory be calculated to a larger precision by using a different numerical route,
+            // but in order to avoid slow row-by-row exp/log calls, this just clips symmetrically
+            // to the nearest integer value that would avoid running into zeros and ones.
+            const algorithmFPType maxSigmoidInput = std::is_same<algorithmFPType, double>::value ? 35 : 16;
+            const algorithmFPType minSigmoidInput = -maxSigmoidInput;
+
+            PRAGMA_FORCE_SIMD
+            for (size_t row = 0; row < nRowsToProcess; row++)
+            {
+                fPtrLocal[row] = std::fmax(std::fmin(fPtrLocal[row], maxSigmoidInput), minSigmoidInput);
```
Vika-F: Please replace std usage here with service_math.h functionality:
https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/src/externals/service_math.h#L45

I understand that it is probably replaced in compile time, but if not - this is a potential cause of illegal instructions.
The same applies to std::is_same.

david-cortes-intel: @Vika-F But this is a .i file, not header. Does it still suffer from the same kind of problems?

david-cortes-intel: @Vika-F I've changed it now to use the functions from service math, but if it's going to be making two function calls per each row, then it sounds like we might was well just use the same non-vectorized computational route that scikit-learn is using (see link at the top).

david-cortes-intel:
@Vika-F Looks like `std::is_same` is being used in other files too, and doesn't seem to cause issues.

Not sure about the math functions from `std::`, but being an `.i` file, I would imagine that there shouldn't be any issue in using them either. Functions `std::max` and `std::min` are used throughout the codebase in `.h`, `.cpp`, and `.i` files and do not seem to cause issues - do you think using those instead of `std::fmax` / `std::fmin` would be good enough?

cpp/daal/src/algorithms/objective_function/logistic_loss/logistic_loss_dense_default_batch_impl.i
Vika-F: @david-cortes-intel all the code that is compiled as part of _cpu.cpp files might suffer from those issues, unfortunately [as shown by recent nightly crashes on SSE4.2].

I hope that those service_math functions will be inlined; another option is to use custom macros, as it is done in some other places in the codebase. But be careful with the naming in order not to clash with Microsoft's MIN/MAX.

david-cortes-intel: @Vika-F Even if those are inlined, it's still a function call to a vector function from MKL with size=1 behind the scenes.

Vika-F:
@david-cortes-intel Ok, I'm just warning. I like to use std:: myself and not to reinvent the wheel.
But I also do not like fixing the illegal instructions issues =\
Probably, in this particular case everything would be inlined, so you could try going with std::.
Maybe someday we could define the scope of "safe" std functionality within our codebase.

cpp/daal/src/algorithms/objective_function/logistic_loss/logistic_loss_dense_default_batch_impl.i
```diff
@@ -325,6 +327,23 @@ services::Status LogLossKernel<algorithmFPType, method, cpu>::doCompute(const Nu
                 applyBeta(xLocal, b, fPtrLocal, nRowsToProcess, p, parameter->interceptFlag);
             }
 
+            // Note: due to the way the function is calculated here, it needs to avoid sigmoid
+            // and '1-sigmoid' values from being exactly zero or exactly one. They could in
+            // theory be calculated to a larger precision by using a different numerical route,
+            // but in order to avoid slow row-by-row exp/log calls, this just clips symmetrically
+            // to the nearest integer value that would avoid running into zeros and ones.
+            const algorithmFPType maxSigmoidInput = std::is_same<algorithmFPType, double>::value ? 35 : 16;
```
icfaust: @Vika-F just wanted your input on using ```std::``` in the .i file here, whether that would cause ISA issues, (sorry for asking about this again since its not concrete in my mind exactly yet).

david-cortes-intel: `std::is_same` is a compile-time templated check. It doesn't link in anything from the standard library.

avolkov-intel: In SYCL implementation we actually clip the outputs of sigmoid, I recall this approach was copied from sklearn, but cannot find the place where it is clipped. https://github.com/uxlfoundation/oneDAL/blob/main/cpp/oneapi/dal/backend/primitives/objective_function/logloss_dpc.cpp#L58
You may want to align the cliping in SYCL and C++ implementation

david-cortes-intel: Modified them to match the DPC++ version.

But by now, scikit-learn doesn't use those thresholds anymore. You can find the numbers and logic they use in the link mentioned in the PR description.

Vika-F: @icfaust @david-cortes-intel I've answered upper in the thread that I am not totally against std:: 
I agree that std::is_same should be inlined here and we can try using it.

Overall, the changes look good to me. I'm just waiting for a green CI.

Vika-F (approved):
Let's have a (quasi-)green CI and LGTM

PR #3296, opened 2025-07-15
Fix illegal instruction on SSE4.2

Vika-F:
## Description

---

**Testing**

**Performance**

### DIFF
--- a/cpp/oneapi/dal/detail/cpu.hpp
+++ b/cpp/oneapi/dal/detail/cpu.hpp
@@ -73,7 +73,7 @@ enum class cpu_feature : uint64_t {
 /// A map of CPU features to their string representations.
 /// This map is used to convert CPU feature bitmasks to human-readable strings.
 /// Keys are bitflags representing CPU features. They are defined in daal::CpuFeature enumeration.
-static const std::map<uint64_t, const std::string> cpu_feature_map = {
+inline const std::map<uint64_t, const std::string> cpu_feature_map = {
     { uint64_t(cpu_feature::unknown), "Unknown" },
 #if defined(TARGET_X86_64)
     { uint64_t(cpu_feature::sstep), "Intel(R) SpeedStep" },


ethanglaser (approved):

cpp/oneapi/dal/detail/cpu.hpp
```diff
@@ -73,7 +73,7 @@ enum class cpu_feature : uint64_t {
 /// A map of CPU features to their string representations.
 /// This map is used to convert CPU feature bitmasks to human-readable strings.
 /// Keys are bitflags representing CPU features. They are defined in daal::CpuFeature enumeration.
-static const std::map<uint64_t, const std::string> cpu_feature_map = {
+inline const std::map<uint64_t, const std::string> cpu_feature_map = {
```
david-cortes-intel, line 76: Could this be moved to the .cpp file instead?

Vika-F: I can do this.
But the positive moment about the current implementation is that `cpu_feature` enum and `cpu_feature_map` are located next to each other and will be modified accordingly (I hope so).
Having them in two different files might lead to inconsistent changes in the future updates.

david-cortes-intel: And does having `inline` actually solve the problem? Otherwise, perhaps another way could be to do away with `std::map` and `std::string` and define them instead as two arrays plus a function for getting the right element.

Vika-F: It should solve the issue, because the problem was that `static const std::map` is compiled multiple times, as part of each .cpp file that includes `cpu.hpp`. Compiling it as part of `_cpu.cpp` files (multiple times for sse4.2, avx2, etc) led to illegal instructions, as AVX2 variant of the code was chosen by the linker.

`inline const std::map` has an external linkage, so it would only be compiled as part of cpu_info_impl.cpp (as it is only used there), only with the default (sse2) compiler flags.

david-cortes-intel (approved):

Vika-F:
CI that was started yesterday to test the solution on SSE4.2:
[CI job](http://intel-ci.intel.com/f0619362-679e-f162-929b-a4bf010d0e2d)

But I couldn't check the results yet.

PR #3300, opened 2025-07-21
Add SLEEF-based SVE-optimized float exponential implementation for postGemmPart function

shubhamsvc:
This pull request adds an SVE-optimized implementation of float exponential function using SLEEF's algorithm and integrates it into the postGemmPart() function of the RBF kernel in HelperKernelRBF<float, sve>.
**Machine**: AWS Graviton3 Processor

**Performance:**
- Avg speedup of postGemmPart() wrt to Scalar code: 2.46 x
- Avg speedup of postGemmPart() wrt to SVE code: 1.92 x

The ULP accuracy of exponential function from SLEEF is 1.


cpp/daal/src/algorithms/kernel_function/kernel_function_rbf_helper.h
```diff
@@ -22,6 +22,11 @@
 #if defined(TARGET_ARM)
     #if (__CPUID__(DAAL_CPU) == __sve__)
         #include <arm_sve.h>
+        #define R_LN2f          1.4426950408889634f
+        #define L2Uf            0.693145751953125f
+        #define L2Lf            1.428606765330187045e-06f
+        #define SLEEF_INFINITYf __builtin_inff()
```
david-cortes-intel: Having 'sleef' in the name is perhaps not the best choice as this is just a regular infinity, and having 'sleef' makes it sound like there might be something else going on. Better to just replace its usage with a direct call to this function.

cpp/daal/src/algorithms/kernel_function/kernel_function_rbf_helper.h
```diff
@@ -22,6 +22,11 @@
 #if defined(TARGET_ARM)
     #if (__CPUID__(DAAL_CPU) == __sve__)
         #include <arm_sve.h>
+        #define R_LN2f          1.4426950408889634f
+        #define L2Uf            0.693145751953125f
```
david-cortes-intel:
```suggestion
        #define L2Uf            0.693145751953125f // log(2)
```

How about putting comments about what these constants are.

david-cortes-intel:
@shubhamsvc Since this is taken from a different library, I guess the terms of their license would require adding their copyright notice to this file:
https://github.com/uxlfoundation/oneDAL/blob/main/third-party-programs.txt

Also, if this function really has 1ULP of precision, might be better to have it as a replacement of the general fp32 scalar exp here:
https://github.com/uxlfoundation/oneDAL/blob/519eafb800200326049786d3631deff5563ed176/cpp/daal/src/externals/service_math_ref.h#L224

cpp/daal/src/algorithms/kernel_function/kernel_function_rbf_helper.h
shubhamsvc: Done

cpp/daal/src/algorithms/kernel_function/kernel_function_rbf_helper.h
shubhamsvc: Done

cpp/daal/src/externals/service_math_ref.h
```diff
@@ -221,8 +229,72 @@ struct RefMath<float, cpu>
         for (SizeType i = 0; i < n; ++i) out[i] = erf(in[i]);
     }
 
+#if defined(TARGET_ARM)
+    #if (__CPUID__(DAAL_CPU) == __sve__)
+    static svfloat32_t exp_sleef(svfloat32_t d)
```
david-cortes-intel: This one could be moved into a helper function outside of the class, so that they would share the same methods across all templates.

Also perhaps the name could be changed to something like 'exp_vectorized' and some comments added there mentioning that it is based on the x86 implementation from the sleef library.

cpp/daal/src/externals/service_math_ref.h
```diff
@@ -221,8 +229,72 @@ struct RefMath<float, cpu>
...
     static void vExp(SizeType n, const float * in, float * out)
     {
+#if defined(TARGET_ARM) && (__CPUID__(DAAL_CPU) == __sve__)
+        // SVE code
+        SizeType i = 0;
+        svbool_t pg;
+
+        for (; i < n; i += svcntw())
+        {
+            pg = svwhilelt_b32(i, n);
```
david-cortes-intel: Is this about the last iterate having potentially less elements than the rest? If so, perhaps it could break down the loop beforehand into 'main' vs. 'last' to avoid checking every time.

shubhamsvc: Done

cpp/daal/src/externals/service_math_ref.h
shubhamsvc: Done

third-party-programs.txt
```diff
@@ -547,4 +547,28 @@ HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
 LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
 OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+=====================================================================================
+6. Boost Software License - Version 1.0 - August 17th, 2003
```
david-cortes-intel: I think this is missing the important part which is the copyright header, as added throughout the sleef library - e.g.
> Copyright Naoki Shibata and contributors 2010 - 2025.

Also elsewhere in this file the title mentions what it applies to, whereas the title here is 'Boost Software License'.

shubhamsvc: Details added as per suggestion. Thank you!

david-cortes-intel:
LGTM. Waiting for some comments from @maria-Petrova or @napetrov about the license handling.

napetrov:
Looks good for me - would wait for @maria-Petrova  final go ahead confirmation here - once this could be integrated without impact on protexIP checks

PR #3303, opened 2025-07-23
MAINT: Avoid instantiation that generates problems under some compilers

david-cortes-intel:
## Description

This instantiation of a map with `const std::string` variables created out of string literals appear to result in memory corruptions under some platforms and compiler options. Note that after this, there would still be other memory corruptions happening, so it's unclear whether this solves an actual issue or just pushes elsewhere in way that evades detection by automated tools.

---

**Testing**

### DIFF
--- a/cpp/oneapi/dal/detail/cpu.hpp
+++ b/cpp/oneapi/dal/detail/cpu.hpp
@@ -73,7 +73,7 @@ enum class cpu_feature : uint64_t {
 /// A map of CPU features to their string representations.
 /// This map is used to convert CPU feature bitmasks to human-readable strings.
 /// Keys are bitflags representing CPU features. They are defined in daal::CpuFeature enumeration.
-inline const std::map<uint64_t, const std::string> cpu_feature_map = {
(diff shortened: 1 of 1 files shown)


icfaust:
i think it would be good be explicit in what tools you were seeing this from in the description in order  to more easily reproduce your steps by future devs

david-cortes-intel:
> i think it would be good be explicit in what tools you were seeing this from in the description in order to more easily reproduce your steps by future devs

The specific issue that this is trying to solve was seen:
* With ICX and CLANG as compilers. Sometimes with GCC too but not deterministically and not entirely sure if the issue was at the exact same section.
* With flags '-O0' or '-Og'.
* On ice lake and lunar lake machines, but not on GNR machines (albeit they do suffer from a similar issue elsewhere that I cannot pinpoint yet).

icfaust:
So it does cause an ABI break, but I think its allowable in this circumstance. Thoughts @Alexandr-Solovev @Vika-F ?

david-cortes-intel:
Closing as this is already included in https://github.com/uxlfoundation/oneDAL/pull/3304

PR #3317, opened 2025-08-06
[Maintenance] track Intel oneAPI apt components via Renovate

icfaust:
## Description

As a way to reduce the maintenance burden of the repository (especially for upcoming releases), adjustments are made to ```.ci/env/apt.sh``` and ```.github/renovate.json``` to track intel-oneapi packages in apt following this guide (https://docs.renovatebot.com/modules/datasource/deb/).  This should lead to all Ubuntu-based public runners not requiring oversight on the latest compiler/MKL/TBB releases. It required the following:

 - Setting the installed packages to be generalized with set major.minor.patch values, rather than the individual package releases
 - Adding renovatebot checks for those major.minor.patch values
 - Setting a renovatebot rule that for major and minor releases the TBB, MKL and DPC++ compilers should be grouped into a single PR, where patch changes are not (which would indicate an important fix that should not break the support of the other 2 packages). This is important to try and alleviate issues of incompatability between MKL/DPC++ or TBB/DPC++ etc.

This accomplishes half of the maintenance of the MKL/DPC++/TBB for public runners not using bazel. For Windows based machines, there is still manual checking required of the intel downloads page for the basekit, which needs to be somehow coded into a script which will check for an endpoint URL for download.

---

**Testing**


.ci/env/apt.sh
```diff
@@ -31,19 +31,19 @@ function add_repo {
 }
 
 function install_dpcpp {
-    sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp-2025.2 intel-oneapi-runtime-libs
+    sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp=2025.2.1-7 intel-oneapi-runtime-libs
```
david-cortes-intel, line 35: Maybe you could add a comment here that these numbers are managed by renovatebot.

.github/renovate.json
ethanglaser, line 1: This seems quite powerful, very cool. @homksei is this something we'd be able to set up in the infra repo? Because we have several hard coded package versions there.

PR #3347, opened 2025-09-03
Fix insufficient memory allocation in csr_table conversion

Vika-F:
## Description

Fix insufficient memory allocation in case 0-based oneDAL csr_table to 1-based DAAL CSRNumericTable conversion.

---

<details>
<summary>Checklist:</summary>

**Testing**

        New test configurations are failing on GPU. Investigation is in progress.

**Performance**

not applicable.

</details>

### DIFF
--- a/cpp/oneapi/dal/algo/basic_statistics/test/batch.cpp
+++ b/cpp/oneapi/dal/algo/basic_statistics/test/batch.cpp
@@ -67,11 +67,14 @@ TEMPLATE_LIST_TEST_M(basic_statistics_batch_test,
                      "[basic_statistics][integration][batch]",
                      basic_statistics_sparse_types) {
     SKIP_IF(this->not_float64_friendly());
-    const auto data = GENERATE_COPY(te::csr_table_builder(5, 5),
-                                    te::csr_table_builder(7, 10),
-                                    te::csr_table_builder(100, 100),
-                                    te::csr_table_builder(1000, 1000),
-                                    te::csr_table_builder(15000, 1000));
+    const float nnz_fraction = 0.05;
+    this->data_indexing_ = GENERATE(sparse_indexing::zero_based, sparse_indexing::one_based);
+    const auto data =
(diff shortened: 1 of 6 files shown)


david-cortes-intel:
The change looks correct to me, but if you put print statements at the end of the method with the right sizes, do the arrays actually have the expected values when you run code that uses it? For example, do the row offsets correspond with the input data?

david-cortes-intel:
@Vika-F Could you perhaps also add a test verifying end results? For example, checking that linear regression produces the exact same coefficients for both inputs. Note that coefficients in linear regression with random data are more variable and higher away from zero when having small data (e.g. 20 rows and 4 columns).

Would also be ideal to have different tests situations with the number of non-zeros being smaller/larger than the number of columns.

Vika-F:
@david-cortes-intel Good idea, but let this be a follow up task, because it will require more code changes, basically I'll need another kind of fixtures to check that 0-bazed and 1-bazed results are equal.

The tests already use different combinations of number of columns to number of non zeros. For example, in basic statistics:

Testcase 0:
Number of columns: 5
Number of non-zeros: 1

Testcase 1:
Number of columns: 10
Number of non-zeros: 3

Testcase 2:
Number of columns: 100
Number of non-zeros: 500

I've also checked the CSR indices after 0-bazed to 1-bazed conversion, they are Ok.

david-cortes-intel:
Failing test looks like it might be taking some CSR as input:
```
//cpp/oneapi/dal/algo/subgraph_isomorphism:test_subgraph_isomorphism_test_host
```

ethanglaser:
> Failing test looks like it might be taking some CSR as input:
> 
> ```
> //cpp/oneapi/dal/algo/subgraph_isomorphism:test_subgraph_isomorphism_test_host
> ```

This test has failed in nightly a couple times this week. No idea where it would be coming from, but fortunately this PR is not causing it.

david-cortes-intel (approved):

Vika-F:
There are also failures in sparse K-means on Xe-LP GPU after adding the new tests.
I'm working on the fix.

PR #3356, opened 2025-09-08
MAINT: Execute debug calculation helpers for random forests at fp64 precision

david-cortes-intel:
## Description

When enabling the debug macro for random forests, the calculations it does at fp32 can get too imprecise, especially if the data has more than 2^23 rows, resulting in some situations where the short calculation is roughly correct but the debug one is not. This PR makes the calculations inside the debug macro always follow double precision.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>


david-cortes-intel:
> I'd leave a comment into the code similar to the description. There is some wonkiness associated with precision and the online computation which you are hitting at that should be highlighted to other developers.

Added a comment.

ethanglaser (approved):
No objections from me

PR #3357, opened 2025-09-08
GitHub: Coverity scan enabling (Lnx)

vyevtyus:
## Description

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,0 +1,112 @@
+#===============================================================================
+# Copyright contributors to the oneDAL project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+name: Coverity Scan
+
+on:
+  schedule:
+    - cron: '0 21 * * *'
+  # allows to reuse this workflow
+  workflow_call:
+
+permissions: read-all
+
+env:
+  COVERITY_PROJECT: uxlfoundation%2FoneDAL
+  COVERITY_PROJECT_ID: 32342
+
+jobs:
+  coverity_linux:
+    name: Coverity Linux
+    if: github.repository == 'uxlfoundation/oneDAL'
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Download Coverity Build Tool (linux64)
+        run: |
+          curl --fail https://scan.coverity.com/download/cxx/linux64 --output ${GITHUB_WORKSPACE}/cov-linux64-tool.tar.gz \
+            --data "token=${{secrets.COVERITY_TOKEN}}&project=${{env.COVERITY_PROJECT}}" || { echo "Download failed"; exit 1; }
+          mkdir cov-linux64-tool
+          tar -xzf cov-linux64-tool.tar.gz --strip 1 -C cov-linux64-tool
+          cd cov-linux64-tool/config
+          git apply --check ${GITHUB_WORKSPACE}/.github/fix.coverity-2024.12.patch
+          patch -p1 < ${GITHUB_WORKSPACE}/.github/fix.coverity-2024.12.patch
+
+      - name: Install DPC++
+        run: .ci/env/apt.sh dpcpp
+
+      - name: Install MKL
+        run: .ci/env/apt.sh mkl
+
+      - name: Make daal (Coverity)
+        run: |
+          source /opt/intel/oneapi/setvars.sh
+          export PATH="${PWD}/cov-linux64-tool/bin:${PATH}"
+          cov-configure --template --compiler icx --comptype intel_icx:linux
+          cov-configure --template --compiler icpx --comptype intel_icpx:linux
+          cov-build --dir cov-int .ci/scripts/build.sh --compiler icx --optimizations avx2 --target daal
+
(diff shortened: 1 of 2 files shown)


napetrov:
for patch -
1. can we contribute it to coverity?
2. if we would still have it - worth dropping it lower level (  .github/patches/coverity/ ?)   and also putting explanatory .md file for why we have it and how to support it going forward?

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,112 @@
...
+
+env:
+  COVERITY_PROJECT: uxlfoundation%2FoneDAL
+  COVERITY_PROJECT_ID: 32342
+
+jobs:
+  coverity_linux:
+    name: Coverity Linux
+    if: github.repository == 'uxlfoundation/oneDAL'
+    runs-on: [ubuntu-latest]
```
icfaust:
```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,112 @@
...
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Download Coverity Build Tool (linux64)
+        run: |
+          curl --fail https://scan.coverity.com/download/cxx/linux64 --output ${GITHUB_WORKSPACE}/cov-linux64-tool.tar.gz \
+            --data "token=${{secrets.COVERITY_TOKEN}}&project=${{env.COVERITY_PROJECT}}" || { echo "Download failed"; exit 1; }
```
icfaust:
```suggestion
            --data "token=${{secrets.COVERITY_TOKEN}}&project=${{env.COVERITY_PROJECT}}" || { echo "::error:: Download failed"; exit 1; }
```
Could you ping a maintainer and have them verify to me that the secret has been properly installed. This small change will make it give a more direct message in github actions.

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,112 @@
...
+          mkdir cov-linux64-tool
+          tar -xzf cov-linux64-tool.tar.gz --strip 1 -C cov-linux64-tool
+          cd cov-linux64-tool/config
+          git apply --check ${GITHUB_WORKSPACE}/.github/fix.coverity-2024.12.patch
```
icfaust: Can you talk about the maintainability of this patch? I expect this will need to be updated on a regular basis, and do we have a contingency plan for testing and verifying changes to the patch?

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,112 @@
...
+
+permissions: read-all
+
+env:
+  COVERITY_PROJECT: uxlfoundation%2FoneDAL
+  COVERITY_PROJECT_ID: 32342
+
+jobs:
+  coverity_linux:
+    name: Coverity Linux
```
icfaust:
```suggestion
    name: Coverity Linux (AVX2)
```
We need to be explicit about what ISA is being used (for clarity).

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,112 @@
...
+      - name: Archive Coverity build results
+        id: check_size
+        run: |
+          tar -czvf cov-int.tgz cov-int
+          size=$(du -m cov-int.tgz | cut -f1)
+          echo "Artifact size: $size MB"
+          echo "size=$size" >> $GITHUB_OUTPUT
+
+      - name: Submit Coverity results for analysis
+        #* if condition is related to Coverity limitation for archives that exceed 500MB
```
icfaust: So what is the downside of using the >500 MB branch for all sizes?

.github/workflows/coverity.yml
vyevtyus: I suppose @homksei could confirm that

.github/workflows/coverity.yml
vyevtyus: There is no problem with using only one variant. Modified.

.github/workflows/coverity.yml
vyevtyus: Taking into account low frequency of client versions update and almost zero difference between compiler configs between versions (according to our experience), I would say that we'll probably need to review the patch relevance once every 6-12 month. As for plan, currently there is no strategy on maintainability and it worth to be discussed. We may have patch application optional - in this case it's irrelevance won't broke the scan in case of tool client version update. Besides, the newer client version we use, the better Intel compiler support is implemented in it (theoretically).

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,99 @@
...
+# limitations under the License.
+#===============================================================================
+
+name: Coverity Scan
+
+on:
+  schedule:
+    - cron: '0 21 * * *'
+  # allows to reuse this workflow
+  workflow_call:
```
icfaust, line 23: Using ```workflow_call``` is the recommended way to trigger this CI step for validating updates to coverity? Still trying to figure out how we will validate new patches.

vyevtyus: `workflow_call` is not a way to trigger the scan. It provides a possibility to include this scan as step into CI/Nightly if needed. If there are no such plans at the moment, we can remove this.

If someone needs or wants to test a new patch, it's enough to create a fork, make necessary changes to the patch, disable results uploading stage, and run a scan for that fork. After the build step completes, the percentage of captured compilation units, displayed at the end of the log, will indicate whether the new patch is good enough.

icfaust (changes requested):
I'd like to have a guideline added to the markdown how to approve new patches and changes to this pipeline

.github/workflows/coverity.yml
icfaust: I think such a pipeline could be added to this PR which checks for changes in the patch file and runs a specially triggered version of the current pipeline which doesn't upload (by additional logic checks in the current PR).

vyevtyus: Potential solution is easier. Upload will not start if branch name is not 'main' - I've added an appropriate check into uploading step.

vyevtyus:
> @vyevtyus any updates here?

.md file has been updated by adding the information requested. Copyright-related questions was discussed internally as well - no additional changes are required. Let me know if you have further questions.

Vika-F (approved):

PR #3360, opened 2025-09-11
FIX: Use float64 for random forest calculations

david-cortes-intel:
## Description

This PR modifies calculations of means / variances / counts / gini in random forests for both regression and classification to use exclusively float64 type (feature borders an values are still taken with the float width they come). They are set as a separate type in line with the current logic in regression forests, so that it'd be easier to change them back to float32 if needed once we have some mechanism to detect when it wouldn't harm quality of results.

Motivation: when the data contains more than a handful million rows, calculations about these quantities done in float32 can get imprecise, since integers are not representable exactly, and these impressions accumulate as trees get deeper, to the point that leafs down a regression tree can end up computing means that are outside of the range of the 'y' variable that was passed, potentially even with different orders of magnitude, which makes the resulting models perform worse than random guessing.

Note that this comes with a performance impact even for float64 data, as currently some calculations for the classification variant are done in float32 regardless (which is currently also a problem if there's many millions of rows).

Picture before/after showing the kind of issue that it intends to fix:
image

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/classification/df_classification_train_dense_default_impl.i
@@ -1332,8 +1332,9 @@ int UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitFewClasses(int nDi
         for (size_t iClass = 0; iClass < K; ++iClass) leftWeights += histLeft[iClass];
     }
 
-    if (!(((n - nLeft) < nMinSplitPart) || ((totalWeights - leftWeights) < minWeightLeaf) || (nLeft < nMinSplitPart) || (leftWeights < minWeightLeaf))
-        && leftWeights)
+    const intermSummFPType rightWeights = totalWeights - leftWeights;
+    if (!(((n - nLeft) < nMinSplitPart) || (rightWeights < minWeightLeaf) || (nLeft < nMinSplitPart) || (leftWeights < minWeightLeaf)) && leftWeights
+        && rightWeights > 0)
     {
         auto histTotal            = curImpurity.hist.get();
         intermSummFPType sumLeft  = 0;
@@ -1346,7 +1347,7 @@ int UnorderedRespHelperRandom<algorithmFPType, cpu>::findSplitFewClasses(int nDi
             sumRight += (histTotal[iClass] - histLeft[iClass]) * (histTotal[iClass] - histLeft[iClass]);
         }
 
-        const intermSummFPType decrease = sumLeft / leftWeights + sumRight / (totalWeights - leftWeights);
+        const intermSummFPType decrease = sumLeft / leftWeights + sumRight / rightWeights;
         if (decrease > bestImpDecrease)
         {
             split.left.hist     = this->_histLeft;
@@ -1692,10 +1693,11 @@ Status TreeThreadCtx<algorithmFPType, cpu>::finalizeOOBError(const NumericTable 
         }
         if (resPerObs) resPerObs[i] = algorithmFPType(maxIdx != classLabel);
     });
-    if (res) *res = nPredicted.get() ? algorithmFPType(nError.get()) / static_cast<intermSummFPType>(nPredicted.get()) : 0;
+    if (res) *res = nPredicted.get() ? static_cast<intermSummFPType>(nError.get()) / static_cast<intermSummFPType>(nPredicted.get()) : 0;
     if (resAccuracy)
-        *resAccuracy = nPredicted.get() ? algorithmFPType(1) - algorithmFPType(nError.get()) / static_cast<intermSummFPType>(nPredicted.get()) :
-                                          algorithmFPType(1);
+        *resAccuracy = nPredicted.get() ?
+                           intermSummFPType(1) - static_cast<intermSummFPType>(nError.get()) / static_cast<intermSummFPType>(nPredicted.get()) :
+                           intermSummFPType(1);
     return Status();
 }
 
--- a/cpp/daal/src/algorithms/dtrees/forest/regression/df_regression_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/regression/df_regression_train_dense_default_impl.i
@@ -724,8 +724,7 @@ void RespHelperBase<algorithmFPType, cpu, crtp>::finalizeBestSplit(const IndexTy
     }
     else
     {
-        divL =
-            isZero<intermSummFPType, cpu>(bestSplit.leftWeights) ? intermSummFPType(1) : (1.0 / static_cast<intermSummFPType>(bestSplit.leftWeights));
+        divL = isZero<intermSummFPType, cpu>(bestSplit.leftWeights) ? intermSummFPType(1) : (1.0 / bestSplit.leftWeights);
 
         bestSplit.left.mean *= divL;
         bestSplit.left.var            = 0;


cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -1153,7 +1162,7 @@ NodeSplitResult TrainBatchTaskBase<algorithmFPType, BinIndexType, DataHelper, Hy
     /* total number of features */
     const size_t maxFeatures = nFeatures();
     /* minimum fraction of all samples per bin */
-    const algorithmFPType qMax = 0.02;
+    const intermSummFPType qMax = 0.02;
```
icfaust, line 1165: Why must this be changed?

david-cortes-intel: Because it's used in operations with data of `intermSummFPType` type, so it would need to get casted either way when it gets used.

icfaust: Please link to locations where qMax is used beyond line 1213

david-cortes-intel: > Please link to locations where qMax is used beyond line 1213

`qMax` is used in line 1213 in operations with `intermSummFPType` type. So if was a float32 and `intermSummFPType` is float64, it would get casted to float64.

icfaust: All the other aspects of that computation are integer type. If you wanted to you could turn that hard coded value into a pure integer computation if you really wanted to (given the value of qMax). I guess I was trying to figure out what the key to that specific change was, and why it was done.

icfaust: So to be clear I am okay with the change, just wondering if we should comment it in some way.

david-cortes-intel: Yes, it could be put down to integer types if you keep only values that could be expressed as `1/x`. But this operation is done only once so there's very little to gain out of it.

david-cortes-intel: > So to be clear I am okay with the change, just wondering if we should comment it in some way.

I don't think it warrants a comment. The variable is only used once and it's easy to see how from the code.

icfaust: So then why was it changed?

david-cortes-intel: Because otherwise it would get casted. Why would you want to define a variable in one type if then it would get promoted to a different type at the moment it gets used?

david-cortes-intel:
Attaching performance comparisons before/after on selected servers:
[report_rf64_gnr.xlsx](https://github.com/user-attachments/files/22643107/report_rf64_gnr.xlsx)
[report_rf64_emr.xlsx](https://github.com/user-attachments/files/22643112/report_rf64_emr.xlsx)

They show slight accuracy gains in this branch at the expense of slightly timing degradations as expected, but it looks like the variances are very high and numbers sometimes show either branch being faster, so I wouldn't put much trust in the stability of those aggregates.

The changes are mostly in random forest classifier as expected, which switches to float64 here.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
icfaust: I do not understand your argument @david-cortes-intel , the only use of `qMax` is the comparison ```(static_cast<intermSummFPType>(n) > qMax * static_cast<intermSummFPType>(_helper.indexedFeatures().numIndices(iFeature))``` where `qMax` times the number of indices can not exceed the value of a float32 (even if 64 bit size_t).  At worst the integer values would be set at float32 and the result would stay as such.

david-cortes-intel: The line seen in terms of data types is as follows:
```
bool result = bool && double > qMax * double
```

If `qMax` is a `float`, it will get casted to `double` to calculate the RHS. If it's a `double`, it will be used as-is.

icfaust: The previous logic converted the two ints to `AlgorithmFPType` implicitly and now you are explicitly casting them to `double` using `static_cast` explicitly, the total number of casts haven't changed.  Again, I still don't see the logic here.

EDIT: I see now that it was previously explicitly casted to `AlgorithmFPType`

david-cortes-intel: image

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -669,7 +677,8 @@ protected:
     services::internal::HostAppHelper _hostApp;
     typename DataHelper::TreeType _tree;
     mutable TVector<IndexType, cpu> _aSample;
-    mutable TArray<algorithmFPTypeArray, cpu> _aFeatureBuf;
+    mutable TArray<intermSummFPTypeArray, cpu> _aFeatureBuf;
+    mutable TArray<algorithmFPType, cpu> _BufFPType;
```
Vika-F: Can you please comment what is the purpose of this array?

Vika-F: This variant is more aligned with the rest of the file:
```suggestion
    mutable TArray<algorithmFPType, cpu> _bufFPType;
```

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -580,25 +583,30 @@ protected:
...
     typename DataHelper::NodeType::Base * buildNode(const size_t level, const size_t nClasses, size_t & remainingSplitNodes, WorkItem & item,
                                                     typename DataHelper::ImpurityData & impurity);
 
-    algorithmFPType * featureBuf(size_t iBuf) const
+    intermSummFPType * featureBuf(size_t iBuf) const
     {
         DAAL_ASSERT(iBuf < _nFeatureBufs);
         return _aFeatureBuf[iBuf].get();
     }
+    algorithmFPType * featureBuf_FPType() const
```
Vika-F: The mix of camel case and snake case looks strange.
It would be also nice to provide the description to `featureBuf` and this function, because they look similar, but do not so similar things.
```suggestion
    algorithmFPType * featureBufFPType() const
```

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
david-cortes-intel: In the main branch, one buffer was being reused for two different purposes where a buffer of the same length was needed. With this PR, that cannot be done anymore, because the two purposes would potentially require different data types for the buffer, so this adds another buffer object with a different type from the other.

david-cortes-intel: Renamed.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
david-cortes-intel: Renamed.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
icfaust: Just as a heads up @david-cortes-intel @Vika-F I had at some point figured out that whoever first wrote it used https://en.wikipedia.org/wiki/Hungarian_notation . Some of the variables are actually opposite to what is expected (via bad translation? no idea). So I'd stick to it if you wanted to be consistent, but I am not a stickler.

PR #3395, opened 2025-10-09
[maintenance] Remove Python dependence from oneDAL building

icfaust:
## Description

Python is only used to generate the package config file for dal shared and static libraries.  The python script is only doing string manipulation and concatenation based on the hardware and operating system of the build.  The logic can be replicated by using macros on a C++ file, where the OS and hardware can be determined for all the supported compilers, hardware and operating systems. This has the added benefit that the logic can be introduced to bazel building as the underlying dependencies are already satisfied.

This also solves an issue for cross-compiled .pc files, which by the nature of the python file were reporting the wrong hardware.  By having this information defined by the compiler, it will match the compiled oneDAL build.

It also adds the proper logic for RISC-V that was missing in setting up destination folders.

This should expand support of oneDAL to better out-of-the-box support in older Linux versions which were limited by the Python3.9 requirement. For example, this should be able to be installed standard g++ on Ubuntu20 or Ubuntu18 without much difficulty.  This should also minimize maintenance, as it is one less dependency to track, especially one that will change like Python.

Following changes have been made:
 - Delete the pkg-config.tpl file
 - Delete the generate_pkgconfig.py file
 - Add a cpp-like file template with macros for pc file generation
 - Adding logic to the makefile to postprocess the preproccessed output into a valid .pc file
 - Remove or shift python installs in various CI runners (sometimes required to get MPI libraries)
 - Remove some mentions of Python from INSTALL.md documentation
 - changed `dev-base` not to include conda

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/makefile
+++ b/makefile
@@ -1025,8 +1025,8 @@ endif
 _release_c: ./deploy/pkg-config/pkg-config.cpp
 	mkdir -p $(RELEASEDIR.pkgconfig)
 # use the compiler's preprocessor to define the pkg-config file as it can handle cross-compilation, OS and ISA determination for all OSes.
-	$(COMPILER.$(_OS).$(COMPILER)) -E -DSTATIC ./deploy/pkg-config/pkg-config.cpp $(if $(COMPILER_is_vc),>,-o) $(WORKDIR.lib)/dal-static-threading-host.pc
-	$(COMPILER.$(_OS).$(COMPILER)) -E ./deploy/pkg-config/pkg-config.cpp $(if $(COMPILER_is_vc),>,-o) $(WORKDIR.lib)/dal-dynamic-threading-host.pc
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -123,6 +116,13 @@ jobs:
         run: |
             source /opt/intel/oneapi/setvars.sh
             .ci/scripts/test.sh --test-kind examples --build-dir __release_lnx --compiler icx --interface oneapi/dpc --build-system cmake
+      - name: Install Python
```
david-cortes-intel: What would this one be used for if the python commands are being removed elsewhere?

icfaust: I thought it was necessary to have a ground python install to get miniconda to start, but seems like that may not be the case.
We need conda for the MPI install (Used to get MPI: https://github.com/icfaust/oneDAL/blob/dev/remove_python/.ci/env/environment.yml). I'll try it without once the CI systems run through once.

david-cortes-intel: Miniconda should be able to install and run without a base python. At least the mamba parts.

icfaust: done

PR #3400, opened 2025-10-15
RFC: oneDAL 2026.0 algorithms deprecations

Vika-F:
## Description

This RFC proposes to deprecate select algorithms available in the DAAL as part of the ongoing transition to the modern oneDAL API.

A rendered variant of this RFC: https://github.com/Vika-F/daal/blob/dev/rfcs/algo_removal/rfcs/20251015-DAAL-algorithms-deprecation/README.md

PR #3408, opened 2025-10-22
RFC: oneDAL 2026.0: DAAL API layers of certain algorithms deprecations 

Vika-F:
## Description

This RFC proposes to deprecate API layers of the select algorithms available in the DAAL as part of the ongoing transition to the modern oneDAL API.

In contrast to PR #3400, this RFC proposes removing only the API layers while preserving all underlying
algorithmic computational kernels.

A rendered variant of this RFC: https://github.com/Vika-F/daal/blob/dev/interfaces_removal/rfcs/20251022-DAAL-interfaces-deprecation/README.md

---

<details>
<summary>Checklist:</summary>

**Testing**

not applicable

**Performance**

not applicable

</details>

### DIFF
--- a/rfcs/20251022-DAAL-interfaces-deprecation/README.md
+++ b/rfcs/20251022-DAAL-interfaces-deprecation/README.md
@@ -68,7 +68,7 @@ with complete API removal scheduled for the next major release, oneDAL 2026.0:
 - [kernel_function::rbf::Batch](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kernel_function/kernel_function_rbf.h#L91)
 - [kernel_function::rbf::Input](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kernel_function/kernel_function_types_rbf.h#L82)
 
-#### CPU Features Dispatching(https://uxlfoundation.github.io/oneDAL/contribution/cpu_features.html#cpu-features-dispatching)
+#### [CPU Features Dispatching](https://uxlfoundation.github.io/oneDAL/contribution/cpu_features.html#cpu-features-dispatching)
 
 - [CpuType](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/services/cpu_type.h#L27)
 - [CpuFeature](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/services/cpu_type.h#L52)


rfcs/20251022-DAAL-interfaces-deprecation/README.md
```diff
@@ -0,0 +1,85 @@
...
+- [kmeans::BatchContainer](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kmeans/kmeans_batch.h#L56)
+- [kmeans::Batch](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kmeans/kmeans_batch.h#L87)
+- [kmeans::init::BatchContainer](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kmeans/kmeans_init_batch.h#L58)
+- [kmeans::init::Batch](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kmeans/kmeans_init_batch.h#L109)
+
+#### [Linear and Ridge Regression](https://uxlfoundation.github.io/oneDAL/daal/algorithms/linear_ridge_regression/index.html)
+
+- [linear_regression::prediction::Batch](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/linear_regression/linear_regression_predict.h#L88)
+- [linear_regression::training::BatchContainer](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/linear_regression/linear_regression_training_batch.h#L56)
+- [linear_regression::training::Batch](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/linear_regression/linear_regression_training_batch.h#L93)
```
david-cortes-intel: @Vika-F Earlier comment is still relevant here: how would daal4py work if this gets removed but there's still a streaming interface, for example?

Vika-F: @david-cortes-intel Yes, removing of the Batch APIs might impact daal4py streaming APIs because in the Python side Batch and Online are attached to the same API, just with different arguments.

But in case of the linear & ridge regression we already have alternatives in sklearnex: IncrementalLinearRegression and IncrementalRidge, so we can suggest to use those instead.

I think I have to check how the removal of Batch API impacts daal4py streaming and distributed APIs.

david-cortes-intel: Does it mean that then it will only be possible to call the daal4py class with `distributed=True`? It sounds like it'd require a lot of modifications in the daal4py code generator.

Vika-F: @david-cortes-intel I've checked the possibility of generating daal4py API for an algorithm that does not have Batch API in DAAL. It looks there are no easy ways to do that, it requires significant changes (I've tried insignificant changes, but that didn't help) in all those jinja2 generators, as those expect that all the algorithms provide Batch API, without exceptions.

It looks like there are 2 options:

1) close this RFC, and exclude DAAL APIs from the list of deprecations for 2026.0 release.
2) provide sklearnex APIs for distributed algorithms on CPU, and then deprecate Batch, Online, Distributed and daal4py APIs for the algorithms that have respective sklearnex APIs. The initial works have been started here: https://github.com/uxlfoundation/oneDAL/pull/3420. This might be feasible for algorithms with rather simple distributed implementations on CPU (K-means, LinReg, PCA). For DBSCAN and kNN it is doubtful due to the implementation complexity.

david-cortes-intel: If we're 100% sure that nothing new will be added to the DAAL interface, then there's also a 3rd option of making some classes be defined from manual cython code outside of the generators. But it'd also be quite complicated to execute.

And in option 1), it'd either way make sense to mark them all as deprecated, even if they aren't removed in the next release.

rfcs/20251022-DAAL-interfaces-deprecation/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+- [kernel_function::linear::BatchContainer](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kernel_function/kernel_function_linear.h#L58)
+- [kernel_function::linear::Batch](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kernel_function/kernel_function_linear.h#L92)
+- [kernel_function::linear::Input](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kernel_function/kernel_function_types_linear.h#L83)
+
+- [kernel_function::rbf::BatchContainer](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kernel_function/kernel_function_rbf.h#L57)
+- [kernel_function::rbf::Batch](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kernel_function/kernel_function_rbf.h#L91)
+- [kernel_function::rbf::Input](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/algorithms/kernel_function/kernel_function_types_rbf.h#L82)
+
+#### Other APIs
```
david-cortes-intel: What about the 'optmization solver' families? Are they already covered by another RFC?

Vika-F: AdaGrad, SGD and SAGA are covered by another RFC (https://github.com/uxlfoundation/oneDAL/tree/rfcs/rfcs/20251015-DAAL-algorithms-deprecation).
Coordinate descent and LBFGS are left unchanged for now, as those are touched by sklex.

theComputeKid:
No objections - though will defer to @rakshithgb-fujitsu from the Arm side due to scope of changes.

PR #3434, opened 2025-11-07
FIX: Fix undefined behavior warning with GCC

david-cortes-intel:
## Description

Fixes a warning from GCC about potential undefined behavior, which cannot be encountered either way but lacks hints for it.

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/forest/regression/df_regression_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/regression/df_regression_train_dense_default_impl.i
@@ -219,6 +219,12 @@ void OrderedRespHelperBest<algorithmFPType, cpu>::calcImpurity(const IndexType *
             {
                 const size_t iStart = iMain * simdBatchSize;
 #if defined(__GNUC__) && !defined(__clang__)
+                // Note: GCC is unable to determine that the pointer offset plus the
+                // vectorized section will be within the limits of pointer ranges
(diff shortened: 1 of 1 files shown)


cpp/daal/src/algorithms/dtrees/forest/regression/df_regression_train_dense_default_impl.i
```diff
@@ -216,7 +217,10 @@ void OrderedRespHelperBest<algorithmFPType, cpu>::calcImpurity(const IndexType *
 
             for (size_t iMain = 0; iMain < itersSimdLoop; iMain++)
             {
-                const size_t iStart         = iMain * simdBatchSize;
+                const size_t iStart = iMain * simdBatchSize;
+#if defined(__GNUC__) && !defined(__clang__)
```
Vika-F, line 221: Please add some comments here. For example:
```suggestion
#if defined(__GNUC__) && !defined(__clang__)
// g++ compiler cannot vectorize the further loops without this attribute
// due to -Werror=aggressive-loop-optimizations, as since GCC 4.8
// "GCC now uses a more aggressive analysis to derive an upper bound
// for the number of iterations of loops using constraints imposed by language standards"
```

david-cortes-intel: Added a comment.

PR #3437, opened 2025-11-17
FIX: DAAL KDtree KNN multi-thread bug fix

DDJHB:
# Problem Description
This PR addresses two critical Use-After-Free (UAF) vulnerabilities identified in the parallel construction phase of the KD-Tree (buildSecondPartOfKDTree).

Both issues arise from pointer invalidation during array reallocation. The algorithm uses raw pointers (bboxCur and curNode) to reference elements inside dynamic arrays (local->bboxes and local->extraKDTreeNodes). When these arrays grow and are reallocated, the raw pointers are left dangling, pointing to freed memory.

1. bboxCur Use-After-Free
Trigger: Occurs when the recursion stack grows deep enough to trigger a resize of the local->bboxes array.
Details: bboxCur is initialized to point to the current node's bounding box. If pushing children onto the stack causes local->bboxes to reallocate, bboxCur becomes a dangling pointer. The subsequent call to copyBBox reads from this freed memory.
```c++
// 1. bboxCur points to local->bboxes
bboxCur = &(local->bboxes[local->bboxPos * xColumnCount]);

// ... (pushing children increments bboxPos) ...

if (local->bboxPos >= local->bboxesCapacity) {
    BBox * const oldBboxes = local->bboxes;
    // 2. Reallocation happens here
    service_scalable_free<BBox, cpu>(oldBboxes);
    // bboxCur is now dangling!
}

// 3. Crash/Corruption: Reading from freed memory
this->copyBBox(bboxLeft, bboxCur, xColumnCount);
```

2. curNode Use-After-Free
Trigger: Occurs when a single thread processes significantly more nodes than its pre-allocated slice, forcing usage of the extraKDTreeNodes overflow buffer.
Details: curNode may point into local->extraKDTreeNodes. If this array fills up and reallocates, curNode is not updated and points to freed memory. Dereferencing it to access rightIndex causes a crash.
```c++
// 1. curNode points into extraKDTreeNodes
curNode = &(local->extraKDTreeNodes[bn.nodePos - firstExtraNodeIndex]);

// ...

if (extraIndex >= local->extraKDTreeNodesCapacity) {
    KDTreeNode * oldNodes = local->extraKDTreeNodes;
    // 2. Reallocation happens here
    daal_free(oldNodes);
    // curNode is now dangling!
}

// 3. Crash: Accessing member of freed object
bnRight.nodePos = curNode->rightIndex;
```
The trigger for this bug is highly data‑ and workload‑dependent, which makes failures sporadic and hard to reproduce. Both UAFs occur only when internal buffers (bboxes or extraKDTreeNodes) need to grow: for bboxCur this happens on unusually deep recursion paths in the KD‑tree, and for curNode it happens when a particular thread exhausts its pre‑allocated node range and starts using the per‑thread overflow buffer.

Significant performance improvement is expected with this change due to the enablement of multiple threads in a compute costly segment of training and has been observed (around ~2.5x) while measuring with scikit-learn_bench.

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/cpp/oneapi/dal/algo/knn/test/batch.cpp
+++ b/cpp/oneapi/dal/algo/knn/test/batch.cpp
@@ -69,9 +69,9 @@ class knn_batch_test : public knn_test<TestType, knn_batch_test<TestType>> {};
                          "[small-dataset][knn][integration][batch][test]", \
                          knn_cls_types)
 
-#define KNN_KDTREE_SINGLE_RUN_TEST(name)                                       \
-    TEMPLATE_LIST_TEST_M(knn_batch_test,                                       \
-                         name,                                                 \
+#define KNN_KDTREE_SINGLE_RUN_TEST(name)                                                \
+    TEMPLATE_LIST_TEST_M(knn_batch_test,                                                \
+                         name,                                                          \
                          "[synthetic-dataset][knn][integration][batch][test][nightly]", \
                          knn_kdtree_cls_float_only)
 


cpp/oneapi/dal/algo/knn/test/batch.cpp
```diff
@@ -191,6 +197,35 @@ KNN_CLS_SYNTHETIC_TEST("knn nearest points test random uniform 16390x20x5") {
     this->exact_nearest_indices_check(x_train_table, x_infer_table, infer_result);
 }
 
+KNN_KDTREE_SINGLE_RUN_TEST("knn nearest points test random uniform 1'000'000x20x10") {
+    // this test triggers issue with KD-Tree training on large datasets
+    // that leads to crash due to use-after-free in multithreaded execution
+    SKIP_IF(this->not_available_on_device());
+    SKIP_IF(this->not_float64_friendly());
+
+    constexpr std::int64_t train_row_count = 1'000'000;
```
david-cortes-intel, line 206: Does it need to be this large? This adds up to the running times which are already long.

Alexandr-Solovev: As far as I know, its the only way to reproduce the issue right now, and this test is enabled only for "[synthetic-dataset][knn][integration][batch][test]**[nightly]**", so does not affect running times for most of the jobs

cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
Vika-F, line 1149: Is my understanding correct that `curNode` can also be invalidated here, as the `local->extraKDTreeNodes` change here?

Wouldn't it be better to make `curNode` resetting after the `if (local->extraKDTreeNodes)`?

cpp/oneapi/dal/algo/knn/test/batch.cpp
DDJHB: @david-cortes-intel yeah, only way afaik to reliably trigger the seg fault here without fixes. it is limited to run only once and only during nightly (as @Alexandr-Solovev mentioned), so should not affect manual job invocations.

cpp/daal/src/algorithms/k_nearest_neighbors/kdtree_knn_classification_train_dense_default_impl.i
DDJHB: @Vika-F The conditions for this else clause to be entered are: 1) `local->nodeIndex >= firstExtraNodeIndex` (we need extra nodes) and 2) `local->extraKDTreeNodes` is `nullptr` (we never allocated extra nodes before). Since curNode is initialized as:
```c++
curNode = (bn.nodePos < firstExtraNodeIndex) ? static_cast<KDTreeNode *>(kdTreeTable.getArray()) + bn.nodePos :
                                                                       &(local->extraKDTreeNodes[bn.nodePos - firstExtraNodeIndex]);
```
, it could not be pointing into `extraKDTreeNodes` (by design) yet. This `else` clause carries the mission to allocate memory for future nodes i.e. children of `curNode` since we increment `local->nodeIndex` when editing `curNode`.
```c++
curNode->leftIndex  = (local->nodeIndex)++;
curNode->rightIndex = (local->nodeIndex)++;
if (local->nodeIndex >= firstExtraNodeIndex) {
    if (...) { 
        // re-allocation 
    }
    else {
        // initialization
    }
}
```
It is technically redundant for the else clause if we move the fix outside the if/else block, so i kept it in to group with the source of the issue.

ethanglaser:
Nightly job with this branch: http://intel-ci.intel.com/f0d1688a-17ba-f103-8b28-a4bf010d0e2d

Vika-F (approved):

ethanglaser:
Observation: this `[nightly]` option functionally does nothing because we do not run this configuration anywhere - neither in our infra for private CI nor anywhere in oneDAL itself

ethanglaser:
If we want the test to run we should coordinate with @homksei to run bazel test with --test_nightly flag enabled in our nightly CI

PR #3440, opened 2025-11-18
Fix high and low priority Coverity hits

Vika-F:
## Description

Integer overflow checks were added to the places where potential overflow can happen, according to Coverity.

Destructors were added to fix rule-of-three violations in `AnalysisContainerIface`, `mrg32k3a::Batch`, `philox4x32x10::Batch`, `kernel_function::polynomial::internal::Batch`.

Assignment operators were fixed to prevent self-assignment in `cross_entopy_loss::Parameter`, `logistic_loss::Parameter`, `mse::Parameter`, `objective_function::Parameter`.

ABI checks were improved to suppress the checking of internal ABI and the ABI from `dal::*::preview` and `dal::*::detail` namespaces.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

not applicable

</details>


cpp/daal/src/services/service_topo.cpp
```diff
@@ -48,6 +48,12 @@ FILE * stderr = __stderrp;
     #define _INTERNAL_DAAL_MEMSET(a1, a2, a3)     __internal_daal_memset((a1), (a2), (a3))
     #define _INTERNAL_DAAL_MEMCPY(a1, a2, a3, a4) daal::services::internal::daal_memcpy_s((a1), (a2), (a3), (a4))
 
+    #define _INTERNAL_DAAL_OVERFLOW_CHECK_BY_ADDING(type, op1, op2)            \
+        {                                                                      \
+            volatile type r = (op1) + (op2);                                   \
+            r -= (op1);                                                        \
```
david-cortes-intel, line 54: Since integer overflow is undefined behavior in C++, the compiler would be free to optimize away this. It'd have to be checked before summing them (e.g. UINT_MAX - op2 <= op1).

Although the only case where this is used is for the core counts, which will not be reaching anywhere close to 2^32, so perhaps it could just be marked as a false positive.

Vika-F: The usage of the macro was incorrect, the wider data type must be used. I replaced `unsigned int` with `unsigned __int64`, it should work correctly after the change.
Also the `r` variable shouldn't be optimized out as it is volatile.

This part is not performance critical, so having one more checks would not be a problem here. I am mostly concerned not with the real overflow, but with potential reading of some garbage from the memory.

.ci/scripts/abi_check.sh
```diff
@@ -35,7 +35,7 @@ for i in "${solibs[@]}"
 do
     name=$(basename $i)
     echo "======== ${name} ========"
-    abidiff $i $release_dir/$name
+    abidiff --suppr .github/.abignore $i $release_dir/$name
```
david-cortes-intel, line 38: Does this mean that the current file `.github/.abiginore` is not actually being used?

cpp/daal/src/services/service_topo.cpp
```diff
@@ -217,7 +223,13 @@ unsigned int _internal_daal_GetMaxCPUSupportedByOS()
         glbl_obj.error |= _MSGTYP_UNKNOWNERR_OS;
         return 0;
     }
-    for (unsigned int i = 0; i < grpCnt; i++) lcl_OSProcessorCount += pSystem_rel_info->Group.GroupInfo[i].ActiveProcessorCount;
+    for (unsigned int i = 0; i < grpCnt; i++)
+    {
+        const auto activeProcCount = pSystem_rel_info->Group.GroupInfo[i].ActiveProcessorCount;
+        _INTERNAL_DAAL_OVERFLOW_CHECK_BY_ADDING(unsigned __int64, lcl_OSProcessorCount, activeProcCount);
```
david-cortes-intel, line 229: Why not `std::uint64_t`?

david-cortes-intel, line 229: Would coverity still flag the operation if using uint64?

cpp/daal/src/services/service_topo.cpp
david-cortes-intel: But then, does it need `volatile`?

cpp/daal/src/services/service_topo.cpp
Vika-F: It would require an additional include - a more invasive change.

.ci/scripts/abi_check.sh
Vika-F: Yes.

cpp/daal/src/services/service_topo.cpp
Vika-F: Is it possible to check this before the commit?
I thought it can be only seen after a nightly Coverity run. I.e. I am trying to fix the hits in this PR, but I will know whether they are actually fixed or not only after the merge.

david-cortes-intel: I don't know if it's possible to trigger it for a PR. @ethanglaser Any comments here?

david-cortes-intel: Since this is C++17, isn't it just a one-liner `#include <cstdint>`? Is `__int64` available on all platforms?

cpp/daal/src/services/service_topo.cpp
Vika-F: I think yes, because an optimizing compiler might think like: 
    line 53: r <- op1 + op2
    line 54: r <- (op1 + op2) - op1 == op2
    line 55: if (!(r == op2)) is the same as if(false) ==> let's optimize this out

volatile should prevent this kind of optimizations.

david-cortes-intel: But the types are different here. `r` is 64-bit, while `op1` and `op2` are 32-bit, so `r` cannot overflow.

Vika-F: Yes, but `op1 + op2` can overflow, making the value assigned to `r` incorrect in some sense.
I.e. in the cases of overflow `op1 + op2 != type(op1) + type(op2)`, because the left one overflows.
This is the whole idea of the check.

cpp/daal/src/services/service_topo.cpp
ethanglaser: I don't think it can be checked on a branch. Quite unfortunate and I still don't really understand why.

PR #3441, opened 2025-11-19
Fix: vectorization pragmas issues in debug (REQDBG=yes) jobs

ethanglaser:
## Description

---

<details>
<summary>Checklist:</summary>

**Testing**

</details>

### DIFF
--- a/cpp/daal/src/services/service_defines.h
+++ b/cpp/daal/src/services/service_defines.h
@@ -54,7 +54,7 @@ DAAL_EXPORT bool daal_check_is_intel_cpu();
     #define PRAGMA_NOVECTOR         _Pragma("novector")
     #define PRAGMA_VECTOR_UNALIGNED _Pragma("vector unaligned")
     // TODO: Temporary workaround. icx fails to vectorize some loops in debug build on Windows or with gcov.
-    #if (defined(_MSC_VER) && defined(_DEBUG)) || defined(GCOV_BUILD)
+    #if (defined(_MSC_VER) && defined(_DEBUG)) || defined(GCOV_BUILD) || defined(DEBUG_ASSERT)
         #define PRAGMA_VECTOR_ALWAYS
         #define PRAGMA_OMP_SIMD
         #define PRAGMA_OMP_SIMD_ARGS(ARGS)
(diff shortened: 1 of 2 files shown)


cpp/daal/src/services/service_defines.h
```diff
@@ -54,7 +54,7 @@ DAAL_EXPORT bool daal_check_is_intel_cpu();
     #define PRAGMA_NOVECTOR         _Pragma("novector")
     #define PRAGMA_VECTOR_UNALIGNED _Pragma("vector unaligned")
     // TODO: Temporary workaround. icx fails to vectorize some loops in debug build on Windows or with gcov.
-    #if (defined(_MSC_VER) && defined(_DEBUG)) || defined(GCOV_BUILD)
+    #if (defined(_MSC_VER) && defined(_DEBUG)) || defined(GCOV_BUILD) || defined(DEBUG_ASSERT)
```
david-cortes-intel, line 57: If the problem is only with MSVC, then better to put this into the first condition, so that it'd still work with other compilers.

Vika-F, line 57: I agree with David here. It would be better like this:
```suggestion
    #if (defined(_MSC_VER) && (defined(_DEBUG) || defined(DEBUG_ASSERT))) || defined(GCOV_BUILD)
```

ethanglaser: It is a problem both on windows and linux: http://intel-ci.intel.com/f0c43b6f-ca2a-f1f6-8669-a4bf010d0e2d, not specific to MSVC

ethanglaser:
> Debug CI scope (REQDBG=yes, different than [previous fix for windows build with MSVC_RUNTIME_VERSION=debug](https://github.com/uxlfoundation/oneDAL/pull/3423)) With main branch: http://intel-ci.intel.com/f0c43b6f-ca2a-f1f6-8669-a4bf010d0e2d With this branch: http://intel-ci.intel.com/f0c4ebb0-f6d3-f1d9-993d-d4f5ef20c6a0



> Actually I am seeing a similar issue on local linux builds about compilation failing due to not being able to vectorize some loop when there are debug symbols, but it appears to be limited to only a few problem cases. @ethanglaser Does the issue with the CI happen only on windows?

Yes sorry if my original message was confusing - it happens on both linux and windows, and is different case than the previous fix for specifically windows debug that Vika made (https://github.com/uxlfoundation/oneDAL/pull/3423)

Vika-F (approved):
I approve, but let's wait for at least a green public CI.

PR #3447, opened 2025-11-24
Suppress vectorization pragmas warnings (REQDBG=yes)

Vika-F:
## Description

Add `-Wno-pass-failed` compiler option to ICX and Clang debug builds to suppress the warnings that cause build failures.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

not applicable

</details>

### DIFF
--- a/dev/make/compiler_definitions/clang.mk
+++ b/dev/make/compiler_definitions/clang.mk
@@ -43,7 +43,7 @@ endif
 
 -Zl.clang =
 
--DEBC.clang = -g
+-DEBC.clang = -g -Wno-pass-failed
 
 -asanstatic.clang = -static-libasan
 -asanshared.clang = -shared-libasan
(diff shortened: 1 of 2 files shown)


dev/make/compiler_definitions/clang.mk
```diff
@@ -43,7 +43,7 @@ endif
 
 -Zl.clang =
 
--DEBC.clang = -g
+-DEBC.clang = -g -Wno-pass-failed
```
david-cortes-intel, line 46: Shouldn't this be on `clang.32e.mk`?

Vika-F: `clang.mk` is included into `clang.32e.mk` and it does not have any additions to `-DEBC.clang` there.

I can move this to `clang.32e.mk` and do `-DEBC.clang += -Wno-pass-failed` there. In this case the warnings would be suppressed only on x86, and not on Arm, for example.
I can do it if you think this is relevant.

david-cortes-intel: I guess it's no big deal, but it's hard to know what needs to be put under each file, and it appears to differ by compiler.

PR #3455, opened 2025-12-05
Implement efficient data layout for modelbuilders and gbt trees

avolkov-intel:
## Description

This PR introduces hybrid layout for GBT modelbuilders. After this fix the first 10 layers of trees will be stored in the form of full binary tree (the same way it was done before). This will allow to maintain condition that children of node idx are stored at positions 2 * idx and 2 * idx + 1 and save time on inference on trees with small depth.

Nodes on deeper levels after this update will be stored in sparse format, meaning that the amount of required memory for storing them will depend linearly on the number of nodes on deeper levels. Previously this dependency was exponential and for deep trees the use of GBT modelbuilders was impractical.

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
@@ -187,8 +187,8 @@ inline void predictForTreeVector(const DecisionTreeType & t, const FeatureTypes 
     const FeatureIndexType * const leftIds  = t.getLeftChildIndexes() - 1;
     const FeatureIndexType * const fIndexes = t.getFeatureIndexesForSplit() - 1;
     const int * const defaultLeft           = t.getDefaultLeftForSplit() - 1;
+    const size_t nNodes                     = t.getNumberOfNodes();
     const FeatureIndexType nFeat            = featTypes.getNumberOfFeatures();
-
     FeatureIndexType i[vectorBlockSize];
     services::internal::service_memset_seq<FeatureIndexType, cpu>(i, FeatureIndexType(1), vectorBlockSize);
     const FeatureIndexType maxLvl         = t.getMaxLvl();
@@ -200,6 +200,10 @@ inline void predictForTreeVector(const DecisionTreeType & t, const FeatureTypes 
             const FeatureIndexType idx          = i[k];
             const FeatureIndexType splitFeature = fIndexes[idx];
             i[k] = updateIndexDense<algorithmFPType>(idx, x[splitFeature + k * nFeat], values, defaultLeft, featTypes, splitFeature, dispatcher);
+            // After itr, node depth is itr + 1
+            // Node indexes on level k are in range [2^k, 2^(k + 1) - 1]
+            DAAL_ASSERT(i[k] >= (static_cast<size_t>(1) << (itr + 1)))
+            DAAL_ASSERT(i[k] < (static_cast<size_t>(1) << (itr + 2)))
         }
     }
     // TODO: Implement performance optimization for deep but sparse trees
@@ -210,6 +214,7 @@ inline void predictForTreeVector(const DecisionTreeType & t, const FeatureTypes 
             const FeatureIndexType idx          = i[k];
             const FeatureIndexType splitFeature = fIndexes[idx];
             i[k] = updateIndex<algorithmFPType>(idx, x[splitFeature + k * nFeat], values, leftIds, defaultLeft, featTypes, splitFeature, dispatcher);
+            DAAL_ASSERT(i[k] <= nNodes)
         }
     }
 
@@ -232,13 +237,17 @@ inline algorithmFPType predictForTree(const DecisionTreeType & t, const FeatureT
 
     const FeatureIndexType maxLvl         = t.getMaxLvl();
     const FeatureIndexType numDenseLayers = t.getNumDenseLayers();
+    const size_t nNodes                   = t.getNumberOfNodes();
 
     FeatureIndexType i = 1;
-
     for (FeatureIndexType itr = 0; itr < std::min(numDenseLayers, maxLvl); itr++)
     {
         const FeatureIndexType splitFeature = fIndexes[i];
         i = updateIndexDense<algorithmFPType>(i, x[splitFeature], values, defaultLeft, featTypes, splitFeature, dispatcher);
+        // After itr, node depth is itr + 1
+        // Node indexes on level k are in range [2^k, 2^(k + 1) - 1]
+        DAAL_ASSERT(i >= (static_cast<size_t>(1) << (itr + 1)))
+        DAAL_ASSERT(i < (static_cast<size_t>(1) << (itr + 2)))
     }
 
     // TODO: Implement performance optimization for deep but sparse trees
@@ -246,6 +255,7 @@ inline algorithmFPType predictForTree(const DecisionTreeType & t, const FeatureT
     {
         const FeatureIndexType splitFeature = fIndexes[i];
         i = updateIndex<algorithmFPType>(i, x[splitFeature], values, leftIds, defaultLeft, featTypes, splitFeature, dispatcher);
+        DAAL_ASSERT(i <= nNodes)
     }
 
     return values[i];


cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
```diff
@@ -96,27 +99,92 @@ inline FeatureIndexType updateIndex(FeatureIndexType idx, algorithmFPType valueF
     }
 }
 
+// In the implementations below leftChildIndexes array is used to determine the position of node's children.
+// They should be used on deeper levels where nodes are stored in sparse format to achieve faster inference.
+// However, they can be used for all levels as leftChildIndexes is filled correctly for all nodes.
+
+template <typename algorithmFPType>
+inline FeatureIndexType updateIndex(FeatureIndexType idx, algorithmFPType valueFromDataSet, const ModelFPType * splitPoints,
```
razdoburdin, line 127: Please consider using a different name for overladed `updateIndex` to simplify code understanding in the future

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
```diff
@@ -96,27 +99,92 @@ inline FeatureIndexType updateIndex(FeatureIndexType idx, algorithmFPType valueF
...
 
-    const FeatureIndexType maxLvl = t.getMaxLvl();
-
-    for (FeatureIndexType itr = 0; itr < maxLvl; itr++)
+    for (FeatureIndexType itr = numDenseLayers; itr < maxLvl; itr++)
     {
         for (FeatureIndexType k = 0; k < vectorBlockSize; k++)
         {
             const FeatureIndexType idx          = i[k];
             const FeatureIndexType splitFeature = fIndexes[idx];
```
razdoburdin, line 215: Please add performance optimization for this part in TODO list.
For these nodes reordering loops. i.e loop over k should be the outer, for loop over depth should be replaced by while(!isleaf) loop.

cpp/daal/src/algorithms/dtrees/gbt/gbt_model_impl.h
```diff
@@ -44,10 +44,11 @@ typedef uint32_t FeatureIndexType;
...
 
-static inline size_t getNumberOfNodesByLvls(const size_t nLvls)
-{
-    return (1 << (nLvls + 1)) - 1;
-}
+// The number of layers for which tree is stored in full binary format (by creating dummy nodes)
+// This is required to maintain condition that children of node idx are located at nodes 2 * idx and 2 * idx + 1
+// Nodes on deeper levels (if they exist) will be stored in sparse format
+// TODO: add an option to control this parameter
+const size_t defaultNumDenseLayers = 10;
```
razdoburdin: please consider making this constant `constexpr`

cpp/daal/src/algorithms/dtrees/gbt/gbt_model_impl.h
```diff
@@ -206,17 +229,20 @@ class GbtDecisionTree : public SerializationIface
...
         arch->setSharedPtrObj(_nodeCoverValues);
         arch->setSharedPtrObj(_defaultLeft);
+        arch->setSharedPtrObj(_leftChildIndexes);
 
         return services::Status();
     }
 
 protected:
     size_t _nNodes;
+    size_t _numDenseLayers;
```
Vika-F, line 241:
```suggestion
    size_t _numDenseLayers;           // The number of layers for which tree is stored in full binary format
```

cpp/daal/src/algorithms/dtrees/gbt/gbt_model_impl.h
```diff
@@ -65,32 +66,39 @@ class GbtDecisionTree : public SerializationIface
     using NodeCoverType              = HomogenNumericTable<ModelFPType>;
     using FeatureIndexesForSplitType = HomogenNumericTable<FeatureIndexType>;
     using defaultLeftForSplitType    = HomogenNumericTable<int>;
+    using leftChildIndexType         = HomogenNumericTable<size_t>;
```
Vika-F: Why not HomogenNumericTable\<unsigned\>? Is it expected to store trees with more than 2^32 nodes?

cpp/daal/src/algorithms/dtrees/gbt/gbt_model_impl.h
```diff
@@ -206,17 +229,20 @@ class GbtDecisionTree : public SerializationIface
...
 
 protected:
     size_t _nNodes;
+    size_t _numDenseLayers;
     FeatureIndexType _maxLvl;
     services::SharedPtr<SplitPointType> _splitPoints;
     services::SharedPtr<FeatureIndexesForSplitType> _featureIndexes;
     services::SharedPtr<NodeCoverType> _nodeCoverValues;
     services::SharedPtr<defaultLeftForSplitType> _defaultLeft;
+    services::SharedPtr<leftChildIndexType> _leftChildIndexes;
```
Vika-F, line 253:
```suggestion
    services::SharedPtr<leftChildIndexType> _leftChildIndexes;   // Homogeneous table of size [the size] which stores indices of the left children of the tree's nodes.
```
Please also add the description about the format in which the indices are stored in the table. I.e. "_leftChildIndexes[0,i] stores ..."

It would be also good to add the descriptions for some other fields as well. While `_featureIndexes` or `_splitPoints` are more or less self descriptive, the meaning of ` _nodeCoverValues` and `_maxLvl` is not so obvious.

cpp/daal/src/algorithms/dtrees/gbt/gbt_model_impl.h
```diff
@@ -274,6 +301,25 @@ class GbtTreeImpl : public dtrees::internal::TreeImpl<TNodeType, TAllocator>
             if (maxLvl < curLvl) maxLvl = curLvl;
         }
     }
+
+    void getMaxLvLAndNumNodes(const typename TNodeType::Base & node, size_t & maxLvl, size_t & numNodes, const size_t numDenseLayers,
```
Vika-F, line 297: Should `getMaxLvl` function be deleted, as this function was added?

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
```diff
@@ -54,6 +54,9 @@ struct PredictDispatcher
     typedef PredictDispatcher<hasUnorderedFeatures, hasAnyMissing> type;
 };
 
+// In the implementations below leftChildIndexes array is not used as we suppose that left child index is idx * 2
+// They shouldn't be uses for cases where this condition is not satisfied, this will lead to incorrect results
+
```
Vika-F, line 63: It would be good to implement some assert here.
For example, `DAAL_ASSERT(idx < (1 << defaultNumDenseLayers));`

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
```diff
@@ -96,27 +99,92 @@ inline FeatureIndexType updateIndex(FeatureIndexType idx, algorithmFPType valueF
...
+// In the implementations below leftChildIndexes array is used to determine the position of node's children.
+// They should be used on deeper levels where nodes are stored in sparse format to achieve faster inference.
+// However, they can be used for all levels as leftChildIndexes is filled correctly for all nodes.
+
+template <typename algorithmFPType>
+inline FeatureIndexType updateIndex(FeatureIndexType idx, algorithmFPType valueFromDataSet, const ModelFPType * splitPoints,
+                                    const size_t * leftChildIndexes, const int * defaultLeft, const FeatureTypes & featTypes,
+                                    FeatureIndexType splitFeature, const PredictDispatcher<false, false> & dispatcher)
+{
+    return leftChildIndexes[idx] + ((idx != leftChildIndexes[idx]) && (valueFromDataSet > splitPoints[idx]));
```
Vika-F, line 132: This function will also be better with assert.

Also it is not clear why this term `(idx != leftChildIndexes[idx])` is required? Does it encode the leaf nodes? I.e. if (idx == leftChildIndexes[idx]) then it's a leaf node.

But in the implementation used for the dense case there were no leaf node checking. Why it is required here?

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
```diff
@@ -96,27 +99,92 @@ inline FeatureIndexType updateIndex(FeatureIndexType idx, algorithmFPType valueF
     }
 }
 
+// In the implementations below leftChildIndexes array is used to determine the position of node's children.
+// They should be used on deeper levels where nodes are stored in sparse format to achieve faster inference.
+// However, they can be used for all levels as leftChildIndexes is filled correctly for all nodes.
```
Vika-F: From the comment it is unclear whether `leftChildIndexes` used now for all nodes or only for the nodes on the deeper levels?

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
```diff
@@ -96,27 +99,92 @@ inline FeatureIndexType updateIndex(FeatureIndexType idx, algorithmFPType valueF
...
+{
+    return leftChildIndexes[idx] + ((idx != leftChildIndexes[idx]) && (valueFromDataSet > splitPoints[idx]));
+}
+
+template <typename algorithmFPType>
+inline FeatureIndexType updateIndex(FeatureIndexType idx, algorithmFPType valueFromDataSet, const ModelFPType * splitPoints,
+                                    const size_t * leftChildIndexes, const int * defaultLeft, const FeatureTypes & featTypes,
+                                    FeatureIndexType splitFeature, const PredictDispatcher<true, false> & dispatcher)
+{
+    return leftChildIndexes[idx]
```
Vika-F, line 141: This and further two functions require some clarifying comments.

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
avolkov-intel: When the function for dense case is called it is guaranteed that the maximum level is not reached (outer loop goes till min(maxLevel, numDenseLayers)), so node will always have a child. Here it's not guaranteed

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
avolkov-intel: leftChildIndexes array is correctly filled for all layers, so using these functions will give you correct results. However, for more efficient execution you can rely on additional restriction (children are at 2 * idx, 2 * idx + 1) for first layers

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
avolkov-intel: Not sure if this is a good assert since the number of dense layers in the tree theoretically can be different from defaultNumDenseLayers, actual number of dense layers should be obtained from tree meta information which isn't possible to get here without introducing additional argument. Introducing additional argument will lead to slow down since this a performance critical place

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
avolkov-intel: I will rename first 4 implementations to updateIndexDense, the last 4 names will be unchanged (they can be used for all layers)

cpp/daal/src/algorithms/dtrees/gbt/gbt_model_impl.h
avolkov-intel: done

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
avolkov-intel: done

cpp/daal/src/algorithms/dtrees/gbt/gbt_predict_dense_default_impl.i
avolkov-intel: done

cpp/daal/src/algorithms/dtrees/gbt/gbt_model_impl.h
avolkov-intel: You are right we can use uint32_t, I changed the data type

razdoburdin (approved):
looks nice for me

PR #3518, opened 2026-02-23
fix: correct MKL strip_prefixes for conda packages in MODULE.bazel

napetrov:
This PR fixes Bazel release build failures caused by incorrect `strip_prefixes` in `mkl_repo` configuration.

### Problem
Build failed with missing MKL static libraries:
- `libmkl_core.a`
- `libmkl_intel_ilp64.a`
- `libmkl_tbb_thread.a`

### Root cause
`mkl_repo` mixes artifacts from both:
- `.whl` packages (have `.data/data/...` layout)
- `.conda` packages (extract directly to `lib/...`)

`strip_prefixes` were configured uniformly as `.data/data`, which is correct for wheels but **misleading for conda packages**.

**Note:** Current `repos.bzl` implementation ignores `strip_prefix` for `.conda` files (they always extract directly to output). However, the non-empty prefixes in `MODULE.bazel` were causing confusion in downstream `mkl.tpl.BUILD` which expects files at `lib/` for conda packages, not at `.data/data/lib/`.

### Fix
In `MODULE.bazel`:
- set **empty** `strip_prefix` (`""`) for conda packages to reflect actual extraction behavior:
  - `mkl-static`
  - all `onemkl-sycl-*` packages (datafitting, blas, dft, lapack, sparse, rng, vm, stats)
- keep `.data/data` prefixes for wheel packages (correctly handled by `download_and_extract`):
  - `mkl_include`
  - `onemkl_sycl_include`
  - `mkl_devel_dpcpp`

This change makes `strip_prefixes` list semantically correct and aligns with actual file layout expectations in `mkl.tpl.BUILD`.

### Validation
Validated on Ubuntu with Bazel 8.5.1:

```bash
./bazel/bin/bazel clean --expunge
./bazel/bin/bazel build //:release --verbose_failures
```

`//:release` builds successfully and produces expected artifacts (`libonedal*.a/.so`).

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,7 +2,7 @@ alabaster==1.0.0
 Babel==2.18.0
 beautifulsoup4==4.14.3
 certifi==2026.1.4
-chardet==5.2.0
+chardet==6.0.0.post1
 click==8.3.1
 colorama==0.4.6
 docutils~=0.21.2


Alexandr-Solovev (approved):
LGTM

PR #3551, opened 2026-03-05
[conda] Split GPU runtime into dal-gpu package

napetrov:
## Summary
- split GPU runtime (`libonedal_dpc.so`) from `dal` into dedicated `dal-gpu` conda package
- keep `dal` as CPU runtime package (`libonedal_core.so`, `libonedal_thread.so`)
- update `dal-devel` test requirements to include `dal-gpu` only for `cxx_compiler == dpcpp`
- update `test-devel.sh` to always run CPU examples and run `oneapi/dpc` examples only when GPU runtime is installed

## Motivation
Need explicit validation of two independent scenarios:
1. CPU-only install (`dal` without `dal-gpu`) still passes CPU examples
2. CPU + GPU install (`dal` + `dal-gpu`) enables DPC++ runtime coverage

This removes coupling between CPU runtime delivery and DPC++ runtime delivery in conda packaging.

## Testing
- local syntax check: `bash -n conda-recipe/test-devel.sh`
- CI conda recipe workflow validates matrix behavior (`gxx` / `clangxx` / `dpcpp`)

## Naming
- package name follows agreed convention: `dal-gpu`


conda-recipe/meta.yaml
```diff
@@ -69,10 +69,24 @@ outputs:
...
+      summary: oneAPI Data Analytics Library (oneDAL) CPU runtime libraries
+
+  - name: dal-gpu
+    build:
+      skip: true  # [cxx_compiler != "dpcpp"]
+    script: pack.sh  # [linux]
+    requirements:
+      run:
+        - {{ pin_subpackage('dal', exact=True) }}
+        - dpcpp_impl_linux-64  # [linux64]
```
david-cortes-intel: Isn't this the compiler? Shouldn't it require the runtimes instead? And shouldn't it depend on MKL DPC packages?

conda-recipe/test-devel.sh
```diff
@@ -58,8 +60,84 @@ run_examples() {
     )
 }
 
+run_dpc_examples() {
+    local linking_type=$1
+
+    if [ "$linking_type" == "dynamic" ]; then
```
Alexandr-Solovev: I believe this part can be hardcoded to dynamic, due to we don't produce static for dpc

napetrov: Agreed. I simplified this in 545a2ae: DPC examples are now hardcoded to dynamic mode in run_dpc_examples() (no linking_type parameter), since conda packaging does not produce static DPC++ artifacts.

conda-recipe/meta.yaml
napetrov: Good question. In this recipe dpcpp_impl_linux-64 is used as the runtime dependency required by libonedal_dpc.so (it provides the DPC++ runtime pieces needed at execution time). The dal-gpu output is intentionally a runtime split package, so we avoid pulling full devel toolchains there.

For MKL DPC coverage, we keep it in dal-devel test requirements (mkl-devel-dpcpp) where we build/run oneapi/dpc examples. That validates the full GPU stack without over-constraining the lightweight runtime split package.

conda-recipe/test-devel.sh
```diff
@@ -58,8 +60,81 @@ run_examples() {
...
+
+            cmake .. $cmake_args
+            make -j$(nproc)
+        )
+
+        for example in _cmake_results/intel_intel64_$library_postfix/*; do
+            echo "================"
+            echo "Running example: oneapi-dpc-$linking_type-$(basename $example)"
+            echo "================"
+            $example
```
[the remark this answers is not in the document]
napetrov: Fixed in 076275388: replaced the literal glob pattern loop with a portable shell expansion check (set -- glob; test if $1 == pattern), failing clearly if nothing is found.

conda-recipe/test-devel.sh
```diff
@@ -58,8 +60,81 @@ run_examples() {
...
+echo "========================================"
+echo "Running CPU examples: daal/cpp static"
+echo "========================================"
 run_examples daal static
+
+# ============================================================
+# GPU/DPC++ tests: oneapi/dpc (requires dal-gpu / libonedal_dpc.so)
+# Skipped if GPU library is not installed.
+# ============================================================
+if [ -f "$CONDA_PREFIX/lib/libonedal_dpc.so" ] && [ -f "$CONDA_PREFIX/lib/libonedal_parameters_dpc.so" ]; then
```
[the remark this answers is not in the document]
napetrov: Fixed in 076275388: GPU detection now uses find with -name 'libonedal_dpc.so*' (matches both versioned and unversioned), so the check works regardless of whether the unversioned symlink is present.

conda-recipe/pack.sh
```diff
@@ -40,9 +40,13 @@ if [ "$PKG_NAME" = "dal-include" ]; then
     mkdir -p "$PREFIX/include"
     cp -r include/* "$PREFIX/include/"
 fi
-# copy libraries
+# copy CPU runtime libraries (excludes all DPC++ runtime libs, moved to dal-gpu)
 if [ "$PKG_NAME" = "dal" ]; then
-    find lib/intel64 -name "libonedal*.so*" -exec cp -P {} "$PREFIX/lib/" \;
+    find lib/intel64 -name "libonedal*.so*" ! -name "libonedal*_dpc*" -exec cp -P {} "$PREFIX/lib/" \;
+fi
```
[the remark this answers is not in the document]
napetrov: Fixed in 076275388: DPC runtime libs are now skipped in the intel64 symlink loop in pack.sh via a case statement.

conda-recipe/test-devel.sh
```diff
@@ -23,6 +23,7 @@ source $CONDA_PREFIX/env/vars.sh
 run_examples() {
     local interface_name=$1
     local linking_type=$2
+    local extra_cmake_args=${3:-""}
 
     if [ "$linking_type" == "dynamic" ]; then
         library_postfix="so"
```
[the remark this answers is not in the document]
napetrov: Fixed in bf4b405: shebang changed to '#!/usr/bin/env bash' — the script uses local, process substitution, and other bash features.

conda-recipe/test-devel.sh
```diff
@@ -58,8 +60,81 @@ run_examples() {
...
+run_dpc_examples() {
+    # DPC++ examples are validated only in dynamic mode.
+    # We do not produce static DPC++ artifacts in conda packaging.
+    local linking_type="dynamic"
+    local library_postfix="so"
+
+    (
+        cd examples/oneapi/dpc
+        mkdir build_$linking_type
+
```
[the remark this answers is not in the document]
napetrov: Fixed in bf4b405: changed to 'mkdir -p' for idempotent behavior on re-runs.

conda-recipe/meta.yaml
david-cortes-intel: > For MKL DPC coverage, we keep it in dal-devel test requirements (mkl-devel-dpcpp) where we build/run oneapi/dpc examples.

But that's not how it's supposed to work. The oneDAL library should be linking to MKL, not the libraries using oneDAL.

napetrov:
Addressed latest review comments in follow-up commits:

- `5b5fff738` — add `mkl` runtime dependency for `dal-gpu` (aligns with dynamic oneMKL linkage from main)
- `360a36f3d` — hardening updates:
  1) `conda-recipe/build.sh`
     - stop creating `libtbb.so` under `$PREFIX/lib`
     - create isolated `TBBROOT` staging under `$SRC_DIR/__tbbroot` and place unversioned `libtbb.so` shim there only
  2) `conda-recipe/test-devel.sh`
     - add `set -euo pipefail`
     - clean `build_*` dirs before configure/build to avoid stale binaries on reruns
  3) `conda-recipe/meta.yaml`
     - tighten `dal-gpu` selector: `skip: true  # [cxx_compiler != "dpcpp" or not linux64]`

This should address the recent Copilot notes and the MKL runtime dependency concern.

conda-recipe/meta.yaml
```diff
@@ -69,10 +69,25 @@ outputs:
...
+
+  - name: dal-gpu
+    build:
+      skip: true  # [cxx_compiler != "dpcpp" or not linux64]
+    script: pack.sh  # [linux]
+    requirements:
+      run:
+        - {{ pin_subpackage('dal', exact=True) }}
+        - dpcpp_impl_linux-64  # [linux64]
+        - mkl  # [linux64]
```
david-cortes-intel: This should be a transitive dependency from MKL DPC libraries. It shouldn't specify it directly.

david-cortes-intel (changes requested):

conda-recipe/meta.yaml
napetrov: You're right on both points. Fixed in ab49f00: replaced `dpcpp_impl_linux-64` (compiler package) with `intel-sycl-rt` (SYCL runtime), and `mkl` with `mkl-dpcpp` (GPU MKL runtime). The base `mkl` will come in transitively through `mkl-dpcpp`.

conda-recipe/meta.yaml
```diff
@@ -69,10 +69,25 @@ outputs:
...
-        - test -f $PREFIX/lib/libonedal_dpc.so.{{ major_binary_version }}  # [linux]
         - test -f $PREFIX/lib/libonedal_thread.so.{{ major_binary_version }}  # [linux]
     about:
-      summary: oneAPI Data Analytics Library (oneDAL) runtime libraries
+      summary: oneAPI Data Analytics Library (oneDAL) CPU runtime libraries
+
+  - name: dal-gpu
+    build:
+      skip: true  # [cxx_compiler != "dpcpp" or not linux64]
+    script: pack.sh  # [linux]
```
david-cortes-intel, line 94: Some of these are 'linux' and others 'linux64'. Shouldn't they all use the same?

conda-recipe/meta.yaml
```diff
@@ -69,10 +69,25 @@ outputs:
...
+    requirements:
+      run:
+        - {{ pin_subpackage('dal', exact=True) }}
+        - intel-sycl-rt  # [linux64]
```
david-cortes-intel: Looks like this should already be a transitive dependency from `mkl-dpcpp`.

conda-recipe/meta.yaml
```diff
@@ -69,10 +69,25 @@ outputs:
...
+        - mkl-dpcpp  # [linux64]
```
david-cortes-intel, line 101: This one should have a constraint on the version based on the compiler that was used.

david-cortes-intel (changes requested):

conda-recipe/meta.yaml
napetrov: Fixed in 557e2f8. The `dal-gpu` output is guarded by `skip: true  # [cxx_compiler != "dpcpp" or not linux64]`, so `linux64` is the right selector throughout — updated `script:`, run dep selectors (dropped them as redundant inside an already-skipped output), and test commands to use `# [linux64]` consistently.

conda-recipe/meta.yaml
napetrov: Correct — `intel-sycl-rt` is already a transitive dependency through `mkl-dpcpp`. However, we're keeping the explicit listing intentionally for clarity: end-users installing `dal-gpu` should see what runtime components are required without having to trace transitive deps. That said — if the preference is to rely fully on the transitive dep, happy to drop it. What do you think?

conda-recipe/meta.yaml
napetrov: The version constraint is already provided transitively: `dpcpp_impl_linux-64` in the build section carries a `run_exports: strong: [intel-sycl-rt, exact]` pin (from intel-compiler-repack-feedstock). However, the recipe-level `ignore_run_exports_from: dpcpp_impl_linux-64` suppresses that propagation for the top-level output — but `dal-gpu` is a separate sub-output and does not inherit this suppression, so the pin does flow through correctly when `dpcpp_impl` is in the build env. If you'd like an explicit constraint anyway, we can add `- intel-sycl-rt ={{ cxx_compiler_version.split('.')[0] }}.*` — just need to confirm that's the right pattern for your build matrix.

napetrov: Fixed in cfaa164: added `intel-sycl-rt =={{ cxx_compiler_version }}` to pin the SYCL runtime to the exact compiler version used at build time.

napetrov: Updated in d9c3d85: using a version range instead of exact match — `intel-sycl-rt >={{ cxx_compiler_version }},<{{ (cxx_compiler_version.split(".")[0]|int + 1) }}`. For a build with compiler 2025.3.2 this resolves to `>=2025.3.2,<2026`, allowing patch/minor updates within the same compiler year while blocking incompatible major upgrades.

napetrov: Updated in 4968284: switched to `pin_compatible` — added `intel-sycl-rt` to the `host` requirements of `dal-gpu` so conda-build can resolve the installed version at build time, then `{{ pin_compatible("intel-sycl-rt", max_pin="x") }}` generates `>=<build_version>,<next_year` automatically (e.g. `>=2025.3.2,<2026`). No version numbers hardcoded.

conda-recipe/meta.yaml
david-cortes-intel: The conda and mamba installers would show exactly what dependencies would need to be pulled when installing a package and ask the user for confirmation, so I don't think it should include transitive dependencies.

conda-recipe/meta.yaml
david-cortes-intel: It looks to me like it pins "level-zero", but not SYCL runtimes:
https://github.com/conda-forge/intel-compiler-repack-feedstock/blob/c0f9782a7c9b08b16716f01f69c0f817d1be68fb/recipe/meta.yaml#L181

david-cortes-intel: On a second look, it does have a pinned dependency for sycl, guess it should be fine:
https://github.com/conda-forge/intel-compiler-repack-feedstock/blob/c0f9782a7c9b08b16716f01f69c0f817d1be68fb/recipe/meta.yaml#L381

napetrov:
> @napetrov If having `dpcpp_impl_linux-64` as a build-time dependency propagates the constraint on runtime packages, shouldn't that package be used only for the `dal` (CPU) package?

we have 
  ignore_run_exports_from:
    - dpcpp_impl_linux-64  # [linux64 or win]
    
    which do not pull dependency down the line. 
    
    we can move it from global section to outputs sections and do this for all but dal-gpu

conda-recipe/meta.yaml
```diff
@@ -34,8 +34,12 @@ build:
   number: {{ buildnumber }}
   string: h{{ git_hash }}_{{ buildnumber }}_{{ cxx_compiler }}
   include_recipe: False
-  ignore_run_exports_from:
-    - dpcpp_impl_linux-64  # [linux64 or win]
+  # Note: ignore_run_exports_from is NOT set globally here.
+  # It is set per-output: dal (CPU) ignores dpcpp_impl run_exports so it
```
david-cortes-intel, line 41: Shouldn't it also apply to `dal-static` and `dal-include`?

napetrov: yes, fixed now for all

conda-recipe/meta.yaml
```diff
@@ -124,6 +130,9 @@ outputs:
       summary: Static libraries for oneAPI Data Analytics Library (oneDAL)
 
   - name: dal-devel
+    build:
+      ignore_run_exports_from:
```
david-cortes-intel, line 137: I think this one is not necessary, because it depends on `dal-gpu`, but it shouldn't make a difference to have it or not.

napetrov: yes, thought was that from clean deps perspective - only gpu package should have it

conda-recipe/build.sh
```diff
@@ -15,9 +15,33 @@
...
 
-export TBBROOT=$PREFIX
 export DPL_ROOT=$PREFIX
 
+# Use isolated TBBROOT staging so we don't create files under $PREFIX that may
+# accidentally end up in output packages.
+# conda-forge tbb-devel may expose only versioned SONAMEs (libtbb.so.<N>,
+# libtbbmalloc.so.<N>, etc.) while oneDAL Make expects unversioned names in
+# TBBROOT/lib prerequisites.  Mirror the full libtbb* tree into a staging dir,
+# adding unversioned symlinks where missing.
```
david-cortes-intel, line 25: If this is how it works, sounds like it's a problem on the oneDAL side.

CC @Alexandr-Solovev  please review.

conda-recipe/pack.sh
```diff
@@ -24,9 +24,21 @@ if [ "$PKG_NAME" = "dal-devel" ]; then
     cp -r "env" "$PREFIX/"
     cp -r "lib/cmake" "$PREFIX/lib/"
     cp -r "lib/pkgconfig" "$PREFIX/lib/"
+    # Install example datasets so dal-devel test can find them via $CONDA_PREFIX/data
```
david-cortes-intel: @Alexandr-Solovev Is it intended to work like this? Shouldn't paths have been updated after the PR with optimizations for dataset usage?

conda-recipe/test-devel.sh
```diff
@@ -1,4 +1,5 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
david-cortes-intel: As far as I am aware, it is not guaranteed that bash will be available if not set as a build requirement. It also doesn't look like the syntax here includes non-posix bashisms.

conda-recipe/test-devel.sh
```diff
@@ -44,22 +55,120 @@ run_examples() {
...
+    # DPC++ examples are validated only in dynamic mode.
+    # We do not produce static DPC++ artifacts in conda packaging.
+    local linking_type="dynamic"
+    local library_postfix="so"
+
+    (
+        cd examples/oneapi/dpc
+        # Keep data path layout consistent with oneapi/cpp examples.
+        if [ ! -d ../data ] && [ -d "$CONDA_PREFIX/share/oneDAL/data" ]; then
+            ln -sf "$CONDA_PREFIX/share/oneDAL/data" ../data
```
david-cortes-intel, line 101: Wasn't the earlier command copying them from this same folder into the folder on the right hand side?

conda-recipe/test-devel.sh
```diff
@@ -44,22 +55,120 @@ run_examples() {
...
+        if [ ! -e data ] && [ -d ../data ]; then
+            ln -s ../data data
+        fi
+        rm -rf build_$linking_type
+        mkdir -p build_$linking_type
+
+        (
+            cd build_$linking_type
+
+            cmake_args="-DONEDAL_LINK=$linking_type"
```
david-cortes-intel, line 112: Shouldn't this hard-code it to dynamic linking, since this is under "DPC" examples?

conda-recipe/test-devel.sh
```diff
@@ -44,22 +55,120 @@ run_examples() {
...
+            # Note: MKL cmake config (incl. SYCL variant) requires TBB to be findable at configure time.
+            # conda-forge tbb-devel does not ship TBBConfig.cmake, so we set TBB as found manually
```
david-cortes-intel, line 114: It does include that file:
Image

conda-recipe/pack.sh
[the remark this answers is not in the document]
napetrov: Good catch — updated the comment to match the actual install location. It now explicitly says datasets are installed under `$CONDA_PREFIX/share/oneDAL/data` for dal-devel tests. Fixed in 382f101.

conda-recipe/test-devel.sh
napetrov: Good point — agreed we should not rely on bash being present implicitly.\n\nAddressed in 65763ca: `test-devel.sh` is POSIX now (`#!/usr/bin/env sh`), with bashisms removed (`source` -> `.`, `local` removed, `==` -> `=`). This avoids introducing an extra bash build/runtime requirement for the test script.

conda-recipe/pack.sh
napetrov: Thanks, aligned this with the dataset-path cleanup.\n\nAddressed in 65763ca: `pack.sh` now prefers the canonical source location (`$SRC_DIR/data`) when installing test CSVs into `$PREFIX/share/oneDAL/data`, with `$SRC_DIR/examples/oneapi/data` and `data/` kept only as fallbacks for layout compatibility.

conda-recipe/test-devel.sh
napetrov: Here they serve two different layouts:\n1) `../data` is the expected location for oneAPI examples (`examples/oneapi/{cpp,dpc}`), and if absent we bridge it from `$CONDA_PREFIX/share/oneDAL/data`;\n2) `data` is a local convenience link used by some example paths/scripts.\n\nSo the second link is intentional (not a duplicate copy), to keep both access patterns working from the example dir.

conda-recipe/test-devel.sh
napetrov: Yes — agreed. This is already hardcoded to dynamic in `run_dpc_examples()` (introduced earlier in 545a2ae), so DPC examples are no longer parameterized by link type and always run with `-DONEDAL_LINK=dynamic`.

conda-recipe/test-devel.sh
napetrov: Right — that screenshot is exactly why I switched the detection to glob checks instead of requiring an unversioned symlink.\n\nCurrent logic checks `libonedal_dpc.so*` and `libonedal_parameters_dpc.so*`, so either versioned or unversioned names are accepted (see 076275388). This avoids false skips when only SONAME files are present.

conda-recipe/build.sh
napetrov: Agreed this is effectively a oneDAL/CMake expectation mismatch (expects unversioned `libtbb*.so` in TBBROOT).\n\nFor this recipe I handled it in build-time staging only (TBBROOT under `$SRC_DIR`), so we avoid creating extra files under `$PREFIX` that could leak into package outputs. The packaging side is contained, while we can still discuss upstream build-system improvements separately.

conda-recipe/meta.yaml
napetrov: Agreed, and this is fixed now.\n\nUpdated in ab49f00: we switched to runtime-oriented deps (`intel-sycl-rt` + `mkl-dpcpp`) so oneDAL links against MKL DPC runtime itself, instead of relying on downstream libraries/apps to bring it in.

conda-recipe/pack.sh
```diff
@@ -40,9 +44,13 @@ if [ "$PKG_NAME" = "dal-include" ]; then
...
 fi
-# copy libraries
+# copy CPU runtime libraries (excludes all DPC++ runtime libs, moved to dal-gpu)
 if [ "$PKG_NAME" = "dal" ]; then
-    find lib/intel64 -name "libonedal*.so*" -exec cp -P {} "$PREFIX/lib/" \;
+    find lib/intel64 -name "libonedal*.so*" ! -name "libonedal*_dpc*" -exec cp -P {} "$PREFIX/lib/" \;
+fi
+# copy GPU/DPC++ runtime libraries
+if [ "$PKG_NAME" = "dal-gpu" ]; then
+    find lib/intel64 -name "libonedal*_dpc*.so*" -exec cp -P {} "$PREFIX/lib/" \;
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/test-devel.sh
```diff
@@ -44,22 +45,112 @@ run_examples() {
...
 }
 
+# ============================================================
+# CPU-only tests: oneapi/cpp + daal (no GPU lib required)
+# These must pass even without dal-gpu installed.
+# ============================================================
+echo "========================================"
+echo "Running CPU examples: oneapi/cpp dynamic"
+echo "========================================"
 run_examples oneapi dynamic
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/build.sh
```diff
@@ -18,6 +18,15 @@
...
 export DPL_ROOT=$PREFIX
 
+# Workaround: some tbb-devel builds expose only versioned SONAME (libtbb.so.<N>)
+# while oneDAL Make expects unversioned libtbb.so path in prerequisites.
+if [ ! -f "$PREFIX/lib/libtbb.so" ]; then
+    tbb_soname=$(find "$PREFIX/lib" -maxdepth 1 -name "libtbb.so.*" | head -1)
+    if [ -n "$tbb_soname" ]; then
+        ln -sf "$(basename "$tbb_soname")" "$PREFIX/lib/libtbb.so"
+    fi
+fi
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/test-devel.sh
```diff
@@ -44,22 +45,112 @@ run_examples() {
             if [ "$linking_type" == "static" ]; then
                 cmake_args="$cmake_args -DTBB_tbb_FOUND=YES"
             fi
+            cmake_args="$cmake_args $extra_cmake_args"
 
             cmake .. $cmake_args
             make -j$(nproc)
         )
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/test-devel.sh
```diff
@@ -44,22 +45,112 @@ run_examples() {
...
+
+            cmake_args="-DONEDAL_LINK=$linking_type"
+            # Note: MKL cmake config (incl. SYCL variant) requires TBB to be findable at configure time.
+            # conda-forge tbb-devel does not ship TBBConfig.cmake, so we set TBB as found manually
+            # (same approach as for the static CPU build above).
+            cmake_args="$cmake_args -DTBB_tbb_FOUND=YES"
+
+            cmake .. $cmake_args
+            make -j$(nproc)
+        )
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/meta.yaml
```diff
@@ -69,10 +69,25 @@ outputs:
...
         - test -f $PREFIX/lib/libonedal_core.so.{{ major_binary_version }}  # [linux]
-        - test -f $PREFIX/lib/libonedal_dpc.so.{{ major_binary_version }}  # [linux]
         - test -f $PREFIX/lib/libonedal_thread.so.{{ major_binary_version }}  # [linux]
     about:
-      summary: oneAPI Data Analytics Library (oneDAL) runtime libraries
+      summary: oneAPI Data Analytics Library (oneDAL) CPU runtime libraries
+
+  - name: dal-gpu
+    build:
+      skip: true  # [cxx_compiler != "dpcpp"]
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/build.sh
```diff
@@ -15,9 +15,33 @@
...
+# conda-forge tbb-devel may expose only versioned SONAMEs (libtbb.so.<N>,
+# libtbbmalloc.so.<N>, etc.) while oneDAL Make expects unversioned names in
+# TBBROOT/lib prerequisites.  Mirror the full libtbb* tree into a staging dir,
+# adding unversioned symlinks where missing.
+export TBBROOT="$SRC_DIR/__tbbroot"
+mkdir -p "$TBBROOT/lib" "$TBBROOT/include"
+
+# Symlink TBB headers (TBBROOT/include -> $PREFIX/include)
+ln -sfn "$PREFIX/include" "$TBBROOT/include"
+
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/test-devel.sh
```diff
@@ -32,7 +34,19 @@ run_examples() {
...
+        # Ensure oneapi example datasets are reachable for get_data_path().
+        # Preferred layout is ../data relative to this directory.
+        if [ ! -d ../data ]; then
+            if [ -d "$CONDA_PREFIX/data" ]; then
+                mkdir -p ../data
+                cp -f "$CONDA_PREFIX"/data/*.csv ../data/ 2>/dev/null || true
+            fi
+        fi
+        if [ ! -e data ] && [ -d ../data ]; then
+            ln -s ../data data
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/test-devel.sh
```diff
@@ -44,22 +58,123 @@ run_examples() {
...
+        cd examples/oneapi/dpc
+        # Keep data path layout consistent with oneapi/cpp examples.
+        if [ ! -d ../data ]; then
+            if [ -d "$CONDA_PREFIX/data" ]; then
+                mkdir -p ../data
+                cp -f "$CONDA_PREFIX"/data/*.csv ../data/ 2>/dev/null || true
+            fi
+        fi
+        if [ ! -e data ] && [ -d ../data ]; then
+            ln -s ../data data
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/meta.yaml
```diff
@@ -109,6 +124,7 @@ outputs:
         - mkl-devel
         - mkl-static
         - mkl-devel-dpcpp  # [ cxx_compiler == "dpcpp" ]
+        - dal-gpu =={{ version }}  # [ cxx_compiler == "dpcpp" ]
```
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/build.sh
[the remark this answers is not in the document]
napetrov: Thanks, addressed in c5ddfe8: applied the requested fix and re-checked the recipe/scripts.

conda-recipe/meta.yaml
```diff
@@ -109,6 +147,7 @@ outputs:
         - mkl-devel
         - mkl-static
         - mkl-devel-dpcpp  # [ cxx_compiler == "dpcpp" ]
+        - {{ pin_subpackage('dal-gpu', exact=True) }}  # [ cxx_compiler == "dpcpp" ]
```
david-cortes-intel: Something here is not working correctly. It still gets pulled when it shouldn't. Either that, or `dal-gpu` is being built when it should.

conda-recipe/meta.yaml
```diff
@@ -62,19 +66,47 @@ requirements:
...
         - test -f $PREFIX/lib/libonedal_core.so.{{ major_binary_version }}  # [linux]
-        - test -f $PREFIX/lib/libonedal_dpc.so.{{ major_binary_version }}  # [linux]
         - test -f $PREFIX/lib/libonedal_thread.so.{{ major_binary_version }}  # [linux]
     about:
-      summary: oneAPI Data Analytics Library (oneDAL) runtime libraries
+      summary: oneAPI Data Analytics Library (oneDAL) CPU runtime libraries
+
+  - name: dal-gpu
+    build:
+      skip: true  # [cxx_compiler != "dpcpp" or not linux64]
```
david-cortes-intel, line 90: This one has:
```
[cxx_compiler != "dpcpp" or not linux64]
```

But `dal-devel` has:
```
{{ pin_subpackage('dal-gpu', exact=True) }}  # [ cxx_compiler == "dpcpp" ]
```

conda-recipe/test-devel.sh
```diff
@@ -57,6 +57,7 @@ run_examples() {
             fi
             cmake_args="$cmake_args $extra_cmake_args"
 
+            # shellcheck disable=SC2086
```
david-cortes-intel, line 60: What is this supposed to do?

david-cortes-intel:
CI is still running into the same error where testing of `dal-devel` fails due to not being able to find `dal-gpu`.

conda-recipe/meta.yaml
```diff
@@ -62,19 +66,47 @@ requirements:
 
 outputs:
   - name: dal
+    build:
+      # Suppress dpcpp_impl run_exports for the CPU-only package:
+      # dal must not pull GPU/SYCL runtime (intel-sycl-rt).
+      ignore_run_exports_from:
+        - dpcpp_impl_linux-64  # [linux64 or win]
```
david-cortes-intel, line 73: The one for windows would be named `dpcpp_impl_win-64`

conda-recipe/meta.yaml
```diff
@@ -108,7 +146,8 @@ outputs:
         - dal-static =={{ version }}
```
david-cortes-intel, line 146: It has:
```
dal-static =={{ version }}
````

and
```
{{ pin_subpackage('dal-gpu', exact=True) }}  # [cxx_compiler == "dpcpp" and linux64]
```

Shouldn't they both use the same syntax, since both are from the same recipe?

**Comment by napetrov** [comment]
Addressed the remaining review follow-ups in `36fbf1deb`:

- split `dpcpp_impl` package selectors by platform: `dpcpp_impl_linux-64` for Linux and `dpcpp_impl_win-64` for Windows
- made `dal-static` use the same `pin_subpackage(..., exact=True)` style as `dal-gpu`
- clarified in `test-devel.sh` that sourcing `env/vars.sh` is needed to expose oneDAL env/pkg-config/CMake helpers during example builds

I did not resolve the threads blindly where CI evidence is still pending, but the code-review items above are now addressed in the branch.

**Comment by david-cortes-intel** [comment]
@napetrov Since this now has some entries to accommodate builds on windows but the recipe as a whole doesn't actually work on windows, would it throw some informative error like "platform not supported" if a user tries to build it on windows?

**Comment by napetrov** [comment]
Addressed in f42d77971: removed Windows selectors/dependencies from this recipe and made the top-level build skip every non-linux64 platform. The recipe is Linux-only now, so Windows builds are skipped up front instead of trying to resolve dpcpp_impl_win-64 for Linux-only build/test scripts.

**Comment by david-cortes-intel** [comment]
Still getting the same error about dependencies for tests:
```
Could not solve for environment specs
The following package could not be installed
└─ dal-gpu =2025.5.0 0 does not exist (perhaps a typo or a missing channel).
```

Also getting these warning:
```
/localdisk2/mkl/dcortes/miniforge3/lib/python3.12/site-packages/conda_build/environ.py:546: UserWarning: The environment variable 'REQCPU' is being passed through with value 'sse2 avx2 avx512'.  If you are splitting build and test phases with --no-test, please ensure that this value is also set similarly at test time.
  warnings.warn(
/localdisk2/mkl/dcortes/miniforge3/lib/python3.12/site-packages/conda_build/environ.py:546: UserWarning: The environment variable 'CONDA_CXX_COMPILER' is being passed through with value 'icx'.  If you are splitting build and test phases with --no-test, please ensure that this value is also set similarly at test time.
```

david-cortes-intel (approved):
I can verify that it works in the sense that conda-build doesn't error out. But I don't know to test the resulting package. Maybe @ethanglaser could review too.

napetrov:
I see that package can be installed.  Though we might need to play more with the pipeline to validate it if we want. 
Would goahead with integration and we can follow-up with pipeline later on

PR #3564, opened 2026-03-17
perf: AMX BF16 dispatch in EuclideanDistances::computeABt

napetrov:
## AMX-BF16 fast path for float32 Euclidean distance GEMM

This PR adds an opt-in AMX-BF16 execution path for the float32 GEMM used inside Euclidean distance computation, specifically `EuclideanDistances::computeABt` (`A * B'`). This path is used by brute-force KNN and DBSCAN distance computation.

By default, behavior is unchanged: oneDAL continues to use the existing full-float32 path. Reduced precision is used only when explicitly requested and when the operation/hardware are eligible.

## User-visible behavior

Default mode is strict float32:

```sh
# default behavior; no BF16
ONEDAL_FLOAT32_MATMUL_PRECISION unset
```

Opt in through the environment:

```sh
ONEDAL_FLOAT32_MATMUL_PRECISION=ALLOW_BF16
```

For experiments/frontends that want to force the implemented BF16 path on supported hardware:

```sh
ONEDAL_FLOAT32_MATMUL_PRECISION=REQUIRE_BF16
```

The same policy can also be set programmatically through the internal API:

```cpp
daal_set_float32_matmul_precision(daal::internal::Float32MatmulPrecision::allow_bf16);
daal_set_float32_matmul_precision(daal::internal::Float32MatmulPrecision::require_bf16);
```

The API is intentionally internal for now. `Float32MatmulPrecision` lives in `cpp/daal/src/services/service_defines.h` under `daal::internal`; it is not exposed through installed public headers yet.

## Precision policy interface

`daal::internal::Float32MatmulPrecision` currently has three modes:

| Mode | Behavior |
|---|---|
| `strict` | Default. Always use the existing float32 path (`sgemm` for float, `dgemm` for double). No reduced-precision arithmetic. |
| `allow_bf16` | Permit oneDAL to use BF16 operands for implemented operations when hardware and the conservative size heuristic allow it. Currently this means float32 Euclidean GEMM on AMX-BF16 hardware with `m`, `n`, and `k` all at least 64. Accumulation and output remain float32. |
| `require_bf16` | Internal/experimental opt-in for benchmarking or frontend-controlled execution. For operations that have an implemented BF16 path and can execute it on the current hardware, use BF16 even when the size heuristic would otherwise skip it. Unsupported algorithms, unsupported hardware, and allocation failures still fall back to the regular path. |

The getter/setter store only the requested precision policy. They do not perform hardware checks.

Hardware availability is queried separately through:

```cpp
daal_has_amx_bf16()
```

AMX-BF16 detection is integrated into the existing CPU feature detection infrastructure in `env_detect_features.cpp`.

## Dispatch behavior

The dispatch logic is encapsulated in `BF16GemmDispatcher` in `cpp/daal/src/algorithms/service_kernel_math.h`.

The primary template keeps the existing BLAS behavior for non-float types. The `float` specialization does the BF16 dispatch when all gates pass:

1. requested precision policy allows BF16 (`allow_bf16` or `require_bf16`),
2. AMX-BF16 is available,
3. the operation is the implemented Euclidean distance GEMM path,
4. for `allow_bf16`, `m`, `n`, and `k` are all at least 64,
5. temporary BF16 buffers can be allocated.

When BF16 is selected, float32 inputs are converted to BF16 with round-to-nearest-even conversion, then dispatched through `cblas_gemm_bf16bf16f32`. Accumulation and output remain float32.

If any gate fails, the code transparently falls back to the existing `sgemm` path. This includes non-AMX hardware, unsupported architectures, small matrices in `allow_bf16` mode, unsupported algorithms, and allocation failures.

## Accuracy and compatibility

The default `strict` mode preserves existing behavior and numerical results.

In BF16 modes, only operands are reduced to BF16; accumulation and output are float32. The BF16 path is limited to Euclidean distance GEMM and is opt-in. On machines without AMX-BF16, the BF16 policy is accepted but execution falls back to the regular float32 path.

## Benchmark results

Hardware: Intel Xeon 6975P-C (Granite Rapids), KVM, AMX-BF16 enabled.  
Method: scikit-learn_bench / sklearnex, float32, CPU backend.

Observed speedups for feature dimensions where the BF16 path is eligible:

| Workload | Speedup range |
|---|---|
| DBSCAN euclidean, feat >= 64 | ~1.2x–1.6x depending on vCPU count and `n_jobs` |
| KNN classification inference, feat >= 128 | ~1.1x–1.4x depending on vCPU count |

No regressions were observed for paths that do not hit the BF16 gate, such as KNN training, KNN regression with low feature count, and DBSCAN with low feature count.

## Files changed

| File | Change |
|---|---|
| `cpp/daal/src/services/service_defines.h` | Internal `Float32MatmulPrecision` enum and precision-control declarations. |
| `cpp/daal/src/services/compiler/generic/env_detect_precision.cpp` | Environment parsing and getter/setter implementation for the requested precision policy. |
| `cpp/daal/src/services/compiler/generic/env_detect_features.cpp` | AMX-BF16 hardware capability detection through the existing CPU feature mechanism. |
| `cpp/daal/src/services/cpu_type.h` | Adds `CpuFeature::amx_bf16`. |
| `cpp/daal/src/algorithms/service_kernel_math.h` | Adds `BF16GemmDispatcher` and wires `EuclideanDistances::computeABt` through it. |

## Summary

This PR keeps oneDAL's default numerical behavior unchanged while adding an internal, opt-in BF16 policy for Euclidean distance GEMM. The implementation centralizes hardware/policy dispatch, keeps fallback behavior transparent, and provides a path for sklearnex or other internal frontends to enable AMX-BF16 acceleration when appropriate.

CI results:
Public - https://github.com/uxlfoundation/oneDAL/pull/3564/checks?check_run_id=82147985037
Private - http://intel-ci.intel.com/f16b119b-3e3f-f14e-bad6-d4f5ef20c6a0


cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -292,8 +293,32 @@ class EuclideanDistances : public PairwiseDistances<FPType, cpu>
 
         return safeStat.detach();
     }
-
-    // compute (A x B')
+#ifndef DAAL_REF
+    // AMX-BF16 capability check (MKL builds only; cross-platform)
+    static bool knn_has_amx_bf16()
```
Vika-F: It would be better to extend the function we already have for CPU features detection, if needed:
https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/src/services/compiler/generic/env_detect_features.cpp#L303

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -122,6 +124,94 @@ class PairwiseDistances
...
+template <CpuType cpu>
+struct BF16GemmDispatcher<float, cpu>
+{
+    static void compute(const float * a, const float * b, DAAL_INT m, DAAL_INT n, DAAL_INT k, float * out)
+    {
+        if (g_use_bf16_gemm && m >= kBF16MinDim && n >= kBF16MinDim && k >= kBF16MinDim)
+        {
+            const size_t szA = static_cast<size_t>(n) * static_cast<size_t>(k);
+            const size_t szB = static_cast<size_t>(m) * static_cast<size_t>(k);
+            MKL_BF16 * a16   = static_cast<MKL_BF16 *>(mkl_malloc(szA * sizeof(MKL_BF16), 64));
```
david-cortes-intel: We have smart pointer classes to use. Would prevent potential errors from this sort of memory management.

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -122,6 +124,94 @@ class PairwiseDistances
...
+            MKL_BF16 * b16   = static_cast<MKL_BF16 *>(mkl_malloc(szB * sizeof(MKL_BF16), 64));
+            if (a16 && b16)
+            {
+                // BF16 = upper 16 bits of IEEE-754 float32; use memcpy to read bits safely.
+                for (size_t i = 0; i < szA; ++i)
+                {
+                    uint32_t bits;
+                    memcpy(&bits, &a[i], sizeof(bits));
+                    a16[i] = static_cast<MKL_BF16>(bits >> 16);
```
david-cortes-intel: I believe MKL itself should have utilities for this sort of thing.

cpp/daal/src/services/compiler/generic/env_detect_features.cpp
```diff
@@ -300,6 +300,18 @@ DAAL_EXPORT int daal_enabled_cpu_detect()
             result |= feature;                                             \
         }
 
+/// Check if AMX-BF16 is available: CPUID(7,0).EDX[22] + XCR0[17:18] tile state enabled by OS.
+static int check_amx_bf16_features()
+{
+    /* CPUID.(EAX=07H, ECX=0H):EDX.AMX-BF16[bit 22]==1 */
+    if (!check_cpuid(7, 0, 3, (1 << 22)))
```
david-cortes-intel, line 307: Would this segfault or otherwise break if executed on AMD? Is there perhaps some check that tells whether it's an intel CPU?

cpp/daal/src/services/compiler/generic/env_detect_precision.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+/// Hardware capability flag — evaluated ONCE at process start.
+/// True iff AMX-BF16 is present and OS-enabled.
+/// AMX-BF16 is x86-only; always false on ARM/RISCV64.
+/// daal_serv_cpu_feature_detect() returns a DAAL_UINT64 bitmask;
+/// amx_bf16 corresponds to bit 5 (see daal::internal::CpuFeature::amx_bf16).
+static bool detect_hw_amx_bf16()
+{
+#if defined(TARGET_X86_64) || defined(__x86_64__) || defined(_M_X64)
+    // AMX-BF16 bit = (1ULL << 5) in daal::internal::CpuFeature (cpu_type.h).
```
david-cortes-intel: Sounds like it should be moved to that file instead.

CC @Vika-F for comments.

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -122,6 +124,94 @@ class PairwiseDistances
...
+            const size_t szB = static_cast<size_t>(m) * static_cast<size_t>(k);
+            MKL_BF16 * a16   = static_cast<MKL_BF16 *>(mkl_malloc(szA * sizeof(MKL_BF16), 64));
+            MKL_BF16 * b16   = static_cast<MKL_BF16 *>(mkl_malloc(szB * sizeof(MKL_BF16), 64));
+            if (a16 && b16)
+            {
+                // BF16 = upper 16 bits of IEEE-754 float32; use memcpy to read bits safely.
+                for (size_t i = 0; i < szA; ++i)
+                {
+                    uint32_t bits;
+                    memcpy(&bits, &a[i], sizeof(bits));
```
david-cortes-intel: @Vika-F Is it a problem to have stdlib calls like this in a header?

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -24,6 +24,7 @@
 #ifndef __SERVICE_KERNEL_MATH_H__
 #define __SERVICE_KERNEL_MATH_H__
 
+#include <cstring>
```
david-cortes-intel: This is including a C++ header, but then calling the functions from it without `std::`.

david-cortes-intel (changes requested):
This would require benchmarks to assess whether it leads to improved runtimes. Less precision could potentially result in requiring more iterations for convergence.

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -46,6 +47,7 @@
 #include "src/externals/service_math.h"
 #include "src/services/service_profiler.h"
 
+#include <mutex>
```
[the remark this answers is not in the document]
Vika-F: Agree on this.

cpp/daal/src/algorithms/service_kernel_math.h
Vika-F: Yes, it's better to use TArray or other classes defined here:
https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/src/services/service_arrays.h#L109

cpp/daal/src/algorithms/service_kernel_math.h
Vika-F: memcpy is deprecated due to security reasons. daal_memcpy_s should be used:
https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/include/services/daal_memory.h#L71

cpp/daal/src/algorithms/service_kernel_math.h
Vika-F: It looks like no. Intrinsics like _mm512_cvtneps_pbh might be faster, but I am Ok to have it like this for now.

david-cortes-intel: Shouldn't it have `#pragma omp simd` then?

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -122,6 +124,94 @@ class PairwiseDistances
...
+                    a16[i] = static_cast<MKL_BF16>(bits >> 16);
+                }
+                for (size_t i = 0; i < szB; ++i)
+                {
+                    uint32_t bits;
+                    memcpy(&bits, &b[i], sizeof(bits));
+                    b16[i] = static_cast<MKL_BF16>(bits >> 16);
+                }
+                // col-major: C = B16 * A16^T  =>  cblas NoTrans B, Trans A
+                cblas_gemm_bf16bf16f32(CblasColMajor, CblasNoTrans, CblasTrans, m, n, k, 1.0f, b16, m, a16, n, 0.0f, out, m);
```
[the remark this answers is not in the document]
Vika-F: This is wrong because xxgemm uses Fortran memory alignment which is column-major, and  cblas_gemm_bf16bf16f32 uses C memory alignment which is row-major.

cpp/daal/src/algorithms/service_kernel_math.h
```diff
@@ -131,7 +221,7 @@ class EuclideanDistances : public PairwiseDistances<FPType, cpu>
         : _a(a), _b(b), _squared(squared), _isSqrtNorm(isSqrtNorm)
     {}
 
-    ~EuclideanDistances() override {}
+    virtual ~EuclideanDistances() override {}
```
Vika-F: Please remove redundant "virtual" here and in other similar places.

cpp/daal/src/services/compiler/generic/env_detect_features.cpp
Vika-F: It is Ok, as it is checked at line 318 below.

david-cortes-intel: @Vika-F Is it possible for the call to `check_cpuid` to segfault?

david-cortes-intel: Actually I see now that it does check for intel CPU.

cpp/daal/src/services/compiler/generic/env_detect_precision.cpp
Vika-F: Yes, it's better to move `detect_hw_amx_bf16()` and `g_hw_amx_bf16` into env_detect_features.cpp.

I am also not sure that the functions in this file really need DAAL_EXPORT. Because they are only used internally in DAAL (i.e. libonedal_core.so).

Vika-F: Regarding AMX-BF16 bit = (1ULL << 5).
I think it's better not to use magic constants here, and use:
static const DAAL_UINT64 amx_bf16_mask = static_cast<DAAL_UINT64>(daal::internal::amx_bf16);

cpp/daal/src/algorithms/service_kernel_math.h
david-cortes-intel: Header shouldn't be needed anymore if using the internal versions of memcpy and related.

napetrov:
Added a third precision-policy mode in 466f40e04: strict / allow_bf16 / require_bf16.\n\nSemantics:\n- strict: existing float32 path only\n- allow_bf16: use AMX-BF16 only when hardware supports it and the conservative performance heuristic passes (currently dims >= 64)\n- require_bf16: for kernels with an implemented BF16 path on AMX-BF16 hardware, use BF16 even when the performance-size heuristic would otherwise skip it; unsupported algorithms/hardware/allocation failures still fall back to the regular path\n\nEnv var now accepts ONEDAL_FLOAT32_MATMUL_PRECISION=ALLOW_BF16 or REQUIRE_BF16; programmatic setter supports Float32MatmulPrecision::require_bf16 as well.

napetrov:
Interface/behavior summary for review:

This PR adds an internal precision-control interface for float32 GEMM-like paths, currently used only by `EuclideanDistances::computeABt` (the `A * B'` part used by brute-force KNN and DBSCAN distance computation).

The new internal enum is `daal::internal::Float32MatmulPrecision` in `cpp/daal/src/services/service_defines.h`. It is intentionally not a public installed-header API yet. The modes are:

| Mode | Behavior |
|---|---|
| `strict` | Default. Always uses the existing float32 path (`sgemm` for float, `dgemm` for double). No reduced-precision arithmetic. |
| `allow_bf16` | Allows oneDAL to use BF16 operands for implemented operations when hardware and the conservative size heuristic allow it. Currently this means float32 Euclidean GEMM on AMX-BF16 hardware with `m`, `n`, and `k` all at least 64. Accumulation and output remain float32. |
| `require_bf16` | Experimental/internal opt-in for frontends or benchmarking. For operations that have an implemented BF16 path and can execute it on the current hardware, use BF16 even if the size heuristic would normally skip it. Unsupported algorithms, unsupported hardware, and allocation failures still fall back to the regular float32 path. |

The requested mode can be set through `ONEDAL_FLOAT32_MATMUL_PRECISION=ALLOW_BF16` or `REQUIRE_BF16`, or programmatically through `daal_set_float32_matmul_precision(...)`. The getter/setter store only the requested policy; they do not perform hardware checks. AMX-BF16 availability is exposed separately through `daal_has_amx_bf16()` and is detected through the existing CPU feature detection path.

Dispatch behavior is encapsulated in `BF16GemmDispatcher` in `service_kernel_math.h`. The primary template keeps the existing BLAS path for non-float types. The `float` specialization checks the precision policy plus AMX-BF16 support, converts float32 operands to BF16 with round-to-nearest-even, calls `cblas_gemm_bf16bf16f32`, and writes float32 output. If the policy is `strict`, AMX-BF16 is unavailable, dimensions are below the `allow_bf16` threshold, or scratch allocation fails, it transparently falls back to the existing `sgemm` path.

So the external behavior is unchanged by default. Reduced precision is opt-in, limited to the Euclidean distance GEMM path, and should be safe to disable or unavailable on non-AMX machines without changing callers.

david-cortes-intel:
@napetrov Is it possible to reach this new code path when calling oneDAL from Python?

napetrov:
> @napetrov Is it possible to reach this new code path when calling oneDAL from Python?

yes, and the plan is to have API in python to set it

napetrov:
i think this is good to  go in this form and can be tunned later on

cpp/oneapi/dal/algo/knn/test/internal.cpp
```diff
@@ -0,0 +1,63 @@
+/*******************************************************************************
+* Copyright 2021 Intel Corporation
```
Vika-F:
```suggestion
* Copyright 2026 Intel Corporation
```

.ci/pipeline/ci.yml
```diff
@@ -543,6 +543,13 @@ jobs:
                  --test_thread_mode=par
     displayName: 'cpp-tests-thread-dev'
 
+  - script: |
+      bazel test //cpp/oneapi/dal:tests \
+                 --config=host-private \
+                 --test_link_mode=dev \
+                 --test_thread_mode=par
+    displayName: 'cpp-tests-thread-dev-private'
+
```
Vika-F: Remove unneeded changes.
```suggestion
```

Vika-F (approved):

david-cortes-intel (approved):

PR #3608, opened 2026-04-13
Experiment: Add linker version scripts to the build for better ABI control 

Vika-F:
## Description

Work in progress!

The scripts are going to have more details; also need to cover Bazel and Windows builds.

Inspired by oneTBB's approach regarding ABI control:
https://github.com/uxlfoundation/oneTBB/tree/master/src/tbb/def

---

<details>
<summary>Checklist:</summary>

- [ ] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.

**Testing**

**Performance**

- [ ] I have provided justification why performance and/or quality metrics have changed or why changes are not expected.

</details>


david-cortes-intel:
@Vika-F Aren't these .def files only recognized by msvc by default? And shouldn't they also have `LIBRARY` and `EXPORTS`?

Vika-F:
> Aren't these .def files only recognized by msvc by default?

You are partially right. I think the extension is misleading, .map is a more common variant for Linux, and .def - for Windows.

For now I am implementing Linux version using ld version script files, the syntax is different for those:
https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html

I will need to implement similar functionality for MSVC, with `LIBRARY` and `EXPORTS` sections, yes. As far as I understand those files are not cross-platform :(

david-cortes-intel:
> > Aren't these .def files only recognized by msvc by default?
> 
> You are partially right. I think the extension is misleading, .map is a more common variant for Linux, and .def - for Windows.
> 
> For now I am implementing Linux version using ld version script files, the syntax is different for those: https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html
> 
> I will need to implement similar functionality for MSVC, with `LIBRARY` and `EXPORTS` sections, yes. As far as I understand those files are not cross-platform :(

Got it. Then let's better name them .map to avoid confusions.

PR #3615, opened 2026-04-29
Enable Bazel build and examples on Windows

napetrov:
Draft Windows Bazel bring-up for oneDAL.

Summary:
- add Windows extra/cc toolchain configuration and register the local MSVC toolchain via rules_cc
- add Windows MKL/TBB prebuilt dependency handling using win_amd64 wheels
- include all Windows MKL/TBB runtime DLLs needed by examples/tests
- make release/pkg-config/vars generation work on Windows
- relax Linux-specific PIC/static-library assumptions for MSVC and handle Windows dynamic/interface libraries
- use response files for Windows lib.exe static-library merging to avoid CreateProcessW command-line limits
- add MSVC /utf-8 where needed for fmt-based test/example builds
- fix DAAL example suite wiring on Windows/Linux Bazel path
- add a WindowsBazel GitHub Actions job that runs release build, example smoke builds, and one host test smoke on windows-2022
- document Windows Bazel setup and usage in `dev/bazel/README.md` and update CPU feature docs to mention Windows CPU/MSVC support

Validation on Windows VM:
- `bazelisk build //cpp/daal:core_dynamic --verbose_failures --jobs=16` passes
- `bazelisk build //cpp/oneapi/dal:dynamic --verbose_failures --jobs=16` passes
- `bazelisk build //:release --verbose_failures --jobs=16` passes
- `bazelisk build //examples/oneapi/cpp:basic_statistics_dense_batch_host //examples/daal/cpp:low_order_moms_dense_batch_host --verbose_failures --jobs=16` passes
- `bazelisk test //cpp/daal/src/algorithms/dtrees/gbt/regression:test_gbt_regression_model_builder_unit_host --verbose_failures --test_output=errors --jobs=16` passes with `BAZEL_SH`, MKL/TBB `bin` on `PATH`, and `MKL_THREADING_LAYER=TBB`

Out of scope for this PR:
- Windows DPC++/SYCL Bazel support. Current failure is missing `sycl/sycl.hpp` / oneAPI DPC++ toolchain wiring. Tracked in the master Make-to-Bazel migration issue: https://github.com/uxlfoundation/oneDAL/issues/3530 . It covers `dpcpp`/`icx-cl`, SYCL headers/libs/runtime DLLs, and device flags separately.

Notes:
- This is intentionally draft/WIP; Linux regression testing and Make-vs-Bazel release artifact comparison are still pending.


MODULE.bazel
```diff
@@ -207,6 +222,21 @@ mkl_repo(
...
+        "https://files.pythonhosted.org/packages/86/60/f979218ad807331524f3cd88c05b603d9ea5a685cffa513304bee8ae012b/mkl_devel-2025.2.0-py2.py3-none-win_amd64.whl",
+        "https://files.pythonhosted.org/packages/06/87/3eee37bf95c6b820b6394ad98e50132798514ecda1b2584c71c2c96b973c/mkl_include-2025.2.0-py2.py3-none-win_amd64.whl",
+    ],
+    win_sha256s = [
+        "b6ec153e4a073421dbb52ef99c7be97e66cde0272e4a1e3569b090b6f0130253",
+        "305745583d7b08d2f8b8b37d20e6fa4b4325627a5989625c74aaaf651b10e9da",
+        "d20305b4adfa36407a808ec6a16dc5d6da6f8b9cb4a96bdcc0e0ab3239c43816",
+    ],
+    win_strip_prefixes = [
+        "mkl-2025.2.0.data/data/Library",
```
david-cortes-intel: Why these particular versions?

docs/source/contribution/cpu_features.rst
```diff
@@ -275,8 +275,9 @@ file contains the compiler options for the GNU compiler for x86-64 architecture
 Bazel
 -----
 
-For now, Bazel build is supported only for Linux x86-64 platform
-It provides ``cpu`` `option <https://github.com/uxlfoundation/oneDAL/tree/main/dev/bazel#bazel-options>`_
+Bazel build supports Linux x86-64 and CPU/MSVC builds on Windows x86-64.
```
david-cortes-intel: Does this mean that ICX is not supported on windows?

dev/bazel/README.md
```diff
@@ -225,13 +273,34 @@ The most used Bazel commands are `build`, `test` and `run`.
...
+### Build release package
+- On Linux:
+  ```sh
+  bazel build //:release --verbose_failures --jobs=$(nproc)
+  ```
+
+- On Windows, run from the Visual Studio Developer Command Prompt described
+  above:
+  ```bat
+  bazelisk.exe build //:release --verbose_failures --jobs=%NUMBER_OF_PROCESSORS%
```
david-cortes-intel: Is there perhaps some argument or variable provided by bazelisk that would signal "all available threads"?

dev/bazel/README.md
```diff
@@ -43,6 +43,54 @@
...
+   ```bat
+   call "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\Common7\Tools\VsDevCmd.bat" -arch=amd64 -host_arch=amd64
+   bazelisk.exe build //:release --verbose_failures --jobs=%NUMBER_OF_PROCESSORS%
+   ```
+
+5. For tests that run executables linked with MKL/TBB, make the downloaded
+   runtime DLLs visible through `PATH` and pass the environment to Bazel tests.
+   Bazel downloads these dependencies into its external repository cache.
+   ```powershell
+   $outputBase = (.\bazelisk.exe info output_base).Trim()
```
david-cortes-intel: This is powershell while other commands were for CMD. Does it require a mixture of the two?

docs/source/contribution/cpu_features.rst
napetrov: not yet - i think it would go as separate PR to not blow this one in size .

MODULE.bazel
napetrov: Just random version selected by agent to get this working.  would be aligned with linux version

dev/bazel/README.md
napetrov: it would take all by default - we can just remove this extra

dev/bazel/README.md
napetrov: yes, make sense to align to cmd

MODULE.bazel
david-cortes-intel: Version 2026.0 is already out. Should probably use that instead, but not sure about the status of bazel builds with latest MKL (CC @Alexandr-Solovev ).

dev/bazel/README.md
napetrov: Addressed in 18666a88b: converted the Windows setup/test snippets to cmd syntax so the README no longer mixes PowerShell and cmd.

dev/bazel/README.md
napetrov: Addressed in 18666a88b: removed the explicit --jobs examples and let Bazel use its default parallelism.

docs/source/contribution/cpu_features.rst
napetrov: Addressed in 18666a88b: clarified that this PR enables regular Windows x86-64 CPU/MSVC Bazel builds; Windows ICX and DPC++/SYCL remain separate follow-ups.

MODULE.bazel
napetrov: Addressed in 18666a88b: moved the Windows classic MKL PyPI wheels to 2025.3.0 (mkl, mkl_devel, mkl_include) with updated sha256s, and documented the scope as Windows CPU/MSVC dependencies.

dev/bazel/cc.bzl
```diff
@@ -198,8 +241,30 @@ def _cc_dynamic_lib_impl(ctx):
...
+                is_windows = is_windows,
+                extra_inputs = [dynamic_outputs.interface_library] if dynamic_outputs.interface_library else [],
+            ))
+        if dynamic_outputs.interface_library:
+            default_files.append(_copy_dynamic_release_file(
+                ctx,
+                dynamic_outputs.interface_library,
+                "{}_dll.lib".format(ctx.attr.lib_name),
+                is_windows = is_windows,
+            ))
```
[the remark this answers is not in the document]
napetrov: Addressed in d4038725a: on Windows the dynamic library is now linked with the versioned DLL name up front, so the generated import library records the same version-suffixed DLL name that is packaged in the release. The copy step is kept only as a fallback if the linker output name differs.

dev/bazel/scripts.bzl
```diff
@@ -87,44 +73,58 @@ def generate_vars_sh(name, out = "env/vars.sh", **kwargs):
...
+            content = """prefix=${{pcfiledir}}/../../
+exec_prefix=${{prefix}}
+libdir=${{exec_prefix}}/lib/intel64
+includedir=${{prefix}}/include
+
+Name: oneDAL
+Description: oneAPI Data Analytics Library
+Version: {major}.{minor}
+URL: https://www.intel.com/content/www/us/en/developer/tools/oneapi/onedal.html
+Libs: {onedal_libs} mkl_core.lib mkl_intel_lp64.lib mkl_tbb_thread.lib tbb12.lib tbbmalloc.lib
```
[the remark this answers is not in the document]
napetrov: Addressed in d4038725a: changed the Windows pkg-config `Libs:` entry from `mkl_intel_lp64.lib` to `mkl_intel_ilp64.lib` to match the Bazel/Make Windows MKL linkage.

docs/source/contribution/cpu_features.rst
```diff
@@ -275,8 +275,10 @@ file contains the compiler options for the GNU compiler for x86-64 architecture
 Bazel
 -----
 
-For now, Bazel build is supported only for Linux x86-64 platform
-It provides ``cpu`` `option <https://github.com/uxlfoundation/oneDAL/tree/main/dev/bazel#bazel-options>`_
+Bazel build supports Linux x86-64. This PR additionally enables the regular
+CPU build with the MSVC compiler on Windows x86-64. ICX and DPC++/SYCL Bazel
+support on Windows are tracked separately and are not enabled yet. Bazel
+provides the ``cpu`` `option <https://github.com/uxlfoundation/oneDAL/tree/main/dev/bazel#bazel-options>`_
```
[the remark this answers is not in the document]
napetrov: Addressed in d4038725a: rephrased the documentation as a timeless statement of current Bazel platform support, without referring to this PR.

dev/bazel/cc/link.bzl
```diff
@@ -126,6 +217,7 @@ def _static(owner, name, actions, cc_toolchain,
     linker_input = cc_common.create_linker_input(
         owner = owner,
         libraries = depset([static_lib_to_link] +
+                           unpacked_linking_context.static_libraries_to_link +
```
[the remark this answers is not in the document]
napetrov: Addressed in d4038725a: dropped the already-merged `static_libraries_to_link` from the propagated linker input to avoid re-exporting the same static archives downstream.

.github/workflows/ci.yml
```diff
@@ -34,6 +34,55 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+
+  WindowsBazel:
```
Alexandr-Solovev: Does it make sense to move the ci bazel in azure, because original LinuxBazel step in azure pipeline

dev/bazel/deps/mkl.bzl
```diff
@@ -33,6 +33,19 @@ mkl_repo = repos.prebuilt_libs_repo_rule(
...
     ],
     build_template = "@onedal//dev/bazel/deps:mkl.tpl.BUILD",
+    win_includes = [
+        "include",
+    ],
+    win_libs = [
+        "lib/mkl_core.lib",
+        "lib/mkl_intel_ilp64.lib",
+        "lib/mkl_tbb_thread.lib",
+        "lib/mkl_rt.lib",
```
Alexandr-Solovev: Not sure that rt lib is necessary here, we dont use it in make

dev/bazel/deps/mkl.tpl.BUILD
```diff
@@ -24,9 +24,9 @@ cc_library(
         # The source libraries have circular symbol dependencies. To successfully build this cc_library,
         # oneMKL requires wrapping the libraries with -Wl,--start-group and -Wl,--end-group.
         "-Wl,--start-group",
-        "$(location lib/libmkl_core.a)",
-        "$(location lib/libmkl_intel_ilp64.a)",
-        "$(location lib/libmkl_tbb_thread.a)",
+        "%{repo_root}/lib/libmkl_core.a",
```
Alexandr-Solovev, line 29: If it somehow impact windows, I think .a should be replaced with smth like .a if linux, .lib if win, as well as $(location) for lin and % for win

dev/bazel/cc/link.bzl
```diff
@@ -43,8 +43,24 @@ def _filter_user_link_flags(feature_configuration, user_link_flags):
...
+        args.set_param_file_format("multiline")
+        args.add("/NOLOGO")
+        args.add("/OUT:" + output_file.path)
+        args.add_all(static_libs)
+        actions.run(
+            executable = "lib.exe",
+            arguments = [args],
+            inputs = static_libs,
+            outputs = [output_file],
+            mnemonic = "MergeStaticLibraries",
```
Alexandr-Solovev, line 60: I am not sure, that in perfect case it sill necessary, because the merging libs has been removed for make. May be worth to remove in the next pr

Alexandr-Solovev (commented):
One more thing about compatibility, https://github.com/uxlfoundation/oneDAL/blob/main/.bazelrc#L35 I think makes sense to extend it for windows with PATH likely

dev/bazel/deps/mkl.bzl
napetrov: Addressed in 7886e87f5: removed `mkl_rt.lib` from the Windows MKL repo inputs and dropped the `mkl_rt` import from the Windows MKL BUILD template. The Windows CPU/MSVC path now uses the same static MKL interface/core/TBB-thread set as Make.

dev/bazel/deps/mkl.tpl.BUILD
napetrov: This template is Linux-only; Windows uses `dev/bazel/deps/mkl_win.tpl.BUILD` through `win_build_template`, so the `.a` paths here do not affect the Windows build. I would keep the Linux absolute-path cleanup separate unless we want to touch the existing Linux Bazel MKL linkage in this PR.

dev/bazel/cc/link.bzl
napetrov: Agree, this can likely be revisited as a follow-up. For this PR I kept the merge helper in place because the current Bazel rules still rely on it for the existing static-library flow; removing it would be a broader behavior change than the Windows bring-up.

.github/workflows/ci.yml
napetrov: Good point. I kept this as a GitHub Actions smoke job for the bring-up because there is no existing Windows Bazel Azure job yet, while the current Azure LinuxBazel coverage stays unchanged. Moving/duplicating this into Azure makes sense as a follow-up once the Windows Bazel environment is available there.

napetrov:
Addressed the latest review notes in 7886e87f5: added `PATH` to Bazel test env for Windows runtime DLL lookup and removed the unused `mkl_rt` dependency from the Windows MKL CPU Bazel setup.

.github/workflows/ci.yml
napetrov: Updated in fc7588d28: moved the Windows Bazel smoke coverage from GitHub Actions into the Azure pipeline next to the existing LinuxBazel job, and removed the temporary GitHub Actions WindowsBazel job.

dev/bazel/deps/mkl.tpl.BUILD
napetrov: Updated in fc7588d28: added an explicit note that `mkl.tpl.BUILD` is the non-Windows MKL template and Windows uses `mkl_win.tpl.BUILD`, so the `.a` paths here are not used by the Windows Bazel path.

napetrov:
Updated directly in this PR with fc7588d28: moved WindowsBazel smoke coverage into Azure CI and removed the temporary GitHub Actions job. Also documented that `mkl.tpl.BUILD` is non-Windows-only; Windows continues to use `mkl_win.tpl.BUILD`.

.ci/pipeline/ci.yml
```diff
@@ -584,6 +584,73 @@ jobs:
...
+    displayName: 'example smoke build'
+
+  - script: |
+      call "$(VSDEVCMD)" -arch=amd64 -host_arch=amd64
+      set BAZEL_SH=$(BAZEL_SH)
+      "$(BAZELISK_EXE)" info output_base > output_base.txt
+      set /p OUTPUT_BASE=<output_base.txt
+      set "OUTPUT_BASE=%OUTPUT_BASE:/=\%"
+      set PATH=%OUTPUT_BASE%\external\+mkl_repo+mkl\bin;%OUTPUT_BASE%\external\+tbb_repo+tbb\bin;%PATH%
+      set MKL_THREADING_LAYER=TBB
```
Alexandr-Solovev: Looks like a redundant part, may be good thing = aligning steps as much as possible with LinuxBazel

.github/workflows/ci.yml
```diff
@@ -34,6 +34,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+
```
Alexandr-Solovev: unnecessary

cpp/daal/BUILD
```diff
@@ -244,6 +244,9 @@ daal_dynamic_lib(
     deps = [
         ":core",
         ":all_algorithms",
+        # Match Make: onedal_core.dll contains the DAAL threading implementation
```
Alexandr-Solovev: Looks like a bug for both os, may be should be done in the follow up pr

cpp/oneapi/dal/BUILD
```diff
@@ -96,6 +96,11 @@ dal_dynamic_lib(
         ":core",
         ":optional",
     ],
+    extra_deps = [
```
Alexandr-Solovev: Not sure, why its necessary in this pr. In my understanding its common for both linux and windows

cpp/oneapi/dal/BUILD
```diff
@@ -106,6 +111,11 @@ dal_dynamic_lib(
         "@onedal//cpp/oneapi/dal/algo:parameters",
         "@onedal//cpp/oneapi/dal/detail/parameters",
     ],
+    extra_deps = [
```
Alexandr-Solovev: the same comment+ we dont support parameters for win

Alexandr-Solovev:
Configuration: 2ed572d13c45899e15c1ad8593364eba47a1825be5b910ca9a724a8b0b7c1193
 Execution platform: @@platforms//host:host
LINK : warning LNK4044: unrecognized option '/Wl,--exclude-libs=libmkl_core.a'; ignored
LINK : warning LNK4044: unrecognized option '/Wl,--exclude-libs=libmkl_intel_ilp64.a'; ignored
Loosk incorrect and must be not called for windows.

dev/bazel/deps/tbb_win.tpl.BUILD
```diff
@@ -0,0 +1,36 @@
...
+
+load("@rules_cc//cc:defs.bzl", "cc_import", "cc_library")
+
+cc_library(
+    name = "headers",
+    hdrs = glob(["include/**/*.h", "include/**/*.hpp"], allow_empty = True),
+    includes = ["include"],
+)
+
+cc_import(
```
Alexandr-Solovev: we used cc_library instead of import for linux. May be the reason of faliure. Makes sense to try with cc_library

Alexandr-Solovev:
The step is green now, can you please extend the job with launching examples with output(similar to LinuxBazel) instead fo smoke tests. Would be nice to have daal-cpp-examples-thread-dev/onedal examples in this step

napetrov:
> The step is green now, can you please extend the job with launching examples with output(similar to LinuxBazel) instead of smoke tests. Would be nice to have daal-cpp-examples-thread-dev/onedal examples in this step

I would prefer to integrate things in base way and then extend to keep PRs relatively simple.

dev/bazel/run_examples_split.sh
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env bash
```
david-cortes-intel: Shouldn't this default to generic POSIX sh like the other shell scripts?

napetrov: Addressed in ecc5b6321: converted `run_examples_split.sh` to POSIX `sh` syntax and switched the shebang to `#!/usr/bin/env sh` to match the generic shell-script style.

Alexandr-Solovev:
Based on the log from the ci job 
[343 / 673] Compiling examples/daal/cpp/source/covariance/cov_dense_online.cpp; 1s local
ERROR: D:/a/1/s/examples/daal/cpp/BUILD:349:24: Linking examples/daal/cpp/cov_dense_batch_host.exe failed: missing input file '@@+onedal_repo+onedal_release//:lib/intel64/onedal_dll.lib'
ERROR: D:/a/1/s/examples/daal/cpp/BUILD:349:24: Linking examples/daal/cpp/cov_dense_batch_host.exe failed: missing input file '@@+onedal_repo+onedal_release//:lib/intel64/onedal_parameters_dll.lib'
ERROR: D:/a/1/s/examples/daal/cpp/BUILD:349:24: Linking examples/daal/cpp/cov_dense_batch_host.exe failed: 2 input file(s) do not exist
[345 / 673] 1 / 112 tests, 1 failed; checking cached actions
ERROR: D:/a/1/s/examples/daal/cpp/BUILD:349:24 Linking examples/daal/cpp/cov_dense_batch_host.exe failed: 2 input file(s) do not exist
INFO: Elapsed time: 98.000s, Critical Path: 7.57s
INFO: 120 processes: 224 action cache hit, 114 internal, 6 local.
ERROR: Build did NOT complete successfully
Likely you need to remove dependency on the parameters lib for win "onedal_parameters_dll.lib". 

And also strange moment, that in the release step the build is done with next libs(close to target one)
image
but the, for some reason in examples it starts to rebuild oneDAL instead of reuse build from previous step

Alexandr-Solovev (approved):

PR #3661, opened 2026-06-12
Fix docs/dalapi CodeFactor issues

napetrov:
## Summary
- move shared generateDS parser support code into `docs/dalapi/doxypy/parser/common.py` to remove duplicated boilerplate from `compound.py` and `index.py`
- use defusedxml-backed parsing in the generated dalapi Doxygen parsers
- add defusedxml to the docs requirements
- remove stale docs/dalapi TODO marker and nearby Ruff findings
- avoid shell=True in the Doxygen process helper
- fix the follow-up CodeFactor complex-method report in `docs/dalapi/doxypy/parser/compound.py` by splitting `compounddefType.__init__`
- make dalapi property extraction deterministic by preserving getter order instead of iterating a set intersection

## CodeFactor
- addressed the original 24 CodeFactor issues reported for this PR
- addressed the new CodeFactor `Complex Method` finding for `docs/dalapi/doxypy/parser/compound.py#L374-L470`

## Verification
- `python3 -m py_compile docs/dalapi/doxypy/parser/common.py docs/dalapi/doxypy/parser/index.py docs/dalapi/doxypy/parser/compound.py docs/dalapi/transformers.py docs/dalapi/utils.py docs/dalapi/generator.py docs/dalapi/extension.py`
- `python3 -m bandit -q docs/dalapi/doxypy/parser/common.py docs/dalapi/doxypy/parser/index.py docs/dalapi/doxypy/parser/compound.py docs/dalapi/transformers.py`
- `python3 -m bandit -q -t B602 docs/dalapi/utils.py`
- `python3 -m ruff check --select F821 docs/dalapi/doxypy/parser/common.py docs/dalapi/doxypy/parser/index.py docs/dalapi/doxypy/parser/compound.py docs/dalapi/transformers.py docs/dalapi/utils.py docs/dalapi/generator.py`
- `python3 -m ruff check docs/dalapi/utils.py docs/dalapi/generator.py docs/dalapi/extension.py`
- `/tmp/onedal-docs-pr3661-venv/bin/python` constructor smoke confirmed `compounddefType()` still creates independent list fields when optional list arguments are omitted
- `/tmp/onedal-docs-pr3661-venv/bin/python` property-order smoke confirmed `PropertyTransformer._get_properties_info()` preserves getter order
- explicit parser-path smoke after XML generation: `Parser.parse('index')` read 1306 compounds from `index.xml`; `Parser.parse('<sample-refid>')` read a real compound XML with `compounddef=1`, confirming both `parser/index.py` and `parser/compound.py` paths still work after the shared `common.py` refactor
- `make -C docs parse-doxygen` with Doxygen 1.9.8 from `/tmp/onedal-doxygen-pr3661` and docs requirements installed in `/tmp/onedal-docs-pr3661-venv`: passed, regenerated `docs/doxygen/oneapi/doxygen/xml` and `docs/build/tree.yaml`
- `make -C docs html SPHINXOPTS='-E -j2'` with the same environment: passed, clean-built all 429 Sphinx sources and produced `docs/build`

## Generated Docs Result Check
- fetched every HTML path from the published site at `https://uxlfoundation.github.io/oneDAL/` that exists in the PR build: 433/433 fetched, 0 missing
- normalized rendered-text comparison against the published site found 15 pages with ordering-only differences; multiset comparison of rendered text lines found 0 added/removed text lines, so the generated documentation content/result is unchanged
- normalized rendered-text multiset comparison against an `origin/main` docs build also found 0 added/removed rendered text lines
- the only observed differences are member/property ordering in generated algorithm API pages; this PR now removes one source of that nondeterminism by preserving getter order in `PropertyTransformer`


docs/dalapi/doxypy/parser/common.py
```diff
@@ -0,0 +1,935 @@
...
+                    self.gds_elementtree_node_.sourceline)
+            return ""
+
+
+    def getSubclassFromModule_(module, class_):
+        '''Get the subclass of a class from a specific module.'''
+        name = class_.__name__ + 'Sub'
+        if hasattr(module, name):
+            return getattr(module, name)
+        return None
```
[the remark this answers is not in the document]
napetrov: Fixed in bd3e04992: moved getSubclassFromModule_ to module scope so it is available whether generatedssuper is imported or the fallback class is used.

Added targeted regression smoke with a temporary generatedssuper.py verifying both index.DoxygenType.factory() and compound.DoxygenType.factory() resolve subclass helpers without NameError.

docs/dalapi/doxypy/parser/index.py
```diff
@@ -35,920 +35,40 @@
...
-    def get_choice(self): return self.choice
-    def set_optional(self, optional): self.optional = optional
-    def get_optional(self): return self.optional
-
-
-def _cast(typ, value):
-    if typ is None or value is None:
-        return value
-    return typ(value)
+from .common import (
```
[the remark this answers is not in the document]
napetrov: Fixed in bd3e04992: restored local import sys in parser/index.py so generated CLI/helper paths using sys.argv, sys.stdout, sys.stderr, and sys.exit have the expected module global.

Verified with py_compile, Ruff F821, Bandit, and parser index CLI smoke against generated Doxygen XML.

docs/dalapi/doxypy/parser/compound.py
```diff
@@ -35,920 +35,40 @@
...
-    def get_choice(self): return self.choice
-    def set_optional(self, optional): self.optional = optional
-    def get_optional(self): return self.optional
-
-
-def _cast(typ, value):
-    if typ is None or value is None:
-        return value
-    return typ(value)
+from .common import (
```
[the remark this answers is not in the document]
napetrov: Fixed in bd3e04992: restored local import sys in parser/compound.py so generated CLI/helper paths using sys.argv, sys.stdout, sys.stderr, and sys.exit have the expected module global.

Verified with py_compile, Ruff F821, Bandit, and parser compound CLI smoke against a generated Doxygen compound XML file.

docs/dalapi/doxypy/parser/common.py
```diff
@@ -0,0 +1,937 @@
...
+#
+# Calls to the methods in these classes are generated by generateDS.py.
+# You can replace these methods by re-implementing the following class
+#   in a module named generatedssuper.py.
+
+try:
+    from generatedssuper import GeneratedsSuper
+except ImportError:
+
+    class GeneratedsSuper(object):
```
david-cortes-intel, line 171: Why is this needed if the docs are using doxygen which already contains similar tooling?

napetrov: Good question. This is not intended to duplicate or replace Doxygen’s tooling.

Doxygen is still the tool that generates the XML. After that, the existing `docs/dalapi` pipeline runs `python -m dalapi.doxypy.cli ...`, which parses that Doxygen XML into the dalapi Python model/YAML used by the Sphinx extension.

The code in `common.py` is the generateDS support/runtime code that was already present in both generated parser modules (`parser/index.py` for `index.xml` and `parser/compound.py` for compound XML files). This PR only moves that duplicated support code into one shared module and keeps the current pipeline behavior the same.

So `common.py` is needed by the current Python parser layer, not by Doxygen itself. Replacing this with Doxygen-provided parsing/tooling could be possible as a separate cleanup, but it would be a larger change to the `dalapi.doxypy` pipeline rather than a CodeFactor-focused refactor.

For the concrete docs-generation path, I also checked the call chain on this PR head:

```text
make -n -C docs parse-doxygen
  cd doxygen/oneapi && doxygen
  mkdir -p build
  python -m dalapi.doxypy.cli doxygen/oneapi/doxygen/xml --compact > build/tree.yaml
```

and the parser modules resolve the shared symbols from `common.py` at runtime:

```text
index.parsexml_ module: dalapi.doxypy.parser.common
compound.parsexml_ module: dalapi.doxypy.parser.common
index.GeneratedsSuper same as common.GeneratedsSuper: True
compound.GeneratedsSuper same as common.GeneratedsSuper: True
```

david-cortes-intel: Do I understand it that  this tries to add a custom sphinx extension to convert doxygen docs into rst? Couldn't it use public packages for that? For example, xgboost uses breathe:
https://github.com/dmlc/xgboost/blob/f872422eb4a3b11e949f205398ccaa1f98aeb304/doc/conf.py#L313
https://github.com/dmlc/xgboost/blob/f872422eb4a3b11e949f205398ccaa1f98aeb304/doc/conf.py#L261

napetrov: Yes, that would be the right direction for a larger docs-tooling cleanup, but this PR is not adding a new Sphinx extension or a new Doxygen-to-RST path.

The custom `dalapi` extension and the `doxypy` XML parser are already the existing oneDAL documentation pipeline. This PR only refactors the duplicated generateDS support code that was already embedded in `parser/index.py` and `parser/compound.py` into `parser/common.py`, then keeps the same generated parser behavior.

Using Breathe/Exhale or another public Doxygen/Sphinx bridge could be a good separate migration, but it would need to replace or adapt the current oneDAL-specific pieces around `tree.yaml`, `dalapi` directives/roles, path resolution, and API page generation. I would avoid mixing that migration into this CodeFactor-focused PR because it changes the docs architecture rather than just cleaning the current implementation.

david-cortes-intel (approved):

PR #3662, opened 2026-06-15
bazel: align Linux MKL release parity

napetrov:
## Summary

This PR aligns the Bazel Linux MKL release with the Make release for the #3530 parity work.

Changes:
- keep oneAPI libraries from absorbing DAAL/MKL objects
- split parameter libraries into parameter-only object sets
- restore Linux shared library metadata/dependency behavior: SONAME, DT_NEEDED, and no Bazel RUNPATH in release shared libs
- hide stray thread-library linker marker exports
- add missing DAAL algorithm coverage used by the Make build
- align Linux release packaging: modulefile, pkg-config files, examples, OS-specific nuspec/sample filtering, and no generated dispatcher header in installed includes

## Validation

Run on `onedal-build` from top of main `d989d07c3ec0d5efa420bd44730e2f5a9453d53a`.

Commands:
- `bazelisk clean`
- `bazelisk build //:release --backend_config=mkl --release_dpc=false --jobs=8 --verbose_failures`

Result:
- clean Bazel MKL release build passed
- time: `327.09s`
- actions: `3179`

Final compare vs Make MKL release:
- release tree: `only_make 0`, `only_bazel 0`
- `libonedal_core`: members `1919/1919`, exports `8545/8545`, MKL exports `0`
- `libonedal_thread`: members `2/2`, exports `319/319`, MKL exports `0`
- `libonedal`: exports `3422/3422`, MKL exports `0`
- `libonedal_parameters`: members `8/8`, exports `71/71`, MKL exports `0`
- SONAME/DT_NEEDED match for checked libraries
- Bazel RUNPATH absent in checked release shared libraries

Remaining size note:
- Make release: `405M`
- Bazel release: `410M`
- remaining delta is static archives; Bazel archives PIC objects (`.pic.o`) while Make archive composition is non-PIC-object based in the compared places. Dynamic library section totals are not the source of the delta.

Local report:
`/home/ubuntu/onedal-mkl-rootcause-20260614-223805/mkl-close-gaps-final-report.txt`

### DIFF
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -415,7 +415,7 @@ jobs:
       # Display oneDAL build configuration
       bazel build @config//:dump
       echo
-      cat bazel-bin/external/+_repo_rules+config/config.json
+      cat bazel-bin/external/+declare_onedal_config+config/config.json
       echo
     displayName: 'bazel-configure'
 
@@ -524,10 +524,12 @@ jobs:
     displayName: 'Clean Bazel cache (expunge)'
 
   - script: |
+      export LD_LIBRARY_PATH="$(DALROOT)/lib/intel64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
       bazel test //examples/daal/cpp:all \
                  --test_link_mode=release_dynamic \
                  --test_thread_mode=par \
-                 --test_env DALROOT
+                 --test_env DALROOT \
+                 --test_env LD_LIBRARY_PATH
     displayName: 'daal-cpp-examples-thread-release-dynamic'
 
   - script: |
@@ -536,10 +538,12 @@ jobs:
     displayName: 'Clean Bazel cache (expunge)'
 
   - script: |
+      export LD_LIBRARY_PATH="$(DALROOT)/lib/intel64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
       bazel test //examples/oneapi/cpp:all \
                  --test_link_mode=release_dynamic \
                  --test_thread_mode=par \
-                 --test_env DALROOT
+                 --test_env DALROOT \
+                 --test_env LD_LIBRARY_PATH
     displayName: 'cpp-examples-thread-release-dynamic'
 
   - script: |
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -593,6 +593,27 @@ jobs:
       fi
     displayName: 'bazel-cache-limit'
 
+- job: 'LinuxReleaseCompare'
+  dependsOn:
+  - LinuxMakeGNU_MKL
+  - LinuxBazel
+  pool:
+    vmImage: '$(VM_IMAGE)'
+  steps:
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      artifact: 'lnx32e build'
+      path: '$(Pipeline.Workspace)/make'
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      artifact: 'lnx32e Bazel build'
+      path: '$(Pipeline.Workspace)/bazel'
+  - script: |
+      python3 dev/release_tests/compare_release_trees.py \
+        --platform linux \
+        --make "$(Pipeline.Workspace)/make/daal/latest" \
+        --bazel "$(Pipeline.Workspace)/bazel"
+    displayName: 'Compare Make and Bazel release trees'
 
 - job: 'WindowsBazel'
   timeoutInMinutes: 0
@@ -849,6 +870,12 @@ jobs:
       call %TEMP%\oneapi\setvars.bat --force
       .\.ci\scripts\build.bat onedal_dpc icx avx2
     displayName: 'make onedal_dpc'
+  - task: PublishPipelineArtifact@1
+    inputs:
+      artifactName: '$(platform.type) icx build'
+      targetPath: '$(Build.Repository.LocalPath)/$(release.dir)'
+    displayName: 'Upload Intel build artifacts'
+    continueOnError: true
   - script: |
       call %TEMP%\oneapi\setvars.bat --force
       .\.ci\scripts\test.bat daal\cpp lib icx cmake
@@ -865,6 +892,30 @@ jobs:
       tree -h -I include __nuget/inteldal*/
       ls -lh __nuget/inteldal*.nupkg
     displayName: 'nuget pkg'
+
+- job: 'WindowsReleaseCompare'
+  dependsOn:
+  - WindowsMakeIntel
+  - WindowsBazel
+  pool:
+    vmImage: '$(WIN_VM_IMAGE)'
+  steps:
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      artifact: 'win32e icx build'
+      path: '$(Pipeline.Workspace)\make'
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      artifact: 'win32e Bazel icx build'
+      path: '$(Pipeline.Workspace)\bazel'
+  - powershell: |
+      python dev/release_tests/compare_release_trees.py `
+        --platform windows `
+        --structure-only `
+        --make "$(Pipeline.Workspace)\make\daal\latest" `
+        --bazel "$(Pipeline.Workspace)\bazel"
+    displayName: 'Compare Make and Bazel release tree structure'
+    continueOnError: true
 
 
 - job: WindowsSklearnex
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -524,7 +524,7 @@ jobs:
     displayName: 'Clean Bazel cache (expunge)'
 
   - script: |
-      export LD_LIBRARY_PATH="$(DALROOT)/lib/intel64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
+      . "$(DALROOT)/env/vars.sh"
       bazel test //examples/daal/cpp:all \
                  --test_link_mode=release_dynamic \
                  --test_thread_mode=par \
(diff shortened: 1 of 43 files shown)


dev/bazel/tests/release_structure_test.sh
```diff
@@ -47,7 +47,8 @@ if ! ${IS_LINUX}; then
     echo "  SKIP: .so versioning checks are Linux-only (detected OS: ${OS})"
 fi
 
-for lib_base in libonedal_core libonedal libonedal_thread; do
+BINARY_MAJOR_VER=""
+for lib_base in libonedal_core libonedal libonedal_thread libonedal_parameters; do
```
david-cortes-intel, line 52: Could this instead have common tests between bazel and conda?
https://github.com/uxlfoundation/oneDAL/blob/main/conda-recipe/test-devel.sh

napetrov: Addressed by factoring the overlapping package metadata checks into `dev/release_tests/package_metadata_test.sh` and calling it from both `dev/bazel/tests/release_structure_test.sh` and `conda-recipe/test-devel.sh`. The Bazel script still keeps Bazel/release-layout-specific checks (SONAME, symlinks, modulefile, leaked headers) local to that test.

dev/bazel/cc.bzl
```diff
@@ -255,7 +255,14 @@ def _copy_dynamic_release_file(ctx, src, out_name, is_windows = False, extra_inp
...
 
 def _cc_dynamic_lib_impl(ctx):
-    toolchain, feature_config = _init_cc_rule(ctx)
+    toolchain, feature_config = _init_cc_rule(ctx, features=[
+        # Keep produced shared libraries free of dynamic Bazel deps on Linux.
+        # Release-dynamic tests provide standalone oneDAL libs through
+        # DALROOT/LD_LIBRARY_PATH while Bazel-provided runtimes such as TBB
+        # live in test runfiles; DT_NEEDED entries on those deps are not
+        # resolved through the executable RUNPATH.
+        "do_not_link_dynamic_dependencies",
```
david-cortes-intel, line 264: I thought the .so files were supposed to contain relative entries in their RPATHs pointing to MKL and TBB as ithey would be found assuming the structure of offline oneAPI toolkit installs, with both oneDAL and dependencies being placed under `/opt/intel/`.

@Alexandr-Solovev Any comments here?

dev/bazel/release.bzl
```diff
@@ -120,7 +120,24 @@ def _try_relativize(path, start):
...
     return path
 
-def _copy_include(ctx, prefix):
+def _copy_version_header(ctx, src_file, dst_path, version_info):
+    dst_file = ctx.actions.declare_file(dst_path)
+    ctx.actions.expand_template(
+        template = src_file,
+        output = dst_file,
+        substitutions = {
+            "#define __INTEL_DAAL_BUILD_DATE 21990101": "#define __INTEL_DAAL_BUILD_DATE {}".format(version_info.build),
```
david-cortes-intel, line 129: Could this use `#ifndef` if the files to make it more clear? Would any kind of build (bazel / make / conda / etc.) leave them with these defaults?

dev/release_tests/package_metadata_test.sh
```diff
@@ -0,0 +1,89 @@
...
+    fi
+
+    if grep -q "^Name:" "$pc_file"; then
+        pass "${pc_name} has Name field"
+    else
+        fail "${pc_name} missing Name field"
+    fi
+
+    if grep -q "^Libs:" "$pc_file"; then
+        pass "${pc_name} has Libs field"
```
david-cortes-intel, line 54: @Alexandr-Solovev Are these files included in the conda builds? Is there any CI job building with conda?

dev/bazel/cc.bzl
napetrov: No, only certain installations would end up in opt.  + we can work with TBB that was build from sources from examples. 

There should be no dependency on some assumed paths

napetrov: The Make release on Linux ships shared libs without a Bazel-injected RUNPATH — relocation relies on the user's LD_LIBRARY_PATH, ld.so.conf, or setvars.sh. We can't bake $ORIGIN/../../mkl/... or /opt/intel/... paths in: TBB built from sources, conda envs, system-package TBB, and oneAPI online installer all violate that assumption. This feature flag matches Make's behaviour (no extra RUNPATH on the produced .so). The Bazel-side TBB is only used for in-tree tests via LD_LIBRARY_PATH/runfiles — the released libonedal*.so deliberately have no RUNPATH.

dev/bazel/release.bzl
napetrov: Good point — the current literal-text expand_template is fragile (any reformat of services/library_version_info.h silently breaks substitution). I'll convert the header to #ifndef-guarded defaults in a follow-up so all builds (Bazel/Make/conda) can either define via -D or leave defaults. Make today writes the values from makefile.ver; once the header is #ifndef-guarded both paths converge and Bazel can drop _copy_version_header entirely. Tracking as a follow-up to keep this PR scoped to release-parity.

dev/release_tests/package_metadata_test.sh
napetrov: Yes — both lib/pkgconfig/*.pc and lib/cmake/oneDAL/*.cmake are installed into the conda devel package. The conda CI is the LinuxMakeGNU_MKL_Conda Azure pipeline job: it builds via the conda recipe and runs conda-recipe/test-devel.sh, which this PR updates to call dev/release_tests/package_metadata_test.sh. So conda gets the same metadata coverage Bazel does, and LinuxMakeGNU_MKL_Conda is green on this PR.

BUILD
```diff
@@ -123,17 +137,21 @@ release(
...
         "//examples/daal/cpp:release_files",
+        "//examples/oneapi/cpp:release_files",
     ],
     extra_files = [
         release_extra_file(":release_vars_sh", "env/vars.sh", windows_dst_path = "env/vars.bat"),
-        release_extra_file(":release_pkgconfig", "lib/pkgconfig/onedal.pc", windows_dst_path = ""),
-        release_extra_file(":release_pkgconfig_dynamic_threading_host", "", windows_dst_path = "lib/pkgconfig/dal-dynamic-threading-host.pc"),
-        release_extra_file(":release_pkgconfig_static_threading_host", "", windows_dst_path = "lib/pkgconfig/dal-static-threading-host.pc"),
+        release_extra_file(":release_pkgconfig", "", windows_dst_path = "lib/pkgconfig/onedal.pc"),
+        release_extra_file(":release_pkgconfig_dynamic_threading_host", "lib/pkgconfig/dal-dynamic-threading-host.pc", windows_dst_path = ""),
```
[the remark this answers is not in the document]
napetrov: Intentional — Make-release parity. Linux ships dal-{dynamic,static}-threading-host.pc (host-tagged, threading-specific) and not the combined onedal.pc; Windows ships the combined onedal.pc and not the threading-host variants. The bot is reading the swap as a regression because release_pkgconfig generates lib/pkgconfig/onedal.pc, but on Linux that's superseded by the threading-host .pc files. LinuxReleaseCompare (release-tree diff vs Make) is green on this PR and confirms the layout matches.

.ci/pipeline/ci.yml
```diff
@@ -524,10 +524,12 @@ jobs:
     displayName: 'Clean Bazel cache (expunge)'
 
   - script: |
+      export LD_LIBRARY_PATH="$(DALROOT)/lib/intel64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
```
Alexandr-Solovev: can we use env/vars.sh from bazel for more coverage in real scenarios instead of exporting? Should have the same effect

.ci/pipeline/ci.yml
```diff
@@ -536,10 +538,12 @@ jobs:
...
```
Alexandr-Solovev: same comment

deploy/local/dal
```diff
@@ -64,12 +64,12 @@ set dalroot "$componentroot"
...
 # Setup environment variables
-setenv          DAL_MAJOR_BINARY   1
-setenv          DAL_MINOR_BINARY   1
+setenv          DAL_MAJOR_BINARY   __DAL_MAJOR_BINARY__
+setenv          DAL_MINOR_BINARY   __DAL_MINOR_BINARY__
 setenv          DALROOT            "$dalroot"
 
-prepend-path    LD_LIBRARY_PATH    "$dalroot/lib"
-prepend-path    LIBRARY_PATH       "$dalroot/lib"
+prepend-path    LD_LIBRARY_PATH    "$dalroot/lib/$daal_target_arch"
```
Alexandr-Solovev: what the reason of extending the path with intel64?

.ci/pipeline/ci.yml
napetrov: Done in 00c15f15. The dynamic example jobs now source `$(DALROOT)/env/vars.sh` instead of exporting `LD_LIBRARY_PATH` manually.

I also made `vars_lnx.sh` choose the library directory from the package layout directly: it uses `$DALROOT/lib/$ARCH_DIR_ONEDAL` when that directory/symlink exists and falls back to `$DALROOT/lib` otherwise. That keeps the CI path package-aware across release zip, conda/apt-style installs, and future layouts.

.ci/pipeline/ci.yml
napetrov: Same fix in 00c15f15: this job now sources `$(DALROOT)/env/vars.sh` as well, so both dynamic example jobs exercise the packaged environment setup instead of carrying a hardcoded `lib/intel64` path in CI.

deploy/local/dal
napetrov: Good point. The reason was the Bazel release zip layout: its real Linux libraries are under `lib/intel64`. But other package layouts differ: apt-style installs put real libraries under `lib` and may expose `lib/intel64` only as a compatibility symlink.

I changed this in 00c15f15 so the modulefile no longer assumes only one layout. It now starts from `$dalroot/lib`, switches to `$dalroot/lib/$daal_target_arch` only when that directory/symlink exists, and otherwise keeps `$dalroot/lib`. This matches the same package-aware setup logic used by `env/vars.sh`.

Alexandr-Solovev (approved):

homksei (approved):

PR #3665, opened 2026-06-18
Add transpose operation for F-order data and fix validation for GPU

avolkov-intel:
## Description

The existing `table2ndarray` function used `row_accessor::pull()` which goes through `pull_rows_impl` — a general-purpose path that handles type conversion, non-contiguous blocks, and heterogeneous tables. For column-major homogeneous tables on device, this created excessive intermediate allocations and SYCL events. When the table is already homogeneous with the correct data type, we can bypass `row_accessor` entirely and work directly with the table's raw pointer — only transposing if needed.

We introduce `homogen_table_to_same_order_ndarray` which directly accesses the homogen table's data pointer, uses `alloc_kind_requires_copy` (from `pull_rows_impl`) to decide copy vs zero-copy, and handles layout mismatch via a single `transpose_copy` kernel. The new `table2ndarray_rm` / `table2ndarray_cm` dispatch to this optimized path for homogeneous tables with matching data types, falling back to `row_accessor` for heterogeneous tables or type mismatches.

### Changes

**Core:**
- `cpp/oneapi/dal/backend/primitives/utils.hpp` — rewrote `table2ndarray_rm`, `table2ndarray_cm`, `table2ndarray`, and added `homogen_table_to_same_order_ndarray` with optimized direct-pointer path
- `cpp/oneapi/dal/table/backend/common_kernels.hpp` — updated `alloc_kind_requires_copy` to require a copy from `usm_shared` to host, preventing concurrent page migration crashes

**Algorithm kernels:**
- `cpp/oneapi/dal/algo/logistic_regression/backend/gpu/train_kernel_dense_batch_dpc.cpp` — refactored `train()` to use `table2ndarray` for both data and responses, ensuring device allocation with correct row-major layout
- `cpp/oneapi/dal/algo/linear_regression/backend/gpu/train_kernel_norm_eq_dpc.cpp` — same pattern (already done)
- `cpp/oneapi/dal/algo/finiteness_checker/backend/gpu/compute_kernel_dense_dpc.cpp` — updated validation kernel to avoid unnecessary transpose when table layout already matches the requested order

**Tests:**
- `cpp/oneapi/dal/algo/linear_kernel/test/batch.cpp` — allocate test tables in `usm_device` to avoid shared memory page migration issues on discrete GPUs
- `cpp/oneapi/dal/algo/correlation_distance/test/batch.cpp` — same
---

<details>
<summary>Checklist:</summary>

- [ x] I have commented my code, particularly in hard-to-understand areas.
- [ x] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.
- [ x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [ x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [ x] I have run it locally and tested the changes extensively.

**Performance**

</details>


cpp/oneapi/dal/backend/primitives/utils.hpp
```diff
@@ -180,11 +182,21 @@ inline ndarray<Type, 2, order> table2ndarray(sycl::queue& q,
                                                  q,
                                                  table.get_row_count());
     [[maybe_unused]] const auto layout = table.get_data_layout();
+
+    if (layout == decltype(layout)::row_major) {
+        std::cerr << "Layout is row major\n";
```
david-cortes-intel: Forgot to remove?

david-cortes-intel:
@avolkov-intel What about logistic regression?

david-cortes-intel:
@avolkov-intel @Vika-F Looks like the 'copy' function is not using omatcopy when doing transposes:
https://github.com/uxlfoundation/oneDAL/blob/05bd0eb634001b6591a1c6e77e4b1ee0a837e48c/cpp/oneapi/dal/backend/primitives/ndarray.hpp#L601

Perhaps that could be improved in a different PR. Omatcopy should do it in parallel and can have good speedups on CPU when transposing matrices.

For the non-transpose case with strides, it could also use 'lacpy' from MKL instead:
https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-fortran/2023-2/lacpy.html

cpp/oneapi/dal/backend/primitives/blas/omatcopy.hpp
```diff
@@ -0,0 +1,35 @@
+/*******************************************************************************
+* Copyright 2024 Intel Corporation
```
david-cortes-intel: Please use the current header:
https://github.com/uxlfoundation/oneDAL/blob/3f7d8b8d793a1be0cbdb7a56c4059f6c313da750/CONTRIBUTING.md?plain=1#L115

avolkov-intel:
Some benchmarking results showing speedup for LinearRegression:

image

cpp/oneapi/dal/algo/linear_kernel/test/batch.cpp
```diff
@@ -34,6 +34,16 @@ class linear_kernel_batch_test : public te::float_algo_fixture<std::tuple_elemen
         return linear_kernel::descriptor<Float, Method>{}.set_scale(scale).set_shift(shift);
     }
 
+    table get_device_table(const te::dataframe& df, const te::table_id& id) {
+        // If test is running in data parallel mode, then the table should be created on device
```
Alexandr-Solovev: Not sure that "parallel" mode is proper name here

cpp/oneapi/dal/algo/linear_regression/backend/gpu/train_kernel_norm_eq_dpc.cpp
```diff
@@ -172,7 +172,24 @@ static train_result<Task> train(const context_gpu& ctx,
...
+    const auto data_nd = pr::table2ndarray<Float, pr::ndorder::c>(queue, input.get_data(), alloc);
+    const auto resp_nd =
+        pr::table2ndarray<Float, pr::ndorder::c>(queue, input.get_responses(), alloc);
+
+    const auto data_rows = data_nd.get_dimension(0);
+    const auto data_cols = data_nd.get_dimension(1);
+    const auto resp_rows = resp_nd.get_dimension(0);
+    const auto resp_cols = resp_nd.get_dimension(1);
+
+    auto data_table = homogen_table::wrap(queue, data_nd.get_data(), data_rows, data_cols);
```
Alexandr-Solovev: Probably makes sense to 
`auto data_table = homogen_table::wrap(queue, data_nd.get_data(), data_nd.get_dimension(0), data_nd.get_dimension(1));`
instead of introducing extra variables, unless they are reused elsewhere.

cpp/oneapi/dal/algo/logistic_regression/backend/gpu/train_kernel_dense_batch_dpc.cpp
```diff
@@ -45,15 +45,24 @@ static train_result<Task> train(const context_gpu& ctx,
...
+    // table2ndarray() ensures the returned ndarray is row-major and accessible from the target device.
+    // Depending on source alloc/layout, it may reuse the original storage or allocate/copy/transpose.
+    const auto data_nd = pr::table2ndarray<Float, pr::ndorder::c>(queue, input.get_data(), alloc);
+    const auto resp_nd =
+        pr::table2ndarray<Float, pr::ndorder::c>(queue, input.get_responses(), alloc);
+
+    const auto data_rows = data_nd.get_dimension(0);
+    const auto data_cols = data_nd.get_dimension(1);
+    const auto resp_rows = resp_nd.get_dimension(0);
+    const auto resp_cols = resp_nd.get_dimension(1);
```
Alexandr-Solovev: same comment

cpp/oneapi/dal/backend/primitives/utils.hpp
```diff
@@ -60,106 +62,188 @@ inline ndarray<Type, 1> table2ndarray_1d(const table& table) {
...
+/// @param alloc The requested USM allocation type.
 ///
-/// @return The new 2D ndarray instance.
+/// @return True if the pointer is already suitable for the requested alloc on the queue's device.
 template <typename Type>
-inline ndarray<Type, 2, ndorder::c> table2ndarray_rm(sycl::queue& q,
-                                                     const table& table,
-                                                     sycl::usm::alloc alloc) {
-    constexpr auto order = ndorder::c;
+inline bool is_ptr_accessible(sycl::queue& q, const Type* ptr, sycl::usm::alloc alloc) {
```
Vika-F: Looks like this function is not used. Maybe you forgot to call it somewhere?

cpp/oneapi/dal/backend/primitives/utils.hpp
```diff
@@ -217,7 +298,10 @@ inline auto table2ndarray_variant(sycl::queue& q, const table& table, sycl::usm:
     return result;
 }
 
-/// Convert a table to a 1D ndarray
+/// Convert a table to a 1D ndarray with row-major element order.
```
Vika-F, line 278: It is not clear how this change affects the performance of the algorithms where `table2ndarray_1d` is used.
The semantics of `table2ndarray_1d` was changed here. I am not sure that in all the cases where this function is  it is more beneficial to convert column-major data to row-major.

Maybe it would be better just to replace `table2ndarray_1d` with `table2ndarray_rm` + `ndarray::wrap` where needed?

cpp/oneapi/dal/backend/primitives/utils.hpp
```diff
@@ -179,13 +263,10 @@ inline ndarray<Type, 2, order> table2ndarray(sycl::queue& q,
     ONEDAL_PROFILER_SERVICE_TASK_WITH_ARGS_QUEUE(service::table2ndarray_queue,
                                                  q,
                                                  table.get_row_count());
-    [[maybe_unused]] const auto layout = table.get_data_layout();
     if constexpr (order == ndorder::c) {
-        ONEDAL_ASSERT(layout == decltype(layout)::row_major);
```
Vika-F, line 184: Can you please comment on this change?

cpp/oneapi/dal/algo/correlation_distance/test/batch.cpp
```diff
@@ -35,12 +35,20 @@ class correlation_distance_batch_test
         return correlation_distance::descriptor<Float, Method>{};
     }
 
+    table get_device_table(const te::dataframe& df, const te::table_id& id) {
```
Vika-F: This function is duplicated here and in `linear_kernel/test/batch.cpp`. Please move it into common `dal/test/engine` directory.

cpp/oneapi/dal/table/backend/common_kernels.hpp
```diff
@@ -55,11 +55,19 @@ inline sycl::usm::alloc alloc_kind_to_sycl(alloc_kind kind) {
 inline bool alloc_kind_requires_copy(alloc_kind src_alloc_kind, alloc_kind dst_alloc_kind) {
 #ifdef ONEDAL_DATA_PARALLEL
     switch (dst_alloc_kind) {
+        // Shared USM on discrete GPUs uses page migration which can cause
+        // crashes when accessed concurrently by multi-threaded host code
```
Vika-F, line 59: Thank you for this clarification!

cpp/oneapi/dal/backend/memory.hpp
```diff
@@ -206,8 +206,8 @@ inline sycl::event memcpy(sycl::queue& queue,
                           std::size_t size,
                           const event_vector& deps = {}) {
     ONEDAL_ASSERT(size > 0);
-    ONEDAL_ASSERT(is_known_usm(queue, dest));
```
Vika-F, line 209: Please add the assertions for null-pointers checking:
```
ONEDAL_ASSERT(dest != nullptr)
```
is_known_usm checks also for null-pointers.

cpp/oneapi/dal/backend/primitives/utils.hpp
Vika-F: I see that `flatten_table_by_layout` was introduced to retain column-major order where needed, but it is now used only in finitness checker. Is it the only place where having array in column-major order is beneficial?

cpp/oneapi/dal/backend/primitives/utils.hpp
```diff
@@ -230,9 +314,35 @@ template <typename Type>
...
+///
+/// @tparam Type The type of the memory block elements within the ndarray.
+///
+/// @param q     The SYCL* queue.
+/// @param table Input table.
+/// @param alloc The requested USM allocation type.
+///
+/// @return A 1D ndarray over the table's data.
+template <typename Type>
+inline ndarray<Type, 1> flatten_table_by_layout(sycl::queue& q,
```
Vika-F: The naming is a bit out if line. Other functions are named 'table2ndarray...'.
Renaming proposals: 'table2ndarray_1d_save_layout', 'table2ndarray_1d_tm' ("table major", analogy with NumPy 'A' order, where it means column-major in case input is column-major, and row-major otherwise)

cpp/oneapi/dal/backend/primitives/utils.hpp
avolkov-intel: I'm not sure, I didn't dare to touch other places as I was not sure about internal logic of the algorithms. With finiteness check it was obvious that *it will have no affect on result plus it was a bottleneck for some use cases

avolkov-intel: Regarding the performance effect, I believe previously we also converted column major to row major, since we used row_accessor with pull_rows

cpp/oneapi/dal/backend/primitives/utils.hpp
avolkov-intel: Forgot to remove it

cpp/oneapi/dal/backend/primitives/utils.hpp
avolkov-intel: This assert was basically failing if  table with F-order was provided, we just didn't see it in CI because this assert is only triggered in debug mode.

cpp/oneapi/dal/algo/linear_kernel/test/batch.cpp
avolkov-intel: We just usually guard sycl code with this #ifdef ONEDAL_DATA_PARALLEL so I think it's fine to call it like this

Vika-F (approved):

PR #3670, opened 2026-06-24
Bazel: align Windows release package with make

napetrov:
## Summary
- build the Windows Bazel release in nightly CI and compare its release manifest with the make release
- fix Windows Bazel DPC dynamic linking by archiving object inputs and passing the archive through the Intel/Windows link path
- package the missing oneAPI DPC examples/samples and versioned Windows import libraries while excluding Bazel-only artifacts from release output

## Testing
- Windows remote: `build.bat onedal_dpc vc avx2` completed successfully
- Windows remote: `bazel build //:release --config=release-dpc --cpu=avx2` completed successfully
- Windows remote: `.ci/scripts/compare_windows_release.ps1 -MakeReleaseDir .\__release_win_vc\daal\latest -BazelReleaseDir .\bazel-bin\release\daal\latest` passed with `make files: 1066`, `bazel files: 1066`
- Local: `git diff --cached --check`
- Local: `bazel query //:release --noshow_progress`

Note: local PowerShell parser check was skipped because `pwsh` is not installed in this Linux workspace; the script itself was exercised on the Windows remote compare gate above.

### DIFF
--- a/.ci/env/bazelisk.ps1
+++ b/.ci/env/bazelisk.ps1
@@ -1,0 +1,63 @@
+#===============================================================================
+# Copyright contributors to the oneDAL project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+$ErrorActionPreference = "Stop"
+
+$bazeliskVersion = "v1.29.0"
+$assetName = "bazelisk-windows-amd64.exe"
+$installDir = Join-Path (Get-Location) "bazel\bin"
+$bazelPath = Join-Path $installDir "bazel.exe"
+
+$headers = @{
+    Accept = "application/vnd.github+json"
+    "X-GitHub-Api-Version" = "2022-11-28"
+}
+
+if ($env:GITHUB_TOKEN) {
+    $headers.Authorization = "Bearer $env:GITHUB_TOKEN"
+}
+
+$release = Invoke-RestMethod `
+    -Headers $headers `
+    -Uri "https://api.github.com/repos/bazelbuild/bazelisk/releases/tags/$bazeliskVersion"
+
+$asset = $release.assets | Where-Object { $_.name -eq $assetName } | Select-Object -First 1
+if (-not $asset) {
+    throw "Could not find $assetName in Bazelisk release $bazeliskVersion"
+}
+
+New-Item -ItemType Directory -Force -Path $installDir | Out-Null
+Invoke-WebRequest -Uri $asset.browser_download_url -OutFile $bazelPath
+
+if (-not $asset.digest.StartsWith("sha256:")) {
+    throw "Unexpected Bazelisk digest format: $($asset.digest)"
+}
+
+$expectedHash = $asset.digest.Substring("sha256:".Length)
+$actualHash = (Get-FileHash -Algorithm SHA256 -LiteralPath $bazelPath).Hash.ToLowerInvariant()
+if ($actualHash -ne $expectedHash.ToLowerInvariant()) {
+    throw "Bazelisk SHA256 mismatch. Expected $expectedHash, got $actualHash"
+}
+
+$pathLine = $installDir
+if ($env:GITHUB_PATH) {
+    $pathLine | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+}
+else {
+    $env:PATH = "$installDir;$env:PATH"
+}
+
+& $bazelPath --version
--- a/.ci/scripts/compare_windows_release.ps1
+++ b/.ci/scripts/compare_windows_release.ps1
@@ -1,0 +1,99 @@
+#===============================================================================
+# Copyright contributors to the oneDAL project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+param(
+    [string] $MakeReleaseDir = "__release_win_vc\daal\latest",
+    [string] $BazelReleaseDir = "bazel-bin\release\daal\latest",
+    [int] $MaxDiffLines = 200
+)
+
+$ErrorActionPreference = "Stop"
+
+function Resolve-ReleaseDir {
+    param([string] $Path)
+
+    $resolved = Resolve-Path -LiteralPath $Path -ErrorAction SilentlyContinue
+    if (-not $resolved) {
+        throw "Release directory does not exist: $Path"
+    }
+    return $resolved.ProviderPath.TrimEnd('\', '/')
+}
+
+function Get-ReleaseManifest {
+    param([string] $Root)
+
+    $rootPrefix = $Root.TrimEnd('\', '/') + [IO.Path]::DirectorySeparatorChar
+    Get-ChildItem -LiteralPath $Root -Recurse -File |
+        ForEach-Object {
+            $_.FullName.Substring($rootPrefix.Length).Replace('\', '/')
+        } |
+        Sort-Object
+}
+
+function Write-DiffBlock {
+    param(
+        [string] $Title,
+        [string[]] $Items
+    )
+
+    Write-Host ""
+    Write-Host $Title
+    if ($Items.Count -eq 0) {
+        Write-Host "  none"
+        return
+    }
+
+    $Items | Select-Object -First $MaxDiffLines | ForEach-Object {
+        Write-Host "  $_"
+    }
+
+    if ($Items.Count -gt $MaxDiffLines) {
+        Write-Host "  ... $($Items.Count - $MaxDiffLines) more"
+    }
+}
+
+$makeRoot = Resolve-ReleaseDir $MakeReleaseDir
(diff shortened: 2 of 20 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -125,6 +129,22 @@ jobs:
...
           call .\.ci\scripts\build.bat onedal_dpc vc avx2
+      - name: Install Bazelisk
+        shell: pwsh
+        run: .\.ci\env\bazelisk.ps1
+      - name: Bazel release
+        shell: cmd
+        run: |
+          set PATH=C:\msys64\usr\bin;%PATH%
+          call .\oneapi\setvars.bat
+          call "%ONEAPI_ROOT%\setvars-vcvarsall.bat" %VS_VER%
```
david-cortes-intel, line 155: Isn't bazel meant to handle these types of dependencies inside its build?

dev/bazel/toolchains/cc_toolchain_win.bzl
```diff
@@ -44,6 +44,64 @@ def _find_tool(repo_ctx, tool_name, mandatory = False):
             tool_path = repo_ctx.path("tool_not_found_{}.bat".format(tool_name))
     return str(tool_path), is_found
 
+def _create_dpc_link_wrapper(repo_ctx, dpcc_path):
```
david-cortes-intel, line 47: Why would this file need to be auto-generated instead of making it a regular committed file?

.ci/scripts/compare_windows_release.ps1
```diff
@@ -0,0 +1,99 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#===============================================================================
+
+param(
+    [string] $MakeReleaseDir = "__release_win_vc\daal\latest",
```
david-cortes-intel, line 18: Is this meant to compare MSVC builds or ICX builds?

Alexandr-Solovev:
The best possible thing to validate that its really compatible - add ci check where bazel build is reused for cmake example. In ideal case it should work

.ci/env/bazelisk.ps1
ethanglaser, line 1: What does this provide that `.ci/env/bazelisk.sh` doesn't?

ethanglaser: Just windows alternative?

dev/bazel/toolchains/tools/dpc_link_win.tpl.bat
```diff
@@ -1,3 +1,19 @@
 @echo off
+rem ============================================================================
+rem Copyright 2026 Intel Corporation
```
david-cortes-intel: Please use the new copyright header.

dev/release_tests/compare_release_trees.py
```diff
@@ -62,12 +62,82 @@
     "windows": (".dll",),
 }
 
+IGNORED_FILES = {
+    "linux": {
+        # Make DPC release does not publish these static DPC archives; keep the
```
david-cortes-intel: Since the makefile doesn't even generate these files, shouldn't bazel also avoid generating them?

dev/release_tests/compare_release_trees.py
```diff
@@ -62,12 +62,82 @@
...
+    "__bss_start",
+    "_edata",
+    "_end",
+    # Intel compiler/runtime math symbols may be pulled into Bazel-built shared
+    # objects while Make hides them with linker options. They are not oneDAL API.
+    "_LIB_VERSIONIMF",
+    "cos",
+    "log",
+    "sqrt",
+    "sqrtf",
```
david-cortes-intel: These shouldn't be getting exported in the first place, and in most cases should come from MKL's vectorized equivalents. Most code shouldn't be calling with functions from the standard library, save for the DPC parts. CC @Vika-F .

dev/release_tests/compare_release_trees.py
```diff
@@ -62,12 +62,82 @@
...
+    # Implementation details instantiated through oneTBB/libstdc++ may vary
+    # between Make and Bazel builds while keeping the public oneDAL ABI intact.
+    "_ZN3tbb6detail",
+    "_ZNK3tbb6detail",
+    "_ZSt27__unguarded_partition_pivot",
+    "_Z28_daal_parallel_sort_template",
+    "_ZN4daal10algorithms8internal5qSort",
+    "__intel_",
+    "__libm_",
+    "__svml_",
```
david-cortes-intel, line 100: These also shouldn't be getting exported. They are from compiler libraries. Are they getting static-linked into the oneDAL .so files? If not, from where do they come? The release builds shouldn't have a dependency on these libraries.

.github/workflows/nightly-build.yml
napetrov: Thanks. I agree that Bazel should own the actual build dependency graph. This workflow step is only runner/bootstrap setup: installing Bazelisk, initializing the oneAPI/VC environment, and making the required Windows tools visible in PATH. The release contents and libraries being validated are still produced/resolved by Bazel, not by ad-hoc workflow dependency wiring.

dev/bazel/toolchains/cc_toolchain_win.bzl
napetrov: Agreed. The current version keeps the wrapper logic as a committed tool script/template instead of embedding the whole body inline in Starlark. The repository/toolchain setup only wires paths/tool locations for the generated launcher; the implementation itself is reviewable as a normal file.

.ci/scripts/compare_windows_release.ps1
napetrov: This comparison is for the Windows release package surface produced by the existing CI path versus the Bazel Windows release package. In the workflow both sides run after the same oneAPI + VC environment setup, so the intent is to validate the same Windows/VC-style release directory layout. Bazel may use the configured icx-based toolchain internally, but the compared artifact is the release package layout, not a compiler-specific object-level output.

.ci/env/bazelisk.ps1
napetrov: Yes — this is the Windows CI counterpart of `.ci/env/bazelisk.sh`. The Windows job runs under PowerShell/cmd, downloads `bazelisk-windows-amd64.exe`, verifies the GitHub release SHA256 digest, and appends the install directory through the Windows/GitHub Actions PATH mechanism.

napetrov: Exactly, it is just the Windows alternative so the Windows nightly job can install and verify Bazelisk without relying on the Unix shell helper.

dev/release_tests/compare_release_trees.py
napetrov: Agreed, addressed in fce8fc411. Bazel no longer includes the extra DPC static archives (`libonedal_dpc.a` / `libonedal_parameters_dpc.a`) in the release package, so the release comparison should not need to ignore those Bazel-only files.

dev/release_tests/compare_release_trees.py
napetrov: Thanks, agreed that broad symbol ignores would hide a real ABI/export issue. I narrowed this back rather than treating compiler/math exports as generally acceptable. The latest fce8fc411 validation is running now; if the export diff still reports unwanted symbols after the package cleanup, I will fix the link/export surface instead of adding a broad ignore.

dev/release_tests/compare_release_trees.py
napetrov: Agreed. Compiler/runtime symbols should not be silently accepted as part of the public export surface. I removed the package-side false positive first in fce8fc411, and I am waiting for the new nightly comparison to show the remaining exact export diff. If those symbols are still exported, the next fix will be in Bazel link/export controls, not in a broad comparison filter.

dev/bazel/toolchains/cc_toolchain_win.bzl
```diff
@@ -44,6 +44,64 @@ def _find_tool(repo_ctx, tool_name, mandatory = False):
...
+                $expandedArgs.Add($line)
+            }
+        }
+    }
+    else {
+        $expandedArgs.Add($arg)
+    }
+}
+
+$objectRsp = [IO.Path]::GetTempFileName() + '.rsp'
```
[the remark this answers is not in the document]
napetrov: Addressed in the current wrapper cleanup: the generated inline wrapper path was replaced with committed wrapper logic, avoiding the `GetTempFileName() + '.rsp'` pattern called out here.

napetrov:
Update: I added the requested validation path in the nightly workflow. It builds the Bazel release package and then configures/builds/runs the CMake example against that Bazel-produced release. I also fixed the validation script in fce8fc411 to build the requested CMake target explicitly and run the executable found in the build tree. The new CI run is still in progress; I will follow up if it exposes another compatibility issue.

.ci/scripts/compare_windows_release.ps1
david-cortes-intel: Would be useful to add comments about it in the file.

homksei (approved):

ethanglaser:
sklearnex builds are failing in private CI: `LINK : fatal error LNK1104: cannot open file '...sources\build\_daal4py.cp311-win_amd64.exp'`

homksei:
> sklearnex builds are failing in private CI: `LINK : fatal error LNK1104: cannot open file '...sources\build\_daal4py.cp311-win_amd64.exp'`

The changes in the pull request don't look like something that could have caused this error. I suspect this is a Visual Studio issue related to the path being too long, and it's likely that the problem will not occur with a shorter branch name.

napetrov:
Created short-branch test PR #3698 at the exact same commit (`bc0dc4052accd3ebde535ec62b2e7c16985334b7`) and requested private CI there. Its source branch is `p3670`. This isolates whether the Windows sklearnex `LNK1104` failure is caused by the long branch/workspace path; the original PR and code are unchanged.

ethanglaser:
Private CI there looks good, indeed branch name related: http://intel-ci.intel.com/f18201e7-f2c0-f16e-89b1-d4f5ef20c6a0

napetrov:
@Alexandr-Solovev  @ethanglaser  @david-cortes-intel  - any objections integrating this ?

Alexandr-Solovev (approved):

ethanglaser (approved):
LGTM, but agreed with David that the .ps1 files would benefit from some comments

napetrov:
I opened follow-up PR #3701: https://github.com/uxlfoundation/oneDAL/pull/3701

The review feedback about comments and developer documentation for the PowerShell helpers will be addressed there.

PR #3681, opened 2026-07-01
fix csr kmeans impl(cpu)

Alexandr-Solovev:
## Summary

Fix the exact objective function value for sparse (CSR) KMeans and add a dense-vs-sparse equivalence test.

The Lloyd CSR post-processing kernel was computing the objective by iterating only over the stored non-zeros of each row and summing `(x_j - c_{a,j})^2`. This drops every `(0 - c_{a,j})^2` contribution for features that are implicitly zero in the CSR row, so on sparse data the reported objective was systematically too small. Centroids and assignments were unaffected — only the exact objective value was wrong — but with `n_init > 1` in scikit-learn-intelex this bug can cause the wrong trial to be picked as "best".

## What changed

**`cpp/daal/src/algorithms/kmeans/kmeans_lloyd_postprocessing.h`** — `PostProcessing<lloydCSR>::computeExactObjectiveFunction`:

Reformulate the per-row objective using the identity
```
||x - c||^2 = ||x||^2 - 2·<x, c> + ||c||^2
```
For a CSR row, `x_j = 0` outside `nnz(row)`, so `||x||^2` and `<x, c>` collapse to sums over the stored entries, while `||c||^2` covers all `p` features. Concretely:

1. Precompute `clSq[k] = ||c_k||^2` once per cluster using an `omp simd` reduction.
2. For each row: accumulate `rowGoal = Σ_stored (x^2 - 2·x·c)` starting from zero with `PRAGMA_OMP_SIMD_ARGS(reduction(+ : rowGoal))`, then add `clSq[assk]` once at the end for better accuracy.

## Test

**`cpp/oneapi/dal/algo/kmeans/test/batch.cpp`** — new Catch2 case `"KMeans dense and sparse agree on same input"`:

- Templated over `kmeans_types_csr` (float / double), sweeps both `zero_based` / `one_based` indexing.
- Uses `csr_make_blobs` with `nnz_fraction = 0.05` (≈95% zeros — matches the reviewer's ask for "a sizable amount of zeros").
- Materializes the same underlying arrays as both a `csr_table` and a dense `homogen_table`.
- Trains `lloyd_csr` on the CSR view and `lloyd_dense` on the dense view from identical initial centroids, with `compute_exact_objective_function` enabled.
- Asserts centroids agree index-wise (labels are stable across methods because both start from the same seeds), responses match exactly, and the objective values agree within `1e-3` in single precision / `1e-10` in double.

The tolerance on the objective is chosen to account for the different summation orderings used by the dense (BLAS-gemm-based) and CSR (per-nnz) paths — pre-fix the mismatch was on the order of the total `||c||^2` mass that was being dropped, not float rounding.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>


david-cortes-intel:
@Alexandr-Solovev Please add a unit test verifying that a dense and sparse representation of the same input give the same result, with data that would have a sizable amount of zeros.

cpp/daal/src/algorithms/kmeans/kmeans_lloyd_postprocessing.h
```diff
@@ -300,12 +315,19 @@ struct PostProcessing<lloydCSR, algorithmFPType, cpu>
                 const size_t jFinish = rowIdx[k + 1] - 1;
 
                 const size_t assk = assignments[k];
+                // ||x - c||^2 = ||x||^2 - 2 * <x, c> + ||c||^2.
+                // In CSR, x_j = 0 for indices j not in nnz(row), so both ||x||^2 and <x, c>
+                // reduce to sums over the stored entries; ||c||^2 is over all p features.
+                algorithmFPType rowGoal = clSq[assk];
```
david-cortes-intel: For better accuracy, you could start this from zero and sum `clSq[assk]` at the end.

cpp/daal/src/algorithms/kmeans/kmeans_lloyd_postprocessing.h
```diff
@@ -300,12 +315,19 @@ struct PostProcessing<lloydCSR, algorithmFPType, cpu>
...
                 PRAGMA_VECTOR_UNALIGNED
```
david-cortes-intel: Perhaps it could turn this into an OMP simd clause with a reduction. Don't know if it will make a difference though.

cpp/oneapi/dal/algo/kmeans/test/batch.cpp
```diff
@@ -396,6 +396,81 @@ TEMPLATE_LIST_TEST_M(kmeans_batch_test,
...
+    // requested in review. The CSR and materialized dense views of the same
+    // underlying arrays are the same input in two representations, so training
+    // with identical initial centroids must give similar centroids, responses
+    // and objective_function_value.
+    constexpr std::int64_t cluster_count = 8;
+    constexpr std::int64_t row_count = 300;
+    constexpr std::int64_t column_count = 40;
+    constexpr std::int64_t max_iter = 20;
+    constexpr Float accuracy_threshold = Float(0);
+    constexpr float nnz_fraction = 0.05f;
```
david-cortes-intel: Running time for tests could be kept shorter by making these kinds of problems smaller.

david-cortes-intel:
@Alexandr-Solovev The tests that you are adding here would pass with the code from the main branch too. Please ensure that they trigger the intended code path.

cpp/oneapi/dal/algo/kmeans/test/batch.cpp
```diff
@@ -396,6 +396,81 @@ TEMPLATE_LIST_TEST_M(kmeans_batch_test,
...
+    const auto input = oneapi::dal::test::engine::csr_make_blobs<Float>(cluster_count,
+                                                                        row_count,
+                                                                        column_count,
+                                                                        nnz_fraction,
+                                                                        this->data_indexing_);
+    const auto csr_data = input.get_data(this->get_policy());
+    const auto dense_data = input.get_dense_data(this->get_policy());
+    const auto initial_centroids = input.get_initial_centroids();
+
+    const auto csr_desc =
```
david-cortes-intel, line 538: Shouldn't this pass a random seed somewhere for the initialization?

cpp/oneapi/dal/algo/kmeans/test/batch.cpp
```diff
@@ -14,6 +14,10 @@
 * limitations under the License.
 *******************************************************************************/
 
+#include <algorithm>
+#include <numeric>
+#include <random>
```
Alexandr-Solovev, line 19: I wanted to extend the test scope with random data test case

david-cortes-intel:
In the CI error for aarch64:
```
+ apt-get install -y clang-18 lldb-18 lld-18 clangd-18
E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 39615 (apt-get)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
```

It appears to be trying to execute installation commands concurrently where it shouldn't.

cpp/oneapi/dal/algo/kmeans/test/batch.cpp
```diff
@@ -396,6 +400,186 @@ TEMPLATE_LIST_TEST_M(kmeans_batch_test,
...
+    // (x_j - c_j)^2 only over the stored entries of a row, which drops the
+    // c_j^2 contribution from features where the row has an implicit zero
+    // and the centroid is non-zero. The previous csr_make_blobs-based test
+    // did not trigger this: rows there share the column support of their
+    // centroid, so c_j = 0 at every unstored index and the missing term is
+    // identically zero — the buggy path returned the same value as dense.
+    //
+    // This hand-crafted case exercises the fix explicitly: rows have a
+    // single stored non-zero at three distinct columns, so both rows
+    // assigned to centroid 0 have unstored indices where c_j != 0. On main
```
david-cortes-intel: This 'main' that it references will become outdated once this PR is merged. Could be reworded for posterity.

cpp/oneapi/dal/algo/kmeans/test/batch.cpp
```diff
@@ -417,12 +417,6 @@ TEMPLATE_LIST_TEST_M(kmeans_batch_test,
     // centroid, so c_j = 0 at every unstored index and the missing term is
     // identically zero — the buggy path returned the same value as dense.
```
david-cortes-intel: Please use ascii-only characters.

Vika-F (approved):
The changes and the tests look good to me.
Initially I was concerned with the dense and sparse results divergence due to FP in the test with 10 iterations, but now the test looks less fragile. Not sure about the exact values of the tolerances though.

Let's wait for the CI and LGTM.

cpp/daal/src/algorithms/kmeans/kmeans_lloyd_postprocessing.h
```diff
@@ -300,12 +315,19 @@ struct PostProcessing<lloydCSR, algorithmFPType, cpu>
...
                 for (size_t j = jStart; j < jFinish; j++)
                 {
-                    const size_t m = colIdx[j] - 1;
-                    goal += (data[j] - inClusters[assk * p + m]) * (data[j] - inClusters[assk * p + m]);
+                    const size_t m          = colIdx[j] - 1;
+                    const algorithmFPType c = inClusters[assk * p + m];
+                    const algorithmFPType x = data[j];
+                    rowGoal += x * x - algorithmFPType(2) * x * c;
                 }
+                goal += rowGoal + clSq[assk];
```
david-cortes-intel: In the formula for dense, the result will be guaranteed to be non-negative regardless of floating point inaccuracies, but the formula here can potentially output negative numbers. Please clip it to zero.

david-cortes-intel:
I see a failure in the DPC job for windows for the test being introduced here:
```
FAIL: //cpp/oneapi/dal/algo/kmeans:test_batch_dpc (Aborted)
```

cpp/oneapi/dal/algo/kmeans/test/batch.cpp
```diff
@@ -555,10 +555,23 @@ TEMPLATE_LIST_TEST_M(kmeans_batch_test,
...
     const auto dense_run = this->train(dense_desc, dense_data, init1);
 
-    // Same seed on the sparse path -> bit-identical objective and iteration
-    // count across independent runs.
+    // Same seed on the sparse path -> reproducible objective and iteration
+    // count across independent runs. CPU Lloyd sums in a fixed partition
+    // order -> bit-identical. GPU sycl::reduction combines partial sums in
+    // an implementation-defined order (atomic fetch_add / non-fixed
+    // workgroup tree) and float addition is non-associative, so a few ULPs
+    // of drift is expected between two runs.
```
david-cortes-intel, line 563: @Alexandr-Solovev @Vika-F Does MKL provide bit-by-bit reproducible results on GPU? If so, then oneDAL should too, since that's how scikit-learn is meant to work.

Alexandr-Solovev: I can ask MKL team about it, but I guess we can do it separately, in the next pr and this issue should not be a blocker for the current PR. @Vika-F

david-cortes-intel: Please go ahead and ask.

david-cortes-intel: > I can ask MKL team about it, but I guess we can do it separately, in the next pr and this issue should not be a blocker for the current PR. @Vika-F

There's a run-to-run stability test in sklearnex checking for exact results, but not sure if it executes on GPU.

Alexandr-Solovev: I believe its not executed, because I didn't introduce any regressions in this pr

david-cortes-intel (approved):

PR #3687, opened 2026-07-14
Fix OpenBLAS thread safety race condition (#3329)

### LINKED ISSUE #3329: Single-threaded BLAS/LAPACK calls create race conditions with OpenBLAS
There are some sections within oneDAL which call BLAS and LAPACK functions in parallel, for which we have separate functions that run those BLAS/LAPACK calls single-threaded to avoid nested parallelism - for example:
https://github.com/uxlfoundation/oneDAL/blob/e19dbc4c041d9fe9e437da2d9e933bfd9beb6403/cpp/daal/src/externals/service_blas_mkl.h#L59

When building with MKL as backend for BLAS/LAPACK, the threading is controlled through functions that have effects only within the thread that calls them:
https://github.com/uxlfoundation/oneDAL/blob/e19dbc4c041d9fe9e437da2d9e933bfd9beb6403/cpp/daal/src/externals/service_blas_mkl.h#L62
https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2025-2/mkl-set-num-threads-local.html

Hence, it works in such a way that every time a single-threaded call is needed, the thread executing it queries its current config, configures itself to not spawn more threads, and then restores its previous config, leaving everything as it was.

However, when oneDAL is built with OpenBLAS as backend, it instead calls functions that have global effects beyond the thread that executes them, since OpenBLAS doesn't have thread-local config functions the same way MKL does:
https://github.com/uxlfoundation/oneDAL/blob/e19dbc4c041d9fe9e437da2d9e933bfd9beb6403/cpp/daal/src/externals/service_thread_declar_ref.h#L44

This has two issues:
* There's a potential bottleneck from multiple threads modifying the same global state at the same time, which very likely involves a mutex.
* It's unlikely that at the end of all of this the global number of threads configured will be restored correctly, because only the last change will have an effect. There's no guarantee that the first thread seeing the initial global state with all threads will be the one finishing last (plus there's multiple calls to these functions under each thread), so potentially the configuration could be left globally single-threaded afterwards.

I guess one potential solution could be to have a separate version of these wrapper classes that would handle the configuration and restoration within its constructor and destructor, and place this appropriately where the single-threaded calls are needed.

Another potentially better way of doing this (not 100% sure if it's feasible though) would be to set a custom threading backend for OpenBLAS using the TBB thread pool that oneDAL uses for parallel executions, which should now be supported as of OpenBLAS>=0.3.30, but might be quite complicated.

CC @keeranroth @rakshithgb-fujitsu

lxcxjxhx:
## Motivation

Fix Issue #3329: OpenBLAS backend has thread safety issues in multi-threaded environments.

### Root Cause
When oneDAL uses OpenBLAS as the BLAS/LAPACK backend, it calls \openblas_set_num_threads()\ to set single-threaded mode. However, OpenBLAS lacks a thread-local API equivalent to MKL's \mkl_set_num_threads_local()\, causing:

1. **Race condition**: Multiple threads modify the global thread count configuration simultaneously
2. **Incorrect state restoration**: Thread save/restore operations overwrite each other, potentially corrupting global state

## Changes

Add a static mutex to the \openblas_thread_setter\ class to serialize the save-set-restore sequence:

- Use Meyers singleton pattern (function-local static) to manage the mutex
- Lock in constructor, unlock in destructor
- Disable copy construction and assignment to prevent accidental copies

### Modified Files
- \cpp/daal/src/externals/service_thread_declar_ref.h\

## Testing

- Maintains API compatibility
- Does not affect MKL backend behavior
- Resolves race condition in multi-threaded environments

## Checklist

- [x] Code follows the project's coding standards
- [x] Changes are minimal and focused on the issue
- [x] API compatibility is maintained
- [x] Commit messages follow conventional format

Closes #3329

### DIFF
--- a/cpp/daal/src/externals/service_thread_declar_ref.h
+++ b/cpp/daal/src/externals/service_thread_declar_ref.h
@@ -24,6 +24,8 @@
 #ifndef __SERVICE_THREAD_DECLAR_REF_H__
 #define __SERVICE_THREAD_DECLAR_REF_H__
 
+#include <mutex>
+
 namespace daal
 {
 namespace internal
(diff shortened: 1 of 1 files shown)


david-cortes-intel:
Thanks but this doesn't solve the problem. Please read the issue description carefully:
image

PR #3694, opened 2026-07-17
Rewrite kernel profiler for correct tree nesting and GPU timing

Alexandr-Solovev:
Rebuild the profiler tree from per-thread stacks tracked in the singleton instead of a global level counter. Regular tasks now record a parent index at push time (the caller thread's innermost open task, with fallback to the innermost regular task across threads for worker-thread threading tasks); level derives from parent + 1. This fixes the case where a task opened on one thread and closed on another (or threading tasks under a regular parent) produced garbled nesting.

Merge phase now keys sibling collapse on (parent_idx, name) via an unordered_map, so it is O(n) and correctly matches only true siblings. Levels-and-names on their own were ambiguous: two tasks at the same depth under different parents could collapse into one.

Cache ONEDAL_VERBOSE once per process in a function-local static instead of a namespace-scope `volatile int`. The previous form gave every TU that included the header its own copy, so `daal_verbose_val` observed different values across .so boundaries.

Move the once-per-process suppression sets for the threading logger and tracer messages onto the profiler singleton so their lifetimes are bounded to the profiler, not the process; hoist unique-name checks under the existing mutex.

On the oneAPI side, replace the `start_task(name, queue)` pattern with a `queue_sync_guard` RAII wrapper. The guard issues `wait_and_throw()` before `start_task()` and again on scope exit so recorded durations reflect device-side execution rather than host enqueue. The pre-region wait happens before the timer, so already-in-flight work on the queue is not charged to the region. When the profiler is disabled the guard is a nullptr no-op and no waits are issued.

Uniquify the local variable name via __LINE__ so multiple profiled scopes in the same block coexist; drop the ternary-of-mixed-types pattern and gate each START_TASK* macro on the enable predicate directly. SERVICE task variants gate their logger side-effects on `is_service_debug_enabled` (ONEDAL_VERBOSE=5) rather than `is_logger_enabled` (=1/4), matching the pre-refactor behavior — otherwise hot paths like `table2ndarray` would flood the output under LOGGER mode.

Validated: make onedal_dpc -j112 clean; CPU + GPU cov_dense_batch and kmeans_lloyd_dense_batch under ONEDAL_VERBOSE=3 produce well-formed analyzer trees with correct parent-child nesting and durations.

## Description

---

<details>
<summary>Checklist:</summary>

- [ ] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.

**Testing**

**Performance**

- [ ] I have provided justification why performance and/or quality metrics have changed or why changes are not expected.

</details>

### DIFF
--- a/cpp/daal/src/services/service_profiler.h
+++ b/cpp/daal/src/services/service_profiler.h
@@ -23,7 +23,10 @@
 #pragma once
 
 #include <chrono>
+#include <thread>
+#include <cerrno>
 #include <cstdint>
+#include <cstdlib>
 #include <cstring>
 #include <vector>
 #include <sstream>
@@ -33,6 +36,7 @@
 #include <algorithm>
 #include <exception>
 #include <unordered_map>
+#include <unordered_set>
 #include "services/library_version_info.h"
 
 #ifdef _WIN32
@@ -123,7 +127,6 @@
         return daal::internal::profiler::start_task(nullptr);                                                                                 \
     }()
 
-static volatile int daal_verbose_val                = -1;
 inline static constexpr int PROFILER_MODE_OFF       = 0;
 inline static constexpr int PROFILER_MODE_LOGGER    = 1;
 inline static constexpr int PROFILER_MODE_TRACER    = 2;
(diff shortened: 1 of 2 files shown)


cpp/daal/src/services/service_profiler.h
```diff
@@ -136,25 +139,27 @@ namespace daal
...
-        errno          = 0;
-        long val       = std::strtol(verbose_str, &endptr, 10);
-        bool parsed_ok = (errno == 0 && endptr != verbose_str && *endptr == '\0');
-        if (parsed_ok && val >= 0 && val <= 5) newval = static_cast<int>(val);
-    }
-    daal_verbose_val = newval;
-}
-
+// Reads ONEDAL_VERBOSE once per process; the C++17 function-local static gives thread-safe
+// single initialization and a single storage instance across all TUs (fixes the previous
```
david-cortes-intel, line 143: It could explain what a TU is.

cpp/daal/src/services/service_profiler.h
```diff
@@ -23,7 +23,10 @@
 #pragma once
 
 #include <chrono>
+#include <thread>
+#include <cerrno>
```
david-cortes-intel, line 27: Are macros from this header used somewhere?

cpp/oneapi/dal/detail/profiler.hpp
```diff
@@ -18,82 +18,183 @@
...
+        if (enabled) {
+            q.wait_and_throw();
+            return queue_sync_guard(&q, daal::internal::profiler::start_task(task_name));
+        }
+        return queue_sync_guard(nullptr, daal::internal::profiler::start_task(nullptr));
+    }
+
+    ~queue_sync_guard() {
+        if (queue_) {
+#if (!defined(DAAL_NOTHROW_EXCEPTIONS))
```
david-cortes-intel, line 50: Throwing on a destructor is undefined behavior. It should be catched regardless of this macro.

PR #3710, opened 2026-07-24
bazel: add Linux LeakSanitizer config

napetrov:
## Description

Add a named Bazel `--config=lsan` for Linux non-DPC++ targets, equivalent to Make `REQSAN=leak`.

This is a focused Make-to-Bazel option-parity improvement toward #3530. It does not claim LeakSanitizer support for DPC++ device compilation or Windows ICX, where the sanitizer is unsupported.

## What changed

- add `build:lsan` with:
  - `-fsanitize=leak` for compile and link actions
  - `-fno-omit-frame-pointer`
  - debug symbols through `--config=dbg-symbols`
- document usage, scope, and the Make-to-Bazel mapping
- add `lsan` to the existing bounded Linux Bazel config-smoke loop

## Validation

Validated on the Linux build system at exact commit `e2148797935805f727d490a19b42291b400a9c53`.

### Static and configuration checks

```bash
git diff --check
python3 -c 'import yaml; yaml.safe_load(open(".ci/pipeline/ci.yml")); print("PASS")'
/home/ubuntu/bin/bazelisk \
  --output_user_root=/tmp/issue3530-lsan-analysis \
  --nohome_rc build --nobuild :release --config=lsan
```

Results:

- diff check: passed
- CI YAML parse: passed
- Bazel analysis: passed; 5,114 targets configured

### Materialized action flags

Representative target:

```text
//examples/oneapi/cpp:column_accessor_homogen_host
```

Queries:

```bash
bazel aquery \
  'mnemonic("CppCompile", deps(//examples/oneapi/cpp:column_accessor_homogen_host))' \
  --config=lsan --output=textproto

bazel aquery \
  'mnemonic("CppLink", deps(//examples/oneapi/cpp:column_accessor_homogen_host))' \
  --config=lsan --output=textproto
```

Results:

- compile actions: 1,584/1,584 contain `-fsanitize=leak`
- compile actions: 1,584/1,584 contain `-fno-omit-frame-pointer`
- link actions: 55/55 contain `-fsanitize=leak`
- debug `-g` is inherited through `dbg-symbols`

### Compiler smoke

A bounded Bazel build/run smoke printed `LSan smoke: 42` with:

- GCC 15.2
- Intel ICX 2026.0

The exact configuration flags also passed direct compile/link/run probes with:

- GCC 15.2
- Clang 21.1.8
- ICPX 2026.0

### Known pre-existing limitation

Selecting the repository's Bazel Clang toolchain currently fails during toolchain analysis because its CPU flag lists are empty. This predates this change; direct Clang LeakSanitizer compile/link/run succeeds. The new config does not modify compiler selection.

## Review

Independent build/config review: approve-with-nits; no critical or should-fix code issue. The documentation wording was tightened from “host builds” to “non-DPC++ targets” after review, and exact proof commands/results are recorded above.

---

<details>
<summary>Checklist</summary>

- [x] The configuration follows neighboring sanitizer patterns.
- [x] Linux/non-DPC++ scope and unsupported Windows/DPC++ cases are documented.
- [x] The commit is signed off.

**Testing**

- [x] `git diff --check`
- [x] CI YAML parse
- [x] full release configuration analysis with `--config=lsan`
- [x] compile/link action inspection
- [x] GCC and ICX Bazel smoke
- [x] direct GCC, Clang, and ICPX probes

**Performance**

- [x] No algorithm or production runtime behavior changes unless users explicitly select `--config=lsan`.

</details>

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -201,6 +201,17 @@ build:asan-static \
     --config=dbg-symbols
 
 
+# Configuration: 'lsan'
+# LeakSanitizer build for supported Linux non-DPC++ targets.
+# Equivalent to Make: REQSAN=leak
+#   bazel test //cpp/oneapi/dal:tests --config=lsan
+build:lsan \
+    --copt=-fsanitize=leak \
+    --copt=-fno-omit-frame-pointer \
+    --linkopt=-fsanitize=leak \
+    --config=dbg-symbols
+
+
 # Configuration: 'tsan'
 # ThreadSanitizer build.
 # Equivalent to Make: REQSAN=thread
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -430,7 +430,7 @@ jobs:
   - script: |
       set -euo pipefail
 
-      for config in dbg dbg-symbols asan asan-static tsan ubsan msan type; do
+      for config in dbg dbg-symbols asan asan-static lsan tsan ubsan msan type; do
         echo "Checking Bazel config: ${config}"
         bazel build --nobuild :release --config="${config}"
       done
--- a/dev/bazel/README.md
+++ b/dev/bazel/README.md
@@ -497,6 +497,16 @@ For static libasan linkage (equivalent to Make `REQSAN=static`):
 bazel test //cpp/oneapi/dal:tests --config=asan-static --config=dbg
 ```
 
+### LeakSanitizer (LSan)
+
+Equivalent to Make `REQSAN=leak`. This configuration is for Linux non-DPC++ targets
+with a compiler that supports LeakSanitizer. Do not use it for DPC++ targets;
+DPC++ device compilation does not support LSan. Windows ICX does not support it.
+
+```sh
+bazel test //cpp/oneapi/dal:tests --config=lsan
+```
+
 ### ThreadSanitizer (TSan)
 
 Equivalent to Make `REQSAN=thread`:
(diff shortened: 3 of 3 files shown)


david-cortes-intel:
@napetrov This is already included in the default ASAN config. Why would you want a separate config for it?

napetrov:
The main goal here is option parity with the Make build: Make supports a standalone leak-only mode through `REQSAN=leak`, while the Bazel side currently has no corresponding named configuration.

You are right that `--config=asan` already includes leak detection, so this does not fill a coverage gap. The separate configuration is intended as a lighter-weight leak-only run for cases where the rest of ASan instrumentation is not needed, while also preserving the existing Make workflow during the Bazel migration.

If we do not consider standalone Make-option parity or the lighter leak-only workflow valuable enough to maintain, then I agree that the ASan configuration is sufficient.

napetrov:
@david-cortes-intel Following up with a more precise answer — checked GCC/Clang's sanitizer flags directly rather than relying on the coverage claim alone.

`-fsanitize=leak` is not "the same leak detection under a different name" — it is a distinct, standalone sanitizer runtime from `-fsanitize=address`. `--config=asan` links the full AddressSanitizer runtime (out-of-bounds, use-after-free, double-free, stack/heap corruption, *and* leak detection at exit) with its associated instrumentation and memory overhead on every access. `--config=lsan` links only the lighter LeakSanitizer runtime — leak detection alone, without ASan's redzone instrumentation, without ASan's ~2-3x memory overhead, and without ASan's shadow-memory requirement. This mirrors Make's own split (`REQSAN=address` vs `REQSAN=leak` produce different `-fsanitize=` values in `makefile:149-151`, not the same flag).

So the two configs are not redundant: `--config=asan` is the superset for full memory-safety validation, `--config=lsan` is the standalone-parity option for a fast leak-only pass, matching Make's existing `REQSAN=leak`. Given that, I'd like to keep this as a separate named config rather than drop it. Let me know if that changes your view, otherwise I'll move this out of draft.

david-cortes-intel (approved):

PR #3714, opened 2026-07-24
bazel: add Linux standard allocator option

napetrov:
## Description

Add a typed Bazel `--stdalloc=true` option for the allocator-selection part of Make `STDALLOC=yes` on Linux DAAL core builds.

This is a focused Make-to-Bazel option-parity improvement toward #3530. The option is disabled by default.

## What changed

- add the typed boolean `--stdalloc=true` interface
- define `USE_STD_ALLOC` only for DAAL core compilation actions
- keep separately built threading objects on TBB scalable allocation, matching Make target scope
- reject non-Linux DAAL configurations clearly
- document the difference from Make+ICX, which also adds `-static-libstdc++`
- add a bounded Bazel configuration-analysis smoke to CI

## Validation

Validated on Linux at exact branch head `8ce584627c78d90ca5d18eb709adff9c85dd1c71`.

### Exact-head GCC build/test

```bash
CC=gcc bazelisk build \
  //cpp/daal:core_static //cpp/daal:core_dynamic \
  //cpp/daal:thread_static //cpp/daal:thread_dynamic \
  --stdalloc=true --cpu=sse2 --jobs=2

CC=gcc bazelisk test \
  //cpp/daal/src/algorithms/dtrees/gbt/regression:test_gbt_regression_model_builder_unit_host \
  --stdalloc=true --cpu=sse2 --jobs=2 --test_output=errors
```

Results: build passed (1,151 actions); 1/1 test passed. CI YAML parse and `git diff --check` also passed.

### ICX proof

Validated with ICX/ICPX 2026.0 and Bazel 9.2:

- enabled core actions: 1,140/1,140 contain `USE_STD_ALLOC`
- enabled threading actions: 0/2 contain `USE_STD_ALLOC`
- default core actions: 0/1,140 contain `USE_STD_ALLOC`
- static/dynamic core and threading build passed
- representative tests passed
- runtime allocation/free probe passed with 64-byte alignment
- core resolves to `aligned_alloc/free`; threading retains TBB scalable allocation

## Scope notes

This option controls allocator selection only. Make builds with `COMPILER=icx STDALLOC=yes` additionally pass `-static-libstdc++`; Bazel link options are intentionally unchanged.

---

**Checklist**

- [x] Changes reviewed locally
- [x] Documentation updated
- [x] Commits include Signed-off-by
- [x] Focused build/test validation completed
- [x] No default behavior changes
- [x] No performance change unless the option is explicitly enabled

### DIFF
--- a/.ci/pipeline/ci.yml
+++ b/.ci/pipeline/ci.yml
@@ -438,6 +438,9 @@ jobs:
       echo "Checking Bazel config: release-dpc"
       bazel build --nobuild :release --config=release-dpc --cpu=all
 
+      echo "Checking Bazel config: stdalloc"
+      bazel build --nobuild :release --stdalloc=true --release_dpc=false
+
       echo "Checking Bazel config: dev"
       bazel build --nobuild //cpp/oneapi/dal:tests --config=dev
     displayName: 'Bazel config smoke checks'
--- a/.bazelrc
+++ b/.bazelrc
@@ -18,6 +18,7 @@ build --flag_alias=release_dpc=@config//:release_dpc
 build --flag_alias=device=@config//:device
 build --flag_alias=cpu=@config//:cpu
 build --flag_alias=enable_assert=@config//:enable_assert
+build --flag_alias=stdalloc=@config//:stdalloc
 
 # Always pass this env variable to test rules, because SYCL
 # OpenCL backend uses it to determine available devices
--- a/cpp/daal/BUILD
+++ b/cpp/daal/BUILD
@@ -156,6 +156,9 @@ daal_module(
 daal_module(
     name = "threading_tbb",
     srcs = glob(["src/threading/**/*.cpp"]),
+    # Match Make STDALLOC=yes: separately built threading objects keep TBB
+    # scalable allocation while core objects switch away from MKL allocation.
+    stdalloc = False,
     visibility_hidden = False,
     local_defines = [
         "__TBB_NO_IMPLICIT_LINKAGE",
--- a/dev/bazel/README.md
+++ b/dev/bazel/README.md
@@ -263,6 +263,14 @@ The most used Bazel commands are `build`, `test` and `run`.
   bazel test --test_disable_fp64 //cpp/oneapi/dal/algo/pca:tests
   ```
 
+- `--stdalloc` Selects standard-library aligned allocation for DAAL core
+  objects. Disabled by default and supported only for Linux targets.
+
+  Example:
+  ```sh
+  bazel build //cpp/daal:core_static --stdalloc=true
+  ```
+
 ## Build recipes for oneDAL
 ### Build release artifacts
 - To build the Bazel release artifacts, run:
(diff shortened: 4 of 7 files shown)


.ci/pipeline/ci.yml
```diff
@@ -438,6 +438,9 @@ jobs:
       echo "Checking Bazel config: release-dpc"
       bazel build --nobuild :release --config=release-dpc --cpu=all
 
+      echo "Checking Bazel config: stdalloc"
+      bazel build --nobuild :release --stdalloc=true --release_dpc=false
```
david-cortes-intel, line 442: @Alexandr-Solovev Would this be a problem with the space constraints that you've been trying to address? And wouldn't this potentially increase running times significantly?

Alexandr-Solovev: For the first question, it should not(especially with disabled dpc) For the second: we can take a look on the job time, because it is hard  to estimate

napetrov: Thanks for raising both — here are the concrete numbers.

**Disk space:** this check is analysis-only. `bazel build --nobuild` stops after the analysis phase, so no compile or link actions execute and no output artifacts are written to `bazel-out`. It only adds one more configured target graph in the analysis cache, which is memory, not disk. With `--release_dpc=false` the DPC++ half of the graph is not configured at all, so it is strictly cheaper than the `release-dpc` smoke check already present in this step. So it should not interact with the space constraints you have been addressing.

**Job time:** measured rather than estimated. `CI (LinuxBazel)` on this PR ran **1h59m02s**, against recent merged PRs on the same job:

| PR | CI (LinuxBazel) |
|---|---|
| **3714 (this PR)** | **1h59m02s** |
| 3727 | 2h00m42s |
| 3726 | 1h57m30s |
| 3724 | 1h59m42s |
| 3711 | 2h00m30s |
| 3710 | 1h57m55s |

That is inside the normal ±2 min spread for this job, so the added check is not measurably distinguishable from baseline. The analysis-only nature is why: it reuses the repository rules and external deps already fetched by the preceding `:release` / `release-dpc` checks in the same step.

Happy to drop the check if you would still rather not carry it, but as measured it appears to be effectively free.

napetrov:
Gentle ping on this one — it is ready for another look.

Status since the last round:

- The two questions on the CI smoke check ([thread](https://github.com/uxlfoundation/oneDAL/pull/3714#discussion_r3672059309)) are answered with measurements rather than estimates: the check is analysis-only (`--nobuild`, no compile/link actions, nothing written to `bazel-out`), and `CI (LinuxBazel)` came in at 1h59m02s against a 1h57m–2h01m baseline on recently merged PRs. @david-cortes-intel please confirm whether that settles your concern, or say the word and I will drop the check.
- Branch was brought up to date with `main` and CI re-ran on the current head: 30 passing.
- The one red check, `oneDAL Windows nightly build`, is pre-existing on `main` and unrelated to this PR. Its failing step is `Compare make and Bazel releases`, and the failure is byte-identical to the nightly on `main` (headSha `3da2496`): the same 7 text mismatches (`config/config.txt`, `include/daal.h`, `daal_kernel_defines.h`, both `oneDALConfig*.cmake`, both `.pc` files) and the same 3 shared-library export mismatches, totalling the same `13 difference(s)`. It has been failing that way on `main` for several consecutive nightlies. This PR cannot affect it: `--stdalloc` is Linux-only and off by default, while the comparison runs the default Windows configuration.

No default behavior changes, and no functional change unless `--stdalloc=true` is passed explicitly.

david-cortes-intel (approved):

PR #3682, opened 2026-07-03
CI: exercise oneDAL Bazel examples under a conda-provided toolchain

Alexandr-Solovev:
## Summary

Extends the existing `LinuxMakeGNU_MKL_Conda` CI job in `.ci/pipeline/ci.yml`
with a second, Bazel-driven pass that builds and runs `//examples/daal/cpp` and
`//examples/oneapi/cpp` targets against the **same conda environment**
(`onedal_env`) that is already used for the make/cmake flow.

The goal is narrow and intentional: prove that oneDAL's Bazel example targets
build and pass with `MKLROOT` / `TBBROOT` / `MPIROOT` pointing at a
`conda-forge` install rather than at a system `/opt/intel/oneapi` toolkit.

No release artifact is produced or uploaded — the existing `LinuxBazel` job
covers packaging/`release_static`/`release_dynamic` link modes, and duplicating
that here would balloon runtime for no additional signal. Only
`--test_link_mode=dev` is used.

## What's added to `LinuxMakeGNU_MKL_Conda`

After the existing `daal/cpp/mpi samples` step, six new steps run in-place
inside the pre-created `onedal_env`:

1. `install-bazel` — `source .ci/env/bazelisk.sh` (reuses the same bazelisk
   installer as `LinuxBazel`).
2. `Install DPC MKL packages (bazel)` — adds
   `onemkl-sycl-{blas,lapack,sparse,rng}` from `conda-forge` on top of the
   MKL/TBB/impi packages already present.
3. `bazel-configure (conda env)` — writes `build --cpu=avx2` into `~/.bazelrc`
   and runs `bazel build @config//:dump` under
   `TBBROOT=MKLROOT=MPIROOT=${CONDA_PREFIX}` so the config repo picks up the
   conda toolchain.
4. `daal-cpp-examples-thread-dev (conda env)` —
   `bazel test //examples/daal/cpp:all --test_link_mode=dev --test_thread_mode=par`.
5. `bazel clean --expunge` between runs (matches `LinuxBazel`'s
   disk-hygiene convention).
6. `cpp-examples-thread-dev (conda env)` —
   `bazel test //examples/oneapi/cpp:all --test_link_mode=dev --test_thread_mode=par`.

Every Bazel-invoking step re-activates the conda env and re-exports
`TBBROOT` / `MKLROOT` / `MPIROOT` from `${CONDA_PREFIX}`. This is required
because Azure spawns a fresh shell per `- script:` block, and Bazel does not
persist those values across invocations.

[...]

PR #3716, opened 2026-07-24
bazel: add code coverage build option

napetrov:
## Description

Add a typed Bazel `--code_coverage=true` option corresponding to Make `CODE_COVERAGE=yes` for supported Linux Intel compiler builds.

This is a focused Make-to-Bazel option-parity improvement toward #3530. The option is disabled by default and does not instrument external dependencies.

## What changed

- add the typed boolean `--code_coverage=true` interface
- apply Make-compatible host compile/link coverage options only to oneDAL-owned actions
- scope `GCOV_BUILD` to Make `CORE.objs_a/y`-equivalent DAAL modules
- keep separately built threading, oneAPI DAL, DPC++, tests, examples, and tools outside the `GCOV_BUILD` define scope
- use `-Xscoverage` for DPC++ link actions while leaving DPC++ compilation unchanged
- preserve coverage runtime linkage for instrumented executable/test actions
- reject unsupported OS/compiler selections clearly
- document the action-level mapping and add a bounded GCC negative smoke to CI

## Validation

Validated on Linux before push; pushed exact head is `d7002469af09b3f84cab54af2d57ed5df9f032c3` (the final two changes after validation were comment/table-whitespace only).

### Focused action matrix

ICX/DPC++ Bazel aquery checks verified:

- DAAL core compile: `GCOV_BUILD` and `-coverage`
- threading compile: `-coverage`, no `GCOV_BUILD`
- oneAPI host compile: `-coverage`, no `GCOV_BUILD`
- DPC++ compile: neither `-coverage` nor `GCOV_BUILD`
- host module/final/executable/test links: `-coverage`
- DPC++ links: `-Xscoverage`
- static archives: no coverage linker option
- default-off actions: no coverage tokens
- no coverage flags leaked into external dependency actions

### Build/test proof

```bash
CC=icx CXX=icpx bazelisk test \
  //cpp/daal/src/algorithms/dtrees/gbt/regression:test_gbt_regression_model_builder_unit_host \
  --code_coverage=true --cpu=sse2 --jobs=2 --test_output=errors
```

Result: **1/1 passed**, producing instrumented objects and `.gcda` counters.

Correction to an earlier revision of this description: the `.gcda` files it quoted were paths under Bazel's sandbox stash, i.e. transient. `.gcno` notes files are undeclared outputs of the compile action, so a sandboxed build discards them, and `.gcda` counters are written to the object's build-time path inside the test sandbox. Producing a gcov report therefore needs `--spawn_strategy=local`; this is now asserted by the CPU smoke and documented in `INSTALL.md` and `dev/bazel/README.md`.

```bash
CC=icx CXX=icpx bazelisk build \
  //cpp/oneapi/dal/table:table_dpc \
  --code_coverage=true --release_dpc=true --cpu=sse2 --jobs=2
```

Result: **passed** with the real DPC++ module link using `-Xscoverage`.

GCC selection was also verified to fail analysis with the intended diagnostic. CI YAML parse, `git diff --check`, and `git show --check` passed.

## Scope notes

Make enables this mode only for `CODE_COVERAGE=yes` on Linux. Bazel is stricter: unsupported targets/compiler selections fail clearly instead of silently disabling coverage. Static archive actions remain flag-free; final executable/test links receive the coverage driver option needed to resolve the runtime from instrumented objects.

## Review

Review comments through 2026-09-15 are addressed in bc80c877: the `schedules:` block is dropped in favour of a manual gate, the DPC++ smoke also checks the released `dynamic_dpc` link, both coverage jobs select the toolchain explicitly, and the `.gcno` survivability gap is closed and documented.

---

**Checklist**

- [x] Changes reviewed locally
- [x] Documentation updated
- [x] Commits include Signed-off-by
- [x] Focused ICX host and DPC++ build/test validation completed
- [x] Default behavior unchanged
- [x] Unsupported selections fail clearly
- [x] No production performance effect unless explicitly enabled


.ci/pipeline/ci.yml
```diff
@@ -388,6 +388,9 @@ jobs:
   - script: |
       .ci/env/apt.sh opencl
     displayName: 'install opencl'
+  - script: |
+      .ci/env/apt.sh dpcpp
+    displayName: 'dpcpp installation'
```
david-cortes-intel: Would this actually get used if the public CI jobs are not running anything with GPUs?

dev/bazel/cc.bzl
```diff
@@ -83,8 +84,58 @@ def _init_cc_rule(ctx, features=[], disable_features=[]):
...
+            owner = linker_input.owner,
+            libraries = depset(linker_input.libraries),
+            user_link_flags = [
+                flag
+                for flag in linker_input.user_link_flags
+                if flag not in coverage_link_flags
+            ],
+            additional_inputs = depset(linker_input.additional_inputs),
+            linkstamps = depset(linker_input.linkstamps),
+        ))
```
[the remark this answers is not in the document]
napetrov: Fixed in 5329e7b: the helper now preserves the existing depsets for libraries, additional_inputs, linkstamps, and user_link_flags. The updated thread is outdated, so resolving.

.ci/pipeline/ci.yml
napetrov: Addressed in d913f3c: public PR CI now runs only the CPU icx coverage smoke, while the DPC++ linker smoke is isolated in a scheduled main-branch nightly job with the full DPC++ runtime. The updated thread is outdated, so resolving.

napetrov:
The single red here (`oneDAL Windows nightly build`) is not from this branch. Its failing step is `Compare make and Bazel releases`:

```
Release comparison failed: 13 difference(s)
```

The same step fails on `main` — six consecutive nightly runs yesterday — and it is the only failing step on #3719, #3735, #3682 and #3740 too. Fixing it is what #3738 is for; that branch now also carries the CRLF staging and comparator work, so once it lands this check should clear here without a change to the coverage code.

Everything else on this PR is green and the review comments are answered.

napetrov:
The two `Nightly-test / oneDAL {Linux,Windows} Bazel test` reds are not from this branch — they fail on every pull request from a fork at `Checkout oneDAL`, because `actions/checkout` refuses fork code in a `workflow_run` job that holds `checks: write`. #3752 is closed; #3757 replaces it with just the fork-checkout fix and the two comparator changes it depends on. Everything else here is green and the review threads are answered.

.ci/env/apt.sh
```diff
@@ -35,6 +35,12 @@ function install_dpcpp {
     sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp=2026.1.1-325 intel-oneapi-runtime-libs
 }
 
+function install_icx {
```
david-cortes-intel, line 47: This could be switching into an argument for the function above, to avoid version mismatches. The one above is at a different version.

.ci/pipeline/ci.yml
```diff
@@ -37,6 +37,16 @@ pr:
...
 
+# DPC++ coverage needs the full compiler/runtime installation. Keep it out of
+# public PR CI; the dedicated job below runs from this scheduled main build.
+schedules:
+- cron: '0 4 * * *'
+  displayName: 'Nightly DPC++ coverage smoke'
+  branches:
+    include:
+    - main
+  always: true
```
david-cortes-intel, line 48: CC @icfaust for comment.

.ci/pipeline/ci.yml
```diff
@@ -608,6 +627,78 @@ jobs:
...
+        df -h .
+        du -sh "${coverage_root}" 2>/dev/null || true
+        rm -rf "${coverage_root}"
+        exit "${result}"
+      }
+      trap cleanup EXIT
+      echo '=== disk before CPU coverage smoke ==='
+      df -h .
+      BAZEL=bazel dev/bazel/tests/code_coverage_test.sh
+    displayName: 'Bazel CPU code-coverage smoke build'
```
david-cortes-intel, line 666: Could this simply be added as an additional output from one of the existing calls?

.ci/env/apt.sh
napetrov: Good catch, and the mismatch was already real: `install_icx` pinned `2026.1.0-235` while `install_dpcpp` three lines above pinned `2026.1.1-325`. Nothing made them agree and nothing would have noticed.

Done in ae55b432 — `install_icx` now calls `install_dpcpp --compiler-only`:

```bash
function install_dpcpp {
    local extra=(intel-oneapi-runtime-libs)
    if [ "${1:-}" == "--compiler-only" ]; then
        extra=()
    fi
    sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp=2026.1.1-325 "${extra[@]}"
}

function install_icx {
    install_dpcpp --compiler-only
}
```

`--compiler-only` is the only thing the separate function was for — icx ships in the DPC++/C++ package, and a CPU-only coverage check does not need `intel-oneapi-runtime-libs`. Side effect, deliberately: the coverage job now gets 2026.1.1-325, the version the rest of CI uses.

One thing worth recording, since it constrains how far this can be de-duplicated: the pin has to stay spelled out literally. Renovate finds it by matching `intel-oneapi-compiler-dpcpp-cpp=<version>` in this file (`.github/renovate.json`, `customManagers`), so hoisting it into a `DPCPP_VERSION` variable would silently drop it from dependency updates. One literal occurrence is both trackable and impossible to desynchronise; two is what produced this.

.ci/pipeline/ci.yml
napetrov: Flagging what @icfaust will want to weigh, because it is larger than it looks and I do not think it should be my call.

`main` has **no `schedules:` block in `ci.yml` at all** — this adds the first one. An Azure cron queues the *whole pipeline*, not a job. `LinuxBazelDpcCoverageNightly` is gated:

```yaml
condition: and(succeeded(), eq(variables['Build.Reason'], 'Schedule'))
```

but none of the other ~14 jobs exclude `Schedule`, so all of them — including `LinuxBazel` (2h) and `WindowsBazel` (up to 5h20m) — would also run at 04:00 every night on `main`, and `always: true` means they run even when nothing changed. On a pool that currently starts roughly one build an hour and was four deep yesterday, that is a large standing cost for one smoke test.

Options, in the order I would rank them:

1. **Drop the cron; gate on `Manual`.** Precedent is `WindowsBazelReleaseAll` in #3740, gated on `Build.Reason` being `Manual` or `Schedule` so it is validated on demand. Cheapest, no pipeline-wide effect, and the DPC++ coverage path is the sort of thing you check when touching coverage rather than nightly.
2. **Hang it off the existing GitHub `Nightly-build`/`Nightly-test`,** which is already a real nightly and already installs DPC++. No Azure cron at all.
3. **Its own pipeline YAML** with its own cron. Cleanest separation, but needs someone with Azure DevOps admin to register the pipeline, so it cannot land as a repo-only change.
4. **Keep the cron and add `ne(variables['Build.Reason'], 'Schedule')` to the other 14 jobs.** Works, but every future job has to remember it, and forgetting is silent.

I have left the block as it is rather than picking for you. Say which and I will make it match — (1) is a two-line change and I can push it immediately.

.ci/pipeline/ci.yml
napetrov: No — and I have measurements for this now rather than just the assertion in the comment.

While diagnosing an unrelated red on #3737 I instrumented the `LinuxBazel` job with `df`. It runs at 92–97% of its 72 Gb agent throughout, with 4.6 Gb free by the time it reaches `cpp-tests-thread-dev`:

```
/dev/root  72G  68G used  4.6G avail  94% /
4.5G  /home/vsts/work/1/.bazel-cache
```

Adding the icx install (~3 Gb) plus a coverage build tree to that job would not fit. It is also why the disk cache is scoped per coverage job here (`BAZEL_OUTPUT_USER_ROOT`/`BAZEL_DISK_CACHE` under `Agent.TempDirectory`, removed by the trap) — a `bazel clean --expunge` does not reclaim a disk cache, so a coverage build sharing `LinuxBazel`'s cache would raise that job's floor for the rest of its run.

The other reason is the flags: the coverage smoke needs `--code_coverage=true` on an `icx` toolchain, which is a different configuration from anything `LinuxBazel` builds. Sharing the job means either building everything twice under two configurations in one workspace, or making the existing steps coverage-instrumented, and neither is what we want.

If the concern is agent count rather than correctness, option (1) in the `schedules:` thread above removes one of these two jobs from every run.

napetrov:
Correction: the apt.sh fix is 2bcd0c22 (pushed to this branch in uxlfoundation), not ae55b432 which went to a stale same-named branch in my fork.

napetrov:
@icfaust @Alexandr-Solovev — this one is waiting on a decision rather than on code, so a one-line answer unblocks it.

The open question is the `schedules:` block at `.ci/pipeline/ci.yml:48`. `main` has no `schedules:` block at all today, so this PR would add the first one, and an Azure cron queues the *whole pipeline* rather than a job. `LinuxBazelDpcCoverageNightly` is gated on `eq(variables['Build.Reason'], 'Schedule')`, but none of the other ~14 jobs exclude `Schedule`, so `LinuxBazel` (2 h) and `WindowsBazel` (up to 5 h 20 m) would run at 04:00 nightly on `main` too, with `always: true` meaning they run when nothing changed.

I deliberately did not pick, since it is a pipeline-wide cost decision. Ranked options are in the thread; my recommendation is **(1) drop the cron and gate on `Manual`**, matching the `WindowsBazelReleaseAll` precedent in #3740. That is a two-line change I can push as soon as you say so. Alternative (2) is hanging it off the existing GitHub `Nightly-build`/`Nightly-test`, which already installs DPC++ and needs no Azure cron.

The second thread, on sharing the existing `LinuxBazel` job instead of adding one, is answered with measurements: `LinuxBazel` already runs at 92–97 % of its 72 Gb agent with 4.6 Gb free, so the icx install (~3 Gb) plus a coverage tree does not fit — quite apart from coverage needing a different toolchain configuration.

Everything else here is green.

.ci/env/apt.sh
```diff
@@ -31,8 +31,23 @@ function add_repo {
...
 function install_dpcpp {
-    # DPC++ compiler version monitored by Renovate and sets exact value available via apt
-    sudo apt-get install -y intel-oneapi-compiler-dpcpp-cpp=2026.1.1-325 intel-oneapi-runtime-libs
+    # DPC++ compiler version monitored by Renovate and sets exact value available via apt.
+    # Pass `--compiler-only` to leave out the oneAPI runtime meta-package. The pin stays
+    # spelled out here, and only here: Renovate finds it by matching
+    # `intel-oneapi-compiler-dpcpp-cpp=<version>` literally in this file
+    # (.github/renovate.json), so moving it into a variable would stop it being tracked,
+    # and repeating it in a second function is how the two drift apart.
+    local extra=(intel-oneapi-runtime-libs)
```
Alexandr-Solovev: local extra=()
if [[ ${1:-} != "--compiler-only" ]]; then
    extra=(intel-oneapi-runtime-libs)
fi
may be its more clear

.ci/pipeline/ci.yml
Alexandr-Solovev: this schedules the whole pipeline, not the coverage job

.ci/pipeline/ci.yml
```diff
@@ -443,6 +453,15 @@ jobs:
...
       bazel build --nobuild //cpp/oneapi/dal:tests --config=dev
+
+      echo "Checking code-coverage rejection with the GCC-only Bazel toolchain"
+      if coverage_error=$(bazel build --nobuild :release --code_coverage=true 2>&1); then
+        echo "ERROR: code coverage unexpectedly accepted the GCC host compiler"
+        exit 1
+      fi
+      printf '%s\n' "${coverage_error}"
+      grep -F -- "--code_coverage=true requires host compiler ID 'icx'" <<< "${coverage_error}"
+      grep -F -- "detected compiler ID: 'gcc'" <<< "${coverage_error}"
```
Alexandr-Solovev, line 461: does it make sense to additionally set/export CC=icx

dev/bazel/tests/code_coverage_dpc_test.sh
```diff
@@ -0,0 +1,57 @@
...
+    startup_args+=("--output_user_root=${BAZEL_OUTPUT_USER_ROOT}")
+fi
+common_args=(--code_coverage=true --release_dpc=true)
+if [[ -n "${BAZEL_DISK_CACHE:-}" ]]; then
+    common_args+=("--disk_cache=${BAZEL_DISK_CACHE}")
+fi
+work="$(mktemp -d)"
+trap 'rm -rf "${work}"' EXIT
+
+target="//cpp/oneapi/dal/table:table_dpc"
```
Alexandr-Solovev, line 34: not sure about the target here, may be makes more sense with //cpp/oneapi/dal:dynamic_dpc

dev/bazel/cc.bzl
```diff
@@ -83,8 +84,58 @@ def _init_cc_rule(ctx, features=[], disable_features=[]):
...
+             "(detected compiler ID: '{}')".format(compiler_id))
+
+    is_dpc = cc_common.is_enabled(
+        feature_configuration = feature_config,
+        feature_name = "dpc++",
+    )
+    # Make adds -coverage to every ICX compilation and to dynamic links.
+    # DPC++ compilations are unchanged; DPC++ dynamic links use -Xscoverage.
+    return struct(
+        compile_flags = [] if is_dpc else ["-coverage"],
```
Alexandr-Solovev, line 116: Got AI comment, may be worth to check 
dev/bazel/cc.bzl:110 — the .gcno files never survive the build
  -coverage is smuggled in through user_compile_flags, so gcov's .gcno notes file is an undeclared output of the
  compile action. With Bazel's default Linux sandboxing it is written inside the sandbox and discarded on teardown;
  there's no --collect_code_coverage, no declared output, and nothing in .bazelrc turning the sandbox off. .gcda has
  the mirror problem: gcov writes it to the object file's build-time path, which doesn't exist when a released
  library runs. Neither test script looks for a .gcno or a .gcda. Has anyone taken this end to end — build, run a
  test, produce a gcov/lcov report? Right now the flag demonstrably produces instrumented objects, and it isn't
  demonstrated that any coverage data can be recovered from them.

Alexandr-Solovev (commented):
Please, address the comments

.ci/env/apt.sh
napetrov: Applied verbatim in bc80c877 — the positive form reads better:

```bash
local extra=()
if [ "${1:-}" != "--compiler-only" ]; then
    extra=(intel-oneapi-runtime-libs)
fi
```

Behavior checked with a stubbed `sudo`: `install_dpcpp` → `intel-oneapi-compiler-dpcpp-cpp=2026.1.1-325 intel-oneapi-runtime-libs`, `install_dpcpp --compiler-only` and `install_icx` → the compiler package alone.

.ci/pipeline/ci.yml
napetrov: Correct, and it is why I left that block for a maintainer to decide rather than picking myself. Removed in bc80c877.

`LinuxBazelDpcCoverageNightly` is now `LinuxBazelDpcCoverage`, gated on `Build.Reason == 'Manual'`, and there is no `schedules:` block — `main` still has none.

Two alternatives I looked at and dropped:

- **A job in `.github/workflows/nightly-build.yml`.** It already crons at 21:00 and already runs `.ci/env/apt.sh dpcpp`, so it looked like the cheapest home. Its own header says the opposite, though: *"This action is only for building oneDAL, implement tests in Nightly-test"*, and *"In all circumstances, this job should stay green"* because sklearnex's `CI` depends on it. A smoke that can go red does not belong there.
- **A job in `nightly-test.yml`.** The right home by that rule, but it is `workflow_run`-triggered with the fork-checkout and check-run plumbing documented at the top of the file, so a new job needs `check_start`/`check_finish` entries and can only be verified by the manual `source_run_id` dispatch dance. Worth doing separately if nightly signal is wanted; too much for this PR.

So: manual gate now, and hanging it off Nightly-test stays available as a follow-up.

.ci/pipeline/ci.yml
napetrov: Done for the coverage jobs in bc80c877 — `export CC=icx` / `export CXX=icpx` after `setvars.sh` in both `LinuxBazelCpuCoverage` and `LinuxBazelDpcCoverage`.

It was not strictly required: `detect_compiler` (`dev/bazel/toolchains/common.bzl:54`) falls through to `detect_default_compiler`, which prefers `icx` whenever `repo_ctx.which("icx")` succeeds, and these jobs source `setvars.sh` first. But that makes the toolchain a property of PATH ordering, and these two jobs are meaningless with anything but icx, so it should be stated.

I did not touch the block this thread is anchored to — that one is the *negative* check, run in `LinuxBazel` where no icx is installed, so `gcc` is what must be detected and asserted. Setting `CC` there would also change the other `bazel build`s in the same script step: `CC` is in the toolchain repo rule's env list, so it refetches the toolchain for the whole step.

dev/bazel/tests/code_coverage_dpc_test.sh
napetrov: Agreed that `dynamic_dpc` is the link that actually matters — it is the released `libonedal_dpc.so`, whereas `table_dpc` is a module link. Building it means building all of oneAPI DAL with DPC++, which is more than a smoke should cost on an agent that is already disk-tight.

Split in bc80c877: the *built* check stays on `table_dpc`, and `dynamic_dpc` is checked from analysis only, so no link executes.

```bash
"${bazel_cmd}" "${startup_args[@]}" aquery "//cpp/oneapi/dal:dynamic_dpc" \
    "${common_args[@]}" --output=jsonproto >"${work}/released_actions.json"
```

then asserting `-Xscoverage` on every `CppLink*` action whose arguments mention `libonedal_dpc`, with the observed mnemonics printed if none match. `bazel query //cpp/oneapi/dal:dynamic_dpc` resolves, so the label is right; the `aquery` itself needs an icx toolchain and so is exercised by the job, not locally.

dev/bazel/cc.bzl
napetrov: That comment is right, and I reproduced it rather than reasoning about it. Thanks for forwarding it.

Minimal workspace, Bazel 9.2.0 (the version in `.bazelversion`), `cc_library` with `copts = ["-coverage"]`, plus a `cc_test` over it:

| build | `.gcno` in output tree | `.gcda` in output tree |
|---|---|---|
| default (sandboxed) | **no** — only `sandbox/sandbox_stash/CppCompile/3/…/a.pic.gcno`, transient | **no** — only `sandbox_stash/TestRunner/32/…` |
| `--spawn_strategy=local` | yes — `bazel-out/k8-fastbuild/bin/_objs/lib/a.pic.gcno` | yes — under `…/t.runfiles/_main/bazel-out/…/_objs/` |

Under local execution it goes end to end: `gcov a.pic.gcda` → `Lines executed:100.00% of 1`, `a.cc.gcov` written. Under default sandboxing both artifacts are discarded and no report is recoverable, exactly as described. `--collect_code_coverage` is a separate mechanism (its own toolchain feature, declared outputs) and this flag neither uses nor changes it.

So the honest scope of the flag is Make *option* parity, and the missing half was that the notes files have to survive for that parity to be worth anything. Addressed in bc80c877:

- `code_coverage_test.sh` now builds with `--spawn_strategy=local` and fails if no `.gcno` is found under `bazel info output_path` — so a regression here is caught in public CI rather than discovered by whoever first tries to produce a report.
- `INSTALL.md` and `dev/bazel/README.md` document the requirement, where the two artifact kinds land, and that `.gcda` must be copied beside its `.gcno` before running `gcov`.

One correction to my own PR description while I am here: the "152 `.gcda` files" I quoted were sandbox-stash paths, i.e. transient, not durable outputs. I have edited the description; the claim should have been that instrumented objects and counters are produced, not that coverage data was recoverable from a default build.

Incidentally the same probe confirmed the link half is load-bearing: a `-coverage`-compiled `.so` linked *without* `-coverage` fails at load with `undefined symbol: __gcov_merge_add`, which is what the link-action flags in this PR prevent.

.ci/env/apt.sh
napetrov: Correction to the commit id above: the fix on this branch is 2bcd0c22, not ae55b432 — the latter went to a stale same-named branch in my fork. The wording of `install_dpcpp` was refined again in bc80c877 per @Alexandr-Solovev's suggestion; the Renovate constraint on keeping the pin spelled out literally still holds.

napetrov:
@Alexandr-Solovev all five addressed, plus latest `main` merged (8bb6f823, clean). Fixes are in bc80c877; per-thread detail is in the replies, summary here:

1. **`apt.sh`** — your positive-form `extra=()` applied verbatim; behavior re-checked with a stubbed `sudo`.
2. **`schedules:`** — removed. You are right that an Azure cron queues the whole pipeline; the DPC++ job is now `LinuxBazelDpcCoverage`, gated on `Build.Reason == 'Manual'`. `main` keeps having no `schedules:` block. I also checked the two GitHub nightly workflows as hosts and explain in the thread why neither fits without a bigger change.
3. **`CC=icx`** — exported explicitly (with `CXX=icpx`) in both coverage jobs. It worked before via `detect_default_compiler` preferring icx off PATH after `setvars.sh`, which is too implicit. The line you commented on is the GCC *negative*, so it stays as is.
4. **DPC++ target** — `dynamic_dpc` is now checked too, from `aquery` analysis only, so the released `libonedal_dpc.so` link is asserted without paying for a full DPC++ build of oneAPI DAL.
5. **The forwarded AI comment about `.gcno`** — it was correct, and I reproduced it on Bazel 9.2.0. Under default sandboxing neither `.gcno` nor `.gcda` reaches the output tree, so a default `--code_coverage=true` build cannot yield a gcov report; with `--spawn_strategy=local` both survive and `gcov` produces one. The CPU smoke now builds that way and fails if no `.gcno` is found, and the requirement is documented in `INSTALL.md` and `dev/bazel/README.md`. I also corrected the misleading `.gcda` count in the PR description.

dev/bazel/cc.bzl
```diff
@@ -83,8 +84,58 @@ def _init_cc_rule(ctx, features=[], disable_features=[]):
...
+        linker_inputs.append(cc_common.create_linker_input(
+            owner = linker_input.owner,
+            libraries = depset(linker_input.libraries),
+            user_link_flags = depset([
+                flag
+                for flag in linker_input.user_link_flags
+                if flag not in coverage_link_flags
+            ]),
+            additional_inputs = depset(linker_input.additional_inputs),
+            linkstamps = depset(linker_input.linkstamps),
```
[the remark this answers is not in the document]
napetrov: Checked this on the pinned Bazel (9.2.0) rather than by inspection, and the premise does not hold: those `LinkerInput` fields are **lists**, not depsets.

A rule that reads a real `CcInfo` and then performs exactly what `_without_coverage_link_flags` performs:

```python
lc = ctx.attr.dep[CcInfo].linking_context
for li in lc.linker_inputs.to_list():
    print("libraries:", type(li.libraries), "user_link_flags:", type(li.user_link_flags),
          "additional_inputs:", type(li.additional_inputs), "linkstamps:", type(li.linkstamps))
    new_li = cc_common.create_linker_input(
        owner = li.owner,
        libraries = depset(li.libraries),
        user_link_flags = depset([f for f in li.user_link_flags if f != "-coverage"]),
        additional_inputs = depset(li.additional_inputs),
        linkstamps = depset(li.linkstamps),
    )
```

Output:

```
DEBUG: probe.bzl:7:14: libraries: list user_link_flags: list additional_inputs: list linkstamps: list
DEBUG: probe.bzl:19:14: rebuilt linker_input ok: struct
```

So `depset(li.libraries)` is well-formed and no nesting occurs. `cc/link.bzl:133-138` is the *construction* side — `create_linker_input` accepts depsets for those parameters — which says nothing about what the provider's getters return; the getters are `Sequence`. The same asymmetry is why the list comprehension over `linker_input.user_link_flags` in this helper is legal: iterating a depset directly is not.

Two further points of evidence: this helper runs on every `cc_module` regardless of the coverage flag, as you note, so the `LinuxBazel` job — which analyzes `//:release` with GCC on every push to this PR — would have failed analysis outright if the call were invalid. And a functionally identical earlier version of this comment (#3652434123) was already resolved, so this looks like a re-report on the same code rather than a new finding.

One thing in it I do agree with: `user_link_flags` is the only field that needs a new depset, and the other three could be passed straight through. That would be marginally clearer, but it is a wash behaviorally, so I am leaving the working code alone while the substantive review items are being settled.

napetrov:
Conflict with `main` resolved and pushed as 23349d45; the PR is `MERGEABLE` again. The conflict came from #3737 (parameters-lib layout) landing on `main` and touching the same three files:

- `.bazelrc` and `dev/bazel/tests/BUILD` — both sides purely additive, both kept.
- `dev/bazel/cc.bzl` — kept #3737's new `_is_windows` helper alongside the coverage helpers, used it in `_cc_exec_impl`, and kept both the coverage (`_linux_constraint`, `_code_coverage`) and `_build_parameters_lib` private attributes on `cc_dynamic_lib`.

Worth noting because it is the kind of thing a merge hides: auto-merge produced a duplicated `ConfigFlagInfo` in the `config.bzl` load list, which is a hard Starlark error —

```
ERROR: dev/bazel/cc.bzl:27:6: load statement defines 'ConfigFlagInfo' more than once
```

Removed, and the resolution then verified by loading the merged rules: `bazel query 'kind("cc_module", //cpp/daal:all)'` resolves on Bazel 9.2.0.

Also replied to the new Copilot comment on `cc.bzl:127`: its premise is wrong on the pinned Bazel — `LinkerInput.libraries`/`user_link_flags`/`additional_inputs`/`linkstamps` are lists, not depsets, verified by a probe rule that reproduces the exact call. Details in the thread.

dev/bazel/cc.bzl
napetrov: CI evidence for the assertion added here: Azure build 63301 (head bc80c877) finished **all green**, and `LinuxBazelCpuCoverage` printed all three checks:

```
2026-09-16T01:01:05Z coverage notes files (.gcno) are present in the output tree
2026-09-16T01:01:35Z error_handling.cpp compile action carries -coverage and -DGCOV_BUILD
2026-09-16T01:01:43Z threading_tbb compile actions do not define GCOV_BUILD
2026-09-16T01:01:43Z code-coverage icx smoke checks passed
```

So the notes files now demonstrably survive on a real icx build in CI, not just in my local probe. `LinuxBazelDpcCoverage` shows as `skipped` in the same build, which is the `Build.Reason == 'Manual'` gate behaving as intended.

napetrov:
@Alexandr-Solovev  - should be good to review

napetrov:
@Alexandr-Solovev  @david-cortes-intel , please review

PR #3719, opened 2026-07-26
Bazel: add AArch64/RISC-V64 cross-compile and OpenRNG backend support

napetrov:
## Summary
- Part of #3530 (Bazel Migration tracker), addressing the P1 gaps: "Add Linux AArch64 Bazel toolchain/build", "Add Linux RISC-V Bazel cross/native scope", and "OpenRNG backend".
- Detects host/target arch from `CC`/`CXX` cross-compiler triple prefixes instead of the hardcoded `intel64`, and registers arch-specific `toolchain()`s and `@config//:linux_aarch64` / `@config//:linux_riscv64` target platforms.
- Extends `cpu_info` to allow `sve`/`rv64` as single-variant ISA values on ARM/RISC-V (no runtime CPU dispatch there, unlike x86's sse2/avx2/avx512), and applies `-march=armv8-a+sve` / `-march=rv64gc_v1p0_zvl128b` plus the matching `DAAL_CPU`/`__CPU_TAG__` defines, mirroring the existing Make `dev/make/compiler_definitions/{gnu,clang}.ref.{arm,riscv64}.mk`.
- Adds `dev/bazel/deps/openrng.bzl` (mirrors `openblas.bzl`) and a `rng_backend` config flag; `--rng_backend=openrng` adds `OPENRNG_BACKEND` and links `@openrng//:openrng` alongside OpenBLAS, matching `dev/make/deps.ref.mk`'s `RNG_OPENRNG`.
- Fixes a real bug in `dev/bazel/cc/compile.bzl` surfaced by `//cpp/oneapi/dal:dynamic` on the new arches: the per-ISA source-file split (`cpu_to_file_dict`, e.g. `algo/jaccard`'s `_skx`-suffixed file) only ever has x86 ISA keys, so `cpu="sve"`/`"rv64"` raised a dict KeyError. Now falls back to the general `_cpu` file list on arches with no such split.
- Extends CI: a Bazel job on the existing GitHub-hosted AArch64 runner (`ci-aarch64.yml`, both `ref` and `openrng` rng backends) and a new `LinuxBazelGNU_OpenBLAS_rv64` job cross-compiling with `riscv64-linux-gnu-gcc` in `.ci/pipeline/ci.yml`. Fixed `.ci/env/bazelisk.sh` to pick the Bazelisk asset matching the CI host arch instead of always downloading the amd64 binary (would exec-format-fail on the arm64 runner).

Only the `ref` backend is supported on ARM/RISC-V, same as Make.

## Test plan
- [x] Downloaded real `aarch64-linux-gnu-gcc`/`riscv64-linux-gnu-gcc` cross toolchains + `qemu-user-static` via `apt-get download` (no root) and validated the full toolchain (compile, link, run under qemu) independent of Bazel first.
- [x] `bazel build --platforms=@config//:linux_aarch64 --backend_config=ref //cpp/daal:core_dynamic //cpp/oneapi/dal:dynamic` with `CC=aarch64-linux-gnu-gcc` — compiles and links; produced `.so`s are valid ARM aarch64 ELF (confirmed via `qemu-aarch64-static` and `readelf`).
- [x] Same for RISC-V64 with `CC=riscv64-linux-gnu-gcc` and `qemu-riscv64-static`.
- [x] `--rng_backend=openrng` with a stub VSL-ABI-compatible `openrng.h`/`libopenrng.a` — confirms the repo/include/define/link wiring end-to-end (compiles and links `core_dynamic`).
- [x] Native x86_64 MKL build (`//cpp/daal:core_dynamic`, `//cpp/oneapi/dal:dynamic`) still builds cleanly — no regression on the existing path, including the x86-specific `_skx` per-ISA source split that the `compile.bzl` fix touches.
- [x] The new CI jobs have now run in this repo's actual CI and pass: `LinuxBazelGNU_OpenBLAS_rv64` (Azure), plus `LinuxBazel_OpenBLAS(SVE)` and `LinuxBazel_OpenBLAS(SVE)_OpenRNG` on the GitHub-hosted AArch64 runner.
- [x] The real OpenRNG library (not the local stub header) is built and linked by the `bazel-build-and-test` lane's `rng: openrng` matrix entry via `.ci/env/openrng.sh`; that lane is green.


napetrov:
## Independent review (4 parallel agents: Bazel/Starlark correctness, Make-parity, CI/infra, security)

**Fixed:** `get_cross_tool_prefix` (`dev/bazel/toolchains/common.bzl`) mishandled versioned cross-compiler binary names like `aarch64-linux-gnu-gcc-13` (installed when only the versioned `gcc-13-aarch64-linux-gnu` package is present, no unversioned meta-package) — it stripped only the trailing `-13` instead of the whole `-gcc-13` tool-name segment, leaving a mangled prefix that made `_find_tools` search for a nonexistent `ar`/`strip` and fail the toolchain repo rule. Fixed and re-verified against real aarch64/riscv64/x86_64 builds.

**Checked and ruled out (no action needed):**
- No other file still hardcodes x86_64/intel64 assumptions that block this PR's scope; `cc_toolchain_win.bzl` confirmed genuinely unaffected; `extra_toolchian_lnx.tpl.BUILD`'s dropped `target_compatible_with` doesn't create resolution ambiguity (verified only one `extra_tools_*` target exists per exec-OS).
- `_get_disabled_cpus`/kernel-defines patching is a no-op for arm/riscv64 (single-element `allowed==enabled`), not a hazard.
- Make-vs-Bazel flag parity for gcc+arm and clang+arm/riscv64 checked flag-by-flag against `dev/make/compiler_definitions/*.mk` — all present, correctly gated, no extras.
- OpenRNG symbol-hiding: Bazel's blanket `-Wl,--exclude-libs=ALL` (pre-existing, PR #3670) is a superset of Make's per-archive `MATH_LIBS_TO_EXCLUDE`, no ABI leak.
- CI: TBB genuinely not needed for the new Bazel jobs (confirmed via grep — `core_dynamic`/`oneapi/dal:dynamic` resolve TBB at runtime on Linux, not link time); `OPENBLASROOT`/`OPENRNGROOT` paths match script defaults exactly; Bazelisk arm64/amd64 asset selection + SHA256 verification hand-traced against live GitHub API JSON, correct for both.
- Security: no novel supply-chain risk. `CC`/`CXX` env values only feed `repo_ctx.which()` (PATH lookup, fails closed, no shell injection). OpenRNG's lack of a URL/sha256 (unlike other deps/*.bzl) matches the pre-existing OpenBLAS/MKL/TBB `ROOT`-env-var trust pattern, not a new gap. Workflow uses `pull_request` (not `pull_request_target`), matrix values hardcoded in trusted YAML — no fork-PR injection path.

**Noted, not blocking:**
- `--rng_backend=openrng` has no arch guard, unlike Make/CI's explicit "OpenRNG only on ARM" check in `build.sh`. Bazel would just proceed and fail later at missing-`OPENRNGROOT`/link time on unsupported arches — fails safe, just less friendly. Left as-is; can tighten in a follow-up if desired.
- Bazel now permits GCC+RISC-V64, a combination Make never exposes (Make's `riscv64.mk` hardcodes clang-only). This was validated end-to-end in this PR (real `riscv64-linux-gnu-gcc` cross-compile+link+qemu-run succeeded), so it's a working superset of Make's coverage rather than a latent bug — flagging for maintainer awareness since it's new build-matrix surface area beyond direct Make parity.

napetrov:
@keeranroth  @rageshh-fj  @rakshithgb-fujitsu  - please take a  look on this PR with bazel support for ARM/RISC

dev/bazel/toolchains/common.bzl
```diff
@@ -67,6 +67,75 @@ def detect_compiler(repo_ctx, os_id):
     elif "icpx" in compiler_path:
         return "icpx"
 
+# oneDAL's arch IDs, matching ARCH_DIR_ONEDAL in dev/make/function_definitions/*.mk.
+_ARCH_ID_BY_TRIPLE_PREFIX = {
+    "aarch64": "arm",
+    "riscv64": "riscv64",
+    "x86_64": "intel64",
+    "amd64": "intel64",
+}
```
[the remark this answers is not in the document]
napetrov: Fixed in bef959d9e. `arm64` is now recognized in host arch detection, but deliberately *not* by adding it to the shared map: `arm64-` is not a GNU triple machine name, so putting it there would make `get_cross_tool_prefix()` match a nonexistent triple prefix. Instead the map is split — `_ARCH_ID_BY_TRIPLE_PREFIX` keeps only real triple prefixes, and `_ARCH_ID_BY_UNAME` is a superset of it with `arm64` added, used only by `detect_host_arch()`.

dev/bazel/flags.bzl
```diff
@@ -103,6 +103,16 @@ def get_default_flags(arch_id, os_id, compiler_id, category = "common"):
             ]
         if compiler_id not in ["icx", "icpx"]:
             flags = flags + ["-fno-strict-overflow"]
+        if arch_id in ["arm", "riscv64"]:
+            # ARM/RISC-V ship a single fixed ISA variant (see cpu_type.h),
+            # unlike x86's runtime-dispatched sse2/avx2/avx512 objects, so
+            # `-march` applies to every compile action, not only the
+            # `_cpu`-suffixed ones that go through get_cpu_flags(). Matches
+            # COMPILER.all.gnu / COMPILER.lnx.clang in
+            # dev/make/compiler_definitions/{gnu,clang}.ref.{arm,riscv64}.mk.
```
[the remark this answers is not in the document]
napetrov: Fixed in bef959d9e. The comment was wrong in two ways: there is no `gnu.ref.riscv64.mk` at all (RISC-V is clang-only in Make), and `clang.ref.riscv64.mk` carries `-march` in `rv64_OPT.clang`, not in `COMPILER.lnx.clang`. Both comments now name the exact variables and files, and state explicitly that for gcc+riscv64 this is Bazel-only coverage rather than Make parity.

dev/bazel/daal.bzl
```diff
@@ -67,6 +69,14 @@ def daal_module(name, features=[], lib_tag="daal",
...
             "//conditions:default": [],
+        }) + select({
+            # Matches dev/make/deps.ref.mk RNG_OPENRNG: swaps the ref RNG's
+            # <random>-based shim for the OpenRNG (VSL-ABI-compatible)
+            # backend in src/externals/service_rng_ref.h.
+            "@config//:rng_backend_openrng": [
+                "OPENRNG_BACKEND",
+            ],
+            "//conditions:default": [],
         }),
```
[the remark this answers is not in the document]
napetrov: Fixed in bef959d9e. Added a two-term `config_setting` `@config//:rng_backend_openrng_ref` requiring `rng_backend=openrng` **and** `backend_config=ref`, and the `OPENRNG_BACKEND` define is now gated on it. This matches Make, where `RNG_OPENRNG` lives in `dev/make/deps.ref.mk` (so unreachable on MKL), and `.ci/scripts/build.sh` gates `--use-openrng yes` on `backend_config == ref` too.

cpp/daal/BUILD
```diff
@@ -35,11 +35,18 @@ daal_module(
...
+    # compiled modules, not just this one.)
     deps = select({
         "@config//:backend_ref": [ "@openblas//:openblas",
                                  ],
         "//conditions:default": [ "@mkl//:mkl_core",
                                 ],
+        }) + select({
+            "@config//:rng_backend_openrng": [ "@openrng//:openrng" ],
+            "//conditions:default": [],
         }),
```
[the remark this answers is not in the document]
napetrov: Fixed in bef959d9e. Both `@openrng//:openrng` and `@openrng//:headers` now select on `@config//:rng_backend_openrng_ref` (the new `rng_backend=openrng` + `backend_config=ref` setting), so `--rng_backend=openrng --backend_config=mkl` no longer requires `OPENRNGROOT` or links libopenrng into a build whose ref RNG shim is never compiled. Verified the aarch64 CI OpenRNG lane passes both flags, so it stays enabled there — and both OpenRNG jobs are green on this commit.

napetrov:
The single red here (`oneDAL Windows nightly build`) is not from this branch. Its failing step is `Compare make and Bazel releases`:

```
Release comparison failed: 13 difference(s)
```

The same step fails on `main` — six consecutive nightly runs yesterday — and is the only failing step on #3716, #3735, #3682 and #3740 as well. #3738 is the fix; it now carries the CRLF staging and comparator work on top of the staged-intermediates changes. Nothing to do in the ARM/RISC-V branch for it.

All four Copilot findings are addressed in bef959d9e; still waiting on a look from @keeranroth / @rageshh-fj / @rakshithgb-fujitsu.

napetrov:
@theComputeKid  - might be you would also be interested in this one

theComputeKid:
Admittedly, I am not a Bazel nor cross compilation expert but I will check within my team if anyone has worked with Bazel.

napetrov:
The two `Nightly-test / oneDAL {Linux,Windows} Bazel test` reds are not from this branch — they fail on every pull request from a fork at `Checkout oneDAL`, because `actions/checkout` refuses fork code in a `workflow_run` job that holds `checks: write`. #3752 is closed; #3757 replaces it with just the fork-checkout fix and the two comparator changes it depends on. Everything else here is green and the review threads are answered.

napetrov:
@theComputeKid @rakshithgb-fujitsu @keeranroth @rageshh-fj — following up on the review request here, and with one thing cleared up that may have been putting people off.

@theComputeKid, you mentioned in August you would ask within your team whether anyone had worked with Bazel. No pressure if that came to nothing — but if it did, this is the PR.

**The two red `Nightly-test / oneDAL {Linux,Windows} Bazel test` checks are not from this branch.** They failed on every pull request from a fork, at `Checkout oneDAL`, because `actions/checkout` refuses fork code in a `workflow_run` job holding `checks: write`. That is fixed by #3757, which is now merged, and by #3782 for the one real difference #3757 then exposed. The checks shown here are stale results from 16 August and predate both. Nothing in this PR is implicated.

For anyone who would rather review a slice than the whole thing, the content splits cleanly:

- **cross-compilation toolchains** — `dev/bazel/toolchains/**`, including the `get_cross_tool_prefix` fix for versioned cross-compiler binary names such as `aarch64-linux-gnu-gcc-13`, which appear when only the versioned `gcc-13-aarch64-linux-gnu` package is installed
- **OpenRNG backend** — the RNG dependency and its wiring
- **CI** — the AArch64 and RISC-V64 job definitions

Reviewing any one of those three is more useful than none, and I am happy to split the PR along those lines if that makes it reviewable.

theComputeKid:
@aditew01 : could you take a look please.

rakshithgb-fujitsu:
hi @napetrov apologies for the late review. From the looks of what's added/changed it looks fine at the moment.

rakshithgb-fujitsu (approved):

.ci/env/bazelisk.sh
```diff
@@ -16,6 +16,23 @@
 #===============================================================================
 
 BAZELISK_VERSION=v1.29.0
+
+# Bazelisk itself always runs on the CI *exec* host (even when the build
```
Alexandr-Solovev, line 20: Could be simplified:
```
host_arch=$(uname -m)

case "$host_arch" in
  x86_64|amd64)  arch=amd64 ;;
  aarch64|arm64) arch=arm64 ;;
  *)
    echo ":error: Unsupported host architecture for Bazelisk: $host_arch" >&2
    exit 1
    ;;
esac

BAZELISK_ASSET="bazelisk-linux-${arch}"
```

.ci/pipeline/ci.yml
```diff
@@ -608,6 +608,70 @@ jobs:
...
+      bazel build --platforms=@config//:linux_riscv64 --backend_config=ref \
+        //cpp/daal:core_dynamic //cpp/oneapi/dal:dynamic
+    displayName: 'bazel build (riscv64 cross)'
+  - script: |
+      set -euo pipefail
+      for lib in bazel-bin/cpp/daal/libonedal_core.so bazel-bin/cpp/oneapi/dal/libonedal.so; do
+        file "${lib}"
+        readelf -h "${lib}" | grep -q "RISC-V" || { echo "##vso[task.logissue type=error]${lib} is not a RISC-V ELF"; exit 1; }
+      done
+    displayName: 'Verify RISC-V ELF output'
```
Alexandr-Solovev, line 797: Really makes sense to somehow validate the build, may be examples/tests?

dev/bazel/config/config.bzl
```diff
@@ -87,40 +88,75 @@ CpuInfo = provider(
     ],
 )
 
-_ISA_EXTENSIONS = ["sse2", "avx2", "avx512"]
-_ISA_EXTENSIONS_MODERN = ["sse2", "avx2", "avx512"]
```
Alexandr-Solovev, line 118: offtop question, why its duplicated here and whats purpose for _MODERN?

dev/bazel/toolchains/common.bzl
```diff
@@ -67,6 +67,84 @@ def detect_compiler(repo_ctx, os_id):
...
+    return _ARCH_ID_BY_UNAME.get(value)
+
+def get_cross_tool_prefix(repo_ctx):
+    """Returns the cross-toolchain triple prefix (e.g. `aarch64-linux-gnu-`)
+    if `CC`/`CXX` point at a cross-compiler, otherwise `""`.
+
+    Used to resolve the matching `ar`/`strip`/etc for the target arch instead
+    of picking up the exec host's native binutils.
+    """
+    for env_var in ["CC", "CXX"]:
```
Alexandr-Solovev: could it be moved in external function, because looks pretty complex and hard to maintenance

Alexandr-Solovev (commented):
Overall looks good with minor comments, would be great if will be extended with some real work case inside the ci job.

.ci/env/bazelisk.sh
napetrov: Adopted as suggested in 88c07817 — `uname -m` is resolved once into an `arch` suffix instead of being called twice.

dev/bazel/toolchains/common.bzl
napetrov: Fixed in 88c07817: the triple parsing is now in two helpers, and both callers are one line each.

- `_detect_cross_compiler(repo_ctx)` — scans `CC`/`CXX` for a known triple prefix and returns `(triple_prefix, arch_id)`, or `(None, None)` for a native build. This loop was previously duplicated between `get_cross_tool_prefix()` and `detect_target_arch()`.
- `_triple_prefix_to_triple(basename)` — the fiddly part: drops the tool-name segment to recover the prefix, tolerating a trailing `-<version>` (`aarch64-linux-gnu-gcc-13` -> `aarch64-linux-gnu-`, which is what Debian installs when only the versioned `gcc-13-aarch64-linux-gnu` package is present).

`get_cross_tool_prefix()` and `detect_target_arch()` are now thin wrappers over `_detect_cross_compiler()`.

Verified after the refactor: `CC=aarch64-linux-gnu-gcc` and `CC=aarch64-linux-gnu-gcc-13` both give `cpu: {"allowed":["sve"],"enabled":["sve"]}`, `CC=riscv64-linux-gnu-gcc` gives `rv64`, and the native x86_64 path still resolves the toolchain (`bazel build --nobuild //cpp/daal:core_dynamic`).

dev/bazel/config/config.bzl
napetrov: Not something this PR introduced — the duplication predates it, and it is dead: `_ISA_EXTENSIONS` and `_ISA_EXTENSIONS_MODERN` on `main` hold the same three values, so `--cpu=all` and `--cpu=modern` already select the same set.

They diverged only while `sse42` was a build variant: `all` included it, `modern` did not. `sse42` was removed in #3487, and the two lists have been identical ever since.

Since this PR had to touch the surrounding code anyway (to split the list per architecture), I removed the duplicate in 88c07817: there is now one `_ISA_EXTENSIONS_X86`, and `all`/`modern` share a branch with a comment recording the history. Behavior is unchanged — verified that the `@config//:dump` output for `--cpu=all`, `--cpu=modern`, `--cpu=auto` and `--cpu=avx2` matches what `main` produces (`{"allowed":["sse2","avx2","avx512"],"enabled":[...]}` identical in all four cases).

.ci/pipeline/ci.yml
napetrov: Agreed, and done for AArch64 in 88c07817 — that lane now runs real workloads instead of only inspecting the produced ELFs:

```
bazel test --platforms=@config//:linux_aarch64 --backend_config=ref --rng_backend=<ref|openrng> \
  --test_link_mode=dev --test_thread_mode=par \
  //examples/daal/cpp:all   # and //examples/oneapi/cpp:all
```

Both matrix legs (`ref` and `openrng`) run it, mirroring the `daal-cpp-examples-thread-dev` / `cpp-examples-thread-dev` steps of the x86 Bazel lane.

Two things had to be fixed to make that possible:

1. **oneTBB.** The `@tbb` Bazel repo defaults to an x86_64 prebuilt wheel. It did not matter for the build-only job (TBB is resolved at runtime, not at link time), but running the tests needs a real aarch64 TBB, so the job now builds it with `.ci/env/tbb.sh` and passes it through `TBBROOT` — the same dependency the Make lanes in this workflow already build.
2. **`examples/daal/cpp/BUILD` hardcoded `@mkl//:mkl_static`** in `_TEST_DEPS`. The DAAL examples link the kernels statically, so they need the math backend explicitly — but with MKL hardcoded they are unbuildable on any non-MKL platform, ARM included. That dep is now a backend select (`@openblas//:openblas` on `@config//:backend_ref`, `@mkl//:mkl_static` otherwise), mirroring the select in `//cpp/daal:BUILD`. Checked with `bazel cquery somepath(...)`: under `--backend_config=ref` the examples reach `@openblas//:openblas` and no longer reach `@mkl//:mkl_static`; the default MKL path is unchanged and `//examples/daal/cpp:cholesky_dense_batch_host` still builds and passes on x86_64.

Two exclusions, both mirroring `EXCLUDE_LIST` in `examples/daal/cpp/CMakeLists.txt`:
- `enable_thread_pinning` — hangs on AArch64; the Make lanes in this same workflow work around it with `timeout 12m ... || true`.
- six `decision_forest` examples on the `openrng` leg — the same ones CMake skips for ref+OpenRNG.

(CMake's ref-backend `brownboost` exclusion has no Bazel counterpart; the boosting examples are not wired into the Bazel build.)

For the RISC-V64 job I left the cross-build + ELF check as-is: there is no native RISC-V runner, so executing anything there means running under `qemu-riscv64-static` on the x86 host. That is how I validated this locally while developing the PR, but wiring it into CI is a separate piece of work (runfiles and loader paths for a cross-linked test), so I would rather do it as a follow-up than hold this PR on it. Happy to do it here instead if you prefer.

Alexandr-Solovev:
ERROR: /home/runner/work/oneDAL/oneDAL/cpp/daal/BUILD:286:16: Action cpp/daal/libonedal_thread.a failed: missing input file '@@+openblas_repo+openblas//:lib/libgfortran.a
Likely need to extend openblas deps with this lib

napetrov:
Pushed `44b25f79` — the `libgfortran.a` failure is fixed by **removing** the entry, not by supplying the library.

Root cause: `openblas_repo` listed `lib/libgfortran.a` in `libs`, but oneDAL builds OpenBLAS with `NO_FORTRAN=1` (`.ci/env/openblas.sh`), so that file never exists under `OPENBLASROOT`. `repos.prebuilt_libs_repo_rule` symlinks non-glob entries unconditionally, so `repo_ctx.symlink` produced a *dangling* symlink: the label `@openblas//:lib/libgfortran.a` resolved fine at analysis time and only blew up when an action tried to consume it —

```
ERROR: cpp/daal/BUILD:286:16: Action cpp/daal/libonedal_thread.a failed:
missing input file '@@+openblas_repo+openblas//:lib/libgfortran.a'
```

That is why only the new AArch64 lanes went red: the rv64 lane builds shared libraries only, which don't consume the OpenBLAS archives, and the x86 lanes use MKL.

Make agrees — `dev/make/deps.ref.mk` links `libopenblas.a` alone, and `-lgfortran` in `openblas.tpl.BUILD` was already commented out. Both the `libs` list and the template `srcs` now carry just `lib/libopenblas.a`.

Verified locally on x86 against a real OpenBLAS v0.3.29 build (`NO_FORTRAN=1`, no `libgfortran.a` in the prefix):
- `bazel build --backend_config=ref //cpp/daal:thread_static` — the exact action that failed in CI — succeeds.
- `bazel test --backend_config=ref --test_link_mode=dev --test_thread_mode=par //examples/daal/cpp:cholesky_dense_batch_host //examples/oneapi/cpp:kmeans_lloyd_dense_batch_host` — 2/2 pass.

napetrov:
Both AArch64 Bazel failures diagnosed and fixed in `f078186b`. They were two unrelated things.

### 1. oneAPI examples: missing non-x86 CPU dispatch macros (real Bazel bug)

`dal_generate_cpu_dispatcher` in `dev/bazel/dal.bzl` only ever wrote `ONEDAL_CPU_DISPATCH_{SSE2,AVX2,AVX512}`. On AArch64 the generated `_dal_cpu_dispatcher_gen.hpp` was therefore empty, so `ONEDAL_IF_CPU_DISPATCH_A8SVE(...)` in `cpp/oneapi/dal/backend/dispatcher_cpu.hpp` expanded to nothing and `dispatch_by_cpu()` fell straight through to

```cpp
throw unsupported_device{ dal::detail::error_messages::sve_not_supported() };
```

Result: **58 of 63** `oneapi/cpp` examples failed; the only 5 that passed (`homogen_table`, `csr_table`, `column_accessor_homogen`, `csr_accessor`, `read_batch`) are the ones that never call an algorithm. `daal/cpp` was unaffected (111/111 pass) because the DAAL side has its own dispatcher.

Make has always emitted these defines — `dev/make/function_definitions/arm.mk:72` and `riscv64.mk:74`. The generator now mirrors them:

```python
("#define ONEDAL_CPU_DISPATCH_A8SVE\n" if sets.contains(cpus, "sve")  else "") +
("#define ONEDAL_CPU_DISPATCH_RV64\n"  if sets.contains(cpus, "rv64") else "")
```

Note this was latent for RISC-V too — the rv64 lane only builds libraries, so nothing exercised it.

Verified locally (the generator is a Starlark write action, so no cross toolchain needed): with `CC=aarch64-linux-gnu-*` the header gets `ONEDAL_CPU_DISPATCH_A8SVE`, with `CC=riscv64-linux-gnu-*` it gets `ONEDAL_CPU_DISPATCH_RV64`, and x86 is byte-identical (`SSE2` + `AVX512` for `--cpu=auto` on this host). x86 examples still pass (`kmeans_lloyd`, `pca_cor`, `df_cls_default_dense`).

### 2. OpenRNG lane: pre-existing decision forest instability (not a Bazel issue)

The OpenRNG lane failed a *different* step — `daal/cpp examples`, 102/105 — on `df_cls_traverse_model`, `df_reg_default_dense_extratrees_batch`, `df_reg_hist_dense_extratrees_batch`.

That is not new. On **main**, the Make ref+OpenRNG AArch64 lane fails `df_reg_default_dense_extratrees_batch`, `df_reg_hist_dense_extratrees_batch` and `df_reg_traverse_model` with `errno 139` (SIGSEGV) — it just goes unnoticed because those steps end in `|| true`. Which of `df_cls_traverse_model` / `df_reg_traverse_model` crashes differs between the Make and Bazel builds, so I excluded the whole training-based forest set from the OpenRNG Bazel lane (CMake's `OPENRNG_BACKEND` `EXCLUDE_LIST` already covers 4 of them) and kept `df_cls_dense_batch_model_builder`, which trains nothing. Non-OpenRNG lane runs the full set — all of them pass there.

Worth a separate issue for the ref+OpenRNG forest segfaults on AArch64; happy to file it.

Also added `--test_output=errors` to both example steps, since the previous failure gave only a path into the runner's Bazel cache.

napetrov:
Both AArch64 Bazel lanes are now green (`LinuxBazel_OpenBLAS(SVE)` 14m15s, `LinuxBazel_OpenBLAS(SVE)_OpenRNG` 17m33s), so `CI AArch64` passes. Summary of the last push and of the two remaining reds/pendings, neither of which comes from this branch.

### `1f7d4509` — four oneAPI examples excluded in the OpenRNG lane

With `--test_output=errors` added in `f078186b`, the four remaining failures printed nothing at all and exited on `Segmentation fault`:

```
//examples/oneapi/cpp:connected_components_batch_host   FAILED
//examples/oneapi/cpp:df_cls_dense_batch_host           FAILED
//examples/oneapi/cpp:df_reg_dense_batch_host           FAILED
//examples/oneapi/cpp:louvain_batch_host                FAILED
```

These are the same pre-existing ref+OpenRNG instability as the decision-forest exclusions in the `daal/cpp` step, not a Bazel wiring problem:

* `df_reg_dense_batch`, `louvain_batch` and `connected_components_batch` are exactly the `OPENRNG_BACKEND` `EXCLUDE_LIST` in `examples/oneapi/cpp/CMakeLists.txt` — CMake has never built or run them in this configuration.
* `df_cls_dense_batch` is not in that list, but it segfaults in `main`'s Make ref+OpenRNG lane too: `.ci/scripts/test.sh: line 294: 70875 Segmentation fault (core dumped)` → `FAILED df_cls_dense_batch with errno 139`. It goes unnoticed because that step ends in `|| true`.

The exclusions apply only when `matrix.rng == openrng`; the ref-RNG lane keeps running all 63 oneAPI examples, and all 63 pass. The Bazel lane deliberately does not use `|| true`, so anything that regresses here will actually turn the job red.

Happy to file a separate issue for the AArch64 ref+OpenRNG decision-forest/graph segfaults so the exclusions have something to point at — say the word and I will, with the Make-lane evidence above.

### `Nightly-test / oneDAL Windows Bazel test` — GitHub API rate limit, not this branch

The failing step is `Install Bazelisk`, at `.ci/env/bazelisk.ps1:49`:

```
API rate limit exceeded for 52.157.32.48. (But here's the good news: Authenticated
requests get a higher rate limit. ...)
```

That is an unauthenticated `api.github.com` call for the Bazelisk release metadata hitting the 60 requests/hour per-IP limit on the runner. The identical failure hit `main` on 2026-09-14 (run 34905140840, same job, same step, same line). This branch touches neither `nightly-test.yml` nor `bazelisk.ps1`, and its `bazelisk.sh` change adds no API call. I re-ran the job; `Install Bazelisk` passed on the second attempt.

A token is not the right fix here: `nightly-test.yml` runs these jobs with read-only permissions and no secrets on purpose, because they check out fork code. The robust fix is to drop the API call — pin the Bazelisk SHA256 in `bazelisk.sh` / `bazelisk.ps1` and download straight from the release URL. I can open that as a separate PR if maintainers want it.

### `CI (LinuxBazel)` pending — stalled Azure build, requeued

Build 63331 sat with `LinuxBazel` unscheduled for 14 hours while every other job in it succeeded, and the build record itself stayed at `notStarted`. Builds queued *later* ran their own `LinuxBazel` to completion in the meantime (63332: 22:26 → 02:20). PR #3740's build 63326 is stuck in the same state, so this is not specific to this branch. Requeued via `/azp run CI` above (build 63341).

PR #3703, opened 2026-07-21
ci: remove debug for public ci jobs

Alexandr-Solovev:
## Summary

Trims the public Linux CI build to a release-mode, disk-frugal configuration:
- Drop `--debug symbols` from all three `build.sh` invocations (`daal`, `oneapi_c`, `oneapi_dpc`).
- Delete the `cp -r __work __work_daal` step that previously preserved the DAAL intermediate build tree on non-`ubuntu-24.04` runners.
- Clean `__work` after the DAAL build step (previously only cleaned after the oneAPI steps).

Net effect on `LinuxMakeDPCPP` (job that produces the shared `__release_lnx` artifact):
- Smaller build artifacts (no `-g`).
- Lower peak disk usage between steps.
- Lower peak memory during icx/DPC++ compile — relevant because the SYCL TUs are the primary OOM risk on GitHub-hosted runners.

**Completeness and readability**

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [x] I have run it locally and tested the changes extensively.
- [x] All CI jobs are green or I have provided justification why they aren't.
- [x] I have extended testing suite if new functionality was introduced in this PR.

**Performance**

- [x] I have measured performance for affected algorithms using [scikit-learn_bench](https://github.com/IntelPython/scikit-learn_bench) and provided at least a summary table with measured data, if performance change is expected.
- [x] I have provided justification why performance and/or quality metrics have changed or why changes are not expected.
- [x] I have extended the benchmarking suite and provided a corresponding scikit-learn_bench PR if new measurable functionality was introduced in this PR.

</details>

PR #3725, opened 2026-07-29
feat: restore debug build for c part of oneDAL

Alexandr-Solovev:
## Description

Restores debug info for the **host** libraries of the public `LinuxMakeDPCPP(avx2)` build, so that the
`ABI Conformance(avx2)` job goes back to being a full, type-level ABI check on them — and makes the part
that stays uncovered explicit instead of silent.

### Background

Debug info was dropped from all three `build.sh` invocations of this job in #3703, because the debug
build of the DPC++ target exhausted the public GitHub runner (RAM, then disk) and was blocking
development on #3593 (HDBSCAN). Many compiler/linker options were tried first (device-code
compression, `-flink-huge-device-code`, reduced job count, split links); none brought peak RSS under
the runner limit.

That change had a side effect that was not documented at the time: this job's `__release_lnx` tree is
the input of the `ABI Conformance(avx2)` job, and `abidiff` needs DWARF for anything beyond
symbol-level comparison. Without it the gate degrades to "was a symbol added or removed", which misses
exactly the breaks the pImpl-heavy public API is most likely to introduce:

* a new data member (or base, or virtual) on an exported class — `sizeof` changes, no mangled name does;
* a changed return type — return types do not participate in Itanium mangling.

Thanks @icfaust for pushing on this and for the concrete `train_result<Task>` example.

### What this PR does

1. **Restores `--debug symbols` for the `daal` and `oneapi_c` targets.** `REQDBG=symbols` maps to
   `-g -fno-system-debug` for icx, so `libonedal_core.so`, `libonedal.so`, `libonedal_thread.so` and
   `libonedal_parameters.so` carry DWARF again and `abidiff` performs full function/variable sub-type
   comparison on them. This is the configuration that was in place before #3703, so it is known to fit
   the runner.
2. **Leaves the `oneapi_dpc` target without debug info**, with the reason and a TODO in the workflow.
   DPC++ device DWARF is what actually blew up per-object size and peak RSS. `libonedal_dpc.so` and
   `libonedal_parameters_dpc.so` therefore keep the reduced, symbol-level check.
3. **Reports the resulting blind spot.** `.ci/scripts/abi_check.sh` now also diffs the set of *newly
   exported* symbols of each host library against its DPC++ counterpart (`libonedal.so` vs
   `libonedal_dpc.so`, `libonedal_parameters.so` vs `libonedal_parameters_dpc.so`). A symbol added only
   to the DPC++ flavor is public surface that `abidiff` cannot type-check in this configuration, so it
   is listed in the job log for manual review. This works on `.dynsym` alone, so it needs no debug info.

### Why the drift report is a warning and not a failure

`libonedal_dpc.so` is built from every host translation unit **plus** the `*_dpc.cpp` ones
(`ONEAPI.srcs.dpc` in the top-level `makefile`), compiled with `-DONEDAL_DATA_PARALLEL`. Its symbol set
is therefore structurally a superset of the host library's: every PR that adds a GPU implementation
legitimately adds DPC++-only symbols. Failing the job on that would push ordinary GPU feature PRs
towards the `API/ABI breaking change` label — which makes the workflow skip the whole job and would
disable the `abidiff` gate along with the message it was meant to silence. The report also filters the
namespaces that `.github/.abignore` already drops (`daal::*::internal`,
`oneapi::dal::{backend,detail,preview}`), so both halves of the script agree on what "public" means.

A genuine `nm` or read failure *does* fail the job: an ABI gate that silently degrades to an empty
symbol set would report "no drift" and pass green.

### Coverage after this PR

| library | debug info | `abidiff` coverage |
|---|---|---|
| `libonedal_core.so` | yes | full (types, layout, signatures) |
| `libonedal.so` | yes | full |
| `libonedal_thread.so` | yes | full |
| `libonedal_parameters.so` | yes | full |
| `libonedal_dpc.so` | no | symbol addition/removal + drift report |
| `libonedal_parameters_dpc.so` | no | symbol addition/removal + drift report |

Since the host/DPC++ split in scikit-learn-intelex means the host libraries carry the interfaces most
consumers link against, this restores the check where it matters most. Debug builds of the DPC++ target
remain covered by private CI.

### Validation

* `abi_check.sh` cross-flavor logic exercised against purpose-built shared objects for: a CPU-only
  addition (present in both flavors → `no drift`), a GPU-only addition (reported, job stays green), an
  internal-namespace addition (filtered, as `.abignore` does), a truncated 60-symbol listing, and an
  unreadable library (job fails with `::error::`).
* `bash -n` clean; `ci.yml` parses.
* Public CI on this PR is the check that the host debug build still fits the runner.

### Follow-ups

* Restoring debug info for `oneapi_dpc` needs either device-DWARF suppression
  (`-Xarch_device -g0` / `-fsycl-device-debug=none`, to be measured) or a larger runner. TODO left in
  `ci.yml` next to the invocation.
* After merge, the `CI` workflow should be re-run on `main` via workflow dispatch to regenerate the
  `__release_lnx` cache. Until that happens, PRs compare a build **with** host DWARF against a cached
  `main` build **without** it.

Related: #3703 (debug removal), #3593 (the build that hit the limit).

---

<details>
<summary>Checklist:</summary>

- [x] I have updated the documentation to reflect the changes: the ABI coverage of this build is described in the `Check ABI conformance` step's own log output, so it is visible to anyone reading a CI run, and in comments next to the build invocation.

**Testing**

- [x] I have extended testing suite if new functionality was introduced in this PR: not applicable, the new logic is a CI check. It was validated against purpose-built shared objects locally (see Validation above); there is no harness in the repo for testing `.ci/scripts` themselves.

**Performance**

- [x] I have provided justification why performance and/or quality metrics have changed or why changes are not expected: CI-only change, no library code is touched. Debug info affects the size of the CI build artifacts, not the shipped release packages.

</details>


.ci/scripts/abi_check.sh
```diff
@@ -40,4 +40,114 @@ do
...
+# therefore disable the abidiff gate above to silence an expected message.
+if ! command -v nm >/dev/null 2>&1; then
+    echo "::error:: nm not found (required for cross-flavor symbol drift report)"
+    exit 1
+fi
+
+# Namespaces that oneDAL does not treat as part of the public ABI. Kept in
+# sync with the [suppress_*] entries of .github/.abignore so that this report
+# and abidiff agree on what "public" means; update both together.
+internal_sym_re='_Z+.*4daal[[:digit:]].*8internal[[:digit:]].*|_Z+N6oneapi3dal[[:digit:]].*(7backend|6detail|7preview)[[:digit:]].*'
```
icfaust, line 71: This looks like a sneaky maintenance burden. Shouldn't we just parse .github/.abignore? I know thats more code and likely to be more annoying up front, but this will likely lead to some issues down the line.

icfaust (commented):
We should note this behavior w.r.t the other ABI checker (specifically dpc symbols). Small thing for maintainability, and then we can bring it in.

PR #3738, opened 2026-08-07
bazel: fix Windows release parity inputs

napetrov:
## Summary
- keep generated Windows `.exp` import-library intermediates outside the staged release tree
- pass Windows module-definition files with `/DEF:` rather than treating them as response files

The nightly Windows comparator had been failing on three staged `.exp` files and broad DLL export divergence. `/DEF:` restores the explicit export contract used by the Make build; no broad export suppression is added.

## Validation
- `git diff --check`
- `bazel query //:release --noshow_progress`

## Follow-up validation
Windows build is required to run Make and Bazel release packaging plus the level-4 comparator. The remaining generated text-file differences must be inspected as actual diffs before any normalization is considered.


.ci/scripts/build.bat
```diff
@@ -44,5 +44,11 @@ IF "%VS_VER%"=="2017_build_tools" (
 echo make %1 -j%NUMBER_OF_PROCESSORS% COMPILER=%2 PLAT=win32e REQCPU=%3
 make %1 -j%NUMBER_OF_PROCESSORS% COMPILER=%2 PLAT=win32e REQCPU=%3 || set errorcode=1
 
-cmake -DINSTALL_DIR=__release_win_vc\daal\latest\lib\cmake\oneDAL -DARCH_DIR=intel64 -P cmake\scripts\generate_config.cmake || set errorcode=1
+rem No cmake config generation here: the makefile stages the oneDALConfig files
+rem itself through `_release_cmake_configs`, added in #2243 shortly after this
```
Alexandr-Solovev: If these # means number of pr, looks like a big mistake, because the latest PRs merged are 3600+, worth to doublecheck

cpp/oneapi/dal/BUILD
```diff
@@ -92,37 +92,64 @@ dal_static_lib(
...
+    # (`ONEAPI.objs_y.dpc.lib += $(PARAMETERS.objs_y.dpc.filtered)`). Those
+    # contexts are tagged "parameters", so the tag has to be accepted here or
+    # the objects never reach the link and their dllexport'ed
+    # `train_parameters_cpu` members are missing from the DLL's exports.
+    # Linking `static_parameters_dpc` instead does not work: an archive is
+    # searched only for unresolved references, while the export table is built
+    # from the whole-archived object bundle.
+    dpc_lib_tags = select({
+        "@platforms//os:windows": ["dal", "parameters"],
+        "//conditions:default": ["dal", "daal", "mkl_embed"],
```
Alexandr-Solovev, line 139: I did not get why for Linux parameters are not in the tags?

.ci/scripts/build.bat
napetrov: Double-checked, the numbers are right — they are deliberately *old* PRs, which is the whole point of the argument:

- #2222 "Updating oneDAL examples", merged 2023-01-04 — this is what added the `generate_config.cmake` call to `build.bat`.
- #2243 "makefile helpers", merged 2023-01-20 — this is what taught the makefile to stage the configs itself via `_release_cmake_configs`.

So the call here has been redundant since two weeks after it landed, over three years ago. I reworded the comment in a4712a7 to include the merge dates explicitly so the low numbers do not read as a typo.

cpp/oneapi/dal/BUILD
napetrov: Because on Linux Make does not put them in `libonedal_dpc.so` at all — it builds them as a separate library.

`makefile:64`:
```make
DEFAULT_BUILD_PARAMETERS_LIB := $(if $(OS_is_win),no,yes)
```

With `yes` (Linux), `makefile:343` emits a standalone `libonedal_parameters.so`. With `no` (Windows), the makefile instead folds the parameter objects straight into the DPC library:

```make
ONEAPI.objs_y.dpc.lib += $(PARAMETERS.objs_y.dpc.filtered)
```

So the tag set has to differ to match: accepting `"parameters"` on Windows is what gets those objects into `onedal_dpc.dll`, and *not* accepting it on Linux is what keeps them out of `libonedal_dpc.so`. If Linux accepted it, the definitions would be duplicated between `libonedal_dpc.so` and `libonedal_parameters.so`, and the release comparator would flag the extra exports.

Added this reasoning as a comment in a4712a7.

cpp/oneapi/dal/BUILD
```diff
@@ -92,37 +92,42 @@ dal_static_lib(
...
+    # tags here instead pulls their objects into onedal.dll via /WHOLEARCHIVE
+    # and widens its export surface far beyond the Make contract. On Linux,
     # libonedal.so resolves DAAL symbols at runtime via DT_NEEDED
-    # libonedal_core.so, so the host-only build keeps just the "dal" tag to
-    # avoid duplicating DAAL into onedal.so.
-    lib_tags = select({
-        "@platforms//os:windows": ["dal", "daal", "mkl_embed"],
-        "//conditions:default": ["dal"],
+    # libonedal_core.so, so the host build keeps just the "dal" tag; the DPC
+    # variant keeps the tag set it has always had.
```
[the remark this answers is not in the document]
napetrov: Fixed in a4712a7. The comment no longer claims the DPC variant keeps its historical tags; it now spells out that only the non-Windows branch does:

> Only the non-Windows DPC branch keeps the historical `["dal", "daal", "mkl_embed"]` set; the Windows one is narrowed below.

and the `dpc_lib_tags` select right underneath carries its own block explaining why Windows needs `["dal", "parameters"]` (`BUILD_PARAMETERS_LIB` is `no` there, so Make folds the DPC parameter objects into `onedal_dpc.dll`) and why Linux must not accept `"parameters"` (Make emits `libonedal_parameters.so` separately, makefile:343).

Alexandr-Solovev (approved):

PR #3723, opened 2026-07-29
feat: sparse support for online oneDAL BasicStatistics

Alexandr-Solovev:
## Summary

Enables `partial_compute` + `finalize_compute` on `csr_table` inputs for the
`basic_statistics` algorithm on both CPU and GPU. Previously only `method::dense`
was wired through the online path; `method::sparse` was batch-only, so sklearnex
`IncrementalBasicStatistics` fell back to the dense path on sparse inputs.

The partial-result schema is unchanged (same six fields as dense: `partial_n_rows`,
`partial_min`, `partial_max`, `partial_sum`, `partial_sum_squares`,
`partial_sum_squares_centered`). For CSR:

- `sum` / `sum_squares` over the stored NNZ values equal the same reductions over
  the implicit densified matrix (missing entries are zeros).
- `sum_squares_centered` is recovered exactly at finalize time via
  `sum2cent = sum_squares - sum^2 / n`, so the intermediate carries no sparsity
  information and the dense finalize math is reused as-is.
- `min` / `max` already clamp against implicit zeros in the existing CSR batch
  kernel, so the per-batch partial values are already correct against a dense
  view.

## Changes

**CPU (routes through the existing DAAL `fastCSR` online kernel)**

- `cpp/oneapi/dal/algo/basic_statistics/backend/cpu/partial_compute_kernel_csr.cpp` — new.
  Dispatches to `daal_lom::internal::LowOrderMomentsOnlineKernel<Float, fastCSR, Cpu>`
  with `is_online = true`; carries the previous partial state in when present.
  Rejects per-row weights (`unimplemented`) because the DAAL fastCSR online
  signature has no weighted overload.
- `cpp/oneapi/dal/algo/basic_statistics/backend/cpu/finalize_compute_kernel_csr.cpp` — new.
  Same DAAL finalize call as the dense CPU path but reading the CSR partials.
- `cpp/daal/src/algorithms/low_order_moments/low_order_moments_csr_fast_online_fpt_cpu.cpp` —
  added `DAAL_EXPORT` on the explicit `LowOrderMomentsOnlineKernel<..., fastCSR, ...>`
  instantiation so the symbol is visible to `libonedal.so` (matches the pattern
  in other `*_fpt_cpu.cpp` files whose kernels are called from oneAPI backends).

**GPU (reuses the dense GPU compute + finalize)**

- `cpp/oneapi/dal/algo/basic_statistics/backend/gpu/partial_compute_kernel_csr_dpc.cpp` — new.
[...]

PR #3740, opened 2026-08-09
bazel: decouple Windows MSVC runtime from --compilation_mode, add `d` suffix

napetrov:
## Problem

On Windows the Bazel build has no notion of the MSVC runtime as a separate axis — it derives the CRT from Bazel's built-in `dbg` feature, i.e. from `--compilation_mode`:

```python
# dev/bazel/toolchains/cc_toolchain_config_win.bzl (before)
dbg_feature = feature(name = "dbg")
runtime_library_feature = feature(name = "runtime_library", enabled = True, flag_sets = [
    flag_set(..., flag_groups=[flag_group(flags=["-MDd"])], with_features=[with_feature_set(features=["dbg"])]),
    flag_set(..., flag_groups=[flag_group(flags=["-MD"])],  not_features=["dbg"]),
])
```

The Makefile keeps these apart: `MSVC_RUNTIME_VERSION=release|debug` picks the CRT (an ABI choice that also switches the `msvcrt`/`tbb` variants) while `REQDBG` adds debug info and assertions. In Bazel they were fused, so `--config=dbg-symbols` gave `-g` with MD, and `--config=dbg` forced `-MDd` together with `-O1`. There was no way to ask for just the debug CRT.

Second, library names were identical in both cases. Make encodes the runtime in the name (`makefile:124`, `makefile:337-347`):

```
onedal_cored.lib   onedal_cored.4.dll   onedal_cored_dll.lib   onedal_cored_dll.4.lib
```

Without that, MD and MDd outputs collide and a single release tree carrying both — which the Make build produces from one `RELEASEDIR` and two `__work/{md,mdd}` — is not expressible.

This was also a latent runtime bug, not only a packaging gap: `-MDd` defines `_DEBUG`, and `cpp/daal/src/externals/core_threading_win_dll.cpp:36-40` then loads `onedal_threadd.4.dll`, which no Bazel build emitted. `bazel build //:release --config=dbg` on Windows produced a `onedal_core.dll` that aborts at startup with *"Cannot load onedal_thread.4.dll"*.

## Changes

**Runtime as its own axis.** New `msvc_runtime_debug` toolchain feature drives `-MDd`; `dbg` no longer does. Because `select()` cannot read toolchain features, dependencies read a paired `@config//:msvc_runtime` build setting. `--config=mdd` sets both; a mismatch between them fails with an explicit message rather than silently publishing MD objects under debug names.

**Names.** `_msvc_runtime_suffix()` in `dev/bazel/cc.bzl` appends `d` on Windows debug-CRT builds, applied to the static lib, the DLL, and the import lib. The suffix goes before the ABI version, as in Make. Linux/macOS paths are untouched (suffix is always empty).

**Dependencies follow the runtime.** `_debug` TBB variants (`makefile:330`) and `sycld.lib` rather than `sycl.lib` (`makefile:811`). The latter moved out of the repository rule — it depends on a per-build feature, not a repository-time constant. Generated pkg-config files carry matching names and CRT flag.

## Usage

```bat
bazelisk build //:release                :: release CRT, unsuffixed names -- DEFAULT, unchanged
bazelisk build //:release --config=mdd   :: debug CRT, `d`-suffixed names
bazelisk build //:release_all            :: both flavours in one release tree
```

`--config=mdd` composes with `--config=dbg` (debug CRT *and* debug info), which was the point of separating them.

`//:release_all` builds the release rule twice under Bazel transitions and merges `lib/intel64` + `redist/intel64`; headers, datasets, examples, env scripts and CMake config come from the release-CRT build only, since they are runtime-independent — the filter is deliberately narrower than `lib/`, because `lib/pkgconfig` and `lib/cmake` hold identically named files in both flavours. On non-Windows platforms `mdd` and `release_all` are no-ops (`release_all` forwards the tree rather than copying, so `.so` version symlinks are not dereferenced).

`oneDALConfig.cmake` needed no change — it already derives `DAL_DEBUG_SUFFIX` from `CMAKE_BUILD_TYPE` on the consumer side.

## Validation

Statically reviewed; no Windows or Bazel available in the authoring environment, so **none of this has been compiled** — the CI Windows job is the first real check. Reviewers should watch for:

- the first `-MDd` link: Make also passes `/nodefaultlib:libucrtd.lib ucrtd.lib` (`makefile:330`), whereas Bazel relies on clang-cl's `/defaultlib` directives;
- `tbb12_debug` / `tbbmalloc_debug` presence in the TBB layout on the agent (globbed in `tbb.bzl`, named explicitly in the `select()` so a miss fails loudly rather than silently dropping the dep);
- `//:release_all` output-collision behaviour on a full Windows tree.

A CI step builds `--config=mdd` and asserts the suffixed names exist. Note `WindowsReleaseCompare` still compares against a Make tree built with the default `MSVC_RUNTIME_VERSION=release`; comparing the MDd tree would need a matching Make build, which this PR does not add.


.ci/pipeline/ci.yml
```diff
@@ -685,6 +685,68 @@ jobs:
...
+      $libDir = Join-Path $root "lib\intel64"
+      $redistDir = Join-Path $root "redist\intel64"
+      if (-not (Test-Path $libDir)) { throw "Missing directory: $libDir" }
+      # The full set of debug-runtime artifacts, matching what the Make build
+      # produces for MSVC_RUNTIME_VERSION=debug (makefile:337-347, 962-970).
+      $expectedLib = @("onedal_cored.lib", "onedal_threadd.lib", "onedald.lib",
+                       "onedal_cored_dll.lib", "onedal_cored_dll.4.lib",
+                       "onedald_dll.lib", "onedald_dll.4.lib")
+      $expectedRedist = @("onedal_cored.4.dll", "onedal_threadd.4.dll",
+                          "onedald.4.dll")
```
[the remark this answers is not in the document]
napetrov: Fair, and it would have failed on names that were correct. Fixed in 86a56014 — the major is read from the constant the build uses:

```powershell
$configBzl = "dev\bazel\config\config.bzl"
$majorMatch = Select-String -Path $configBzl -Pattern '^_BINARY_MAJOR\s*=\s*"(\d+)"'
if (-not $majorMatch) { throw "Could not read _BINARY_MAJOR from $configBzl" }
$major = $majorMatch.Matches[0].Groups[1].Value
```

then interpolated into `onedal_cored_dll.$major.lib`, `onedal_cored.$major.dll` and the rest. It throws if the constant cannot be found, rather than quietly comparing against an empty string and passing.

Note `_BINARY_MAJOR` is defined twice — `dev/bazel/config/config.bzl:19` and `dev/bazel/repos.bzl:19` — which is a pre-existing duplication I have not touched here. The check reads the config one.

dev/bazel/scripts.bzl
```diff
@@ -154,15 +154,24 @@ def _generate_pkgconfig_impl(ctx):
...
     if ctx.target_platform_has_constraint(ctx.attr._windows_constraint[platform_common.ConstraintValueInfo]):
+        # A debug MSVC runtime renames every oneDAL library (`onedal_cored.lib`)
+        # and switches both the CRT flag and the TBB variant, so the .pc file has
+        # to follow. See `_msvc_runtime_suffix` in dev/bazel/cc.bzl.
+        is_rt_debug = ctx.attr._msvc_runtime[ConfigFlagInfo].flag == "debug"
+        d = "d" if is_rt_debug else ""
         # Keep this list in sync with the `_WIN32` branch of
         # `deploy/pkg-config/pkg-config.cpp`: the dynamic package exposes only
         # onedal and onedal_core, and Windows has no parameters library at all
         # (`makefile` errors out on `BUILD_PARAMETERS_LIB=yes` there).
```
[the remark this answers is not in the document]
napetrov: Right, the comment overstated it. Reworded in 86a56014.

What has to stay in sync is *which* libraries appear — the dynamic package exposing only onedal and onedal_core, and Windows having no parameters library. The names are deliberately not a 1:1 match and cannot be: the deploy-side generator knows nothing about the MSVC runtime, while this one carries the `d` suffix, `/MDd` and the debug TBB variant. The comment now says that explicitly, so nobody tries to make the two identical.

.ci/pipeline/ci.yml
```diff
@@ -685,6 +685,75 @@ jobs:
...
+      # Read the ABI major from the same constant the build uses, so an ABI bump
+      # does not have to be repeated here to keep this check honest.
+      $configBzl = "dev\bazel\config\config.bzl"
+      $majorMatch = Select-String -Path $configBzl -Pattern '^_BINARY_MAJOR\s*=\s*"(\d+)"'
+      if (-not $majorMatch) { throw "Could not read _BINARY_MAJOR from $configBzl" }
+      $major = $majorMatch.Matches[0].Groups[1].Value
+      Write-Host "ABI major from ${configBzl}: $major"
+      # The full set of debug-runtime artifacts, matching what the Make build
+      # produces for MSVC_RUNTIME_VERSION=debug (makefile:337-347, 962-970).
+      $expectedLib = @("onedal_cored.lib", "onedal_threadd.lib", "onedald.lib",
```
Alexandr-Solovev, line 731: should it be extended with dpc libs? I guess it build dpc libs as well

dev/bazel/deps/mkl.bzl
```diff
@@ -39,7 +39,12 @@ mkl_repo = repos.prebuilt_libs_repo_rule(
     win_libs = [
         "lib/mkl_core.lib",
         "lib/mkl_intel_ilp64.lib",
-        "lib/mkl_tbb_thread.lib",
+        # Globbed so the debug-CRT threading layer (`mkl_tbb_threadd.lib`,
+        # see dev/make/deps.mkl.mk:51) is symlinked too when the MKL layout
+        # ships it. `mkl_win.tpl.BUILD` names both variants explicitly in its
+        # select(), so a debug-runtime build with the file absent fails with a
+        # plain missing-input error instead of silently linking the release one.
+        "lib/mkl_tbb_thread*.lib",
```
Alexandr-Solovev, line 47: just wondering where mkl_intel_thread_dll which is used as deps for sycl?

.ci/pipeline/ci.yml
napetrov: Yes — `--release_dpc` defaults to `True` (`dev/bazel/config/config.tpl.BUILD:150`) and this job never turns it off, so `//:release --config=mdd` does build `onedal_dpc` via `:release_dpc_windows` in `BUILD`. The check was blind to it: `onedal_dpcd` could have lost its suffix entirely and this step would still have passed.

Extended in 5508ed93 — `onedal_dpcd_dll.lib`, `onedal_dpcd_dll.$major.lib` and `onedal_dpcd.$major.dll` are now required, and a leftover `onedal_dpc_dll.lib` joins the unexpected list alongside `onedal_core.lib` and friends.

Only the dynamic DPC++ flavour is expected: `BUILD` ships `cpp/oneapi/dal:dynamic_dpc` on Windows and no static DPC++ library, and no parameters libraries at all on Windows. So there is no `onedal_dpcd.lib` or `onedal_parameters_dpcd.lib` to look for — the release-CRT tree has no counterpart to those either. Noted in a comment next to the list so it does not read as an omission.

dev/bazel/deps/mkl.bzl
napetrov: It is still picked up — `mkl.bzl:50` symlinks it and `mkl_dpc_utils` links it at `dev/bazel/deps/mkl_win.tpl.BUILD:54`, unchanged by this PR.

It is deliberately not globbed like the TBB threading layer above it, because MKL ships no debug-CRT variant of it and Make does not ask for one: `dev/make/deps.mkl.mk:54` is `mkl_intel_thread_dll.$a` with no `$d`, while the static TBB threading archive on line 51 is `mkl_tbb_thread$d.$a`. Only the static threading layer has two flavours. The `_dll` names are import libraries, so the CRT is baked into the MKL DLL rather than into what we link, and there is nothing runtime-dependent to select. Globbing it would only widen the pattern to a filename that cannot exist.

napetrov:
Merged `main` (`ed7e0e07`) into the branch — the conflict was with #3737, which landed the `BUILD_PARAMETERS_LIB` layout control. Three files, all additive collisions rather than semantic ones:

* `.bazelrc` — both sides appended a `--flag_alias`; kept both (`build_parameters_lib` and `msvc_runtime`).
* `dev/bazel/config/config.tpl.BUILD` — both added a `config_flag` at the same spot; kept both. Note `config_flag` generates the per-value settings, so `@config//:msvc_runtime_debug` and `@config//:build_parameters_lib_*` both still exist.
* `dev/bazel/cc.bzl` — `main` factored out an `_is_windows(ctx)` helper where this branch had the inline `target_platform_has_constraint` call, so the local copies are gone and the runtime validation now runs off the shared helper. `cc_static_lib` gained both private attributes (`_msvc_runtime`, `_build_parameters_lib`); a duplicated `ConfigFlagInfo` in the `load()` came out of the merge and is removed.

`BUILD` merged cleanly and #3737's version of the parameters-library selection wins outright — this branch's earlier Windows `select()` there is superseded, which is the right outcome.

Checked locally with Bazel 9.2.0 (analysis only, Linux host):

```
bazel build //:release     --nobuild --cpu=avx2   # Analyzed target //:release (244 packages, 4778 targets)
bazel build //:release_all --nobuild --cpu=avx2   # Analyzed target //:release_all (2543 targets)
bash dev/bazel/tests/parameters_layout_test.sh    # BUILD_PARAMETERS_LIB configured-graph checks passed
```

So #3737's own configured-graph test still passes on the merge result. The `-MDd` path itself is Windows-only and is on CI.

Also, per @Alexandr-Solovev's point above, `Verify debug-runtime library names` now covers the DPC++ library — `onedal_dpcd_dll.lib`, `onedal_dpcd_dll.$major.lib`, `onedal_dpcd.$major.dll`, with a leftover `onedal_dpc_dll.lib` treated as unexpected.

napetrov:
CI (LinuxBazel) failure is the `install-bazel` flake, not this branch:

```
:error: Failed to fetch Bazelisk release information from GitHub API.
##[error]Bash exited with code '1'.
```

`.ci/env/bazelisk.sh` queries `api.github.com/repos/bazelbuild/bazelisk/releases/tags/...` unauthenticated to read the release asset digest, and a hosted agent that lands on a rate-limited egress IP fails the step outright. Same step failed on unrelated builds today — #3723 in builds `63312` and `63284`. Nothing was compiled before it died.

/azp run

napetrov:
Conflict was `dev/bazel/README.md` — #3719 (AArch64/RISC-V cross-compile + OpenRNG) appended rows to the same Make→Bazel flag table this branch appends the `MSVC_RUNTIME_VERSION` rows to. Kept both, MSVC rows first. Merged `main` at `40e35d88` in `51d1f170`; everything else auto-merged (`.bazelrc`, `.ci/pipeline/ci.yml`, `config.tpl.BUILD`, `flags.bzl`, `scripts.bzl`, `onedal_win.tpl.BUILD`).

On the comparison failure in `CI (LinuxReleaseCompare)`: it is job-to-job source skew, not a difference this branch introduces.

```
! lib/pkgconfig/dal-static-threading-host.pc Make=1280B Bazel=1332B
-Libs: ... -lmkl_core -lmkl_intel_ilp64 -lmkl_tbb_thread ...
+Libs: -Wl,--start-group ... -l:libmkl_intel_ilp64.a -l:libmkl_tbb_thread.a -l:libmkl_core.a -Wl,--end-group ...
```

The link group is #3789, which changed both generators together. Build `63326` was queued 2026-09-16 15:35 UTC and ran its jobs 16 hours apart on `refs/pull/3740/merge`:

* `LinuxMakeGNU_MKL` — 2026-09-16 18:49, before #3789 landed in `main` (2026-09-17 04:08 UTC), so its `.pc` has the plain `-l` form.
* `LinuxBazel` — 2026-09-17 07:14, after it, so its `.pc` has the group.

`LinuxReleaseCompare` then compared a pre-#3789 Make tree against a post-#3789 Bazel tree and reported the one line that #3789 changed. The build's pinned `sourceVersion` (`36ab1f51`) has the plain form in *both* generators, and `main` build `63348` at `40e35d88` — which contains #3789 — passes the comparison with no failed tasks. So neither tree is wrong; they are from different sources.

`51d1f170` now contains #3789, so both trees in the next run come from a source where the two generators agree.

Checked locally, Bazel 9.2.0 on Linux:

```
bazel build //:release //:release_all --nobuild --cpu=avx2   # Analyzed 2 targets (250 packages, 6060 targets)
bazel build //:release_pkgconfig_static_threading_host //:release_pkgconfig_dynamic_threading_host
#   static:  -Wl,--start-group ... -l:libmkl_intel_ilp64.a -l:libmkl_tbb_thread.a -l:libmkl_core.a -Wl,--end-group
#   dynamic: ... -lmkl_core -lmkl_intel_lp64 -lmkl_tbb_thread
```

i.e. #3789's shape survives the merge on both files; this branch only touches the `_WIN32` branch of that generator.

Earlier `CI (LinuxBazel)` red on `install-bazel` was the GitHub API rate-limit flake and passed on the rerun.

Alexandr-Solovev (approved):

PR #3743, opened 2026-08-11
[fix] return Nightly-build to operation, add comments and prevent agent modification

icfaust:
## Description

Testing was introduced into `.github/workflows/nightly-build.yml` even though there were code comments added to prevent it. Extensive comments are added to the `nightly-build.yml` and `nightly-test.yml` github action workflows in order to explain what their uses are. Secondly, an AGENTS.md is added along with inline code comments to prevent agentic violation of a hard-set intra-repository dependence which may not be obvious to agentic development. Finally, all testing in `nightly-build.yml` was transferred to `nightly-test.yml`. An action (`github-script`) is used to report the success or failure of jobs from `nightly-test.yml` to pull requests via the check boxes.

This should help to prevent breakage in the future, and limit downstream impacts on sklearnex.

NOTE: private CI not required (due to only github-related code)

---

<details>
<summary>Checklist:</summary>

**Completeness and readability**

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.
- [x] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [x] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [x] I have run it locally and tested the changes extensively.
- [x] All CI jobs are green or I have provided justification why they aren't.
- [x] I have extended testing suite if new functionality was introduced in this PR.

**Performance**

- [x] I have measured performance for affected algorithms using [scikit-learn_bench](https://github.com/IntelPython/scikit-learn_bench) and provided at least a summary table with measured data, if performance change is expected.
- [x] I have provided justification why performance and/or quality metrics have changed or why changes are not expected.
- [x] I have extended the benchmarking suite and provided a corresponding scikit-learn_bench PR if new measurable functionality was introduced in this PR.

</details>

PR #3744, opened 2026-08-11
ci: publish nightly artifacts before comparing releases

napetrov:
## Problem

The Windows nightly has failed **17 nights running** — every night from 2026-07-26 (run [30220918643](https://github.com/uxlfoundation/oneDAL/actions/runs/30220918643)) through 2026-08-10 (run [31433209459](https://github.com/uxlfoundation/oneDAL/actions/runs/31433209459)) — always at step 11, `Compare make and Bazel releases`.

The three upload steps sat *after* that comparison. GitHub skips every later step once one fails, so `__release_win`, `intel_oneapi_basekit` and `opencl_rt_installer` have not been published since **2026-07-24** (run [30127774523](https://github.com/uxlfoundation/oneDAL/actions/runs/30127774523)).

This workflow declares at the top that it "is a dependency of uxlfoundation/scikit-learn-intelex's `CI` GitHub action" — so that repository has had no fresh Windows oneDAL build for over two weeks.

## Nothing regressed in the release

The 2026-07-24 run recorded as green also reported:

```
Check level: 4
Only Bazel files: 3
Text content mismatches: 5
Shared library export mismatches: 3
Release comparison failed: 11 difference(s)
```

The step passed anyway, because Windows PowerShell 5.1 does not turn a native command's nonzero exit into a failure. #3701 added `exit $LASTEXITCODE` to `.ci/scripts/compare_windows_release.ps1`, which made the gate start working from the next nightly. The differences were always present; only their consequence changed.

Closing those differences is the subject of #3738. This PR is only about not making downstream pay for them.

## Change

- Packaging and uploading move ahead of the comparison, on both the Linux and Windows jobs. The comparison still fails the job, so the signal is unchanged.
- `Build and run CMake example from Bazel release` stays after the comparison — it validates the Bazel tree rather than producing an artifact — but gets `if: ${{ !cancelled() }}` so it still reports. It is the only end-to-end check that the Bazel release is consumable (headers, cmake package files, import libraries, runtime DLLs), and it has not run since 2026-07-24.
- Windows job timeout 180 → 240 minutes. The last run that got past the comparison took 165m, runs that fail fast at the comparison already reach 173m, and the steps that only execute once the gate passes cost ~5m — so the first green nightly would likely have died on the timeout instead.

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -83,14 +83,9 @@ jobs:
         run: |
           source /opt/intel/oneapi/setvars.sh
           bazel build //:release --config=release-dpc --cpu=avx2
-      - name: Compare make and Bazel releases
-        run: |
-          source /opt/intel/oneapi/setvars.sh
-          python dev/release_tests/compare_release_trees.py \
-            --platform linux \
-            --make ./__release_lnx/daal/latest \
-            --bazel ./bazel-bin/release/daal/latest \
-            --check-level 4
+      # Upload before validating. Downstream consumers depend on these artifacts,
+      # so a release-parity difference must not also cost them a build. The
+      # comparison below still fails the job, which is the signal we want.
       - name: Archive build
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -101,12 +96,23 @@ jobs:
         with:
           name: oneDAL_env
           path: .ci/env
+      - name: Compare make and Bazel releases
```
icfaust, line 99: Please see #3743, due to logic on sklearnex side we really should not be testing in this file (failure of any step will lead to a failure in nightly-build and lead to problems in discovery on sklearnex side).

.github/workflows/nightly-build.yml
```diff
@@ -160,22 +166,11 @@ jobs:
...
-        run: |
-          call .\oneapi\setvars.bat
-          call "%ONEAPI_ROOT%\setvars-vcvarsall.bat" %VS_VER%
-          powershell -NoProfile -ExecutionPolicy Bypass -File .\.ci\scripts\test_bazel_release_cmake_example.ps1 ^
-            -ReleaseDir .\bazel-bin\release\daal\latest
+      # Upload before validating. scikit-learn-intelex's CI consumes
+      # `__release_win` (see the note at the top of this file), so a
+      # release-parity difference must not also cost that repository its Windows
+      # build. The comparison below still fails the job, which is the signal we
+      # want, but the artifacts are already published by then.
```
icfaust, line 173: Unfortunately insufficient without changes on sklearnex side: https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/.github/workflows/ci.yml#L53

We don't check for the latest artefacts, we check for the latest run which succeeds and then grab the artefacts (as this otherwise would be problematic on PRs, which may be more recent).

PR #3747, opened 2026-08-12
bazel: stage the documentation tree into the release, as make does

napetrov:
## Description

The makefile copies a documentation tree into `daal/latest/documentation` (`makefile:442-444`, `:1051-1058`). Its source, `$(DIR)/../documentation`, lives *outside* the repository and the recipe is guarded by `if [ -d ... ]`, so the Make package ships those files only in build trees that have them.

That is where the release BOM's `documentation/en/common/license.txt`, `third-party-programs.txt`, `third-party-programs-mkl.txt` and `third-party-programs-onetbb.txt` come from. The Bazel release rule has **no** documentation step at all (`grep -r documentation dev/bazel/` finds nothing), so the BOM generator reports every one of them missing when packaging a Bazel-produced release:

```
WARNING - Unable to find .../daal/latest/documentation/en/common/license.txt for component lnx_license_en_redist
WARNING - Unable to find .../daal/latest/documentation/en/common/third-party-programs.txt for component lnx_licensing_redist
...
```

## Changes

* new `:release_documentation` filegroup, `glob(["documentation/**"], allow_empty = True)`, added to `release(data = ...)`. `_copy_data` already keys destinations off `short_path`, so files land at `daal/latest/documentation/...` with structure preserved.
* `.gitignore` entry for the staged tree.

## Note for the packaging side

Bazel cannot glob above the workspace, so unlike make (which reads `../documentation`) the tree has to be visible **inside** the workspace as `documentation/`. The build that provides the docs must symlink or copy it there; when it is absent the glob expands to nothing, exactly like the makefile's `if [ -d ]` guard. Alternatively the packaging job can copy the tree into the release directory after `bazel build //:release` and this hook is not needed — happy to drop it if that is the preferred split.

## Verification

Not built locally. `bazel build //:release` with no `documentation/` present must behave exactly as today (empty glob).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,7 @@ MODULE.bazel.lock
 # Configurations from file explorers
 .directory
 .DS_Store
+
+# Documentation tree staged by the build (see the `release_documentation`
+# filegroup in the root BUILD file); generated outside the repository.
+/documentation/
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,5 @@ MODULE.bazel.lock
 .directory
 .DS_Store
 
-# Documentation tree staged by the build (see the `release_documentation`
-# filegroup in the root BUILD file); generated outside the repository.
-/documentation/
+# Documentation tree staged for the release build
+documentation/
--- a/dev/bazel/README.md
+++ b/dev/bazel/README.md
@@ -303,41 +303,33 @@ The most used Bazel commands are `build`, `test` and `run`.
 
 ### Documentation tree in the release
 
-The release package is expected to carry a `documentation/` tree — the release
-BOM lists `documentation/en/common/license.txt` and the
-`third-party-programs*.txt` files from it. That tree is **not** part of this
-repository. It is produced by a separate documentation build, and both the Make
-and the Bazel packaging step only copy it when it happens to be present:
+The release package carries a `documentation/` tree — the release BOM lists
+`documentation/en/common/license.txt` and the `third-party-programs*.txt` files
+from it. That tree is **not** part of this repository; it comes from a separate
+documentation build. Make takes it from beside the repository and skips it when
+it is not there:
 
-| | Make | Bazel |
-|---|---|---|
-| Source | `$(DIR)/../documentation`, i.e. next to the repository | `documentation/` inside the workspace |
-| Destination | `daal/latest/documentation` | `daal/latest/documentation` |
-| When absent | recipe is guarded by `if [ -d ... ]` (`makefile:442-444`, `:1051-1058`) | `glob(..., allow_empty = True)` expands to nothing |
+```make
+DOC.srcdir:= $(DIR)/../documentation
+release.DOC := $(shell if [ -d $(DOC.srcdir) ]; then find $(DOC.srcdir) -not -wholename '*.svn*' -type f ;fi)
+```
 
-Bazel cannot glob above the workspace, so the tree has to be reachable from
-inside it. Stage it before building the release, by symlink or copy:
+Bazel cannot glob above the workspace, so the tree has to be staged inside it as
+`documentation/`, by symlink or copy, before building the release:
 
 ```sh
 ln -sfn /path/to/documentation documentation
 bazel build //:release
 ```
 
-`documentation/` is in `.gitignore` so a staged tree is never committed. The
-`release_documentation` filegroup in the root `BUILD` file picks it up, and the
-release rule copies it with its directory structure preserved, so nothing has to
-enumerate the documentation layout.
+The `release_documentation` filegroup in the root `BUILD` file globs it with
+`allow_empty = True`, so an absent tree is not an error, and the release rule
+copies it to `daal/latest/documentation` with its structure preserved.
+`/documentation/` is listed in `.gitignore`, so a staged tree is never
+committed.
 
-Nothing in CI stages the tree today, which means the Bazel release built by CI
-ships without documentation, exactly as the Make release does there. Whoever
-produces an official package is responsible for staging it first — the same
-responsibility the Make build already places on them.
-
-If a packaging job would rather copy the tree into the release directory *after*
-`bazel build //:release`, this hook is unnecessary and can be dropped; keeping it
-means the Bazel release rule owns the whole package, so the release tree is
-complete the moment the build finishes and the Make/Bazel comparator sees the
-same set of files on both sides.
+No CI job stages the tree, so releases built in CI ship without documentation on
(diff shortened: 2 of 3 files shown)


dev/bazel/README.md
```diff
@@ -301,6 +301,44 @@ The most used Bazel commands are `build`, `test` and `run`.
...
+BOM lists `documentation/en/common/license.txt` and the
+`third-party-programs*.txt` files from it. That tree is **not** part of this
+repository. It is produced by a separate documentation build, and both the Make
+and the Bazel packaging step only copy it when it happens to be present:
+
+| | Make | Bazel |
+|---|---|---|
+| Source | `$(DIR)/../documentation`, i.e. next to the repository | `documentation/` inside the workspace |
+| Destination | `daal/latest/documentation` | `daal/latest/documentation` |
+| When absent | recipe is guarded by `if [ -d ... ]` (`makefile:442-444`, `:1051-1058`) | `glob(..., allow_empty = True)` expands to nothing |
```
david-cortes-intel: These line numbers will get outdated over time. Please copy the actual command.

dev/bazel/README.md
```diff
@@ -301,6 +301,44 @@ The most used Bazel commands are `build`, `test` and `run`.
...
+
+`documentation/` is in `.gitignore` so a staged tree is never committed. The
+`release_documentation` filegroup in the root `BUILD` file picks it up, and the
+release rule copies it with its directory structure preserved, so nothing has to
+enumerate the documentation layout.
+
+Nothing in CI stages the tree today, which means the Bazel release built by CI
+ships without documentation, exactly as the Make release does there. Whoever
+produces an official package is responsible for staging it first — the same
+responsibility the Make build already places on them.
```
david-cortes-intel: This paragraph is redundant.

dev/bazel/README.md
```diff
@@ -301,6 +301,44 @@ The most used Bazel commands are `build`, `test` and `run`.
...
+
+If a packaging job would rather copy the tree into the release directory *after*
+`bazel build //:release`, this hook is unnecessary and can be dropped; keeping it
+means the Bazel release rule owns the whole package, so the release tree is
+complete the moment the build finishes and the Make/Bazel comparator sees the
+same set of files on both sides.
```
david-cortes-intel: Could be simplified.

dev/bazel/README.md
```diff
@@ -301,6 +301,44 @@ The most used Bazel commands are `build`, `test` and `run`.
...
+
+Bazel cannot glob above the workspace, so the tree has to be reachable from
+inside it. Stage it before building the release, by symlink or copy:
+
+```sh
+ln -sfn /path/to/documentation documentation
+bazel build //:release
+```
+
+`documentation/` is in `.gitignore` so a staged tree is never committed. The
```
david-cortes-intel: This is not correct:
Image

dev/bazel/README.md
napetrov: Done in e6123d70 — the two lines that carry the meaning are quoted instead:

```make
DOC.srcdir:= $(DIR)/../documentation
release.DOC := $(shell if [ -d $(DOC.srcdir) ]; then find $(DOC.srcdir) -not -wholename '*.svn*' -type f ;fi)
```

No line numbers left in the section.

dev/bazel/README.md
napetrov: Removed in e6123d70. It restated what the sentence above it and the quoted recipe already say.

dev/bazel/README.md
napetrov: Simplified in e6123d70 — that paragraph is gone from the guide entirely. It was the open design question (whether a packaging job should copy the tree into the release directory *after* `bazel build //:release`, making this hook unnecessary), which belongs in this PR rather than in the Bazel documentation. Restating it here so it is not lost:

Keeping the hook means the release rule owns the whole package, so the tree is complete when the build finishes and the Make/Bazel comparator sees the same file set on both sides. Dropping it means packaging owns that step, and `//:release` stops looking for `documentation/` at all. I have no stake in which — it is a call for whoever runs official packaging. Happy to drop the hook and close this PR if that is the preference.

dev/bazel/README.md
napetrov: I think this one is looking at the base branch. That screenshot is `.gitignore` on `main` — 31 lines, 395 bytes, ending at `.DS_Store`. This PR adds the entry, which takes it to 35 lines:

```diff
+# Documentation tree staged by the build (see the `release_documentation`
+# filegroup in the root BUILD file); generated outside the repository.
+/documentation/
```

So the statement is true once this merges, but not verifiable from `main`, which is a fair thing to trip over.

I have kept the sentence, since the README describes the state after merge, but reworded it to name the entry rather than the mechanism: "`/documentation/` is listed in `.gitignore`, so a staged tree is never committed" (e6123d70). If you would rather the guide not mention `.gitignore` at all, say so and I will drop the sentence — the `.gitignore` entry itself should stay either way, or a staged tree shows up as untracked in every `git status`.

Let me know if you meant something else about that paragraph and I have misread which part you flagged.

.gitignore
```diff
@@ -29,3 +29,7 @@ MODULE.bazel.lock
 # Configurations from file explorers
 .directory
 .DS_Store
+
+# Documentation tree staged by the build (see the `release_documentation`
+# filegroup in the root BUILD file); generated outside the repository.
+/documentation/
```
david-cortes-intel:
```suggestion
documentation/
```

napetrov: Applied in 985d68c.

Every other entry in the file is unanchored, and there is no other directory named `documentation` anywhere in the tree, so the anchored and unanchored forms select the same paths today — the anchor bought nothing and broke with the house style. I also cut the two-line comment down to one; the rest of the file uses short labels like `# Bazel directories`, and naming the filegroup was more detail than a `.gitignore` needs.

Two follow-ons in the same commit:

- The README sentence referred to the entry by name, so it now says `documentation/`.
- Your earlier line-number point applied to `BUILD` as well as the README — the `release_documentation` comment still carried `makefile:442-444, :1051-1058`. It now quotes the recipe instead, same as the README section.

The `CI (LinuxBazel)` red on the previous head was an `install-bazel` step failing on a GitHub API call while resolving Bazelisk, not anything in this PR; this push re-queues it.

Alexandr-Solovev (approved):

Alexandr-Solovev (changes requested):
Drop — let's handle documentation in the packaging step instead.

PR #3749, opened 2026-08-14
rng: default to philox4x32x10, fix host/device stream mirroring and the GPU random_split

Alexandr-Solovev:
# RNG: make `philox4x32x10` the default engine, fix host/device stream mirroring, rework the GPU `random_split`

## Summary

* Make **`philox4x32x10`** the library-wide default RNG engine (host and device) and centralize the
  DAAL host-engine construction so every backend agrees on it.
* Fix two host/device **stream desynchronization bugs**: `shuffle` under-advanced the device mirror
  by 2x, and the device `partial_fisher_yates_shuffle` never advanced it at all.
* Make `partial_fisher_yates_shuffle` **composable** on host and device by taking an engine
  reference; the old `(seed, method)` form stays as a documented one-shot wrapper.
* Rework the decision forest GPU `random_split` kernel so it is correct when a node holds more rows
  than the work group has work items, and so the regression statistics are reduced over the whole
  work group instead of one sub-group.
* Remove dead RNG code (`device_engine_collection.hpp`, `rnd_seq.hpp`, `reduce_hist_over_group`) and
  document the decisions that were previously only implied by the code.

## Why `philox4x32x10`

`oneapi::mkl::rng::mt2203` — the previous default — has **no `skip_ahead`**. mt2203 is a
parametrized *family* of generators: independent streams come from the `engine_idx` constructor
argument (oneMKL provides 6024 of them), not from jumping inside one stream. Everything in the
library that separates streams by skipping — per-rank offsets in distributed decision forest,
per-tree offsets inside a rank — was therefore a silent no-op on GPU with the old default, so every
rank produced the same trees. `philox4x32x10` is counter-based: `skip_ahead` is an O(1) jump on both
host and device, which is exactly what the skip-based separation needs.

### RNG defaults and centralization

`cpp/oneapi/dal/backend/primitives/rng/rng_types.hpp`

* Add `constexpr std::int64_t default_seed = 777;` and
  `constexpr engine_type_internal default_engine_type_internal = engine_type_internal::philox4x32x10;`
  with the rationale above recorded on the constant.
* Add `make_daal_engine(seed, method)` for both the internal and the public engine enum. This
  replaces the identical 5-branch switch that used to be duplicated in `host_engine`,
  `device_engine` and `host_engine_collection`; the DAAL engine headers move here with it.
* Move the misplaced `#include "oneapi/dal/rng.hpp"` below `#pragma once` (it was above it).

`host_engine.hpp`, `device_engine.hpp`, `host_engine_collection.hpp`

* Defaults are now `default_seed` / `default_engine_type_internal` instead of open-coded `777` and
  `engine_type_internal::mt2203`; all three route construction through `make_daal_engine`.

There is deliberately **no** public `default_engine_type` constant: `decision_forest/common.cpp`
must not pull the backend RNG headers into the public descriptor (its bazel `:core` target does not
depend on `primitives:rng`), so it keeps the literal `engine_type::philox4x32x10` with a comment
pointing at `default_engine_type_internal`.

### Host/device stream mirroring (bug fixes)

A `device_engine` owns a DAAL host engine and an oneMKL device engine that must stay at the same
position in the stream: a routine that draws on the host has to `skip_ahead_gpu` by exactly as many
values as it consumed, or a later on-device `generate` replays values the host already used.

`device_engine.hpp`, `device_engine_dpc.cpp`

* `shuffle` draws **two** uniforms per iteration but advanced the mirror by `count`
  → now `skip_ahead_gpu(2 * count)` (both the header and the `_dpc` overload).
* `partial_fisher_yates_shuffle` (device) draws one uniform per iteration and never advanced the
  mirror at all → now `skip_ahead_gpu(count)`.

### `partial_fisher_yates_shuffle` — composable

`host_engine.hpp`, `device_engine.hpp`, `device_engine_dpc.cpp`

* New canonical overload takes `host_engine&` / `device_engine&`, so the caller owns the stream:
  consecutive calls return different samples and the draw can be interleaved with other draws.
  Previously the routine built a throw-away engine from `(seed, method)` internally, so the same
  seed always returned the same sample and the routine could not be composed.
* The `(seed, method)` form is kept as an inline one-shot wrapper (documented as such) so existing
  callers such as `kmeans_init`'s distributed random-dense kernel are unaffected.
* Both implementations lose a dead `continue` / `k` counter / unreachable assert, and the entry
  assert becomes `count <= top` — drawing exactly `top` distinct indices out of `[0, top)` is valid.

### mt2203 `skip_ahead_gpu` documented as a no-op by design

`device_engine.hpp` (`gen_mt2203::skip_ahead_gpu`) — replaces the bare `//skip;` body with an
explanation of *why* it cannot be implemented and what callers should do instead (distinct
`engine_idx`, or a counter-based engine).

`decision_forest/backend/gpu/train_kernel_hist_impl_dpc.cpp` — spells out the arithmetic of the
per-rank skip (`tree_count * selected_row_total_count` bootstrap values plus
`selected_ftr_count * tree_count * 2` feature/threshold values) and notes that the whole pattern
depends on a counter-based engine.

### Decision forest — GPU `random_split`

`decision_forest/backend/gpu/train_splitter_impl_dpc.cpp`

* **Correctness:** the kernel used one work item per row (`local_id < row_count`), so any node with
  more rows than the work-group size silently dropped the surplus rows from the min/max bin range,
  the class histogram and the regression statistics. All three passes now walk the rows with a
  `local_size` stride.
* **Correctness:** the regression statistics were reduced with `reduce_over_group(sub_group, ...)`,
  i.e. over one sub-group rather than the whole work group. They are now reduced over the group.
* Regression now uses a single Welford pass per work item plus three group reductions, and combines
  the per-item deviations exactly (`sum2cent + n * (mean_i - mean)^2`), instead of a first pass for
  the sum and a second gather to recompute the deviations against the group mean. This matches the
  existing `add_val_to_hist` regression idiom.
* Classification accumulates the left histogram directly in SLM with a work-group-scoped
  `sycl::atomic_ref`, replacing the `class_count - 1` sequential group reductions.
* SLM shrinks by `local_size * sizeof(Float)`: the extra float scratch buffer and its
  `reduce_hist_over_group` helper are gone (`train_helpers.hpp` drops the helper and the now-unused
  `merge_stat` overload).
* Barriers: the best-split slot is only ever touched by work item 0, so it clears its own slot and
  the barrier after the setup is not needed; the barrier at the end of a feature iteration is only
  needed on the classification path, where the whole group rebuilds the trial histogram.

### Dead code removal

* `rng/device_engine_collection.hpp` — unused repo-wide, and it did not compile if it ever had been
  included (`std::vector<device_engine<EngineType>>` with a non-template `device_engine` and an
  undeclared `EngineType`, plus the engine index passed as the `method` argument).
* `rng/rnd_seq.hpp` — had a single consumer. `selection/kselect_by_rows_quick` now allocates its own
  `ndarray<Float, 1>` and fills it with a fixed-seed `host_engine` + `uniform` in the constructor,
  which also drops three DAAL internal includes (`service_rng.h`, `mcg59.h`, `engine_batch_impl.h`)
  from a `primitives` header. The dead `rnd_seq_size_` member goes with it.
* `train_helpers.hpp` — `reduce_hist_over_group` and the `merge_stat(Float&, Float&, Float&, ...)`
  overload it was the only user of.

### Answering the review comment on louvain

`louvain/backend/cpu/louvain_data.hpp` pins `engine_type_internal::mt2203` explicitly. There is
nothing mt2203-specific about the algorithm — it only calls `uniform` on the host and never needs
`skip_ahead`. It is pinned because louvain draws a random vertex visit order, and that order decides
which local optimum the modularity optimization settles in, so the community labels it reports
depend on the exact value stream. mt2203 is the engine it has been using so far (it was the
library-wide default), and pinning it keeps the labels bit-identical while the default moves. Louvain
does not expose the engine on its descriptor, so changing it here would silently reshuffle the labels
of every existing caller. The comment in the file now says all of this.

### Test that needed pinning

`primitives/optimizers/test/newton_cg_dpc.cpp` pins the same engine for its synthetic logistic
dataset. The test asserts accuracy thresholds on that exact dataset, and for `n = 1000, p = 50`
(float) the generalization score sits two samples above the bar: with the philox stream the fit is
still perfect on the training set (700/700) but scores 286/300 on the held-out part against a
`0.96 * 300 = 288` requirement. Newton-CG never touches the RNG, so pinning the data generator keeps
the assertion at its original strength rather than relaxing the threshold.

## Behavior changes to be aware of

* Any algorithm that did not pin its engine now consumes a different value stream. Results stay
  statistically equivalent but are not bit-identical to `main`; conformance tests that hard-code
  expected random output may need updating.
* Distributed decision forest on GPU now actually produces different trees per rank, which is the
  intended behavior the `skip_ahead` calls always described.

## Testing

Local Bazel run (Intel Data Center GPU Max 1100, `icpx` 2026.0):

* `//cpp/oneapi/dal/backend/primitives/...` — 48 tests pass. The remaining entries are unrelated to
  this PR: three perf targets that do not build on `main` either
  (`reduction:test_reduction_rm_cw_perf_dpc`, `reduction:test_reduction_rm_rw_perf_dpc`,
  `stat:test_perf_cov_dpc`) and five heavy `blas`/`search` perf tests that hit the 900 s Bazel
  timeout when run five-wide on a single GPU.
* `decision_forest`, `kmeans_init`, `louvain`, `connected_components` suites pass.
* Two pre-existing failures on this machine, both reproduced at `upstream/main` in a clean worktree,
  so they are not from this PR: `decision_forest:test_spmd_dpc` aborts with a GPU page fault, and
  every `kmeans` `_dpc` target times out (including `test_badarg_dpc`). The `logistic_regression`
  `_dpc` targets also exceed the Bazel timeout here; no `logistic_regression` or optimizer source is
  touched by this PR (only the `newton_cg` test's data generator).
* `clang-format` 20 clean on every changed file.


cpp/oneapi/dal/algo/louvain/backend/cpu/louvain_data.hpp
```diff
@@ -123,7 +123,12 @@ struct louvain_data {
     // Total link weight in the network
     value_type m;
 
-    host_engine eng;
+    // Louvain is an order-sensitive heuristic: the random visit order of the
+    // vertices decides which local optimum the modularity optimization settles
+    // in. The engine is pinned explicitly so that the partition returned to the
+    // user stays stable when the library-wide default engine
+    // (`default_engine_type_internal`) changes.
+    host_engine eng{ default_seed, engine_type_internal::mt2203 };
```
david-cortes-intel, line 131: Why mt2203 in particular?

Alexandr-Solovev: Nothing about the algorithm needs mt2203 — louvain only calls `uniform` on the host and never needs `skip_ahead`, so any of the supported engines would work.

It is pinned for reproducibility of the *output*, not for statistical reasons. Louvain draws a random vertex visit order, and that order decides which local optimum the modularity optimization settles in, so the community labels it reports depend on the exact value stream of the engine. mt2203 is the engine louvain has been using so far (it was the library-wide default), and pinning it keeps the reported labels bit-identical while this PR moves the default to philox4x32x10. Louvain does not expose the engine on its descriptor, so switching it here would silently reshuffle the labels of every existing caller with no way to opt out — and `louvain_test` asserts exact per-vertex labels and community counts, so the change would have been visible as a test failure rather than as a deliberate decision.

I have expanded the comment in the file to say all of the above instead of just pointing at `default_engine_type_internal`. If you would rather louvain follow the new default, that is a one-line change plus updated expected labels in the test — happy to do it in a follow-up, but it seemed wrong to fold a user-visible output change into an RNG-plumbing PR.

david-cortes-intel: That sounds like a test that's not well designed.

cpp/oneapi/dal/algo/decision_forest/backend/gpu/train_splitter_impl_dpc.cpp
```diff
@@ -200,46 +203,80 @@ sycl::event train_splitter_impl<Float, Bin, Index, Task>::random_split(
...
+                        const Index id = tree_order_ptr[row_ofs + i];
+                        const Index bin = data_ptr[id * column_count + ts_scal.ftr_id];
+                        if (bin <= ts_scal.ftr_bin) {
+                            const Float response = response_ptr[id];
+                            ++local_count;
+                            const Float delta = response - local_mean;
+                            local_mean += delta / Float(local_count);
+                            local_sum2cent += delta * (response - local_mean);
+                        }
+                    }
```
david-cortes-intel, line 280: Am I understanding it correctly that this is changing the algorithm?

Alexandr-Solovev: Resolved in the merge with main (35e5a713d). The weighted accumulation that landed upstream was written against the one-row-per-work-item shape this branch removes, so it walked only the first `local_size` rows of a node. It now strides the rows with `local_size` the same way the bin/histogram loops below it do, and `total_weight_sum` is read from `node_weight_list_ptr[node_id]` rather than being re-derived.

cpp/oneapi/dal/algo/decision_forest/backend/gpu/train_kernel_hist_impl_dpc.cpp
```diff
@@ -1845,6 +1845,15 @@ train_result<Task> train_kernel_hist_impl<Float, Bin, Index, Task>::operator()(
 
     pr::ndarray<Float, 1> node_imp_decrease_list;
     if (ctx.distr_mode_) {
+        // Every rank starts from the same seed, so the streams are separated by skipping the
+        // values the lower-numbered ranks consume: `tree_count * selected_row_total_count`
+        // values for the bootstrap draws plus `selected_ftr_count * tree_count * 2` values for
+        // the feature and threshold draws.
+        //
```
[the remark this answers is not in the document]
Alexandr-Solovev: Good catch, and the arithmetic was wrong for a second reason too: `gen_feature_list()` and `gen_random_thresholds()` run once per node per tree level, not once per tree, so the old `selected_ftr_count * tree_count * 2` term under-counted by a factor of the node count and ranks could overlap. Fixed in db7f45e3f by sizing each rank range from upper bounds - one bootstrap value per row, `selected_ftr_count` feature values plus as many threshold values for each of the at most `2 * row_count` nodes, and `2 * row_count` values per column when MDA is on - with `check_mul_overflow()` on each product. I kept the bound data-derived rather than using a fixed power-of-two block because non-counter-based engines have an O(nSkip) skip, where a huge constant jump would hang instead of just leaving a gap.

cpp/oneapi/dal/algo/louvain/backend/cpu/louvain_data.hpp
Alexandr-Solovev: Dropped in db7f45e3f. Louvain only calls `uniform` on the host and never needs `skip_ahead`, so there was nothing engine specific about the pin - it just froze a particular vertex visit order. It now takes the library default like every other algorithm. The visit order decides which local optimum the modularity optimization settles in, so two expectations in `louvain_test.cpp` moved to the new order (20-vertex random graph: 5 communities instead of 4; karate club: one vertex changes community). All 17 host test cases pass.

PR #3755, opened 2026-08-19
build: apply -Wno-pass-failed to all clang/icx configurations

Alexandr-Solovev:
## Description

The clang build of the DAAL core fails to compile the HDBSCAN kernels:

```
./cpp/daal/src/algorithms/hdbscan/hdbscan_ball_tree_batch_impl.i:225:17: error: loop not vectorized:
  the optimizer was unable to perform the requested transformation; the transformation might be
  disabled or specified as part of an unsupported transformation ordering
  [-Werror,-Wpass-failed=transform-warning]
```

with the same error repeated for `hdbscan_kd_tree_batch_impl.i`, `hdbscan_distance_utils.h`, and — via
inlining — for `threading.h:228` (`threader_func`), `knnQuery`, `knnQueryBallTree`,
`nearestMrdBoruvkaQuery` and `nearestMrdBoruvkaQueryBallTree`. All three of
`hdbscan_dense_batch_fpt_cpu.cpp`, `hdbscan_kd_tree_batch_fpt_cpu.cpp` and
`hdbscan_ball_tree_batch_fpt_cpu.cpp` fail to build.

This is a build-configuration bug, not a defect in the kernels: the HDBSCAN sources are **not** changed
by this PR.

## Root cause

Every reported loop is a floating-point `reduction(min:)` / `reduction(max:)` under
`PRAGMA_OMP_SIMD_ARGS`. Clang's loop vectorizer only recognizes an `fcmp` + `select` min/max reduction
when the no-nans and no-signed-zeros fast-math flags are in effect, which this build does not set, so
the requested transformation always fails and clang reports it.

The report is only *fatal* because the suppression never reaches the compile line.
`-Wno-pass-failed` was defined as part of `-DEBC.clang` in
`dev/make/compiler_definitions/clang.mk`, and the top-level makefile expands `-DEBC` only when
`REQDBG` is set (`makefile:142`):

```make
-DEBC := $(if $(REQDBG),$(if $(filter symbols,$(REQDBG)),$(-DEBC.$(COMPILER)),...)) ...
```

The conda-recipe clang build (`conda-recipe/build.sh:59`) runs `make daal oneapi COMPILER=clang`
without `REQDBG`, so it gets `-Werror` from `COMPILER.lnx.clang` but not `-Wno-pass-failed`. The same
gap existed for icx (`-DEBC.icx`) and in the bazel toolchain, where `-Wno-pass-failed` was only part of
`dbg_compile_flags` (`dev/bazel/toolchains/cc_toolchain_lnx.bzl:412`) while `-Werror` lives in
`lnx_cc_common_flags`.

## Changes

**Build system — make the suppression unconditional:**

| File | Change |
| --- | --- |
| `dev/make/compiler_definitions/clang.mk` | New `warn.opts.clang = -Werror -Wno-empty-body -Wreturn-type -Wno-pass-failed`; `-DEBC.clang` reduced to `-g`. |
| `dev/make/compiler_definitions/clang.32e.mk` | `COMPILER.lnx.clang` / `COMPILER.mac.clang` use `$(warn.opts.clang)`. |
| `dev/make/compiler_definitions/clang.ref.arm.mk`, `clang.ref.riscv64.mk` | Same, replacing the inline copies of the three warning flags. |
| `dev/make/compiler_definitions/icx.mkl.32e.mk` | `-Wno-pass-failed` moved from `-DEBC.icx` to `COMPILER.lnx.icx` and `COMPILER.win.icx`. |
| `dev/bazel/flags.bzl` | `-Wno-pass-failed` added to the Linux common flags for `clang`, `icx` and `icpx`, next to the existing `-Werror`. |

The four clang compile lines each repeated `-Werror -Wno-empty-body -Wreturn-type` verbatim, so they
now share one variable instead of gaining a fourth copy of the new flag.

**Sources — drop the workarounds the flag makes unnecessary:**

`#ifndef __clang__` guards around the affected pragmas are removed in

- `cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i` (`maxWeight` in `normalizeWeights`)
- `cpp/daal/src/algorithms/dtrees/gbt/classification/gbt_classification_train_dense_default_impl.i` (`maxArg` in `getSoftmax`)
- `cpp/daal/src/algorithms/objective_function/cross_entropy_loss/cross_entropy_loss_dense_default_batch_impl.i` (`maxArg` in `softmax`)

Besides being redundant now, those guards had two problems: `__clang__` is defined by icx/icpx as well,
so they silently dropped the pragma for the Intel compilers too, and their comment misspelled
"vectorize". No occurrence of the misspelling is left in the tree.

No functional change: the pragmas are vectorization hints only, the loop bodies are untouched, and
min/max reductions are order-independent, so results are bit-identical on every compiler.

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

</details>

### DIFF
--- a/cpp/daal/src/algorithms/hdbscan/hdbscan_ball_tree_batch_impl.i
+++ b/cpp/daal/src/algorithms/hdbscan/hdbscan_ball_tree_batch_impl.i
@@ -274,7 +274,9 @@ static DAAL_INT buildBallTree(const algorithmFPType * data, DAAL_INT * pointIndi
     algorithmFPType maxR = algorithmFPType(0);
     // `d2` is `TArrayScalable`-backed (see d2Arr above); its start is aligned to
     // DAAL_MALLOC_DEFAULT_ALIGNMENT. Reduction body uses `?:` per OMP conformance.
+#ifndef __clang__ // TODO: Temporary workaround. Clang fails to vectoize this simple loop
     PRAGMA_OMP_SIMD_ARGS(reduction(max : maxR) aligned(d2 : DAAL_MALLOC_DEFAULT_ALIGNMENT))
+#endif
     for (DAAL_INT i = 0; i < count; i++)
     {
         maxR = (d2[i] > maxR) ? d2[i] : maxR;
@@ -415,7 +417,9 @@ static algorithmFPType computeMinCoreDistsBallTree(const BallNode<algorithmFPTyp
     if (node.left < 0)
     {
         algorithmFPType minCD = daal::services::internal::MaxVal<algorithmFPType>::get();
+#ifndef __clang__ // TODO: Temporary workaround. Clang fails to vectoize this simple loop
         PRAGMA_OMP_SIMD_ARGS(reduction(min : minCD))
+#endif
         for (DAAL_INT i = node.pointBegin; i < node.pointEnd; i++)
         {
             const algorithmFPType cd = coreDistances[pointIndices[i]];
--- a/cpp/daal/src/algorithms/hdbscan/hdbscan_ball_tree_batch_impl.i
+++ b/cpp/daal/src/algorithms/hdbscan/hdbscan_ball_tree_batch_impl.i
@@ -274,9 +274,7 @@ static DAAL_INT buildBallTree(const algorithmFPType * data, DAAL_INT * pointIndi
     algorithmFPType maxR = algorithmFPType(0);
     // `d2` is `TArrayScalable`-backed (see d2Arr above); its start is aligned to
     // DAAL_MALLOC_DEFAULT_ALIGNMENT. Reduction body uses `?:` per OMP conformance.
-#ifndef __clang__ // TODO: Temporary workaround. Clang fails to vectoize this simple loop
     PRAGMA_OMP_SIMD_ARGS(reduction(max : maxR) aligned(d2 : DAAL_MALLOC_DEFAULT_ALIGNMENT))
-#endif
     for (DAAL_INT i = 0; i < count; i++)
     {
         maxR = (d2[i] > maxR) ? d2[i] : maxR;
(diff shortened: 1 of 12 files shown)


cpp/daal/src/algorithms/hdbscan/hdbscan_ball_tree_batch_impl.i
```diff
@@ -274,7 +274,9 @@ static DAAL_INT buildBallTree(const algorithmFPType * data, DAAL_INT * pointIndi
     algorithmFPType maxR = algorithmFPType(0);
     // `d2` is `TArrayScalable`-backed (see d2Arr above); its start is aligned to
     // DAAL_MALLOC_DEFAULT_ALIGNMENT. Reduction body uses `?:` per OMP conformance.
+#ifndef __clang__ // TODO: Temporary workaround. Clang fails to vectoize this simple loop
```
david-cortes-intel: This macro is also defined by ICX. If it's going to get there, please narrow it down to clang-only.

david-cortes-intel:
```suggestion
#ifndef __clang__ // TODO: Temporary workaround. Clang fails to vectorize this simple loop
```

Please fix typo in all the cases where it occurs.

david-cortes-intel: And along the way, please update other occurrences too:
https://github.com/uxlfoundation/oneDAL/blob/e755d9ac14db3e9c1567965b05482204097f1a9d/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i#L69

dev/make/compiler_definitions/icx.mkl.32e.mk
```diff
@@ -64,18 +64,22 @@ ifeq ($(STDALLOC), yes)
 endif
 
 -Zl.icx = $(if $(OS_is_win),-Zl,) $(-Q)no-intel-lib
--DEBC.icx = $(if $(OS_is_win),-debug:all -Z7,-g) -fno-system-debug -Wno-pass-failed
```
david-cortes-intel, line 67: @Alexandr-Solovev These pragmas are meant to work with ICX in non-debug mode. Do you see any issue with ICX builds?

Alexandr-Solovev: I wanted to align flags usage, reverted now

david-cortes-intel (approved):
LGTM, pending CI results.

napetrov (approved):

PR #3757, opened 2026-08-20
ci: let the nightly Bazel tests run on fork pull requests and report real differences

napetrov:
## Description

Split out of #3752, which bundled three unrelated CI fixes and stalled. This carries only what is needed to stop the two `Nightly-test / oneDAL {Linux,Windows} Bazel test` checks from failing on every pull request from a fork. The `timeoutInMinutes` change for `FormatterChecks` that #3752 also carried is left for a separate pull request.

Both checks are currently red on every open Bazel pull request — #3716, #3719, #3737, #3740, #3746, #3747 — for reasons that have nothing to do with any of them. A reviewer opening any of those sees red and has to read a comment to learn it is noise.

### 1. Fork pull requests never get past checkout

```
##[error]Refusing to check out fork pull request code from a 'workflow_run' workflow.
```

`actions/checkout` is right to refuse. Under `workflow_run` a job carries the base repository's token scopes, secrets and default-branch cache, and building a fork's commit in that context is the "pwn request" shape. The only thing that put these jobs in it was `checks: write`, held solely so a job could report its own check run.

So the reporting moves out. `check_start` creates both check runs, `check_finish` completes them from `needs.*.result`, and neither job checks anything out. The two test jobs keep `contents: read` plus the `actions: read` that `download-artifact` needs to reach the Nightly-build artifacts. They reference no secrets and use no cache action, which leaves them no more privileged than the `pull_request` run that already builds the same commit, and lets them set `allow-unsafe-pr-checkout: true`.

Side effect worth noting: a cancelled or skipped test job now reports as cancelled or skipped instead of as a failure.

### 2. Linux: 24 "differences" that are 12 dereferenced symlinks

Every `libonedal*.so` and `libonedal*.so.4` is reported as a Make-only file and a Bazel-only symlink. Both builds stage them identically (`makefile:975` — one `cp` plus two `ln -sf`), but the Make tree reaches the test job through `actions/upload-artifact`, which stores the contents a symlink points at. `--make-symlinks-dereferenced` tells the comparator the Make side travelled through such a transport: those pairs leave level 1 and are checked the only way left, by requiring the Make file to be byte-identical to the target the Bazel symlink resolves to. A path missing outright, or holding different bytes, still fails. Azure's `LinuxReleaseCompare` never saw this because it downloads both trees as pipeline artifacts, so both sides are dereferenced alike.

### 3. Windows: 26 exports that only `cl` emits

All 26 are `??0` constructors of exception classes whose body is `using Base::Base;`. `cl` materialises the inherited constructors of a `__declspec(dllexport)` class whether the DLL uses them or not; icx and clang-cl emit them only where they are odr-used, and this nightly is the only comparison that builds the Make release with `cl`. They cannot be part of a consumer's link surface — `ONEDAL_EXPORT` is defined only under `__ONEDAL_ENABLE_EXPORT__`, which only oneDAL's own builds set, so a consumer instantiates the constructor it uses in its own object file. Listed one by one, like the neighbouring list, so a constructor of any other class disappearing still fails. `spmd::error_holder`, which declares its constructor out of line, is deliberately absent.

## Verification

The workflow change cannot verify itself here: `workflow_run` always uses the copy of the workflow on `main`, so an edit to `nightly-test.yml` has no effect on the pull request carrying it. `workflow_dispatch` only offers branches that exist in this repository, and this branch lives in a fork. **It needs someone with push access to push the branch here and dispatch Nightly-test on it, or verification on the first scheduled nightly after merge.** Please do not hold review on that.

What was verified:

- `allow-unsafe-pr-checkout` exists as an input on `actions/checkout` at the pinned SHA `9c091bb`.
- The comparator changes are covered by the existing `dev/release_tests` tests; `--make-symlinks-dereferenced` is off by default, so Azure's `LinuxReleaseCompare` and `WindowsReleaseCompare` are unaffected.
- Both jobs that check out the tested commit hold read-only permissions and reference no secrets; both jobs that hold `checks: write` check out nothing.

## Checklist

- [x] PR title and description follow the contribution guide
- [x] Changes are limited to a single goal
- [x] No secrets or credentials added


dev/release_tests/__pycache__/compare_release_trees.cpython-312.pyc
icfaust, line 1: Please remove this file.

dev/release_tests/compare_release_trees.py
```diff
@@ -238,9 +238,66 @@ def is_ignored_linux_export(symbol, library_path=None):
...
+# Not part of a consumer's link surface. `ONEDAL_EXPORT` is defined only under
+# `__ONEDAL_ENABLE_EXPORT__` (cpp/oneapi/dal/common.hpp:35), which only oneDAL's
+# own builds set (makefile:709,723 and dev/bazel/dal.bzl:532). For a consumer the
+# macro expands to nothing, so `throw oneapi::dal::invalid_argument{ msg }`
+# instantiates the constructor in the consumer's own object file and never emits
+# a reference the DLL has to satisfy.
+#
+# Listed one by one, like WINDOWS_IGNORED_EXPORTS above, so that a constructor of
+# a *different* class disappearing from a released library still fails.
+WINDOWS_IGNORED_INHERITED_CTORS = frozenset([
```
david-cortes-intel: Can this list be generated programmatically?

dev/release_tests/__pycache__/compare_release_trees.cpython-312.pyc
napetrov: Removed in `e184a81`, along with adding `__pycache__/` and `*.pyc` to `.gitignore` — the absence of that entry is why it got committed in the first place. `dev/release_tests` is back to the two files it had, and this branch now adds no new paths at all.

dev/release_tests/compare_release_trees.py
napetrov: Yes, and it should be — done in `a76af43`. The 26 names were an expansion of a one-line rule, so the rule is now what is in the file:

```python
WINDOWS_INHERITED_CTOR_PATTERN = re.compile(
    r"^\?\?0(?:"
    + "|".join(re.escape(name) for name in WINDOWS_INHERITED_CTOR_CLASSES)
    + r")@v1@(?:spmd@preview@)?dal@oneapi@@"
)
```

over eleven class names instead of 26 mangled signatures. The class list is the durable half: which *overloads* exist is whatever the standard library base declares, so `std::system_error` gaining or losing a constructor would have left the old list stale and turned the nightly red for a reason with nothing to do with oneDAL.

Going fully programmatic — synthesising the mangled names from the headers — would mean reimplementing MSVC name mangling in Python, and it would have to be exactly right about `std::basic_string` template arguments and back-reference numbering to be worth anything. Not a good trade. Instead the comment carries the awk that rederives the class list from the headers, so the set can be checked against them rather than trusted:

```
awk '/class ONEDAL_EXPORT/ {
         match($0, /class ONEDAL_EXPORT [a-z_]+/)
         c = substr($0, RSTART + 20, RLENGTH - 20)
     }
     /using std::[a-z_]+::[a-z_]+;/ { if (c) print c }' \
    cpp/oneapi/dal/exceptions.hpp cpp/oneapi/dal/spmd/exceptions.hpp | sort
```

That prints exactly the eleven names in the tuple — I ran it rather than eyeballing it.

Verified three ways before pushing:

- all 26 symbols the old list held are still ignored (26/26);
- the eight probes that must **not** match are all still reported: `spmd::error_holder`'s own constructor, a destructor `??1invalid_argument@v1@dal@oneapi@@`, a vftable `??_7invalid_argument@...`, `??0invalid_argument@std@@` in the standard library's own namespace, the same class name under a different namespace, and constructors of `table` and `host_bad_alloc`;
- the eleven-name tuple equals the set the awk derives from the headers, with no entry on either side only.

So the guarantee is unchanged where it counts: a constructor of any *other* class disappearing from a released library still fails, and anything about these eleven that is not a constructor still fails. What it now tolerates that it did not before is a *new* inherited-constructor overload of one of these eleven appearing or disappearing — which is the intended behaviour, since none of them is on a consumer's link surface. `ONEDAL_EXPORT` is defined only under `__ONEDAL_ENABLE_EXPORT__`, which only oneDAL's own builds set, so `throw oneapi::dal::invalid_argument{ msg }` in consumer code instantiates the constructor in the consumer's own object file and never emits a reference the DLL has to satisfy.

icfaust (approved):
I'd like to see this in now.

PR #3772, opened 2026-09-01
fix(dtrees): use the real right-child impurity for best-first leaf priority

### LINKED ISSUE #3771: RandomForest best-first growth (max_leaf_nodes) picks a worse leaf-expansion order than sklearn
### Describe the bug

When `max_leaf_nodes` is set (best-first tree growth), a single `onedal`/`sklearnex` `RandomForestClassifier` tree reaches a **worse training fit** than the equivalent `sklearn.ensemble.RandomForestClassifier` tree for the *same* leaf budget, even with `bootstrap=False`, `max_features=None`, and `max_bins=n_samples` (so bootstrap sampling, feature subsampling, and histogram binning are all ruled out as causes).

With no leaf limit (grow to purity) or with a depth-first constraint (`max_depth`, `min_samples_split`), sklearnex and sklearn trees match essentially exactly. The discrepancy appears specifically when best-first growth is triggered by `max_leaf_nodes`.

This looks like a distinct issue from #3648 (which was about the *regression* random splitter's impurity weighting), this one affects `RandomForest` too (best splitter), and is in the best-first growth / leaf-selection logic shared by classification and regression.

### To Reproduce

```python
import numpy as np
from sklearn.datasets import make_classification
from sklearn.metrics import brier_score_loss, log_loss
import sklearn.ensemble
import sklearnex.ensemble

n_samples, n_features, n_informative = 1000, 20, 10
X, y = make_classification(
    n_samples=n_samples,
    n_features=n_features,
    n_informative=n_informative,
    n_redundant=0,
    n_classes=2,
    random_state=123,
)

def fit_eval(est):
    est.fit(X, y)
    p = est.predict_proba(X)[:, 1]
    return brier_score_loss(y, p), log_loss(y, p)

# every source of randomness removed: single tree, no bootstrap, no feature
# subsampling, and max_bins == n_samples so the CPU histogram is at full
# per-sample resolution.
common = dict(n_estimators=1, bootstrap=False, max_features=None, random_state=0)

for label, constraint in [
    ("unconstrained", {}),
    ("max_depth=4 (depth-first)", {"max_depth": 4}),
    ("min_samples_split=8 (depth-first)", {"min_samples_split": 8}),
    ("max_leaf_nodes=42 (best-first)", {"max_leaf_nodes": 42}),
]:
    b_skl, ll_skl = fit_eval(sklearn.ensemble.RandomForestClassifier(**common, **constraint))
    b_ex, ll_ex = fit_eval(
        sklearnex.ensemble.RandomForestClassifier(max_bins=n_samples, **common, **constraint)
    )
    print(f"{label}")
    print(f"  sklearn  : brier={b_skl:.4f} logloss={ll_skl:.4f}")
    print(f"  sklearnex: brier={b_ex:.4f} logloss={ll_ex:.4f}")
```

### Expected behavior

For the depth-first-constrained and unconstrained cases, sklearnex already matches sklearn to floating-point precision — the same is expected for `max_leaf_nodes`, since it's still supposed to be growing the single best-first tree that maximizes impurity decrease per leaf used.

### Actual behavior

```text
unconstrained
  sklearn  : brier=0.0000 logloss=0.0000
  sklearnex: brier=0.0000 logloss=0.0000

max_depth=4 (depth-first)
  sklearn  : brier=0.1171 logloss=0.3738
  sklearnex: brier=0.1171 logloss=0.3738

min_samples_split=8 (depth-first)
  sklearn  : brier=0.0087 logloss=0.0248
  sklearnex: brier=0.0083 logloss=0.0238

max_leaf_nodes=42 (best-first)
  sklearn  : brier=0.0562 logloss=0.2136
  sklearnex: brier=0.0691 logloss=0.2325
```

The `max_leaf_nodes` case is consistently and substantially worse for sklearnex (≈+23% Brier, ≈+9% logloss here), while every depth-first case matches. `avg_n_leaves` is identical (42) in both cases — the *shape* of the tree differs (sklearnex trees come out shallower on average for the same leaf count across repeated runs), which points at the best-first leaf-expansion order rather than at individual split quality.

### Root cause & fix

Traced to `buildBestFirst`/`buildNode` in [`cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i`](https://github.com/uxlfoundation/oneDAL/blob/main/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i), shared by RandomForest/ExtraTrees, classification/regression. Two independent bugs, both now fixed.

**[FIXED in #3772] Bug 1 — wrong leaf-selection priority formula.** The priority used to rank pending leaves was computed as `imp * item.totalWeights - impLeft * item.leftWeights - (item.totalWeights - item.leftWeights) * (imp - impLeft)`. This had two compounding issues: `item.leftWeights` was read before it was ever set for the split being evaluated (every freshly built `WorkItem` initializes it to `0.0`), which collapses the formula to `totalWeights * impLeft` and ignores the right child (and its weight) entirely; and, independently, the right-child impurity was approximated as `(imp - impLeft)` instead of the actual computed value, which is only exact when the parent's impurity equals the weighted average of the children's — not true in general for Gini or variance impurity.

**Bug 2 — leaf budget spent at the wrong time.** Even with the formula fixed, a real (but smaller) gap to sklearn remained. The cause: `buildBestFirst`'s leaf-count budget (`remainingSplitNodes`) was being spent at node-*birth* time — the instant a child is created, in a fixed left-then-right order, immediately after its parent is popped from the priority queue — instead of at node-*selection* (pop) time. This let a low-priority node consume the last available split slot simply by being evaluated first, even when a genuinely higher-priority node elsewhere in the frontier hadn't had its turn yet. sklearn's `BestFirstTreeBuilder` avoids this by re-checking its own budget fresh every time it pops a node, decrementing only on an actual commit-to-split. A structural tree diff confirmed this was a real, one-sided bug rather than floating-point noise: the nodes sklearn split that sklearnex left as leaves had ground-truth `impurity_improvement` 0.0029–0.0035, while the nodes sklearnex split instead (that sklearn left as leaves) had only 0.0017–0.0025. A Python simulation of both variants (birth-order-eager vs. pop-time-lazy budget checking) confirmed the lazy version reproduces sklearn's real tree for the repro above exactly (identical node count, leaf count, max depth, and full leaf-depth histogram), while the eager version reproduces sklearnex's actual buggy shape. PR with a proposed fix: #3773

**Result:** with both fixes applied, the `max_leaf_nodes=42` case from the repro above now matches sklearn within ordinary cross-implementation floating-point noise (`sklearn brier=0.056150413952262807` vs `sklearnex brier=0.055019580618929476` — sklearnex is now marginally *better*, i.e. no longer a one-sided systematic miss), down from the original +23% gap reported above. Both PRs include a regression test (a small toy dataset each) that was numerically verified to fail on the corresponding pre-fix code and pass with the fix.

### Environment

```
scikit-learn-intelex (importlib.metadata): 2026.1.0
daal4py._get__daal_link_version__():       20260100b'P'_20260601
scikit-learn:                              1.8.0
OS:                                        Linux-6.17.0-1032-oem-x86_64-with-glibc2.39 (Ubuntu)
```

Installed via pixi / conda, CPU only.

cakedev0:
## Summary

Partially fixes #3771 — see #3773 for the rest (a second, independent bug in the same area, found while validating this fix).

`buildNode` (used by `buildBestFirst` to rank pending leaves under `max_leaf_nodes`) computed the leaf-expansion priority as:

```cpp
intermSummFPType improve = imp * item.totalWeights - impLeft * item.leftWeights
                            - (item.totalWeights - item.leftWeights) * (imp - impLeft);
```

Two bugs stacked here:

1. `item.leftWeights` was read *before* it was ever set for the split being evaluated (it's only assigned `split.leftWeights` further down). Every freshly built `WorkItem` initializes `leftWeights` to `0.0`, so `improve` always collapsed to `totalWeights * impLeft`, ignoring the right child entirely.
2. The right-child impurity was approximated as `(imp - impLeft)` instead of the actual computed right-child impurity — only exact if the parent's impurity equals the weighted average of the children's, which does not hold in general for Gini/variance impurity.

This made `buildBestFirst` pick a different (and lower-quality) leaf to expand next than sklearn's `BestFirstTreeBuilder` would, for the same `max_leaf_nodes` budget — same tree size, worse fit.

## Fix

Use `split.leftWeights` (not `item.leftWeights`), and compute the real right-child impurity via the existing `convertLeftImpToRight` helper before the priority check instead of approximating it. `nLeft`/`leftWeights`/left-impurity are captured before that call since it mutates `split` in place to hold the right-child data.

`buildDepthFirst`'s termination check (`split.totalWeights * split.impurityDecrease < minImpurityDecrease`) doesn't go through this code path, which is why depth-first growth (`max_depth`, `min_samples_split`, unconstrained) already matched sklearn exactly and only `max_leaf_nodes` was affected.

## Test plan

- [x] Repro from #3771: `max_leaf_nodes=42` Brier error vs. sklearn drops from +23.0% to +8.7% with this fix alone (remaining gap is #3773, fixed separately).
- [x] Added `cpp/oneapi/dal/algo/decision_forest/test/best_first_leaf_selection.cpp` — a 10-point toy case where the buggy vs. fixed formula disagree on accepting the root's only candidate split. Verified it fails on pre-fix code.
- [ ] Existing decision forest unit/system tests — not run in this environment (no CI access); deferred to CI.

### DIFF
--- a/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
@@ -952,9 +952,6 @@ typename DataHelper::NodeType::Base * TrainBatchTaskBase<algorithmFPType, BinInd
         const intermSummFPType rightWeights            = item.totalWeights - leftWeights;
         typename DataHelper::ImpurityData impurityLeft = split.left;
 
-        // Use the actual right-child impurity, not (imp - impLeft): that substitution
-        // is only exact when imp equals the weighted average of the children's
-        // impurities, which Gini/variance impurity does not satisfy in general.
         _helper.convertLeftImpToRight(item.n, impurity, split);
         const intermSummFPType impRight = split.left.var;
 
--- a/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
@@ -945,11 +945,11 @@ typename DataHelper::NodeType::Base * TrainBatchTaskBase<algorithmFPType, BinInd
     DAAL_ASSERT(split_result.status.ok());
     if (split_result.bSplitSucceeded)
     {
-        const intermSummFPType imp          = impurity.var;
-        const intermSummFPType impLeft      = split.left.var;
-        const size_t nLeft                  = split.nLeft;
-        const intermSummFPType leftWeights  = split.leftWeights;
-        const intermSummFPType rightWeights = item.totalWeights - leftWeights;
+        const intermSummFPType imp                     = impurity.var;
+        const intermSummFPType impLeft                 = split.left.var;
+        const size_t nLeft                             = split.nLeft;
+        const intermSummFPType leftWeights             = split.leftWeights;
+        const intermSummFPType rightWeights            = item.totalWeights - leftWeights;
         typename DataHelper::ImpurityData impurityLeft = split.left;
 
         // Use the actual right-child impurity, not (imp - impLeft): that substitution
--- a/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
@@ -945,11 +945,22 @@ typename DataHelper::NodeType::Base * TrainBatchTaskBase<algorithmFPType, BinInd
     DAAL_ASSERT(split_result.status.ok());
     if (split_result.bSplitSucceeded)
     {
-        const intermSummFPType imp     = impurity.var;
-        const intermSummFPType impLeft = split.left.var;
+        const intermSummFPType imp          = impurity.var;
+        const intermSummFPType impLeft      = split.left.var;
+        const size_t nLeft                  = split.nLeft;
+        const intermSummFPType leftWeights  = split.leftWeights;
+        const intermSummFPType rightWeights = item.totalWeights - leftWeights;
+        typename DataHelper::ImpurityData impurityLeft = split.left;
+
+        // Use the actual right-child impurity, not (imp - impLeft): that substitution
+        // is only exact when imp equals the weighted average of the children's
+        // impurities, which Gini/variance impurity does not satisfy in general, and
+        // was throwing off leaf-selection priority in best-first growth.
+        _helper.convertLeftImpToRight(item.n, impurity, split);
+        const intermSummFPType impRight = split.left.var;
 
         // check impurity decrease
-        intermSummFPType improve = imp * item.totalWeights - impLeft * item.leftWeights - (item.totalWeights - item.leftWeights) * (imp - impLeft);
+        intermSummFPType improve = imp * item.totalWeights - impLeft * leftWeights - impRight * rightWeights;
         if (improve < _minImpurityDecrease)
         {
             return makeLeaf(_aSample.get() + item.start, item.n, impurity, nClasses);
@@ -961,11 +972,10 @@ typename DataHelper::NodeType::Base * TrainBatchTaskBase<algorithmFPType, BinInd
                 addImpurityDecrease(iFeature, item.n, impurity, split);
             }
 
-            item.nLeft        = split.nLeft;
-            item.leftWeights  = split.leftWeights;
-            item.improvement  = improve;
-            item.impurityLeft = split.left;
-            _helper.convertLeftImpToRight(item.n, impurity, split);
+            item.nLeft         = nLeft;
+            item.leftWeights   = leftWeights;
+            item.improvement   = improve;
+            item.impurityLeft  = impurityLeft;
             item.impurityRight = split.left;
 
             if (!(item.node = makeSplit(iFeature, split.featureValue, split.featureUnordered, nullptr, nullptr, impurity.var)))
--- a/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
+++ b/cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
@@ -954,8 +954,7 @@ typename DataHelper::NodeType::Base * TrainBatchTaskBase<algorithmFPType, BinInd
 
         // Use the actual right-child impurity, not (imp - impLeft): that substitution
         // is only exact when imp equals the weighted average of the children's
-        // impurities, which Gini/variance impurity does not satisfy in general, and
-        // was throwing off leaf-selection priority in best-first growth.
+        // impurities, which Gini/variance impurity does not satisfy in general.
         _helper.convertLeftImpToRight(item.n, impurity, split);
         const intermSummFPType impRight = split.left.var;
 
--- a/cpp/oneapi/dal/algo/decision_forest/test/best_first_leaf_selection.cpp
+++ b/cpp/oneapi/dal/algo/decision_forest/test/best_first_leaf_selection.cpp
@@ -30,10 +30,10 @@ public:
     // on inline x/y arrays and returns the predicted response for every training
     // point, in order.
     std::vector<double> train_and_predict_all(const float* x,
-                                               const float* y,
-                                               std::int64_t n,
-                                               std::int64_t max_leaf_nodes,
-                                               double min_impurity_decrease = 0.0) {
+                                              const float* y,
+                                              std::int64_t n,
+                                              std::int64_t max_leaf_nodes,
+                                              double min_impurity_decrease = 0.0) {
         const te::dataframe x_df{ array<float>::wrap(x, n), n, 1 };
         const te::dataframe y_df{ array<float>::wrap(y, n), n, 1 };
 
@@ -54,7 +54,8 @@ public:
         const auto train_result = this->train(desc, x_table, y_table);
         const auto infer_result = this->infer(desc, train_result.get_model(), x_table);
 
-        const auto responses = dal::row_accessor<const float_t>(infer_result.get_responses()).pull();
+        const auto responses =
+            dal::row_accessor<const float_t>(infer_result.get_responses()).pull();
         std::vector<double> result(n);
         for (std::int64_t i = 0; i < n; i++) {
             result[i] = static_cast<double>(responses[i]);
@@ -67,8 +68,11 @@ using df_best_first_types = _TE_COMBINE_TYPES_3((float, double),
                                                 (df::method::dense, df::method::hist),
                                                 (df::task::regression));
 
-#define DF_BEST_FIRST_TEST(name) \
-    TEMPLATE_LIST_TEST_M(df_best_first_test, name, "[df][integration][best-first]", df_best_first_types)
+#define DF_BEST_FIRST_TEST(name)                          \
+    TEMPLATE_LIST_TEST_M(df_best_first_test,              \
+                         name,                            \
+                         "[df][integration][best-first]", \
+                         df_best_first_types)
 
 // Regression test for https://github.com/uxlfoundation/oneDAL/issues/3771 /
 // https://github.com/uxlfoundation/oneDAL/pull/3772.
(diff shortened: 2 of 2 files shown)


cakedev0:
## Validation results

Built oneDAL from this branch and scikit-learn-intelex against it, then re-ran the repro script from #3771 (single tree, `bootstrap=False`, `max_features=None`, `max_bins=n_samples`).

**Depth-first / unconstrained cases: unaffected, still exact**, as expected (`buildDepthFirst` never goes through the fixed code path):

```
unconstrained:                      sklearn brier=0.0000 logloss=0.0000 | sklearnex brier=0.0000 logloss=0.0000
max_depth=4 (depth-first):          sklearn brier=0.1171 logloss=0.3738 | sklearnex brier=0.1171 logloss=0.3738
min_samples_split=8 (depth-first):  sklearn brier=0.0208 logloss=0.0602 | sklearnex brier=0.0208 logloss=0.0602
```

**`max_leaf_nodes=42` (best-first), A/B against the parent commit:**

```
                    sklearn                pre-fix (buggy)        this PR (fixed)
brier:              0.056150               0.0691                 0.061058
logloss:            0.213574               0.2325                 0.221938
```

This PR's fix cuts the gap roughly in half (Brier relative error vs. sklearn: **+23.0% → +8.7%**), confirming the `item.leftWeights`-read-before-set bug and the `(imp - impLeft)` approximation were real and were driving the discrepancy described in #3771.

**However, this is not yet bit-exact with sklearn.** I dug further:

- Ruled out floating-point drift from the subtraction-based `convertLeftImpToRight` propagating through descendants: tried also special-casing singleton right children to compute their impurity directly from raw data (mirroring `buildDepthFirst`'s existing `singleSwap` path) — this produced **zero change** in the result, so that's not the cause here (not included in this PR, since it made no measurable difference).
- Both trees end up with identical `node_count=83` / `n_leaves=42`, but different shapes: sklearn's tree reaches `max_depth=11` with leaf depths `[4,4,5,5,5,5,5,6,6,6,6,6,6,7×11,8,8,9,9,9,9,10×7,11,12,12]`, while sklearnex's reaches `max_depth=10` with leaf depths `[5×9,6×6,7×10,8,8,9×6,10,10,11×4]` — consistently shallower/more balanced, matching the shape difference #3771 already called out.

My current read: this residual is a separate, harder-to-pin-down divergence in *tie-breaking/expansion order* between oneDAL's `BinaryHeap` and sklearn's `PriorityHeap` when candidate leaf "improvement" values are equal or extremely close — not a further bug of the same kind this PR fixes. I don't have a concrete fix for it yet, so I'm not blocking this PR on it. Happy to open a follow-up issue for it if maintainers agree it's worth chasing further.

cakedev0:
Added a regression test for this fix: `cpp/oneapi/dal/algo/decision_forest/test/best_first_leaf_selection.cpp` (commit c9bb85fa4) — a 10-point toy dataset where the old buggy formula and the fixed one disagree on whether to accept the root's only candidate split at all (a pure accept/reject threshold check, isolated from the leaf-selection-order bug fixed separately in #3773). Verified against an actual rebuild of the pre-fix code that the test would have failed there.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -945,11 +945,22 @@ typename DataHelper::NodeType::Base * TrainBatchTaskBase<algorithmFPType, BinInd
...
+        const intermSummFPType impLeft      = split.left.var;
+        const size_t nLeft                  = split.nLeft;
+        const intermSummFPType leftWeights  = split.leftWeights;
+        const intermSummFPType rightWeights = item.totalWeights - leftWeights;
+        typename DataHelper::ImpurityData impurityLeft = split.left;
+
+        // Use the actual right-child impurity, not (imp - impLeft): that substitution
+        // is only exact when imp equals the weighted average of the children's
+        // impurities, which Gini/variance impurity does not satisfy in general, and
+        // was throwing off leaf-selection priority in best-first growth.
```
david-cortes-intel: I think this comment can be left out since it won't be relevant after this PR is merged.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -945,11 +945,22 @@ typename DataHelper::NodeType::Base * TrainBatchTaskBase<algorithmFPType, BinInd
     DAAL_ASSERT(split_result.status.ok());
     if (split_result.bSplitSucceeded)
     {
-        const intermSummFPType imp     = impurity.var;
-        const intermSummFPType impLeft = split.left.var;
+        const intermSummFPType imp          = impurity.var;
+        const intermSummFPType impLeft      = split.left.var;
+        const size_t nLeft                  = split.nLeft;
```
david-cortes-intel: Would some fix be required here too?
https://github.com/uxlfoundation/oneDAL/blob/55c35a1ff88f717ea64aa14618226875ba5e57e4/cpp/daal/src/algorithms/dtrees/forest/regression/df_regression_train_dense_default_impl.i#L671

cpp/oneapi/dal/algo/decision_forest/test/best_first_leaf_selection.cpp
```diff
@@ -0,0 +1,115 @@
...
+// wildly overestimating it since it ignores the right side entirely. Setting
+// min_impurity_decrease_in_split_node to 260 (so the internal threshold,
+// scaled by totalWeights=10, is ~2600) sits strictly between these two
+// values: the correct formula rejects the split (single-leaf tree, every
+// prediction equal to the root's mean, 3.6), while the buggy formula would
+// have wrongly accepted it (verified against a build of the pre-fix code).
+DF_BEST_FIRST_TEST(
+    "best-first split priority uses the real impurity decrease, not a same-side approximation") {
+    SKIP_IF(this->not_available_on_device());
+    SKIP_IF(this->not_float64_friendly());
```
david-cortes-intel: I think these are for GPU tests only, while CPU ones instead parameterize by the float type. Leaving it to @Vika-F to comment.

cpp/oneapi/dal/algo/decision_forest/test/best_first_leaf_selection.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    static const float x[] = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
+    static const float y[] = { 1, 2, 3, 4, 5, 6, 7, 8, -50, 50 };
+
+    const auto predictions =
+        this->train_and_predict_all(x, y, 10, /*max_leaf_nodes*/ 2, /*min_impurity_decrease*/ 260.0);
+
+    for (double p : predictions) {
+        REQUIRE(p == Approx(3.6).epsilon(1e-6));
```
david-cortes-intel: This Approx is triggering a compilation error:
```
cpp/oneapi/dal/algo/decision_forest/test/best_first_leaf_selection.cpp: In member function 'void oneapi::dal::decision_forest::test::CATCH2_INTERNAL_TEMPLATE_TEST_30<TestType>::test()':
cpp/oneapi/dal/algo/decision_forest/test/best_first_leaf_selection.cpp:111:22: error: there are no arguments to 'Approx' that depend on a template parameter, so a declaration of 'Approx' must be available [-fpermissive]
  111 |         REQUIRE(p == Approx(3.6).epsilon(1e-6));
```
I _think_ it might need `Catch::Approx`.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
cakedev0: Hum I'm not sure what you mean, this looks correct.

david-cortes-intel: If, on entry, `split.nLeft` had the actual number of rows on the left, wouldn't `split.nLeft` have the number on the right on exit? Or does it have something else on entry?

david-cortes-intel: Nevermind, was quite confused by the names. I see now that it flips a split.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
cakedev0: Removed the "historical" part of the comment.

cpp/daal/src/algorithms/dtrees/forest/df_train_dense_default_impl.i
```diff
@@ -945,11 +945,21 @@ typename DataHelper::NodeType::Base * TrainBatchTaskBase<algorithmFPType, BinInd
...
+        const intermSummFPType imp                     = impurity.var;
+        const intermSummFPType impLeft                 = split.left.var;
+        const size_t nLeft                             = split.nLeft;
+        const intermSummFPType leftWeights             = split.leftWeights;
+        const intermSummFPType rightWeights            = item.totalWeights - leftWeights;
+        typename DataHelper::ImpurityData impurityLeft = split.left;
+
+        // Use the actual right-child impurity, not (imp - impLeft): that substitution
+        // is only exact when imp equals the weighted average of the children's
+        // impurities, which Gini/variance impurity does not satisfy in general.
```
david-cortes-intel: I still think this comment can be left out entirely.

david-cortes-intel (commented):
CI failure is on a GPU machine with no float64 support. Leaving it to @Vika-F to comment about how to handle it.

david-cortes-intel (approved):
CI issue is not related to these changes. Still leaving it for other people to review before merging.

avolkov-intel (approved):
Looks good to me as well

PR #3764, opened 2026-08-28
Ci/release parity fixes

Alexandr-Solovev:
## Description

---

<details>
<summary>Checklist:</summary>

**Completeness and readability**

- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes or created a separate PR with updates and provided its number in the description, if necessary.
- [ ] Git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/uxlfoundation/scikit-learn-intelex/blob/main/CONTRIBUTING.md#pull-requests) for details)_.
- [ ] I have resolved any merge conflicts that might occur with the base branch.

**Testing**

- [ ] I have run it locally and tested the changes extensively.
- [ ] All CI jobs are green or I have provided justification why they aren't.
- [ ] I have extended testing suite if new functionality was introduced in this PR.

**Performance**

- [ ] I have measured performance for affected algorithms using [scikit-learn_bench](https://github.com/IntelPython/scikit-learn_bench) and provided at least a summary table with measured data, if performance change is expected.
- [ ] I have provided justification why performance and/or quality metrics have changed or why changes are not expected.
- [ ] I have extended the benchmarking suite and provided a corresponding scikit-learn_bench PR if new measurable functionality was introduced in this PR.

</details>

PR #3782, opened 2026-09-10
ci: fix release comparison failing on every dereferenced Make symlink

napetrov:
## Problem

The Linux Bazel nightly has been red on plain `main`, in "Compare make and Bazel releases":

```
Make:  266 dirs, 1092 files, 0 links
Bazel: 266 dirs, 1080 files, 12 links

=== level 1: release tree entries ===
Dereferenced symlink content mismatches: 12
  ! lib/intel64/libonedal.so: Make file differs from Bazel lib/intel64/libonedal.so -> lib/intel64/libonedal.so.4.0
  ... (all 12 are .so / .so.4 version aliases)
Make files reconciled against Bazel symlinks: 12 (--make-symlinks-dereferenced)
=== level 2: symlink targets ===       Compared symlink targets: 0
=== level 3: text file contents ===    Compared text files: 1065
=== level 4: shared library exports === Compared shared library exports: 6
Release comparison failed: 12 difference(s)
```

Reproduced on nightly-test runs of plain-`main` commits `34289321138` (`e3a28f0b`) and `34414847486` (`36949e32`), both zero commits ahead of `main`. Levels 2-4 are clean; the 12 differences are entirely from level 1.

## Cause

`absorb_dereferenced_links` reconciles the shared-object aliases that `actions/upload-artifact` flattens into copies on the Make side, and checked each flattened copy against the file the **Bazel** symlink resolves to:

```python
target = (bazel_root / path).resolve()          # Bazel's libonedal.so.4.0
if not filecmp.cmp(make_root / path, target, shallow=False):
```

`make/lib/intel64/libonedal.so` is a copy of **Make's** `libonedal.so.4.0`. Byte-comparing it to **Bazel's** build of the same library asks two independently built shared objects to be byte-identical — which nothing else in this script asks: `.so` is in `BINARY_SUFFIXES` so level 3 skips it, and level 4 compares exported symbols precisely *because* the bytes differ.

So the check cannot pass on any commit. 6 libraries x 2 aliases = 12, all failing, which is why the mismatch count and the reconciled count are the same number. Azure's `LinuxReleaseCompare` is green because it does not pass `--make-symlinks-dereferenced`: both of its trees travel through pipeline artifacts, so neither has symlinks left and nothing is reconciled. The GitHub nightly is the only consumer of the flag.

## Fix

Compare the flattened copy against the file the **Make** tree holds at the path the Bazel symlink resolves to. That is exactly what the symlink asserted before the transport flattened it (`libonedal.so` and `libonedal.so.4` carried the bytes of `libonedal.so.4.0`), and it is checkable from what survives transport.

```python
target_path = rel(target, bazel_root)
make_target = make_root / target_path
if not make_target.is_file():
    missing_targets.append((path, target_path))
elif not filecmp.cmp(make_root / path, make_target, shallow=False):
    mismatches.append((path, target_path))
```

Also in this change:

* New `missing_targets` bucket — a Bazel alias whose target the Make release does not ship has nothing to be checked against, and is reported rather than silently reconciled.
* An entry that fails is no longer also printed under "reconciled", and that line now reads `N of M`, so a tolerated difference cannot read as full coverage.
* The docstring records why the cross-tree byte comparison cannot hold, so it does not come back.

The pre-existing caveat is unchanged and still documented: `resolve()` collapses the whole chain, so the *shape* of the Bazel link chain (whether `libonedal_core.so` hops through the SONAME-carrying `.so.4`) is still not asserted anywhere.

## Validation

Synthetic trees in the nightly's shape — 3 libraries, real `.so.4.0` plus `.so`/`.so.4` symlinks on the Bazel side, flattened copies with different bytes on the Make side:

| case | before | after |
|---|---|---|
| healthy trees (nightly shape) | `Release comparison failed: 6` | **passes**, `reconciled 6 of 6` |
| Make copy stale w.r.t. its Make target | caught | caught, 1 mismatch |
| Bazel alias retargeted at another library | lost in the 6 false mismatches | caught, 1 mismatch |
| Bazel alias target absent from Make | not distinguished | caught, 2 missing targets + level-1 set diff |

Real-CI confirmation needs a nightly run on this branch (`workflow_dispatch` with a `source_run_id`), since a local simulation cannot produce genuine Make and Bazel artifacts.

## Relation to #3764

#3764 adds `reconcile_dereferenced_symlinks` on a base that predates `absorb_dereferenced_links`, and its version compares the Make copy against `(bazel_root / path).resolve()` too — the same cross-tree byte comparison, so it would reconcile 0 of 12 and leave the aliases to be counted at level 1. Whichever lands second should keep the Make-side target comparison.

### DIFF
--- a/dev/release_tests/compare_release_trees.py
+++ b/dev/release_tests/compare_release_trees.py
@@ -346,10 +346,23 @@ def absorb_dereferenced_links(make_root, make_files, bazel_root, bazel_links, li
 
     Where Bazel has a symlink and Make a plain file at the same path, the pair is
     removed from level 1 and checked the only way the transport still permits:
-    the Make file must be byte-identical to the target the Bazel symlink resolves
-    to. Both collections are mutated in place; the number of reconciled entries
-    is printed, because a silently tolerated difference reads like a comparison
-    that covered everything.
+    the Make file must be byte-identical to the file the *Make* tree holds at the
+    path the Bazel symlink resolves to. That is precisely what the symlink
+    asserted before the transport flattened it -- `libonedal.so` and
+    `libonedal.so.4` carried the bytes of `libonedal.so.4.0` -- and it still
+    catches a Bazel tree whose link points somewhere else.
+
+    Comparing the Make copy against *Bazel's* build of the target instead, as
+    this did originally, demands that two independently built shared objects be
+    byte-identical. Nothing else in this script demands that: `.so` is a
+    BINARY_SUFFIX so level 3 skips it, and level 4 compares exported symbols
+    exactly because the bytes differ. The comparison therefore failed on every
+    reconciled entry on plain main (12 of 12, runs 34289321138 and 34414847486),
+    reporting the same paths as mismatches and as reconciled.
+
+    Both collections are mutated in place; the number of reconciled entries is
+    printed, because a silently tolerated difference reads like a comparison that
+    covered everything.
 
     What this cannot check, and no other comparison covers either: the *shape* of
     the Bazel link chain. `resolve()` collapses every hop, so a Bazel tree that
@@ -373,6 +386,7 @@ def absorb_dereferenced_links(make_root, make_files, bazel_root, bazel_links, li
     errors = 0
     absorbed = []
     mismatches = []
+    missing_targets = []
 
     for path in sorted(set(bazel_links).intersection(make_files)):
         target = (bazel_root / path).resolve()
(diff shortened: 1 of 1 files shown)


dev/release_tests/compare_release_trees.py
```diff
@@ -373,31 +386,50 @@ def absorb_dereferenced_links(make_root, make_files, bazel_root, bazel_links, li
...
-            f"Make files reconciled against Bazel symlinks: {len(absorbed)}"
-            " (--make-symlinks-dereferenced)"
+            f"Make files reconciled against Bazel symlinks: {len(reconciled)}"
+            f" of {len(absorbed)} (--make-symlinks-dereferenced)"
         )
-        for path in absorbed[:limit]:
-            print(f"  ~ {path}")
+        for path, target_path in absorbed[:limit]:
+            if path not in flagged:
+                print(f"  ~ {path} -> {target_path}")
```
Alexandr-Solovev, line 432: absorbed[:limit] is sliced before the if path not in flagged filter, so flagged entries eat slots in the list. Collect reconciled in the loop (else: branch of the two checks) and slice that instead; flagged on line 411 then becomes unneeded. Latent at the default limit of 50 with 12 entries.

dev/release_tests/compare_release_trees.py
```diff
@@ -373,31 +386,50 @@ def absorb_dereferenced_links(make_root, make_files, bazel_root, bazel_links, li
...
+        # flattened copy against, so the entry is not reconciled, only dropped.
+        if not make_target.is_file():
+            missing_targets.append((path, target_path))
+        elif not filecmp.cmp(make_root / path, make_target, shallow=False):
+            mismatches.append((path, target_path))
+
+    flagged = {path for path, _ in mismatches + missing_targets}
+
+    if missing_targets:
+        errors += len(missing_targets)
```
Alexandr-Solovev, line 414: errors += len(missing_targets) double-counts. Your comment on 402-405 says level 1 already reports the absent target as Only Bazel files; one missing .so.4.0 with two aliases gives 3 errors.

dev/release_tests/compare_release_trees.py
```diff
@@ -373,31 +386,50 @@ def absorb_dereferenced_links(make_root, make_files, bazel_root, bazel_links, li
...
+        for path, target_path in missing_targets[:limit]:
+            print(f"  ! {path}: Bazel {path} -> {target_path}, absent from Make release")
     if mismatches:
         errors += len(mismatches)
         print(f"Dereferenced symlink content mismatches: {len(mismatches)}")
-        for path, target in mismatches[:limit]:
-            print(f"  ! {path}: Make file differs from Bazel {path} -> {target}")
+        for path, target_path in mismatches[:limit]:
+            print(f"  ! {path}: Make file differs from Make {target_path}"
+                  f" (Bazel {path} -> {target_path})")
```
Alexandr-Solovev, line 423: likely redundant here,  ! {path}: Make file differs from Make {target_path}" is enough in my understanding

Alexandr-Solovev (commented):
Overall looks good with minor comments

dev/release_tests/compare_release_trees.py
napetrov: Fixed in 3491e43a. `reconciled` is now collected in the loop (`else:` after the two checks) and sliced directly; `flagged` is gone. Verified with a synthetic tree at `--summary-limit 1`, one mismatch and one good alias: before, the mismatch consumed the slot and nothing was listed; now `~ lib/libonedal.so.4 -> lib/libonedal.so.4.0` prints.

dev/release_tests/compare_release_trees.py
napetrov: Fixed in 3491e43a — `errors += len(missing_targets)` dropped. The list is still printed, since it explains why those aliases are absent from the reconciled list, with a comment saying explicitly that it is not counted because level 1 reports the target. Synthetic tree with `libonedal.so.4.0` missing from Make and two Bazel aliases: 3 errors before, 1 now (`Only Bazel files: 1  + lib/libonedal.so.4.0`).

dev/release_tests/compare_release_trees.py
napetrov: Agreed — suffix dropped in 3491e43a. Line is now `! lib/libonedal.so: Make file differs from Make lib/libonedal.so.4.0`.

Alexandr-Solovev (approved):
LGTM! Lets wait for ci and then merge!

PR #3790, opened 2026-09-16
Drop the empty Windows onedal_dynamic_dpc Bazel target

napetrov:
Second of the follow-ups approved in https://github.com/uxlfoundation/oneDAL/pull/3737#issuecomment-5696098644 (the first is #3789).

`dev/bazel/deps/onedal_win.tpl.BUILD` declared:

```python
cc_library(
    name = "onedal_dynamic_dpc",
    srcs = [],
    deps = [":headers"],
)
```

The Windows side of `onedal_repo` (`dev/bazel/deps/onedal.bzl`) declares no DPC files at all — `win_libs` has no `onedal_dpc_dll.lib`, `win_bins` has no `onedal_dpc.<major>.dll`, and there is no `@mkl//:mkl_dpc` dependency the way the Linux template has one. So the target could only ever be an empty library: a consumer depending on it would link successfully and get a binary with no oneDAL in it.

Nothing in the tree references it (`grep -rn onedal_dynamic_dpc` finds only the two template definitions), so removing it turns a silent no-op into `no such target`. A comment in its place records why Windows has no such target, so it does not get re-added as a symmetry fix.

Making it real is the other option, but it would mean declaring the Windows DPC import library and DLL in the repository rule first, i.e. asserting that a Windows Bazel DPC release exists — a bigger change than this cleanup, and not something I can verify.


dev/bazel/deps/onedal_win.tpl.BUILD
```diff
@@ -111,10 +111,9 @@ filegroup(
...
 )
 
-cc_library(
-    name = "onedal_dynamic_dpc",
-    srcs = [],
-    deps = [
-        ":headers",
-    ],
-)
+# No `onedal_dynamic_dpc` here, unlike `onedal.tpl.BUILD`: the Windows side of
```
Alexandr-Solovev: comment looks too big, may be makes sense to make it shorter

Alexandr-Solovev (approved):
Minor comment, overall looks good

dev/bazel/deps/onedal_win.tpl.BUILD
```diff
@@ -111,10 +111,9 @@ filegroup(
...
+# `onedal_repo` declares no DPC files at all -- neither `onedal_dpc_dll.lib` in
+# `win_libs` nor `onedal_dpc.<major>.dll` in `win_bins` -- so the target can
+# only ever be an empty library that links nothing and quietly produces a
+# binary with no oneDAL in it. A consumer asking for it now gets "no such
```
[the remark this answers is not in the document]
napetrov: Checked this against the configured graph rather than by reading the macros, and the mechanism is real but the conclusion is not: this target is unreachable in every configuration CI exercises, and its static sibling does not exist at all.

`--test_link_mode=release_static` on the same DPC++ example already fails on `main`, because `_get_dpc_deps` asks for a target no template declares:

```
$ bazel cquery 'filter(":onedal_", deps(//examples/oneapi/dpc:_basic_statistics_dense_batch_dpc))' \
      --test_link_mode=release_static
ERROR: no such target '@@+onedal_repo+onedal_release//:onedal_static_dpc' ...
  and referenced by '//examples/oneapi/dpc:_basic_statistics_dense_batch_dpc'
```

`release_dynamic` resolves only because `onedal_dynamic_dpc` happens to exist:

```
$ bazel cquery 'filter(":onedal_(dynamic|static)", deps(//examples/oneapi/dpc:_basic_statistics_dense_batch_dpc))' \
      --test_link_mode=release_dynamic
@onedal_release//:onedal_dynamic_dpc (431669a)
```

Nothing in CI hits either path. The release-mode example steps in `.ci/pipeline/ci.yml` are `//examples/daal/cpp:all` and `//examples/oneapi/cpp:all`; neither package has a `_dpc` `cc_test`/`cc_binary` — the only `dpc` entries are the `example_util_dpc` header module — and `//examples/oneapi/dpc:all` is never built in a release link mode on either platform.

On Windows specifically, the target could not have worked even if it were reached: `srcs = []`, no `@mkl//:mkl_dpc`, and `onedal_repo` declares no Windows DPC files, so it would have linked an executable with no oneDAL in it and no diagnostic.

The asymmetry you point at is worth fixing on its own terms, so I filed it as #3791 with both queries and the two candidate fixes (declare the missing release targets, or stop `_dpc`-mangling labels that come from `_test_link_mode_deps`) plus the missing CI step that would keep the configuration honest. Keeping this PR as the removal.

dev/bazel/deps/onedal_win.tpl.BUILD
napetrov: Shortened to two lines in `b6807642` — it now just says `onedal_repo` declares no Windows DPC files, so the target could only be an empty library. The longer reasoning lives in the commit message and in #3791.

PR #3794, opened 2026-09-17
perf(dtrees): stop scaling the GPU distributed row buffers with the total row count

Alexandr-Solovev:
## Description

The GPU histogram Decision Forest training kernel allocated its per-tree row-order
buffers against the **all-ranks** row count, so in SPMD mode every rank paid for the
rows owned by all the other ranks and the per-rank footprint grew linearly with the
rank count. This PR makes every allocation scale with the **rank-local** row count only.

### The problem

* `tree_order_lev_` and `tree_order_lev_buf_` were sized
  `selected_row_total_count_ * tree_in_block_` on every rank.
* `required_mem_size_for_one_tree` and the OOB row-list stride used the same global count.
* The bootstrap path generated indices over `[0, row_total_count_)` and then ran a
  compaction kernel that kept only the ones falling in
  `[global_row_offset_, global_row_offset_ + row_count_)`, leaving the rest of the buffer
  as padding.
* The non-bootstrap path gave rank *i*
  `min(selected_row_total_count_ - global_row_offset_, row_count_)` rows, which is **zero**
  for the later ranks whenever `observations_per_tree_fraction < 1.0` — those ranks
  contributed nothing to any tree.

### The change

* `selected_row_count_ = observations_per_tree_fraction * row_count_` in **both** batch and
  distributed mode, clamped to at least one row for a rank that owns rows, and
  `selected_row_total_count_ = allreduce(selected_row_count_)` through the renamed
  `get_total_count(bool, Index)`.
* The row-order buffers, the one-tree memory estimate and the OOB row-list stride all use
  the local count, so nothing scales with the rank count.
* The bootstrap path draws directly in `[0, row_count_)`. The global draw plus compaction
  kernel is gone, together with `get_global_row_offset`, `row_total_count_`,
  `global_row_offset_`, and `gen_initial_tree_order`'s now-unneeded `node_list` /
  `engine_offset` parameters.
* `selected_row_total_count_` survives as a scalar only: input validation, the root
  `ind_grc`, and the per-rank RNG substream stride. The initial statistics are still
  reduced over the ranks, so the root impurity remains a global one and the split
  quality of the existing distributed scheme is unchanged.
* Asserts that pinned buffer sizes to the global stride were re-derived against the local
  one (`train_service_kernels_dpc.cpp`, `train_splitter_impl_dpc.cpp`), and the
  `global_row_count` parameter of `mark_present_rows` / `get_oob_row_list` was renamed
  `row_list_stride` to match what it actually is.

### Drive-by cleanup

Removed ten private declarations in `train_kernel_hist_impl.hpp` that had no definition
anywhere in the tree (`compute_histogram`, `compute_histogram_distr`,
`compute_partial_histograms`, `reduce_partial_histograms`, `compute_partial_count_and_sum`,
`compute_partial_sum2cent`, `sum_reduce_partial_histograms`, `fin_histogram_distr`,
`get_part_hist_required_mem_size`, `get_part_hist_elem_count`) and three unused members
(`selected_row_host_`, `selected_row_global_host_`, `node_imp_decr_list_`).

### Testing

`cpp/oneapi/dal/algo/decision_forest/test/spmd.cpp` gains:

* `df cls / reg observations per tree fraction over several ranks` — rank count x
  `observations_per_tree_fraction` x `bootstrap`, with accuracy checks on real datasets.
  These are `[external-dataset]` tests, so **they need CI to run**; I have no dataset
  directory available locally.
* `df cls observations per tree fraction with a small number of rows` — a
  non-external-dataset case on the 6-row built-in fixture (ranks 2/3, fractions
  0.3/0.7/1.0, bootstrap on/off) asserting the trained model shape. Fraction 0.3 over
  3 ranks is exactly the case where the fraction truncates a rank's local sample to a
  single row, i.e. the new clamp.

Locally, all 8 `decision_forest` `_dpc` targets and all 8 `_host` targets pass. Worth
flagging: `test_spmd_dpc` currently aborts on `main` on my machine with
`Segmentation fault from GPU ... banned: 1` out of `drm_neo.cpp`, and it passes with this
change — that abort was a real out-of-bounds write in these buffers, not a driver issue.

### Known limitations, unchanged by this PR

* Only the root statistics are reduced across ranks; `calc_imp_dec` then derives the right
  child as `global_parent - local_left`, so descendant `ind_grc` values drift. Making the
  splitter fully local was tried and reverted: it regresses the small SPMD accuracy checks,
  because the root allreduce is what carries the other ranks' class counts. A real fix needs
  the per-(node, feature, bin) histograms reduced, and `best_split` builds those in SYCL
  local memory without ever materializing them globally. Out of scope here.
* The `skip_ahead` per-rank budget `rank * selected_ftr_count_ * tree_count_ * 2` is not
  depth-aware, so per-rank RNG substreams can overlap for very deep trees. This affects the
  statistical independence of the per-rank randomness only, not correctness, and predates
  this PR.

---

<details>
<summary>Checklist:</summary>

**Testing**

**Performance**

- [x] I have provided justification why performance and/or quality metrics have changed or why changes are not expected. This is a memory-footprint change in the distributed GPU path: per-rank row-order allocations go from `O(total_rows)` to `O(local_rows)`. Batch mode allocates and computes exactly as before. Quality metrics are expected to change only for distributed runs with `observations_per_tree_fraction < 1.0`, where ranks that previously received zero sampled rows now contribute to every tree.

</details>


avolkov-intel:
I would postpone this enhancements, since currently algorithm has some issues with correctness that have higher priority

avolkov-intel (changes requested):
Hold until issues with correctness are fixed

Alexandr-Solovev:
@avolkov-intel understood on priority, and I am not asking to jump the queue -- but I would like to check that this one is actually in the queue's way, because I do not think it touches the code the correctness problems live in.

What this PR changes is **buffer sizing only**: the distributed tree-order buffers are allocated from `selected_row_total_count_` (the all-ranks row count) instead of the rank's local row count, so every rank carries `rank_count` times the rows it will ever index. The change makes the allocation local and does the row sampling per rank. It does not touch `calc_imp_dec`, `best_split`, the histogram reduction, or anything that decides a split -- so it cannot move a split decision, and it cannot fix or hide a wrong one.

Two things that argue for landing it rather than holding it:

1. **It removes a failure rather than adding risk.** All 8 DF GPU bazel targets pass locally on this branch, `test_spmd_dpc` included -- and that target currently **segfaults at `upstream/main`** on this machine. Holding the PR keeps that segfault in place.
2. **Holding it makes the correctness work harder, not easier.** Anyone debugging the distributed splitters is doing it against buffers that are `rank_count` times larger than the data they describe, which is extra noise in exactly the code under investigation.

So: which correctness issue do you have in mind, and is there one you would like me to pick up instead? The one I have written down is the mixed global/local scale in `calc_imp_dec` -- only the root statistics are allreduced (`class_hist_list_` / `sum_list` / `sum2cent_list`, root `ind_grc = selected_row_total_count_`), and the right child is then derived as `global_parent - local_left`, so descendant `ind_grc` values drift. I tried making it fully local and **reverted it**: it regresses `test_spmd_dpc` "df cls base check with default params bootstrap disabled", because on that 6-row dataset rank 0 sees only class-0 rows and the harness validates rank 0's model, so the root allreduce is what keeps the small SPMD tests usable. A real fix has to reduce the per-(node, feature, bin) histograms, which `best_split` currently keeps in SYCL local memory and never materializes. Happy to take that on as the prerequisite if it is the blocker you mean.

If you would still rather park this one regardless, say so and I will convert it to a draft with a note pointing at whichever fix it is waiting for -- I just want the dependency written down rather than implied.

PR #3718, opened 2026-07-26
Feature/win arm64 support

nithin-aashik-mcw:
## Add Windows ARM64 (winarm) build support

Adds native build support for oneDAL on Windows ARM64 using clang/lld-link, extending the existing `win32e` (MSVC) and `lnxarm` targets.

### Build system
- New `winarm` platform target with clang/lld-link/llvm-lib toolchain (`winarm.mk`, `clang.ref.arm.mk`, `arm.mk`, `lnxarm.mk`, `vc.mkl.32e.mk`, `common.mk`)
- Registered `winarm` in top-level `PLATs`; made Windows flags (`-Fo`, `-Q`, `-EHsc`, link rules) conditional on MSVC vs. clang
- Fixed TBB/OpenBLAS path resolution to be per-architecture (`$(_IA)`) instead of hardcoded to `intel64`, in both `makefile` and `deps.ref.mk`
- Wired OpenBLAS into the same release-copy machinery as TBB
- Added VS2026 build tools support with arch-aware `vcvarsall.bat` (replacing x64-only `vcvars64.bat`)

### Core / algorithm compatibility (clang-cl ≠ real MSVC)
- Guarded `_MSC_VER`-only code paths with `!defined(__clang__)` across `daal_defines.h`, `service_defines.h`, `service_topo.cpp`, `env_detect_features.cpp`, KD-tree KNN kernel, and `oneapi/dal/{backend,detail}/common.hpp`
- Added ARM64 Windows SVE feature detection via `KUSER_SHARED_DATA`
- Switched to `_aligned_malloc`/`_aligned_free` on Windows instead of POSIX `aligned_alloc`/`free`

### CI / tooling
- `build.bat` / `test.bat`: detect host arch (`AMD64`/`ARM64`), set `PLAT=winarm`, propagate `ARCH_DIR` through TBB/OpenBLAS paths
- Fixed robocopy bug in `tbb.bat` deleting source dirs after move
- Fixed OpenBLAS archive path casing (`OpenBlas-*` → `OpenBLAS-*`)
- `deploy/local/vars_win.bat`: derive `DAAL_IA` from host arch instead of hardcoding `intel64`
- `examples/cmake/setup_examples.cmake` / `samples/cmake/setup_samples.cmake`: recognize `ARM64`, fix `intel_intel64` → `intel64` naming

### Testing
- [✅] Built and ran examples/samples natively on Windows ARM64
- [✅] CI passing on `feature/win-arm64-support`

PR #3795, opened 2026-09-17
build(bazel): support Windows ARM64 (Make PLAT=winarm parity)

napetrov:
## Description

The Bazel build gained AArch64/RISC-V64 cross-compilation for Linux (#3719) and Make gained a Windows ARM64 platform (#3718), but the Bazel *Windows* lane was still x86_64-only in every layer. This closes that gap, mirroring Make's `PLAT=winarm` (`dev/make/function_definitions/winarm.mk`, `dev/make/compiler_definitions/clang.ref.arm.mk`).

### What was missing

| Gap | Location |
|---|---|
| Windows host arch hardcoded `intel64` | `dev/bazel/toolchains/common.bzl` |
| No clang-cl toolchain (icx has no AArch64 target, MSVC `cl` is not what Make uses) | `cc_toolchain_win.bzl` |
| Toolchain reqs + exec/target platform constraints pinned to `intel64`/`x86_64` | `cc_toolchain_win.bzl`, `cc_toolchain_win_icx.tpl.BUILD` |
| `extra_toolchain_win` pinned to `cpu:x86_64` | `extra_toolchain_win.tpl.BUILD` |
| No `windows_arm64` platform | `config.tpl.BUILD` |
| No Windows clang flag set (`--target=`, `-march=armv8-a+sve`, `-fms-runtime-lib`) | `flags.bzl` |
| `@openblas` Linux-only (`lib/libopenblas.a`) | `deps/openblas.bzl` |
| x86_64 default download URLs silently used on ARM64 hosts | `repos.bzl`, `MODULE.bazel` |
| `bazelisk.ps1` hardcodes the amd64 asset | `.ci/env/bazelisk.ps1` |
| x86 `cpudetect` probe compiled on every Windows host | `config/config.bzl` |

### Changes

**Arch detection.** `detect_host_arch` reads `PROCESSOR_ARCHITEW6432` / `PROCESSOR_ARCHITECTURE` on Windows (as `.ci/scripts/build.bat` does), preferring the former so an emulated x86_64 shell does not misreport an ARM64 machine. `detect_target_arch` additionally honors an explicit `--target=<triple>` in `CC`/`CXX`, which is how clang-cl selects its target; that parsing is deliberately kept out of `_detect_cross_compiler`/`get_cross_tool_prefix`, since unlike a triple-prefixed binary name it says nothing about tool *names*. The `cpudetect.cpp` x86 ISA probe now runs only for `intel64` targets.

**Toolchain.** The custom icx toolchain is generalized into a clang-cl-*family* toolchain (icx or upstream clang-cl); `cc_toolchain_win_icx.tpl.BUILD` is renamed to `cc_toolchain_win_llvm.tpl.BUILD` and takes its exec/target CPU constraints from the detected arch. `cc_toolchain_config_win.bzl` needed no functional change — `lld-link` and `llvm-lib` infer the machine type from the objects, and the triple arrives through the compile flags. clang's resource dir is added to the builtin includes, since `INCLUDE` never carries it.

Selection: an ARM64 target always uses clang-cl, and a conflicting `ONEDAL_WIN_COMPILER` **fails with a diagnostic** rather than silently resolving to a toolchain that cannot build the target. On x86_64, clang-cl becomes selectable (`ONEDAL_WIN_COMPILER=clang` or `CC=clang-cl`); the existing default (MSVC `cl` auto-config, icx when present) is unchanged.

**Dependencies.** `@openblas` gains a Windows layout — `.ci/env/openblas.bat` builds a shared library, so the consumable pair is `lib/openblas.lib` + `bin/openblas.dll`, exactly what `OPENBLASDIR.libia.win`/`OPENBLASDIR.soia.win` select in `dev/make/deps.ref.mk`. Since every default URL in `MODULE.bazel` is an x86_64 package, `prebuilt_libs_repo_rule` now fails when a download would supply the wrong architecture (naming the root env var to set) instead of producing binaries that break at link time.

**CI.** `.ci/env/bazelisk.ps1` picks the asset matching the host arch, as `bazelisk.sh` already does. `ci-win.yml` gains a Bazel job on the native `windows-11-vs2026-arm` runner that builds ARM64 OpenBLAS and oneTBB, builds `//cpp/daal:core_dynamic //cpp/oneapi/dal:dynamic` with `--platforms=@config//:windows_arm64 --backend_config=ref`, and verifies the PE machine type — the counterpart of the AArch64 Linux lane's `readelf` check.

### Not covered

Windows ARM64 *release packaging*: `dev/bazel/release.bzl` and `dev/bazel/scripts.bzl` hardcode `lib/intel64` / `redist/intel64` instead of honoring `ARCH_DIR_ONEDAL` (`ARM64` for winarm, `arm` for lnxarm). This is pre-existing and shared with the Linux AArch64 target added in #3719, so it is left for a separate change; the new CI job builds libraries, not `//:release`.

## Testing

Locally (Linux x86_64) verified no regression: `bazel cquery //cpp/daal:core_static` and `//cpp/oneapi/dal:dynamic` analyze clean, and `dev/bazel/tests/parameters_layout_test.sh` (which includes cross-platform Windows-platform analysis) passes. The ARM64 build itself is exercised by the new CI job — I have no Windows ARM64 machine locally, so that lane's first run in CI is the real validation.

## Checklist

- [x] Mirrors the Make behavior it claims to (each non-obvious flag/path cites the Make file and variable in a comment)
- [x] No change to the existing x86_64 Windows or Linux defaults
- [x] Documentation updated (`dev/bazel/README.md`)


napetrov:
@Alexandr-Solovev  @david-cortes-intel , please review

.ci/env/bazelisk.ps1
```diff
@@ -32,7 +32,16 @@ updated, which also makes the script convenient for local CI reproduction.
 $ErrorActionPreference = "Stop"
 
 $bazeliskVersion = "v1.29.0"
-$assetName = "bazelisk-windows-amd64.exe"
+# Bazelisk runs on the CI exec host, so pick the asset matching that host's
+# architecture: `bazelisk-windows-amd64.exe` would run under emulation at best
+# on the ARM64 runners (`.github/workflows/ci-win.yml`).
```
david-cortes-intel: I think this comment is redundant.

.github/workflows/ci-win.yml
```diff
@@ -120,4 +130,135 @@ jobs:
...
+  bazel-build:
+    name: WindowsBazel_OpenBLAS(SVE)-GitHub hosted
+    if: github.repository == 'uxlfoundation/oneDAL'
+    runs-on: windows-11-vs2026-arm
+    timeout-minutes: 180
+    steps:
+      - name: Checkout oneDAL
+        uses: actions/checkout@v6
+
+      # Bazel needs a Bourne shell on Windows for shell-based actions; the
```
david-cortes-intel, line 127: Could this whole yaml call the shell scripts with these same commands instead?

dev/bazel/config/config.tpl.BUILD
```diff
@@ -45,6 +45,22 @@ config_setting(
     },
 )
 
+# Windows + ref backend. A Windows DLL has to resolve every symbol at link
+# time, so `onedal_core.<major>.dll` needs the math backend linked in directly
+# instead of relying on the loader to bind BLAS/LAPACK calls to
+# `onedal_thread.<major>.dll` the way ELF does. Matches makefile:528, where
```
david-cortes-intel: Could have a permalink. Line numbers will change what's pointed at once the file is modified further.

dev/bazel/deps/openblas.bzl
```diff
@@ -29,4 +29,18 @@ openblas_repo = repos.prebuilt_libs_repo_rule(
...
+    # OPENBLASDIR.soia.win in dev/make/deps.ref.mk.
+    win_includes = [
+        "include",
+    ],
+    win_libs = [
+        "lib/openblas.lib",
+    ],
+    win_bins = [
+        "bin/openblas.dll",
+    ],
```
david-cortes-intel, line 44: Isn't openblas supposed to be statically-linked? If not, wouldn't that later on cause conflicts if this openblas-openmp is loaded into the same process as an openblas-pthreads from a scikit-learn pip install?

.ci/env/bazelisk.ps1
napetrov: Removed in 995f977c — the `switch` right below says the same thing.

dev/bazel/config/config.tpl.BUILD
napetrov: Fixed in 995f977c, though not with a permalink: a permalink freezes a SHA, so it keeps resolving but starts pointing at a stale tree. Instead the comment now names the thing it refers to — the makefile rule for `$(WORKDIR.lib)/$(core_y)` and `daaldep.math_backend.shared_link_deps` in `dev/make/deps.ref.mk` — which is greppable and survives edits. Happy to switch to permalinks if you prefer them; there are other `makefile:<line>` references in this PR and in the pre-existing Bazel comments that would want the same treatment for consistency.

.github/workflows/ci-win.yml
napetrov: Done in 995f977c. MSYS2, LLVM and CMake/Ninja are now `.ci/env/msys2.ps1`, `.ci/env/llvm.ps1` and `.ci/env/cmake_ninja.ps1`, called from both the Make job and the Bazel job the same way they already call `openblas.bat` / `tbb.bat` / `bazelisk.ps1`. They take the arch as a parameter (defaulting to the host) and derive the asset names from it, so the `llvm_arch: woa64` matrix entry is gone — that spelling is an LLVM asset naming detail, not a property of the runner.

It also fixed a latent bug in the Make job: `echo "%ProgramFiles%\LLVM\bin" | Out-File $env:GITHUB_PATH` appended that literal string, because cmd-style expansion does not happen in PowerShell. Harmless so far only because every later step sets `PATH` itself.

dev/bazel/deps/openblas.bzl
napetrov: Statically on Linux, yes — `@openblas` there consumes `lib/libopenblas.a`, and that has not changed. Windows is shared, but that is not a decision of this PR: `.ci/env/openblas.bat` has configured OpenBLAS with `-DBUILD_SHARED_LIBS=ON` since b791e691 (#3718, the Make winarm PR), and `dev/make/deps.ref.mk` consumes exactly that pair — `OPENBLASDIR.libia.win := $(OPENBLASDIR)/lib` (the `openblas.lib` import library, so `releaseopen_blas.LIBS_A` on Windows is not an archive) plus `OPENBLASDIR.soia.win := $(OPENBLASDIR)/bin` for `openblas.dll`. The Bazel repo rule just consumes the same install tree, so the two build systems link the same thing; making it static would be a change to the Make lane as much as to this one.

On the conflict: Windows has no ELF-style symbol interposition. `openblas.dll` and a wheel-shipped `libopenblas*.dll` are distinct modules resolved per-import, so neither binds to the other, unlike the `libopenblas.so.0` case where whichever copy is loaded first wins. What does remain is two independent thread pools oversubscribing the machine, and that is equally true of a statically linked copy inside `onedal_core` — this build is plain OpenBLAS threading, not `USE_OPENMP`, so there is no OpenMP runtime to clash either.

If you would still rather have Windows static, I would do it as a separate PR flipping `BUILD_SHARED_LIBS` in `.ci/env/openblas.bat` and dropping the `bin/openblas.dll` plumbing on both sides, so the Make and Bazel lanes move together. Happy to file it.

david-cortes-intel: Would that still hold if another wheel ships the file with name `openblas.dll` instead of `libopenblas.dll`? Note that on windows, the python package would call `add_dll_directory`:
https://github.com/uxlfoundation/scikit-learn-intelex/blob/a0fc42c38ebd67d361fdeab334edc21120a38166/onedal/__init__.py#L100

.github/workflows/ci-win.yml
david-cortes-intel: But this is now adding new files. What about the existing .bat files that were added in the PR that introduced winarm support?

dev/bazel/deps/openblas.bzl
napetrov: Good catch, and no — it would not hold. Our Windows DLL *is* named `openblas.dll` (CMake with `BUILD_SHARED_LIBS=ON` emits `bin/openblas.dll` + the `lib/openblas.lib` import stub), and Windows resolves an implicit import by **base name** against the modules already present in the process, so whichever `openblas.dll` got loaded first wins. With `add_dll_directory` in play there is no way to pin it. That is worse than what I described earlier, so I switched it rather than argue the odds.

`bf726028` builds OpenBLAS statically on Windows (`-DBUILD_SHARED_LIBS=OFF`), matching `.ci/env/openblas.sh`. That also restores what `dev/make/deps.ref.mk` already documents — "Symbols from OpenBLAS are copied into libonedal_core library" — which the shared build silently did not deliver, since `daaldep.math_backend.static_link_deps` was resolving to an import library.

Fallout, all in that commit:
- `releaseopen_blas.LIBS_Y` is now empty on Windows; otherwise `.release.t` in the top-level makefile would chase `bin/openblas.dll`, which no longer exists.
- `@openblas` loses `win_bins` and `openblas_runtime` is empty, so no DLL travels with the test binaries.
- `openblas.lib` is removed from `_WINDOWS_IMPORT_LIBRARIES` in `dev/bazel/cc/common.bzl` — it is a genuine archive now and belongs in the archive merge, like the MKL static libs.

.github/workflows/ci-win.yml
napetrov: The existing `.bat` files are untouched and still called straight from the yaml — `.ci/env/openblas.bat` and `.ci/env/tbb.bat` build dependencies from source, which is a different job from installing a toolchain, and nothing in them is duplicated by the new files.

What moved out of the yaml is the three installer steps (MSYS2, LLVM, CMake+Ninja). They download over HTTP, run `msiexec`, and append to `GITHUB_PATH`, so PowerShell is the natural shell — and `.ci/env/bazelisk.ps1` (#3670) is the existing precedent for a `.ps1` installer in that directory. Writing them as `.bat` would mean `curl` plus hand-rolled arch mapping and no `$ErrorActionPreference = "Stop"`.

They are new files rather than additions to an existing script because both jobs in this workflow need all three, and each takes the arch as a parameter (defaulting to the host, so they are runnable locally as well). The extraction also fixed a real bug that was invisible inline: the Make job did `echo "%ProgramFiles%\LLVM\bin" | Out-File $env:GITHUB_PATH` under `shell: pwsh`, where `%ProgramFiles%` is not expanded, so the literal string went onto `PATH`.

If you would rather see them as `.bat` files, or folded into a single `.ci/env/win_toolchain.ps1` with a switch, say which and I will redo it.

dev/bazel/deps/openblas.bzl
napetrov: The static flip is in and green now, so here is what it took beyond the linkage switch itself, in case any of it looks wrong to you:

* `dev/make/deps.ref.mk` — `releaseopen_blas.LIBS_Y` is empty on Windows. `.ci/env/openblas.bat` no longer produces a DLL, and `_release_common` would otherwise stage a file that does not exist.
* `dev/bazel/cc/common.bzl` — `openblas.lib` left `_WINDOWS_IMPORT_LIBRARIES`. It is a real archive now, so it has to take part in the archive merge instead of being diverted as an import library.
* `.ci/env/openblas.bat` — `NOFORTRAN=ON` so the archive does not carry `/DEFAULTLIB:flang_rt.runtime.dynamic` into every consumer (that is `NO_FORTRAN=1` on Linux), and `USE_THREAD=OFF USE_LOCKING=ON` because the CMake build defaults `USE_THREAD` to 1 on any machine with two cores. That second one was not academic: with a static OpenBLAS both `onedal_core` and `onedal_thread` embed the archive, so the threaded build put two OpenBLAS pools in the process underneath TBB and the example tests stopped finishing.
* `.github/workflows/ci-win.yml` — the released DLLs are now checked with `dumpbin /dependents` for any `openblas`/`flang` import, so the property you asked about is asserted on the artifact rather than assumed from the link line. The two test steps also got `timeout-minutes`, since `.ci/scripts/test.bat` runs each example with no timeout of its own.

One thing I found while doing this and did **not** change, because it is a wider decision than this PR: the Windows OpenBLAS is built LP64, while `DAAL_INT` is `__int64` on `_WIN64`/`TARGET_ARM` (`cpp/daal/include/services/daal_defines.h`) and `service_blas_declar_ref.h` passes `const DAAL_INT *`. The Linux script builds `INTERFACE64=1` by default (`.ci/env/openblas.sh --ilp64`). The examples pass today because a 64-bit argument slot read as its low 32 bits is the right value on little-endian, but integer *outputs* (LAPACK `info`) leave the upper half undefined. The fix is not just a flag: with `INTERFACE64=ON` the OpenBLAS CMake build renames the artifacts to `openblas_64.lib` and `include/openblas64` (`CMakeLists.txt`), so `deps.ref.mk`, `openblas.bzl`, `openblas_win.tpl.BUILD` and the new workflow check all need a Windows-specific name. Happy to do it here or in a follow-up — your call.

PR #3800, opened 2026-09-21
bazel: compare release library linkage, not just exported symbols

napetrov:
## Problem

The Make-vs-Bazel release comparison passed at its **highest** check level on a Bazel Linux release that no consumer can load (the defect #3799 fixes). Two structural gaps let that through:

1. Level 4 compared exported symbols only (`nm -D --defined-only`). Nothing looked at `DT_NEEDED` or at undefined dynamic symbols, so a library that resolves `tbb::detail::…` and `std::…` at runtime while recording no dependency on the libraries providing them was reported as equivalent. Measured on `main` with gcc 13.3 on both sides: all four Linux shared objects were short of dependencies the Make release records — `libtbb.so.12`/`libtbbmalloc.so.2` on `libonedal_thread`, `libstdc++.so.6` everywhere — and the comparison found no differences.
2. The comparison is rooted at `<release>/daal/latest`, so Make's sibling `<release>/tbb/latest/lib` (`makefile:274-279,1101`) was out of scope: a Bazel release staging no oneTBB at all was indistinguishable from one staging it correctly.

## Change

Level 4 now additionally compares, per shared library, the `DT_NEEDED` list and the undefined dynamic symbols, and checks that a runtime which a released oneDAL library *requires* and Make stages in a sibling component tree exists somewhere in the Bazel release. The staged-dependency rule is deliberately narrow — all three conditions must hold — because the two release roots differ in layout by design and a per-entry comparison of the sibling trees would fail on a correct tree.

### Undefined symbols get their own ignore rules

Not the export ones. Which oneTBB or libstdc++ instantiation a build emits as a *definition* is not part of the oneDAL surface, but an undefined oneTBB *reference* is exactly the signal this check exists for: it is what makes a missing `DT_NEEDED libtbb.so.12` provably a defect rather than a dependency the library does not need. Reusing the export list would drop all 30 `tbb::` references in `libonedal_thread` and all 20 in `libonedal_core` — i.e. hide the defect again.

So only compiler and standard-library internals are filtered, by prefix — libstdc++ templates and streams, `operator new`/`delete`, the Itanium ABI runtime (`__cxa_`, `_Unwind_`, `__cxxabiv1` vtables, stack-protector and TLS helpers), and the vectorizer's SIMD clones spelled per ISA letter (`_ZGVbN`, `_ZGVdN`, …) rather than as a bare `_ZGV`, which would also swallow a cross-library reference to a function-local static's guard variable — plus one named symbol: Make's `libonedal` keeps a local, unreferenced copy of `decision_forest::{classification,regression}::internal::ModelImpl::clear`, a thunk tail-jumping to `dtrees::internal::ModelImpl::clear` in `libonedal_core`, and that copy drags in an undefined reference to the base which Bazel's link drops along with the copy. Both trees define the base in `libonedal_core`, both keep the thunks local, and the exports are identical — dead-code placement, not a difference in what either release needs from its environment.

Measured on the four Linux libraries: **1261** raw references, **5** differences, 4 covered by the prefixes and 1 by the named ignore; the **1015** references that survive the filter — including all 64 `tbb::` ones — match entry for entry.

The SYCL libraries are exempt from the undefined-symbol half: Make links them with `icpx -fsycl … -lgomp` (`dev/make/compiler_definitions/dpcpp.mk`) and Bazel with its own DPC flag set, and no CI pairing builds them with the same compiler on both sides — the pre-existing *export* comparison already needs a DPC-specific allowance for the same reason. The skip is printed, not silent.

Libraries are deduplicated by content hash, so a tree whose `.so`, `.so.4` and `.so.4.0` are symlinks and one where they are three identical copies (what artifact publishing produces) both read as four libraries rather than four and twelve.

### Checks that refuse to answer rather than answer wrongly

* The staged-dependency check **refuses to run when both release roots share a base directory**. There each root enumerates the same siblings, so every component of one is trivially a component of the other and nothing can ever be reported missing — a green line that answers nothing. Two real release trees never share a base.
* `readelf -d` and `objdump -p` exit 0 on a file that is not an ELF object, so the dependency reader checks the header itself and treats a missing dynamic section as a read error rather than as an empty dependency list that would compare equal on both sides.
* A DT_NEEDED verdict already reached survives a later failure of the symbol read, and an unreadable staged component tree is charged as one error instead of aborting the run and discarding the differences already found.

The linkage half lives in a new `dev/release_tests/release_linkage.py`: the comparison script was already 814 lines, and these checks are a self-contained concern with their own ignore rules and external tools.

## Coverage and stated gaps

The checks go into level 4 rather than a new level 5, so the four existing CI invocations — which ask for level 4 or take it as the default — exercise them with no pipeline edits. Two caveats, both stated in output rather than silent:

* The staged-dependency check needs both release roots to be `<base>/<component>/latest`. The Azure `LinuxReleaseCompare` job passes a published artifact that is not that shape, so there the check prints that it is skipped and only the GHA nightly exercises it. (#3799 changes what that artifact contains, which also fixes the shape.)
* Windows reports the whole linkage comparison as skipped: the DLL equivalent needs `dumpbin /DEPENDENTS` and `/IMPORTS` parsing this script does not implement yet. A silent pass would be worse than a stated gap — the consequence is that a Windows release staging no `tbb12.dll` still goes undetected.

## Cross-compiler pairings (`--cross-toolchain`)

Both new checks are equalities, and an equality between two trees only holds when the same compiler built both. Azure's `LinuxReleaseCompare` pairs gcc Make with gcc Bazel; the GHA nightly pairs an **icx** Make release with a gcc Bazel one, so the nightly gets a new `--cross-toolchain` flag: undefined symbols are not compared at all, and DT_NEEDED differences are printed without failing the run. Read failures stay errors under either pairing.

This was measured, not assumed — an icx Make release was built on this box with the nightly's own script (`.ci/scripts/build.sh --compiler icx`, ISA set matched to the Bazel tree) and compared against the same Bazel tree that the gcc Make release passes against:

* **8 differences without the flag.** `libstdc++.so.6` is recorded by all four Bazel libraries and by none of the icx Make ones — `dev/make/compiler_definitions/icx.mkl.32e.mk` links with the C driver `icx` plus `-no-intel-lib`, so nothing adds it, while the gnu path links with `g++`. And `libm.so.6` is recorded by icx Make on `libonedal_thread`/`libonedal_parameters`, where it resolves none of their undefined symbols (measured: empty intersection with libm's exports). Neither is a Bazel defect; the first is arguably a real defect in the shipped icx Make tree, which is a separate conversation.
* **40 differing undefined symbols**, 29 of them in the `daal::` namespace, because icx and gcc inline and place code differently — which references cross a library boundary is a property of the compiler. No prefix list can express that, so the undefined half is skipped rather than filtered.
* **With the flag: 5 differences, all pre-existing on `main`** — one level-3 `dal-static-threading-host.pc` difference and four level-4 *export* mismatches from icx-vs-gcc typeinfo and vtable placement. The new staged-dependency check contributes 0 there.
* Isolating the compiler: the **gcc** Make tree against that same Bazel tree gives 0 DT_NEEDED and 0 undefined-symbol differences, so the whole delta above is icx-specific.

`LINUX_IGNORED_UNDEFINED_PREFIXES` also gains `_ZTVNSt`/`_ZTINSt` — the `__cxx11` stream classes nest one level deeper than `_ZTVSt` matches. A gcc-vs-gcc pair cannot expose that gap.

## Validation

Against real release trees built from this commit:

* Bazel release **without** the linkage fix → 6 differences, naming every missing `DT_NEEDED` entry and both unstaged oneTBB runtimes.
* Bazel release **with** the fix → green.
* Make release compared against a copy of itself → green.
* icx Make release vs the fixed Bazel release, `--cross-toolchain` → 5 differences, all pre-existing on `main` (without the flag: 13).
* A fully dereferenced copy of both trees (the artifact shape) → 12 library paths deduplicate to 4 and the verdict matches the symlinked trees.
* The unfixed Bazel release placed under the **same base directory** as the Make one → the four DT_NEEDED differences are still reported and the staged check states it cannot run there (before this was a silent green, because Make's own `tbb/latest` was credited to the Bazel release).

Levels 1-3 and the pre-existing level-4 export output are byte-identical to before.

### Known gaps, deliberately left

* No unit test pins the ignore lists or the measured numbers in the comments, and `dev/release_tests/` currently holds only shell tests. A typo'd prefix (the kind that makes an entry match nothing) would go unnoticed.
* Symbol *versions* are stripped (`@GLIBC_2.x`), matching the existing export reader, so a Bazel library requiring a newer version of the same symbol — raising the release's minimum glibc — compares equal. Catching that needs a baseline, not the other release.
* DT_NEEDED is compared as a set: order decides resolution precedence, but the two build systems order libm/libstdc++ differently on libraries that otherwise agree, so comparing sequences would fail on every correct tree.


napetrov:
Depends on #3799: without it the Azure `LinuxReleaseCompare` artifact is not a `<base>/<component>/latest` tree, so the new staged-dependency check prints that it is skipped there, and the DT_NEEDED comparison fails on every released library (which is the defect #3799 fixes).

napetrov:
Status note on the two red checks, both understood and neither a surprise from this branch's code:

**`Nightly-test / oneDAL Linux Bazel test` — 10 differences.** A `workflow_run` job runs the workflow definition from `main`, so the nightly used main's `nightly-test.yml`, i.e. *without* the `--cross-toolchain` flag this PR adds, while running this branch's `compare_release_trees.py`. That is the trap the header comment in `nightly-test.yml` warns about, and it cannot be verified from a fork branch — `workflow_dispatch` needs the ref to exist in this repository. Reading the run against that: 6 `DT_NEEDED` differences and 2 undefined-symbol differences would have been printed-not-counted under the flag, and the 2 remaining are the real finding — `libtbb.so.12` and `libtbbmalloc.so.2` are required by a released Bazel library and staged by Make only (`Staged sibling components: Make ['tbb'], Bazel []`). That is exactly #3799's defect, detected by this PR against a tree that does not have the fix.

**`CI (LinuxReleaseCompare)` — 6 differences.** Same shape: 4 `DT_NEEDED` mismatches, all Make-only entries (`libstdc++.so.6` everywhere, `libtbb.so.12`/`libtbbmalloc.so.2` on `libonedal_thread`), plus 2 undefined-symbol differences. With #3799 in the tree the `DT_NEEDED` half matches entry for entry, measured locally.

The undefined-symbol half of that job turned up something worth a fix rather than an explanation: the Bazel `libonedal.so`/`libonedal_core.so` reference `dlopen`/`dlsym`/`dlclose`/`dlerror`/`printf` and the Make ones do not, because the two jobs do not use the same oneMKL — apt oneMKL 2026.1.0 vs the pinned conda-forge mkl-static 2025.2.0, linked statically. Undefined-symbol differences are now printed rather than counted, with `--strict-undefined` to gate on them where both toolchain and dependency versions are controlled. `DT_NEEDED` is unaffected by that skew and stays an equality, so the check still fails a release that does not record its oneTBB dependency. Pushed as `f3434e8`, and the description has a new section on it.

